PR #9, opened 2020-04-16
Update prop metadata and emit summary for RSS

psavkar:


components/rss/rss.js
```diff
@@ -11,37 +11,43 @@ module.exports = {
...
+    for (let idx = feed.items.length - 1; idx >= 0; idx--) {
+      const item = feed.items[idx]
+      const key = this.itemKey(item)
+      // XXX throw if !key?
+      if (seenKeysMap[key]) continue
+      seenKeys.unshift(key)
+      seenKeysMap[key] = true // just in case of dupes
+      this.$emit(item, {
+        summary: item.title, 
+        ts: (item.pubDate ? new Date(item.pubDate).getTime() : undefined),
```
dylburger, line 44: You can remove the parens around this one. 

You should also be able to do: `+new Date()` in place of the getTime call - that also returns ms since the epoch.

psavkar: done

PR #11, opened 2020-04-17
updates to props and emit summary

psavkar:


components/http/http.js
```diff
@@ -1,20 +1,53 @@
...
+      body: this.resBody,
+      headers: {
+        "content-type": this.resContentType,
+      },
     });
-    // Emit the whole event, which contains
-    // the HTTP payload, headers, and more
-    this.$emit(event);
+
+    if(this.bodyOnly === true) {
```
tjk, line 47: just `if (this. bodyOnly)` also to be more explicit `emitBodyOnly`

components/http/http.js
```diff
@@ -1,20 +1,53 @@
...
+      description: "The status code to return in the HTTP response.",
+      optional: true,
+      default: '200',
+    },
+    resContentType: {
+      type: "string", 
+      label: "Response Content-Type",
+      description: "The content-type of the body returned in the HTTP response.",
+      optional: true,
+      default: `text/plain`,
```
tjk, line 26: a bit weird to have default be text/plain but default body be json no?

PR #18, opened 2020-05-02
More Twitter sources

psavkar:


components/twitter/my-liked-tweets.js
```diff
@@ -0,0 +1,25 @@
+const twitter = require('https://github.com/PipedreamHQ/pipedream/components/twitter/twitter.app.js')
+
+module.exports = { 
+  name: "my-liked-tweets", 
+  version: "0.0.1",
+  props: {
+    db: "$.service.db",
```
tjk, line 7: not used

components/twitter/my-liked-tweets.js
```diff
@@ -0,0 +1,25 @@
...
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+  dedupe: "unique",
+  async run(event) {     
+    (await this.twitter.getLikedTweets()).reverse().forEach(tweet => {
```
tjk, line 18: can you set `ts`? if so don't need reverse

components/twitter/user-tweets.js
```diff
@@ -0,0 +1,96 @@
...
+  async run(event) {
+    const from = `from:${this.from.replace('@','')}`
+    
+    let q = ''
+
+    if(this.q) {
+      q = this.q
+    }
+    
+    const since_id = this.db.get("since_id") || 0
```
tjk, line 38: this is basically "greatest" dedupe strategy -- right? i guess it just needs to pass you the checkpoint so you don't need db?

components/twitter/my-liked-tweets.js
psavkar: unfortunately, there's no timestamp when the "like" action occurred. the only timestamp is when the liked tweet was initially authored.

components/twitter/my-liked-tweets.js
psavkar: ack

components/twitter/twitter.app.js
```diff
@@ -1,5 +1,7 @@
 const axios = require('axios')
-const querystring = require("querystring")
+const _ = require('lodash')
```
tjk, line 2: Lodash is quite large, so avoid using it if you don't need it.

You are just doing 

```
if (_.get(response, 'status', 'Error') === 200) {
```

which is basically synonymous with `if ((response && response.status) || 'Error') === 200)`
which has no point in having that default since it will not change how the bool evaluates...

```
if (response && response.status === 200)
```

components/twitter/twitter.app.js
```diff
@@ -20,16 +28,49 @@ module.exports = {
...
+      type: "string",
+      label: "Max API Requests per Execution (advanced)",
+      description: "The maximum number of API requests to make per execution (e.g., multiple requests are required to retrieve paginated results). **Note:** Twitter [rate limits API requests](https://developer.twitter.com/en/docs/basics/rate-limiting) per 15 minute interval.",
+      optional: true,
+      default: "5",
+    },
+    from: {
+      type: "string",
+      label: "From",
+      description: "The screen name of the user (e.g., `pipedream`)",
```
tjk, line 46: I see `screen_name` is what the api uses but does the user-facing language use this? I've only heard of `handle`.... but what do i know

components/twitter/twitter.app.js
```diff
@@ -810,6 +870,13 @@ module.exports = {
...
     async _makeRequest(config) {
       if (!config.headers) config.headers = {}
+      if (config.params) {
+        const query = querystring.stringify(config.params)
+        delete config.params
+        const sep = config.url.indexOf('?') === -1 ? '?' : '&'
+        config.url += `${sep}${query}`
+        config.url = config.url.replace('?&','?')
+      }
       const authorization = await this._getAuthorizationHeader(config)
```
tjk, line 880: This is the call I think is 502ing. You should try/catch and decide if it fails whether you should just bail on this request or retry it. Bailing only works if the other methods are okay with this not returning a proper axios response

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+          count,
+          tweet_mode,
+        }
+      })).data
+    },
+    async lookupUsers(userIdArray) {   
+      return (await this._makeRequest({
+        url: `https://api.twitter.com/1.1/users/lookup.json`,
+        params: {
+          user_id: userIdArray.join(),
```
tjk, line 916: Would I had no idea and would not have predicted the default separator to be a comma.

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+        }
+      })).data
+    },
+    async search(q, since_id, tweet_mode, count, result_type, lang, locale, geocode, max_id) {
```
tjk, line 920: Is there a reason you don't just do `async search(params)`? You aren't really checking any of the keys for anything and that way at the call site it's more explicit what fields are passed (and easier to pass separate fields which I'm assuming can be optional)

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+      }
+  
+      if(includeRetweets === false) {
+        q = `${q} -filter:nativeretweets`
+      }
+  
+      const response = await this.search(q, since_id, tweet_mode, count, result_type, lang, locale, geocode, max_id)
+      
+      //console.log(response)
+      if (_.get(response, 'status', 'Error') === 200) {
```
tjk, line 981: see my previous comment

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+        if (!max_id) {
+          max_id = since_id || 0
+        }
+    
+        for (let tweet of response.data.statuses) {
+          if (tweet.id_str !== since_id && tweet.id_str !== max_id) {
```
tjk, line 987: why are you comparing string and numbers here? (or why are you setting max_id / since_id to `0` by default)

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+            if (enrichTweets) {
+              tweet.created_at_timestamp = moment(tweet.created_at, 'ddd MMM DD HH:mm:ss Z YYYY').valueOf()
+              tweet.created_at_iso8601 = moment(tweet.created_at, 'ddd MMM DD HH:mm:ss Z YYYY').toISOString()
+            }
+            tweets.push(tweet)
+            if (tweet.id_str > max_id) {
```
tjk, line 993: Please make sure they are always strings and then also this should be a tiny edge case, but you might need to check their size...

```
'100' > '20'
false
```

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
         console.log(err) // TODO
```
tjk, line 885: This is not a great pattern... but since you return undefined in these cases, you should just also return undefined in the 502 cases above

components/twitter/twitter.app.js
tjk: Yes this is an extra poor check because you are actually testing that `response` is not undefined since in any non-2xx case, it will throw and _makeRequest will return undefined -- so no point really in checking status (i doubt they return 2xx that is not 200)

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+        return {
+          tweets,
+          max_id,
+          min_id,
+          count,
+          resultCount: response.data.statuses.length,
+          statusCode: response.status, 
+        }
+      } else {
+        console.log(`Last request was not successful.`)
```
tjk, line 1014: See comment above about status check... should just more explicitely do

```
const resp = await this.search(...)
if (!resp) {
  // this code block
  return { statusCode: "Error" } // or whatever
}
// handle the stuff above but not nested in else
```

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+          result_type, 
+          enrichTweets,
+          includeReplies, 
+          includeRetweets,
+        })
+        
+        // increment the count of requests to report out after all requests are complete
+        totalRequests++
+        
+        if (response.statusCode !== 200) {
```
tjk, line 1066: same as above -- just check `if (!response)` right afterwards...

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+          break
+        } else {
+          response.tweets.forEach(tweet => {
```
tjk, line 1069:
```
tweets.push(...response.tweets)
```

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+            tweets.push(tweet)
+          })
+
+          //console.log(`total requests: ${totalRequests} max requests: ${maxRequests} results: ${response.resultCount} count: ${response.count}`)
+
+          if (totalRequests * 1 === maxRequests * 1 && response.resultCount === response.count) {
```
tjk, line 1075: what in the world is this advanced mathematics

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+            console.log(`The last API request returned the maximum number of results. There may be additional tweets matching your search criteria. To return more tweets, increase the maximum number of API requests per execution.`)
+          }
+
+          if (response.length === 0 || response.resultCount < response.count) {
```
tjk, line 1079: is count remaining from

components/twitter/twitter.app.js
```diff
@@ -818,19 +885,212 @@ module.exports = {
...
+            break
+          } else {
```
tjk, line 1081: no need for `else` if you are `break`ing in the `if`

components/twitter/user-tweets.js
```diff
@@ -0,0 +1,85 @@
...
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+  methods: {},
+  async run(event) {
+    const from = `from:${this.from.replace('@','')}`
+    const since_id = this.db.get("since_id") || 0
```
tjk, line 34: Why are you defaulting it? What is wrong with `undefined` as a falsy value?

components/twitter/user-tweets.js
```diff
@@ -0,0 +1,85 @@
...
+      tweets.sort(function(a, b){return a.id - b.id})
+
+      tweets.forEach(tweet => {
+        this.$emit(tweet, {
+          ts: moment(tweet.created_at, 'ddd MMM DD HH:mm:ss Z YYYY').valueOf(),
+          summary: tweet.full_text || tweet.text,
+          id: tweet.created_at_timestamp,
+        })
+
+        if (tweet.id_str > max_id || !max_id) {
```
tjk, line 76: Makes more sense to swap the order of the conditions

components/twitter/user-tweets.js
```diff
@@ -0,0 +1,85 @@
...
+
+    // emit array of tweet objects
+    if(tweets.length > 0) {
+      tweets.sort(function(a, b){return a.id - b.id})
+
+      tweets.forEach(tweet => {
+        this.$emit(tweet, {
+          ts: moment(tweet.created_at, 'ddd MMM DD HH:mm:ss Z YYYY').valueOf(),
+          summary: tweet.full_text || tweet.text,
+          id: tweet.created_at_timestamp,
```
tjk, line 73: Are the created_at_timestamp meant to be unique??

components/twitter/twitter.app.js
psavkar: updated. need to test.

components/twitter/twitter.app.js
psavkar: can we discuss after standup? i'm not sure what I need to do.

components/twitter/twitter.app.js
psavkar: I updated this

components/twitter/twitter.app.js
psavkar: done

components/twitter/twitter.app.js
psavkar: i changed this to parseInt() -- lmk if there's a better option

components/twitter/twitter.app.js
psavkar: note: add sleep in the catch

components/twitter/twitter.app.js
```diff
@@ -808,29 +867,237 @@ module.exports = {
...
+      }
+  
+      const response = await this.search({ q, since_id, tweet_mode, count, result_type, lang, locale, geocode, max_id })
+
+      if(!response) {
+        console.log(`Last request was not successful.`)
+        return {
+          statusCode: "Error",
+        }
+      } else {
```
psavkar, line 997: remove this else

components/twitter/twitter.app.js
psavkar: don't try catch

PR #20, opened 2020-05-03
Pipedream SQL Source

dylburger:


components/pipedream/sql/sql.js
```diff
@@ -0,0 +1,82 @@
...
+      label: "Emit each record as its own event",
+      description:
+        "If `true`, each record in your results set is emitted as its own event. Defaults to emitting results as a single event (an array of records)",
+      optional: true,
+      default: false,
+    },
+    pipedream,
+  },
+  async run() {
+    const results = await this.pd.runSQLQuery(this.sqlQuery, this.resultType);
```
tjk, line 39: do you need the prop to be defined `pd: pipedream` for this to work?

components/pipedream/sql/sql.js
```diff
@@ -0,0 +1,82 @@
...
+    console.log(results);
```
tjk, line 40: Should we be concerned about result size? Both in terms of emits and also log observations? Does `runSQLQuery` limit? Looking at that code now.

components/pipedream/pipedream.app.js
```diff
@@ -0,0 +1,70 @@
...
+
+        if (format === "csv") {
+          formattedResults = convertArrayToCSV(formattedResults);
+        }
+
+        results.results = formattedResults;
+
+        return results;
+      } catch (err) {
+        console.log(`Error in SQL query: ${err}`);
```
tjk, line 67: Do you not get observability without this? Ideally should not do this try/catch just to re-throw...

components/pipedream/sql/sql.js
dylburger: Talked with G about this and noted the limitations in the README. Thinking about a better way to enforce on the rt / Athena side, too.

PR #26, opened 2020-05-06
Fauna

dylburger:


components/faunadb/faunadb.app.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+    },
+  },
+  methods: {
+    // Fetches events (changes) for all documents in a collection
+    // made after a specific epoch timestamp
+    getClient() {
+      return new faunadb.Client({ secret: this.$auth.secret });
+    },
+    async getCollections() {
```
tjk, line 25: Does this get used by anything? Should it be called `getCollectionIds`? (and the var inside should maybe be renamed too?)

tjk: Oh nvm ^F didn't work... can you return [{label,value}] ?

tjk: (i see it in async options now)

components/faunadb/faunadb.app.js
```diff
@@ -0,0 +1,61 @@
...
+
+module.exports = {
+  type: "app",
+  app: "faunadb",
+  propDefinitions: {
+    collection: {
+      type: "string",
+      label: "Collection",
+      description: "The collection you'd like to watch for changes",
+      optional: false,
```
tjk, line 13: Shouldn't have to specify `optional: false` (that's the default)

components/faunadb/changes-to-collection.js
```diff
@@ -0,0 +1,63 @@
...
+    if (this.emitEventsInBatch) {
+      this.$emit(events, {
+        summary: `${events.length} new event${events.length > 1 ? "s" : ""}`,
+        id: cursor,
+      });
+    } else {
+      for (const event of events) {
+        this.$emit(event, {
+          summary: `${event.action.toUpperCase()} - ${event.instance.id}`,
+          id: `${event.action}-${event.instance.id}`, // dedupes events based on this ID
```
tjk, line 55: Need event.action in here?

components/faunadb/changes-to-collection.js
```diff
@@ -0,0 +1,63 @@
...
+
+module.exports = {
+  name: "changes-to-collection",
+  version: "0.0.1",
+  dedupe: "unique", // Dedupe events based on the concatenation of event + document ref id
+  props: {
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        cron: "*/15 * * * *",
```
tjk, line 11: Is this a better default than using `intervalSeconds`?

components/faunadb/changes-to-collection.js
```diff
@@ -0,0 +1,63 @@
...
+    } else {
+      for (const event of events) {
+        this.$emit(event, {
+          summary: `${event.action.toUpperCase()} - ${event.instance.id}`,
+          id: `${event.action}-${event.instance.id}`, // dedupes events based on this ID
+        });
+      }
+    }
+
+    // Finally, set cursor for the next run
```
tjk, line 60: Shouldn't this be the `ts` of the last event? This is a bit contrived but

ts = A
make request for events since time A
during that time 1 new event (at time A+1)
save A
so then we load that cursor and we'd get A+1 again

components/faunadb/changes-to-collection.js
dylburger: Yeah if you add and delete the same document, instance ID is the same, so I think (action, instance) is the right unique tuple here

components/faunadb/faunadb.app.js
dylburger: done

sergio-eliot-rodriguez:
8 days follow up

sergio-eliot-rodriguez:
8 days follow up

sergio-eliot-rodriguez:
8 days follow up

sergio-eliot-rodriguez:
15 days follow up

sergio-eliot-rodriguez:
15 days follow up

sergio-eliot-rodriguez:
15 days follow up

PR #30, opened 2020-05-07
Airtable update

psavkar:


components/airtable/new-or-modified-records-in-view.js
```diff
@@ -0,0 +1,62 @@
...
+    const { data } = await axios(config)
+
+    if (!data.records.length) {
+      console.log(`No new or modified records.`)
+      return
+    }
+
+    let newRecords = 0, modifiedRecords = 0
+    for (let record of data.records) {
+      if(moment(record.createdTime) > moment(lastTimestamp) || !lastTimestamp) {
```
tjk, line 47: Reverse order of these checks! How does this make sense like this?? Plus the `!lastTimestamp` check is much faster.

components/airtable/new-or-modified-records.js
```diff
@@ -0,0 +1,58 @@
...
+    const { data } = await axios(config)
+
+    if (!data.records.length) {
+      console.log(`No new or modified records.`)
+      return
+    }
+
+    let newRecords = 0, modifiedRecords = 0
+    for (let record of data.records) {
+      if(moment(record.createdTime) > moment(lastTimestamp) || !lastTimestamp) {
```
tjk, line 43: same

PR #32, opened 2020-05-11
Google drive changes to file

dylburger:


components/google-drive/changes-to-file.js
```diff
@@ -0,0 +1,202 @@
...
+// https://developers.google.com/drive/api/v3/push .
+//
+// This source has two interfaces:
+//
+// 1) The HTTP requests tied to changes in the user's Google Drive
+// 2) A timer that runs on regular intervals, renewing the notification channel as needed
+
+const { uuid } = require("uuidv4");
+const includes = require("lodash.includes");
+const googleDrive = require("https://github.com/PipedreamHQ/pipedream/blob/google-drive-changes-to-file/components/google-drive/google-drive.app.js");
```
dylburger, line 12: will update this ref once review is complete

components/google-drive/changes-to-any-file-in-drive.js
```diff
@@ -0,0 +1,203 @@
...
+        "The Google Drive API requires occasionaly renewal of push notification subscriptions. **This runs in the background, so you should not need to modify this schedule**.",
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 60 * 6,
+      },
+    },
+  },
+  hooks: {
+    async activate() {
+      // Called when a componenent is created or updated. Handles all the logic
```
tjk, line 41: componenent -> component

components/google-drive/changes-to-any-file-in-drive.js
```diff
@@ -0,0 +1,203 @@
...
+    if (event.interval_seconds) {
+      if (!subscription || !subscription.resourceId) {
+        return;
+      }
+      console.log(
+        `Checking for resubscription on resource ${subscription.resourceId}`
+      );
+      // If the subscription for this resource will expire before the next run,
+      // stop the existing subscription and renew
+      if (subscription.expiration < +new Date() + event.interval_seconds) {
```
tjk, line 105: `+new Date()` is epoch in milliseconds which you are adding seconds to so this seems wrong. What are the units for subscription.expiration ?

components/google-drive/changes-to-any-file-in-drive.js
```diff
@@ -0,0 +1,203 @@
...
+      );
+      // We use and increment the pageToken as new changes arrive, in run()
+      this.db.set("pageToken", startPageToken);
+
+      // Save metadata on the subscription so we can stop / renew later
+      this.db.set("subscription", { resourceId, expiration });
+      this.db.set("channelID", channelID);
+    },
+    async deactivate() {
+      const channelID = this.db.get("channelID");
```
tjk, line 64: maybe so that db is cleared in all cases?

```
const channelID = ...
const { resourceId } = ...
if (channelID && resourceId) {
  await this.googleDrive...
}
this.db.set("subscription", null);
this.db.set("channelID", null);
// also clear pageToken ?
```

also why the ID vs Id ?
```

File: components/google-drive/changes-to-any-file-in-drive.js [hunk 5]
```diff
@@ -0,0 +1,203 @@
...
+    ) {
+      console.log("Request missing necessary headers: ", headers);
+      return;
+    }
+
+    const incomingChannelID = headers["x-goog-channel-id"];
+    if (incomingChannelID !== channelID) {
+      console.log(
+        `Channel ID of ${incomingChannelID} not equal to deployed component channel of ${channelID}`
+      );
```
tjk, line 148: missing return here

components/google-drive/changes-to-any-file-in-drive.js
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    if (!(headers["x-goog-resource-id"] === subscription.resourceId)) {
```
tjk, line 151: use !== instead ?

components/google-drive/changes-to-file.js
```diff
@@ -0,0 +1,202 @@
...
+
+    if (!includes(this.updateTypes, headers["x-goog-resource-state"])) {
+      console.log(
+        `Update type ${headers["x-goog-resource-state"]} not in list of updates to watch: `,
+        this.updateTypes
+      );
+      return;
+    }
+
+    // We observed false positives where a single change to a document would trigger two changes:
```
tjk, line 168: 💰

components/google-drive/changes-to-file.js
```diff
@@ -0,0 +1,202 @@
...
+      file,
+      change: {
+        state: headers["x-goog-resource-state"],
+        resourceURI: headers["x-goog-resource-uri"],
+        changed: headers["x-goog-changed"], // "Additional details about the changes. Possible values: content, parents, children, permissions"
+      },
+    };
+
+    this.$emit(eventToEmit, {
+      summary: `${headers["x-goog-resource-state"].toUpperCase()} - ${
```
tjk, line 196: leaving in case getFileMetadata can return null/undefined

components/google-drive/google-drive.app.js
```diff
@@ -0,0 +1,159 @@
...
+      return (await drive.changes.getStartPageToken({})).data.startPageToken;
+    },
+    async listFiles(pageToken) {
+      const drive = this.drive();
+      const resp = await drive.files.list({ pageToken });
+      const { files, nextPageToken } = resp.data;
+      const options = files.map((file) => {
+        return { label: file.name, value: file.id };
+      });
+      return {
```
tjk, line 93: nitpicky but the {options,context} is specific to the options function so it feels a bit weird to leak that model to this method, but no biggy (my suggestion would be to just return resp.data from here and do the rest of the logic inside of options

components/google-drive/google-drive.app.js
```diff
@@ -0,0 +1,159 @@
...
+    drive() {
+      const auth = new google.auth.OAuth2();
+      auth.setCredentials({ access_token: this.$auth.oauth_access_token });
+      const drive = google.drive({ version: "v3", auth });
+      return drive;
+    },
+    // Google's push notifications provide a URL to the resource that changed,
+    // which we can use to fetch the file's metadata. So we use axios here
+    // (vs. the Node client) to get that.
+    async getFileMetadata(url) {
```
tjk, line 58: ok this looks like it will either succeed or throw so we gucci (*should* have no null/undefined issue)

components/google-drive/changes-to-any-file-in-drive.js
dylburger: Not sure how you catch this shit

components/google-drive/google-drive.app.js
```diff
@@ -0,0 +1,159 @@
...
+      optional: true,
+      default: false,
+    },
+  },
+  methods: {
+    // Returns a drive object authenticated with the user's access token
+    drive() {
+      const auth = new google.auth.OAuth2();
+      auth.setCredentials({ access_token: this.$auth.oauth_access_token });
+      const drive = google.drive({ version: "v3", auth });
```
tjk, line 52: nitpicky but clean up the unused local? just `return google.drive...`

components/google-drive/changes-to-any-file-in-drive.js
dylburger: `go lint` fucked me: https://github.com/golang/lint/issues/124

dylburger: I use `resourceId` because that's the name of the param Google accepts, but otherwise I use ID when I control it. I should probably change to be consistent

tjk: lol

components/google-drive/changes-to-any-file-in-drive.js
dylburger: Boops, you're right. expiration is ms so will convert `interval_seconds`

components/google-drive/changes-to-any-file-in-drive.js
dylburger: lol yup

PR #39, opened 2020-05-19
Adding configuration data to event

psavkar:


components/airtable/new-or-modified-records-in-view.js
```diff
@@ -51,6 +51,12 @@ module.exports = {
         record.type = "record_modified"
         modifiedRecords++
       }
+
+      record.metadata = {}
```
dylburger, line 55:
```
const { baseId, tableId, viewId } = this
record.metadata = {
  baseId,
  tableId,
  viewId
}
```

Technically this will be the same for all events, so you can also move this code up, defining a `metadata` object once above the `for` loop, and then just set `record.metadata = metadata` here.

components/airtable/new-or-modified-records-in-view.js
```diff
@@ -3,7 +3,7 @@ const moment = require('moment')
 const axios = require('axios')
 
 module.exports = {
-  name: "new-or-modified-records-in-view",
+  name: "New or modified records in view",
   version: "0.0.1",
```
dylburger, line 7: increment this and other versions

components/airtable/new-or-modified-records.js
```diff
@@ -47,6 +47,11 @@ module.exports = {
         record.type = "record_modified"
         modifiedRecords++
       }
+
+      record.metadata = {}
```
dylburger, line 51: same feedback as above re: how this gets defined

components/airtable/new-records-in-view.js
```diff
@@ -31,7 +31,7 @@ module.exports = {
 
     let maxTimestamp
     const lastMaxTimestamp = this.db.get("lastMaxTimestamp")
-    if (!lastMaxTimestamp) {
+    if (lastMaxTimestamp) {
```
dylburger, line 34: was the `!lastMaxTimestamp` a prior bug you fixed here?

psavkar: yes

sergio-eliot-rodriguez:
1 week&nbsp; follow up

PR #40, opened 2020-05-19
Hello! I'm Dmitriy. I finally found the solution to upload files to repos I don’t own. Please check and response to me.  Also I hope release m..

kornienko-elvira:


kornienko-elvira:
I got to know Github better.  Thank you.

dylburger:
Hi Dmitriy, thanks for adding the images. Can you change two things?

- First, can you create a new, `twitter` directory in the `images` directory, and then `git mv` the new images you added into the new `twitter` directory? So the `images` dir will look like:

```text
images/
  twitter/
    1.png
    2.png
    ...
```

That'll just help with the organization of images.

- It looks like the `README` isn't referencing the correct path to the images - see [this link](https://github.com/PipedreamHQ/pipedream/blob/710171e530018d01a6228a8573ac2a15060bd079/components/twitter/ReadMe.md) to see how the `README` looks right now.

I'd recommend changing the path to the images to `/images/twitter/<image>`, which should reference the correct path. For example, I'd change this:

```html
<img width="80" height="80" src="../../images/Generic_Feed-icon.svg.png">
```

to:

```html
<img width="80" height="80" src="/images/twitter/Generic_Feed-icon.svg.png">
```

Let me know if that makes sense.

PR #41, opened 2020-05-19
dropbox apphook sources

danhsiung:
basic dropbox sources

### DIFF
--- a/components/dropbox/dropbox.app.js
+++ b/components/dropbox/dropbox.app.js
@@ -46,15 +46,13 @@ module.exports = {
           }
           // TODO break after a certain number of folders has been found??
 
-          if (has_more = files.has_more) {
+          if (has_more) {
             files = await sdk.filesListFolderContinue({ cursor })
           }
         } while(has_more)
         options = options.sort((a, b) => { return a.toLowerCase().localeCompare(b.toLowerCase()) })
-        if (path && path != "") {
-          let parent = path.slice(0, path.lastIndexOf('/'))
-          if (parent == "") parent = "/"
-          options.unshift(parent)
+        if (path) {
+          options.unshift(require("path").dirname(path))
         }
         options.unshift(path)
       } catch (err) {
--- a/components/dropbox/dropbox.app.js
+++ b/components/dropbox/dropbox.app.js
@@ -1,0 +1,66 @@
+const Dropbox = require('dropbox').Dropbox
+const fetch = require('isomorphic-fetch')
+
+module.exports = {
+  type: "app",
+  app: "dropbox",
+  propDefinitions: {
+    path: {
+      type: "string",
+      label: "Path",
+      description: "Path to the folder you want to watch for changes.",
+      optional: false,
+      useQuery: true, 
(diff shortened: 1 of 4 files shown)


components/dropbox/dropbox.app.js
```diff
@@ -0,0 +1,66 @@
...
+          ({ entries, has_more, cursor } = files)
+
+          for(entry of entries) {
+            if (entry[".tag"] == "folder") {
+              options.push(entry.path_display)
+            }
+          }
+          // TODO break after a certain number of folders has been found??
+
+          if (has_more = files.has_more) {
```
tjk, line 49: What is the point of this if? has_more already destructured above

components/dropbox/dropbox.app.js
```diff
@@ -0,0 +1,66 @@
...
+            files = await sdk.filesListFolderContinue({ cursor })
+          }
+        } while(has_more)
+        options = options.sort((a, b) => { return a.toLowerCase().localeCompare(b.toLowerCase()) })
+        if (path && path != "") {
```
tjk, line 54: there is no point of the second part of this conditional (if path = "" it is not truthy)

components/dropbox/dropbox.app.js
```diff
@@ -0,0 +1,66 @@
...
+          let parent = path.slice(0, path.lastIndexOf('/'))
```
tjk, line 55: maybe use https://nodejs.org/api/path.html#path_path_dirname_path ?

tjk: idk if trailing slash is handled correctly here?

tjk: maybe dropbox never returns that

components/dropbox/dropbox.app.js
```diff
@@ -0,0 +1,66 @@
...
+          if (parent == "") parent = "/"
+          options.unshift(parent)
+        }
+        options.unshift(path)
+      } catch (err) {
+        options.push("/")
```
tjk, line 61: Does this error handling make sense??

components/dropbox/dropbox.app.js
danhsiung: for the while below

danhsiung: good catch

components/dropbox/dropbox.app.js
danhsiung: removing

components/dropbox/dropbox.app.js
danhsiung: done

components/dropbox/dropbox.app.js
danhsiung: was thinking makes sense to default to the user's home dir

danhsiung:
@tjk did some refactoring today. Any thoughts on current state? Also, did you take a look at the firehose component code (god mode as pd@pipedream.com - https://pipedream.com/sources/dc_zqu3kx)?

PR #44, opened 2020-05-20
Hello! I'm Dmitriy. I changed it as you recommend. Please quick check this and reply at fr...

kornienko-elvira:
I want to get response from you at fr.... 
And also I hope you release mi...


dylburger:
Hi @kornienko-elvira , images look great. Did you see my comment on the last PR about the path to the images? I'd love if you could change

```html
<img width="124" height="100" src="../../images/twitter/TwitterLogo.png">
```

to

```html
<img width="124" height="100" src="/images/twitter/TwitterLogo.png">
```

If we move this README anywhere else in the repo, the current, relative references (the `../../`) will break. The absolute reference, starting from `/`, will not.

Everything else looks great, so if you don't mind making that change, I'll merge!

kornienko-elvira:
Thank you.
But I hope get response from you at fr...

sergio-eliot-rodriguez:
1 week&nbsp; follow up

PR #49, opened 2020-06-02
multipart support, deactivate bug, local require

psavkar:


components/jotform/jotform.app.js
```diff
@@ -67,8 +67,13 @@ module.exports = {
     async deleteHook(opts = {}) { 
       const { formId, endpoint } = opts
       const webhooks = (await this.getWebhooks({ formId })).content
-      const webhookIdx = webhooks.findIndex(w => w === ensureTrailingSlash(endpoint))
-      if(webhookIdx !== -1) {
```
casret, line 71: Seems like the findIndex logic was correct, it was just that this predicate was wrong?   `webhookIdx == -1`?

components/jotform/jotform.app.js
```diff
@@ -67,8 +67,13 @@ module.exports = {
     async deleteHook(opts = {}) { 
       const { formId, endpoint } = opts
       const webhooks = (await this.getWebhooks({ formId })).content
```
adolfo-pd, line 69: Um how about...
```
const result = await this.getWebhooks({ formId }) 
webhooks = result && result.content || []
webhooks = webhooks.map(ensureTrailingSlash)
if (!webhooks.find(ensureTrailingSlash(endpoint)))
```
...?

components/jotform/jotform.app.js
psavkar: no -- i was getting `findIndex` is not a function because they return an object, not an array. their api is really bad.

psavkar: image

components/jotform/jotform.app.js
adolfo-pd: Per Slack discussion
```
webhooks = Object.values(result && result.content || {})
```

components/jotform/jotform.app.js
psavkar: no -- the response is an object, not an array. that was a separate bug.

sergio-eliot-rodriguez:
1 week&nbsp; follow up

PR #228, opened 2020-06-25
GitHub updates

psavkar:


components/github/github.app.js
```diff
@@ -16,6 +16,28 @@ module.exports = {
...
+          page: page + 1, // pipedream page 0-indexed, github is 1
+        })
+        return orgs.map(org => org.login)
+      },
+      optional: true,
+    },
+    branches: {
+      type: "string",
+      label: "Branches",
+      async options({ page }) {
```
tjk, line 33: This needs repoFullName to be passed in...

tjk: doesn't make sense without

components/github/github.app.js
```diff
@@ -32,15 +54,126 @@ module.exports = {
...
       })).data
     },
+    async getCommits(opts) {
+      const {
+        repoFullName,
+      } = opts
+      return (await this._makeRequest({
+        path: `/repos/${repoFullName}/commits`,
+        params: {
+          per_page: 100,
```
tjk, line 80: no pagination support here but there is for getBranches... should be more consistent?

components/github/github.app.js
```diff
@@ -32,15 +54,126 @@ module.exports = {
...
+        },
+      })).data
+    },
+    async getRepos(opts = {}) {
+      const {
+        page,
+        org,
+      } = opts
+      if(!org) {
+        return (await this._makeRequest({
```
tjk, line 113: instead of this big else/if, for now can just do `path: org ? `/orgs/${org}/repos` : '/user/repos'`

components/github/github.app.js
```diff
@@ -16,6 +16,28 @@ module.exports = {
         return repos.map(repo => repo.full_name)
       },
```
tjk, line 17: Did you consider replacing a lot of this with some github sdk?

components/github/github.app.js
psavkar: fixed

components/github/github.app.js
psavkar: we can look at migrating to octokit in the future

PR #275, opened 2020-07-27
Slack

danhsiung:
Example Workflow: https://pipedream.com/@dan/slack-to-sms-p_OKClbN/edit

### DIFF
--- a/components/slack/new-message-in-conversation.js
+++ b/components/slack/new-message-in-conversation.js
@@ -13,7 +13,7 @@ module.exports = {
       async options({ prevContext }) {
         let { types, cursor, userNames } = prevContext
         if (types == null) {
-          scopes = await this.slack.scopes()
+          const scopes = await this.slack.scopes()
           types = ["public_channel"]
           if (scopes.includes("groups:read")) {
             types.push("private_channel")
--- a/components/slack/new-message-in-conversation.js
+++ b/components/slack/new-message-in-conversation.js
@@ -52,6 +52,12 @@ module.exports = {
         return this.conversations || []
       },
     },
+    ignoreMyself: {
+      type: "boolean",
+      label: "Ignore myself",
+      description: "Ignore messages from me",
+      default: true,
+    },
     resolveNames: {
       type: "boolean",
       label: "Resolve names",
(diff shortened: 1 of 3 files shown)


components/slack/new-message-in-conversation.js
```diff
@@ -0,0 +1,109 @@
...
+    slack,
+    conversations: {
+      type: "string[]",
+      label: "Conversations",
+      description: "Conversations you'd like to receive notifications for.",
+      optional: true,
+      async options({ prevContext }) {
+        let { types, cursor, userNames } = prevContext
+        if (types == null) {
+          scopes = await this.slack.scopes()
```
tjk, line 16: declare scopes rather than using global?

components/slack/slack.app.js
```diff
@@ -0,0 +1,45 @@
...
+    },
+    sdk() {
+      return new WebClient(this.$auth.oauth_access_token)
+    },
+    async scopes() {
+      const resp = await this.sdk().auth.test()
+      if (resp.ok) {
+        return resp.response_metadata.scopes
+      } else {
+        console.log("Error getting scopes", resp.error)
```
tjk, line 18: why not throw here?

tjk: yea generally either errors should be handled better (which they aren't) or just throw so errors properly propagate

components/slack/new-message-in-conversation.js
```diff
@@ -0,0 +1,109 @@
...
+      // Due to inconsistencies with the shape of message_changed and message_deleted
+      // events, we are ignoring them for now. If you want to handle these types of
+      // events, feel free to change this code!!
+      console.log("Ignoring message with subtype.")
+      return
+    }
+    if (this.ignoreMyself && event.user == this.slack.mySlackId()) {
+      return
+    }
+    if (this.resolveNames) {
```
tjk, line 80: would it make sense to use `db` to cache resolved names? it's bit more code but possibly much, much less wasteful

tjk: although if cache could be out of date, might be a class of issues we'd rather avoid... for high volume slacks though would rather at least limit how often we end up resolving the same thing

components/slack/new-message-in-conversation.js
danhsiung: fixed

components/slack/slack.app.js
danhsiung: fixed by throwing resp.error

components/slack/new-message-in-conversation.js
danhsiung: do you see this as a blocker? what would you propose? refresh once a day?

PR #329, opened 2020-08-18
YouTube

michelle0927:


components/youtube/new-videos.js
```diff
@@ -0,0 +1,65 @@
+const axios = require('axios')
+const youtube = {
```
dylburger, line 2: I'd recommend separating this into its own app file so that future sources can refer to that. Then you'll want to require that app file like we do in other sources.

components/youtube/new-videos.js
```diff
@@ -0,0 +1,65 @@
...
+    db: "$.service.db",
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+  methods: {
+    async getVideos(pageToken=null) {
```
dylburger, line 23: When you separate the app file, I'd recommend moving this method into the app file so that other sources can benefit from it.

Then you can call the method in this file using `this.youtube.getVideos()` ([see Github example here](https://github.com/PipedreamHQ/pipedream/blob/master/components/github/new-commit.js#L29))

components/youtube/new-videos.js
```diff
@@ -0,0 +1,65 @@
...
+      return await axios.get('https://www.googleapis.com/youtube/v3/search', {
+        headers: {
+          'Authorization' : 'Bearer '+this.youtube.$auth.oauth_access_token
```
dylburger, line 26: Prefer template literals over concatenation:

```
`Bearer ${this.youtube.$auth.oauth_access_token}`
```

(when you move this into the app file, the reference to the access token will also change to just `this.$auth.oauth_access_token`)

components/youtube/new-videos.js
```diff
@@ -0,0 +1,65 @@
...
+        },
+        params: {
+          'part' : 'snippet',
+          'type' : 'video',
+          'forMine' : true,
+          'pageToken' : pageToken
+        }
```
dylburger, line 33: You can simplify this a bit by removing quotes around property names and using object shorthand notation (since `pageToken` holds the same property name and variable reference) to get:

```
params: {
  part: 'snippet',
  type: 'video',
  forMine: true,
  pageToken,
}
```

components/youtube/new-videos.js
```diff
@@ -0,0 +1,65 @@
...
+      });
+    },
+  },  
+
+  async run(event) {
+  	var videos = [];
```
dylburger, line 39: Do you happen to use ESLint / Prettier on your local machine? We're planning to add our own linting / formatting rules to the repo, but if you setup those tools with your text editor, it should automatically format the file on save. That way you'll never get inconsistent spacing like on this line.

dylburger: Moreover, we generally prefer `let` instead of `var` for variables whose value will change in the same block. I'd replace each `var` with `let` below.

components/youtube/new-videos.js
```diff
@@ -0,0 +1,65 @@
...
+      results = await this.getVideos(nextPageToken);
+      totalResults = results.data.pageInfo.totalResults;
+      nextPageToken = results.data.nextPageToken;
+      results.data.items.forEach(function (video) {
+        videos.push(video);
+        count++;
+      });
+    }
+
+    let t = this;
```
dylburger, line 55: You can avoid the `this` scoping problem in a couple of ways:

* Use [arrow functions](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Arrow_functions) - see the "No separate this" section.
* I'd recommend using a `for ... of` loop instead of `forEach`, which I find simpler to read:

```
for (const video of videos) {
  thia.$emit(video, {
    id: video.id.videoId,
    summary: video.snippet.title,
    ts: Date.now()
  })
}); 
```

components/youtube/new-videos.js
michelle0927: Updated this.

components/youtube/new-videos.js
michelle0927: Updated this.

components/youtube/new-videos.js
michelle0927: Updated var to let.

components/youtube/new-videos.js
michelle0927: Created separate youtube.app.js file.

components/youtube/new-videos.js
michelle0927: Created the app file with the getVideos method in it and am calling it from the component.

components/youtube/new-videos.js
michelle0927: Updated this.

components/youtube/new-videos-in-playlist.js
```diff
@@ -0,0 +1,57 @@
...
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+  methods: {},
+
+  async run(event) {
+    let videos = [];
```
dylburger, line 27: Another thing I wanted to ask: are you planning to keep track of the last page token / video ID you pulled the last time the source ran? That is: when the source runs the first time it pulls the last N videos. Since you've already processed those N videos, the next time you can just pull videos starting at N + 1 moving forward.

You may have noticed that, by deduping on video ID in the call to `this.$emit`, if you pull duplicates, we won't emit them. But I was just curious if that was intentional, or if you planned to handle this in the API request to Youtube.

michelle0927: The only issue there is that if someone deletes a video, then the page count would be off. There is a publishedAfter parameter I could include though to cut down on requests.

dylburger: Great, yeah have you seen our built-in `db` ([docs here](https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-API.md#db))? I think you've used `$checkpoint` in workflows in the past - this is a similar API where you could store the video of the last ID you fetched in the prior run, etc.

components/youtube/new-videos-by-username.js
```diff
@@ -0,0 +1,84 @@
+const axios = require("axios");
+const youtube = require("https://github.com/PipedreamHQ/pipedream/components/youtube/youtube.app.js");
+
+module.exports = {
```
dylburger, line 4: Would you mind fixing the indentation on this file? Note that the other files use an indentation of 2 spaces, this one indents with 4 spaces.

We're planning to add global ESLint / Prettier rules to this repo so that your editor should inherit those. That way future updates should be easier.

components/youtube/new-videos.js
```diff
@@ -0,0 +1,52 @@
...
+      youtube,
+      db: "$.service.db",
+      timer: {
+         type: "$.interface.timer",
+         default: {
+            intervalSeconds: 60 * 15,
+         },
+      },
+   },
+   methods: {},
```
dylburger, line 18: You can feel free to remove this line when you don't have any methods defined.

components/youtube/new-videos-by-location.js
```diff
@@ -0,0 +1,75 @@
...
+  name: "Youtube - New Videos by Location",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    youtube,
+    location: {
+      type: "string",
+      label: "Location",
+      description: "The location parameter, in conjunction with the locationRadius parameter, defines a circular geographic area and also restricts a search to videos that specify, in their metadata, a geographic location that falls within that area. The parameter value is a string that specifies latitude/longitude coordinates e.g. (37.42307,-122.08427).",
+      default: "",
```
dylburger, line 14: You can remove this line if you'd like. Params default to `required: true`, and they'll hold a blank value by default, as well, so you can just leave out the empty default and it'll mean the same thing.

components/youtube/youtube.app.js
```diff
@@ -0,0 +1,28 @@
+module.exports = {
```
dylburger, line 1: Two things on this file:

* You'll want to require `axios` in this file
* Can you also adjust the indentation of this file?

components/youtube/new-videos.js
```diff
@@ -0,0 +1,52 @@
+const axios = require("axios");
```
dylburger, line 1: You can remove all instances of these requires, since you don't appear to use `axios` in these files. Instead you can just require it once in the `youtube` app file.

components/youtube/new-videos-by-location.js
```diff
@@ -0,0 +1,75 @@
+const axios = require('axios')
+const youtube = require("https://github.com/PipedreamHQ/pipedream/components/youtube/youtube.app.js");
+
+module.exports = {
+  name: "Youtube - New Videos by Location",
```
dylburger, line 5: We're also adding this to all of our components, but can you add a `description` to these, as well? The format we like to use is something like:

"Emits an event for each new [X]", so in this case it can be something like "Emits an event for each new Youtube video tied to a location", then you can adjust for each source below.

PR #371, opened 2020-08-27
Spotify

michelle0927:


components/spotify/new-playlists.js
```diff
@@ -0,0 +1,50 @@
+const spotify = require("https://github.com/PipedreamHQ/pipedream/components/spotify/spotify.app.js");
+
+module.exports = {
+  name: "Spotify - New Playlists",
```
dylburger, line 4: I forgot to mention this in the other PR, but you can leave out the name of the app in the name of the source, so this can just be "New Playlists".

(I updated the YouTube sources yesterday)

components/spotify/new-playlists.js
```diff
@@ -0,0 +1,50 @@
...
+    let limit = 20;
+    let offset = 0;
+
+    let params = {
+      limit,
+      offset,
+    };
+
+    while (count < total) {
+      results = await this.spotify.getPlaylists(params);
```
dylburger, line 33: Does Spotify allow you to pass a timestamp to the get playlists API endpoint so that, like YouTube, you can filter on playlists created since the last time the source runs?

Again this isn't a huge deal since the deduper will handle duplicate playlists, but just wanted to see.

components/spotify/new-tracks-in-playlist.js
```diff
@@ -0,0 +1,67 @@
...
+      params.offset += limit;
+    }
+
+    this.db.set("lastEvent", now);
+
+    for (const track of tracks) {
+      this.$emit(track, {
+        id: track.track.id,
+        summary: track.track.name,
+        ts: Date.now(),
```
dylburger, line 63: Maybe pass the track's `addedAt` property here?

components/spotify/new-playlists.js
michelle0927: No, the only parameters you can pass to it are limit and offset.

components/spotify/new-playlists.js
michelle0927: Thanks. I've updated the names.

components/spotify/new-tracks-in-playlist.js
michelle0927: Updated this in new-saved-tracks.js and new-tracks-in-playlist.js

PR #375, opened 2020-08-28
Add event source for GCP Pub/Sub

jverce:
* Construct GCP parameters object to be used across all GCP
  integrations.
* Create Pub/Sub integration skeleton with `activate` and `deactivate`
  hooks.
* Create utility method to check/adjust the Pub/Sub topic name
  according to GCP requirements.

### DIFF
--- a/components/gcp/gcp.app.js
+++ b/components/gcp/gcp.app.js
@@ -1,0 +1,17 @@
+module.exports = {
+  type: "app",
+  app: "gcp",
+  methods: {
+    sdkParams() {
+      const { projectId, clientEmail, privateKey } = this.$auth;
+      const credentials = {
+        client_email: clientEmail,
+        private_key: privateKey,
+      };
+      return {
+        credentials,
+        projectId,
+      };
+    },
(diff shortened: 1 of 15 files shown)


components/gcp/gcp.app.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  type: "app",
+  app: "gcp",
```
dylburger, line 3: This isn't well-documented - we're planning to add each of the "app name slugs" to this repo so it's more clear - but the slug here should be `google_cloud` to reference the Google Cloud app.

jverce: Done. I also renamed the containing directory to `google_cloud` as well.

jverce:
Manually tested:
* Activation and deactivation of sources
  * Topics were created and destructed
  * Subscriptions were created and destructed
* Sent test events
  * JSON-formatted messages were properly parsed and emitted
  * Non-JSON messages were passed as `{ "rawMessage": "..." }`

PR #387, opened 2020-09-01
Trello

michelle0927:


components/trello/card-updates.js
```diff
@@ -0,0 +1,53 @@
+const spotify = require("https://github.com/PipedreamHQ/pipedream/components/trello/trello.app.js");
+
+module.exports = {
+  name: "Card Updates",
+  description: "Emits an event for each update to a Trello card.",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    trello,
+    cardId: {
```
dylburger, line 10: We should use async options to list the cards available to the auth user

dylburger: I think it would also make sense for this prop to be an array of strings, i.e.

```
type: "string[]`
```

That way a user can watch multiple cards for updates without having to create a source per card.

components/trello/cards-due.js
```diff
@@ -0,0 +1,67 @@
...
+  description:
+    "Emits an event at a specified time before a card is due.",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    trello,
+    boardId: { propDefinition: [trello, "boardId"] },
+    timeBefore: {
+      type: "integer",
+      label: "Time Before",
```
dylburger, line 14: Consider changing this name to something like "Time Before Due Time", which is a mouthful but slightly more clear.

components/trello/cards-due.js
```diff
@@ -0,0 +1,67 @@
...
+      description: "How far before the due time the event should trigger.",
+      default: 5,
+    },
+    timeBeforeUnit: {
+      type: "string",
+      label: "Time Before (Unit)",
+      description: "Unit of time for Time Before.",
```
dylburger, line 21: Since you set a default above does it make sense to have a default here, as well?

components/trello/card-updates.js
```diff
@@ -0,0 +1,73 @@
...
+    },
+  },
+
+  async run(event) {
+    this.http.respond({
+      status: 200,
+    });
+
+    const body = _.get(event, "body");
+    if (body) {
```
dylburger, line 47: I'd use the pattern:

```
if (!body) {
  return
}

// rest of your code here
```

that way you can return early and keep the rest of the code at the top-level of indentation, instead of nesting under `if (body)`.

components/trello/card-updates.js
```diff
@@ -0,0 +1,73 @@
...
+      const eventType = _.get(body, "action.type");
+      const cardId = _.get(body, "action.data.card.id");
+      const boardId = this.db.get("boardId");
+      const cardIds = this.db.get("cardIds");
+      let emitEvent = false;
+      let card;
+
+      if (eventType && eventType == "updateCard") {
```
dylburger, line 55: Does Trello have any mechanism for creating a webhook that will only notify us of specific events? i.e. can we set the webhook to only send `updateCard` events instead of receiving all events?

I think it makes sense to do the filter here either way but just curious if we can reduce the number of incoming HTTP requests we get.

components/trello/card-updates.js
```diff
@@ -0,0 +1,73 @@
...
+        card = await this.trello.getCard(cardId);
+        if (!boardId) emitEvent = true;
```
dylburger, line 57: I'd recommend also just returning early here in the cases where you don't want to emit. Then by the time you get to the emit, you'll know you're supposed to emit (otherwise you would have returned), so you can avoid the conditional check.

components/trello/cards-due.js
```diff
@@ -0,0 +1,67 @@
...
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
+    let cards = [];
+    let results = [];
+    let due, notifyAt, n;
+    const modelId = this.boardId ? this.boardId : "me";
```
dylburger, line 44: You use a different pattern in `card-updates.js` to get the model id using `await this.trello.getMember("me");` in the absence of a board Id - do you need that logic here, too?

components/trello/cards-due.js
```diff
@@ -0,0 +1,67 @@
...
+    const now = new Date();
+
+    results = await this.trello.getCards(modelId);
+    for (var i=0; i<results.length; i++) {
```
dylburger, line 48: Generally I prefer `for... of` loops so you can avoid the index, e.g.

```
for (const result of results) {
  if (result.due) {
    ...
```

components/trello/cards-due.js
```diff
@@ -0,0 +1,67 @@
...
+          cards.push(results[i]);
+        }
+      }
+    }
+
+    for (const card of cards) {
+      this.$emit(card, {
+        id: card.id,
+        summary: card.name,
+        ts: now,
```
dylburger, line 63: `now` is defined as a Date but needs to be an epoch timestamp here (you correctly pass epochs as the `ts` in other components)

components/trello/new-boards.js
```diff
@@ -0,0 +1,58 @@
...
+  name: "New Boards",
+  description: "Emits an event for each new board added.",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    trello,
+    db: "$.service.db",
+    http: "$.interface.http",
+  },
+
```
dylburger, line 14: remove extraneous newline

components/trello/new-boards.js
```diff
@@ -0,0 +1,58 @@
...
+    },
+  },
+
+  async run(event) {
+    this.http.respond({
+      status: 200,
+    });
+
+    const body = _.get(event, "body");
+    if (body) {
```
dylburger, line 38: Same feedback as above (just return early)

components/trello/new-boards.js
```diff
@@ -0,0 +1,58 @@
...
+      const eventType = _.get(body, "action.type");
+      const boardId = _.get(body, "action.data.board.id");
+      let emitEvent = false;
+      let board;
+
+      if (eventType && eventType == "createBoard") {
```
dylburger, line 44: Same feedback as above re: returning early vs. setting `emitEvent`.

components/trello/new-labels.js
```diff
@@ -0,0 +1,67 @@
...
+
+      if (eventType && eventType == "createLabel") {
+        const labelId = _.get(body, "action.data.label.id");
+        label = await this.trello.getLabel(labelId);
+        emitEvent = true;
+      }
+
+      if (emitEvent) {
+        let summary = label.color;
+        summary += label.name ? ` - ${label.name}` : "";
```
dylburger, line 58: Any reason you didn't combine this with the initialization of `summary` in the line above?

components/trello/card-updates.js
```diff
@@ -0,0 +1,73 @@
+const trello = require("https://github.com/PipedreamHQ/pipedream/components/trello/trello.app.js");
+const _ = require("lodash");
```
dylburger, line 2: You can save on deploy time by just [requiring `lodash.get` directly](https://www.npmjs.com/package/lodash.get):

```
const get = require('lodash.get');

```

components/trello/new-notifications.js
```diff
@@ -0,0 +1,46 @@
...
+    let results = [];
+
+    let since = this.db.get("lastNotificationId") || null;
+
+    let params = {
+      since,
+    };
+
+    results = await this.trello.getNotifications("me", params);
+    results.forEach(function (notification) {
```
dylburger, line 31: Wouldn't results = notifications in this case since you're just adding the same elements to a new array?

components/trello/new-notifications.js
```diff
@@ -0,0 +1,46 @@
...
+    });
+
+    if (results.length > 0)
+      this.db.set("lastNotificationId", results[results.length - 1].id);
+
+    for (const notification of notifications) {
+      this.$emit(notification, {
+        id: notification.id,
+        summary: `${notification.type} : ${notification.data.card.name}`,
+        ts: notification.date,
```
dylburger, line 42: Is this an epoch timestamp?

components/trello/trello.app.js
```diff
@@ -0,0 +1,240 @@
+const axios = require("axios");
+const events = `addAttachmentToCard
+addChecklistToCard
```
dylburger, line 3: Since we're displaying this to users, it might be nice to present human-readable options to the user:

```
[
  { label: "Add Checklist To Card", value: `addChecklistToCard` }
  ...
]
```

components/trello/trello.app.js
```diff
@@ -0,0 +1,240 @@
...
+      // options needs to support standardized opts for pagination
+      async options(opts) {
+        const boards = await this.getBoards(this.$auth.oauth_uid);
+        // XXX short hand where value and label are same value
+        const activeBoards = boards.filter((board) => board.closed === false);
+        return activeBoards.map((board) => {
+          return { label: `${board.name} (${board.id})`, value: board.id };
+        });
+      },
+      // XXX validate
```
dylburger, line 68: Did you still need to do some validation?

components/trello/webhook.js
```diff
@@ -0,0 +1,85 @@
+const trello = require("https://github.com/PipedreamHQ/pipedream/components/trello/trello.app.js");
```
dylburger, line 1: I think we can remove this from the PR - this appears to be an example that the original author was working through.

components/trello/card-updates.js
```diff
@@ -0,0 +1,73 @@
...
+    async deactivate() {
+      console.log(this.db.get("hookId"));
+      await this.trello.deleteHook({
+        hookId: this.db.get("hookId"),
+      });
+    },
+  },
+
+  async run(event) {
+    this.http.respond({
```
dylburger, line 42: We should [validate the signature](https://developer.atlassian.com/cloud/trello/guides/rest-api/webhooks/#webhook-signatures) of the incoming event. You can add this method to the trello app file and re-use it here.

components/trello/card-updates.js
michelle0927: It allows you to filter by idModel (id of a board, card, list), but not by the event type that I can tell. https://developer.atlassian.com/cloud/trello/rest/api-group-webhooks/#api-webhooks-post

components/trello/cards-due.js
michelle0927: In this file, modelId is being used to poll for the cards on a board, but you can use "me" as the id to get all of the user's boards. I probably should have called this boardId.
In card-updates.js, the modelId is being used to create a webhook which doesn't allow using "me" as the id. So if there's no boardId, it calls getMember("me") to retrieve the user's member id.

components/trello/trello.app.js
michelle0927: Not sure. That was written by @psavkar

components/trello/webhook.js
michelle0927: This actually works for retrieving new activity, which is one of the sources we want. I updated it some and changed the name to new-activity.js

components/trello/card-updates.js
```diff
@@ -0,0 +1,84 @@
...
+      });
+    },
+  },
+
+  async run(event) {
+    // validate signature
+    if (
+      !this.trello.verifyTrelloWebhookRequest(
+        event,
+        this.trello.$auth.oauth_refresh_token,
```
dylburger, line 46: technically since you have access to `this.$auth` within the app file's methods, you don't need this as an argument - you can just reference `this.$auth.oauth_refresh_token` within the `verifyTrelloWebhookRequest` method.

components/trello/card-updates.js
```diff
@@ -0,0 +1,84 @@
...
+    if (eventType !== "updateCard") {
+      return;
+    }
+
+    const cardId = get(body, "action.data.card.id");
+    const boardId = this.db.get("boardId");
+    const cardIds = this.db.get("cardIds");
+    const card = await this.trello.getCard(cardId);
+
+    if (boardId) {
```
dylburger, line 71: You can combine to `if (boardId && boardId !== card.idBoard)`

components/trello/card-updates.js
```diff
@@ -0,0 +1,84 @@
...
+      if (boardId !== card.idBoard) return;
+    }
+    if (cardIds) {
```
dylburger, line 74: Similar to above, you should be able to combine these two conditional statements.

PR #409, opened 2020-09-10
Gitlab base

jverce:
* Add utility methods to the Gitlab app file to create and delete
  webhooks in projects/repositories.
* Add other utility methods to make code clearer.


components/gitlab/gitlab.app.js
```diff
@@ -0,0 +1,113 @@
...
+  type: "app",
+  app: "gitlab",
+  propDefinitions: {
+    projectId: {
+      type: "string",
+      label: "Project ID",
+      description: "The project ID, as displayed in the main project page",
+      async options({ page, prevContext }) {
+        let url;
+        let requestParams = this._makeRequestParams();  // Basic request parameters
```
dylburger, line 15: Consider renaming this to `requestConfig` (and maybe the method to `_makeRequestConfig`) or something more generic. Initially I thought it was the actual `params` expected by `axios`.

components/gitlab/gitlab.app.js
```diff
@@ -0,0 +1,113 @@
...
+    _userProjectsEndpoint() {
+      const baseUrl = this._apiUrl();
+      const userId = this._gitlabUserId();
+      return `${baseUrl}/users/${userId}/projects`;
+    },
+    _hooksEndpointUrl(projectId) {
+      const baseUrl = this._apiUrl();
+      return `${baseUrl}/projects/${projectId}/hooks`;
+    },
+    _gitlabAuthToken() {
```
dylburger, line 65: I like the wrapping of `this.$auth` with methods like this.

components/gitlab/gitlab.app.js
jverce: Yeah, I agree, I had the same issue when telling them apart with the query string params. I see that [axios refers to them as `config`](https://github.com/axios/axios#request-method-aliases) so I'll stick to that.

dylburger (approved):

Issue #376, opened 2020-08-28
[SOURCE] Gitlab


puzanov:
**Describe the event source. What app, what event?**

I'd like to receive events from Gitlab.com projects, as like we have now in Github

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**

* https://docs.gitlab.com/ee/api/README.html
* https://docs.gitlab.com/ee/user/project/integrations/webhooks.html

dylburger:
@puzanov thanks for the request! 

Would you be open to developing this source yourself? It may not have been obvious, but all event sources are just Node.js modules that implement [a specific API](https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-API.md). We developed a [Quickstart](https://github.com/PipedreamHQ/pipedream/blob/master/QUICKSTART.md) that shows you how to develop a basic event source, and you can take a look at all existing sources in the [`components` directory](https://github.com/PipedreamHQ/pipedream/tree/master/components) of this repo as examples. No worries if you don't have the time - we are happy to develop it ourselves - but I wanted to mention this as an option!

Since the Gitlab source would ingest data delivered via webhooks, I'd recommend taking a look at two other, similar examples:

* [Github](https://github.com/PipedreamHQ/pipedream/blob/master/components/github/custom-events.js)
* [Stripe](https://github.com/PipedreamHQ/pipedream/blob/master/components/stripe/custom-events.js) 

Both of these sources work by using your linked account to make an API request to create a webhook in the source service when the source is first created (see the code in the `activate` hook, [here](https://github.com/PipedreamHQ/pipedream/blob/master/components/github/custom-events.js#L19) and [here](https://github.com/PipedreamHQ/pipedream/blob/master/components/stripe/custom-events.js#L21), for example).
When you create the webhook via API, you'll want to pass `this.http.endpoint` as the value of the target webhook URL. This should configure the webhook in Gitlab to send webhook requests to the current component.

An event source's `run` method is triggered by incoming webhook requests. This is where you'll want to handle incoming events. Typically, the `run` method just:

* Validates the incoming event, and
* Emits the event data you'd like to send to listening workflows.

Again, the `run` methods in the [Github](https://github.com/PipedreamHQ/pipedream/blob/master/components/github/custom-events.js#L37) and [Stripe](https://github.com/PipedreamHQ/pipedream/blob/master/components/stripe/custom-events.js#L37) sources are good examples to start from.

[...]

PR #426, opened 2020-09-12
Add event source for new branch events from Gitlab

jverce:
Gitlab currently does not issue specific events for branch creations,
but such events can be deduced from generic "push" events.

This change creates webhooks in Gitlab for "push events", and
implements an internal filter that will make an event be
emitted down to other steps only if such event correspond to a branch
being created.

Related to issue #376

### DIFF
--- a/components/gitlab/new-branch.js
+++ b/components/gitlab/new-branch.js
@@ -1,9 +1,27 @@
 const gitlab = require("https://github.com/PipedreamHQ/pipedream/components/gitlab/gitlab.app.js");
 
+function generateMeta(data) {
+  const id = data.checkout_sha;
+
+  const newBranchName = data.ref;
+  const userFullName = data.user_name;
+  const userLogin = data.user_username;
+  const summary = `New Branch: ${newBranchName} by ${userFullName} (${userLogin})`;
+
+  const ts = +new Date();
+
+  return {
+    id,
+    summary,
+    ts,
+  };
+}
+
 module.exports = {
   name: "New Branch (Instant)",
   description: "Emits an event when a new branch is created",
   version: "0.0.1",
+  dedupe: "unique",
   props: {
     gitlab,
     projectId: { propDefinition: [gitlab, "projectId"] },
@@ -20,9 +38,11 @@ module.exports = {
         hookParams,
         projectId: this.projectId,
       };
-      const response = await this.gitlab.createHook(opts);
-
-      const { hookId, token } = response;
+      const { hookId, token } = await this.gitlab.createHook(opts);
+      console.log(
+        `Created "push events" webhook for project ID ${this.projectId}.
+        (Hook ID: ${hookId}, endpoint: ${hookParams.url})`
+      );
       this.db.set("hookId", hookId);
       this.db.set("token", token);
     },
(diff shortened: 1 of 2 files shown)


components/gitlab/new-branch.js
```diff
@@ -0,0 +1,68 @@
...
+    async activate() {
+      const hookParams = {
+        push_events: true,
+        url: this.http.endpoint,
+      };
+      const opts = {
+        hookParams,
+        projectId: this.projectId,
+      };
+      const response = await this.gitlab.createHook(opts);
```
dylburger, line 23: Just to shorten a bit, maybe combine this and the next line to

```javascript
const { hookId, token } = await this.gitlab.createHook(opts);
```

components/gitlab/new-branch.js
```diff
@@ -0,0 +1,68 @@
...
+
+      const { hookId, token } = response;
+      this.db.set("hookId", hookId);
```
dylburger, line 26: I also like to add a little user-facing logging once the webhook is created, just to show them what's going on behind the scenes. You might consider adding something like:

```javascript
console.log(`Webhook created for project ${this.projectId}: ${hookId}`)
```

components/gitlab/new-branch.js
```diff
@@ -0,0 +1,68 @@
...
+      this.db.set("token", token);
+    },
+    async deactivate() {
+      const hookId = this.db.get("hookId");
+      const opts = {
+        hookId,
+        projectId: this.projectId,
+      };
+      this.gitlab.deleteHook(opts);
```
dylburger, line 35: Do you need to `await` this?

components/gitlab/new-branch.js
```diff
@@ -0,0 +1,68 @@
...
+      return before === expectedBeforeValue;
+    },
+  },
+  async run(event) {
+    const { headers, body } = event;
+
+    // Reject any calls not made by the proper Gitlab webhook.
+    if (!this.gitlab.isValidSource(headers, this.db)) {
+      this.http.respond({
+        status: 404,
```
dylburger, line 52: Does Gitlab expect a 404 here?

components/gitlab/new-branch.js
```diff
@@ -0,0 +1,68 @@
...
+
+    // Acknowledge the event back to Gitlab.
+    this.http.respond({
+      status: 200,
+    });
+
+    // Gitlab doesn't offer a specific hook for "new branch" events,
+    // but such event can be deduced from the payload of "push" events.
+    if (this.isNewBranch(body)) {
+      this.$emit(body);
```
dylburger, line 65: Can you also set a `summary`, `id` and `ts` for these emits? https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-API.md#emit

In addition to the `id`, you'll want to add a [`dedupe` strategy](https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-API.md#dedupe-strategies).

dylburger (changes requested):
Looks great in general, added a few comments

components/gitlab/new-branch.js
jverce: Short answer: yes, I think I switched branches and the old `await` got lost.

components/gitlab/new-branch.js
jverce: No, but Gitlab won't get that error. This is more of a security mechanism that Gitlab offers to ensure that we only process legitimate Gitlab webhook calls: https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#secret-token

dylburger: Makes sense, I think a 404 is fine for illegitimate clients.

components/gitlab/new-branch.js
jverce: Done

PR #431, opened 2020-09-13
Airtable Source With Row Deletion Trigger

siliconchild:
The default trigger only works with new and modified rows, when a row is deleted on Airtable no events were triggered. So I wrote this extra event trigger that works with row deletion as well. This is my first time contributing to open-source and if this code somehow doesn't meet he contribution guidelines please guide on how to do this properly. Thank You.

### DIFF
--- a/components/airtable/new-modified-or-deleted-records.js
+++ b/components/airtable/new-modified-or-deleted-records.js
@@ -55,36 +55,7 @@ module.exports = {
       modifiedRecordsCount = 0,
       deletedRecordsCount = 0;
 
-    if (!data.records.length) {
-      delete config.params.filterByFormula;
-
-      while (allRecordIds.length === 0 || config.params.offset) {
-        const { data } = await axiosRateLimiter(config);
-        if (!data.records.length || data.records.length === 0) return;
-        allRecordIds = [...allRecordIds, ...data.records.map(record => record.id)];
-        if (data.offset) {
-          config.params.offset = data.offset;
-        } else {
-          delete config.params.offset;
-        }
-      }
-
-      if (prevAllRecordIds && prevAllRecordIds.length > allRecordIds.length) {
-        let deletedRecordIds = prevAllRecordIds.filter(prevRecord => !allRecordIds.includes(prevRecord));
-        for (let recordID of deletedRecordIds) {
-          deletedRecordsCount++;
-          deletedRecordObj = {
-            metadata,
-            type: 'record_deleted',
-            id: recordID,
-          };
-          this.$emit(deletedRecordObj, {
-            summary: 'record_deleted',
-            id: recordID,
-          });
-        }
-      }
-    } else {
+    if (data.records) {
       for (let record of data.records) {
         if (!lastTimestamp || moment(record.createdTime) > moment(lastTimestamp)) {
           record.type = 'new_record';
@@ -103,6 +74,40 @@ module.exports = {
       }
     }
 
+    delete config.params.filterByFormula;
+
+    while (allRecordIds.length === 0 || config.params.offset) {
+      const { data } = await axiosRateLimiter(config);
+      if (!data.records.length || data.records.length === 0) return;
+      allRecordIds = [
+        ...allRecordIds,
+        ...data.records.map(record => record.id)
+      ];
+      if (data.offset) {
+        config.params.offset = data.offset;
+      } else {
+        delete config.params.offset;
+      }
+    }
+
+    if (prevAllRecordIds) {
+      let deletedRecordIds = prevAllRecordIds.filter(
+        prevRecord => !allRecordIds.includes(prevRecord)
+      );
+      for (let recordID of deletedRecordIds) {
+        deletedRecordsCount++;
+        deletedRecordObj = {
+          metadata,
+          type: "record_deleted",
+          id: recordID
+        };
+        this.$emit(deletedRecordObj, {
+          summary: "record_deleted",
+          id: recordID
+        });
+      }
+    }
+
     console.log(
       `Emitted ${newRecordsCount} new records(s) and ${modifiedRecordsCount} modified record(s) and ${deletedRecordsCount} deleted records.`
     );
@@ -110,4 +115,3 @@ module.exports = {
     this.db.set('lastTimestamp', timestamp);
   },
 };
-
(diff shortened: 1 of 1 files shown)


components/airtable/new-modified-or-deleted-records.js
```diff
@@ -0,0 +1,111 @@
...
+const axios = require('axios');
+const Bottleneck = require('bottleneck');
+
+const limiter = new Bottleneck({
+  minTime: 200, // 5 requets per second
+});
+const axiosRateLimiter = limiter.wrap(axios);
+
+module.exports = {
+  name: 'New, Modified or Deleted Records',
```
dylburger, line 12: We like to add a `description` to event sources, as well (we need to add these to the other Airtable event sources!), something like "Emits an event each time a record is added, updated, or deleted in an Airtable table"

components/airtable/new-modified-or-deleted-records.js
```diff
@@ -0,0 +1,111 @@
...
+  version: '0.0.2',
```
dylburger, line 13: Maybe set to `0.0.1` for the initial release?

components/airtable/new-modified-or-deleted-records.js
```diff
@@ -0,0 +1,111 @@
...
+    const config = {
+      url: `https://api.airtable.com/v0/${encodeURIComponent(this.baseId)}/${encodeURIComponent(this.tableId)}`,
+      params: {},
+      headers: {
+        Authorization: `Bearer ${this.airtable.$auth.api_key}`,
+      },
+    };
+
+    const lastTimestamp = this.db.get('lastTimestamp');
+    const prevAllRecordIds = this.db.get('prevAllRecordIds');
```
dylburger, line 43: We currently don't impose limits on the size of `this.db`, but we may in the future. I could see this list growing large for some tables. Chatting with the team about this.

components/airtable/new-modified-or-deleted-records.js
```diff
@@ -0,0 +1,111 @@
+const airtable = require('https://github.com/PipedreamHQ/pipedream/components/airtable/airtable.app.js');
+const moment = require('moment');
+const axios = require('axios');
+const Bottleneck = require('bottleneck');
+
+const limiter = new Bottleneck({
```
dylburger, line 6: Great use of rate-limiting functionality here

dylburger:
@siliconchild this is great, I really appreciate you spending so much time on this PR. Your approach generally looks good, I added a couple of small comments.

I'm chatting with the team about the `this.db` note re: size, and sharing the PR with some of our Airtable experts for review, as well. We'll get back to you with other comments ASAP and get this merged!

components/airtable/new-modified-or-deleted-records.js
```diff
@@ -0,0 +1,111 @@
...
+            type: 'record_deleted',
+            id: recordID,
+          };
+          this.$emit(deletedRecordObj, {
+            summary: 'record_deleted',
+            id: recordID,
+          });
+        }
+      }
+    } else {
```
dylburger, line 85: Even in this case, you could have records that were deleted correct? If I add one record and delete another record in the same 5-minute period, won't I hit this `else` block and only process the added record?

components/airtable/new-modified-or-deleted-records.js
siliconchild: have added the description field like you suggested.

components/airtable/new-modified-or-deleted-records.js
siliconchild: Yes. Thanks for pointing that out. I have set it to 0.0.1.

siliconchild: Yes, have been set  to 0.0.1

components/airtable/new-modified-or-deleted-records.js
siliconchild: Since we are only saving the ids, would this amount to a significant amount of disk space.

components/airtable/new-modified-or-deleted-records.js
siliconchild: Thank You.

components/airtable/new-modified-or-deleted-records.js
siliconchild: Yes that's true. For my use case I just wanted to have either one event trigger so i can rebuild my Gatsby site on data change. Let me think about how to address this properly. Do you have any suggestions?

components/airtable/new-modified-or-deleted-records.js
dylburger: Yes, each record ID in Airtable is a string of ~17 bytes, so if a user had a base / table with 50,000 records, for example (limit on the number of records of the Pro plan), that's 850Kb of data, which may end up being higher than the limit we enforce for the total size of `this.db` for a given component.

But for now, this is fine - I have another suggested change to the `description` that should warn users with large bases.

components/airtable/new-modified-or-deleted-records.js
dylburger: Yeah in this case, I'd recommend _always_ running the logic to check for deleted records, instead of separating the logic that runs when `!data.records.length` or the `else` case.

components/airtable/new-modified-or-deleted-records.js
dylburger: When we talked about this source as a team, two questions re: limits came up:

1) The `this.db` limits we discussed below
2) Listing all records for a table, like you're doing below, is the best way we know of to check for deleted records. That approach is fine, but for large tables, that may also yield issues (for example, making so many HTTP requests to list records at each offset may cause the component to timeout - currently that timeout is 30 seconds).

Overall, we think it's fine to ship this source as-is, but note an initial record limit for this source in the description so that users with large tables understand that they may hit issues.

I'd change the description to read:

"Emits an event each time a record is added, updated, or deleted in an Airtable table. Supports tables up to 10,000 records".

10,000 is just an estimate, but we can start there and if we see any issues with the source, we can revisit.

dylburger:
@siliconchild changes look great. I resolved some of the remaining discussions and left notes on changing the `description` based on some conversations we had internally, and added a suggestion for the case where you have both additions and deletes.

Once you get changes in for those, I think we can merge this!

components/airtable/new-modified-or-deleted-records.js
siliconchild: have updated the description as you suggested to let users know of the limits.

components/airtable/new-modified-or-deleted-records.js
siliconchild: Yes deleted the if condition. It triggers on both events now.

dylburger:
Looks great, I just tested end to end. Thanks again for the contribution!

PR #447, opened 2020-09-16
Asana

michelle0927:


components/asana/asana.app.js
```diff
@@ -0,0 +1,251 @@
...
+      };
+    },
+    async _getAuthorizationHeader({ data, method, url, headers }) {
+      const token = {
+        key: this.$auth.oauth_access_token,
+        secret: this.$auth.oauth_refresh_token,
+      };
+      return await axios({
+        method: "POST",
+        url: `https://app.asana.com/api/1.0/webhooks`,
```
dylburger, line 75: Since you reference this base URL in a few places, you might create a method like this:

```javascript
_getBaseURL() {
  return "https://app.asana.com/api/1.0/"
}
```

and reference that everywhere you use the URL.

components/asana/asana.app.js
```diff
@@ -0,0 +1,251 @@
...
+        await axios(config);
+      } catch (err) {
+        console.log(err);
+      }
+      return authorization.data;
+    },
+    async createHook(body) {
+      const resp = await this._makeWebhookRequest({
+        method: "post",
+        url: `https://app.asana.com/api/1.0/webhooks`,
```
dylburger, line 94: Can you move some of this logic to the `_makeWebhookRequest` method itself, since each webhook request will use the same URL and headers?

components/asana/asana.app.js
```diff
@@ -0,0 +1,251 @@
...
+    },
+    async verifyAsanaWebhookRequest(request) {
+      let secret = this.$auth.oauth_refresh_token;
+      var base64Digest = function (s) {
+        return crypto.createHmac("sha1", secret).update(s).digest("base64");
+      };
+      var content = JSON.stringify(request.body);
+      var doubleHash = base64Digest(content);
+      var headerHash = request.headers["x-hook-secret"];
+      return doubleHash == headerHash;
```
dylburger, line 126: Prefer `===` for a strict equality check.

components/asana/asana.app.js
```diff
@@ -0,0 +1,251 @@
...
+    },
+    async getWorkspace(workspaceId) {
+      const workspace = await axios.get(
```
dylburger, line 129: Similar to Github and other apps, I think you could abstract some of the HTTP request logic into a `_makeRequest` method so that you don't have to pass the full URL or headers in every case. Instead, you can include those attributes of the config in that single method and then just pass the path to the `_makeRequest` method. You're already close to this with the `_makeWebhookRequest` method, so I think there's the opportunity to consolidate that further.

components/asana/completed-task.js
```diff
@@ -0,0 +1,78 @@
+const asana = require("https://github.com/PipedreamHQ/pipedream/components/asana/asana.app.js");
+const get = require("lodash.get");
+
+module.exports = {
+  name: "Task Completed",
```
dylburger, line 5: Can you change this to `Task Completed (Instant)` for this and other components that utilize webhooks? We're using this pattern now in the title to highlight the fact that the trigger should work (close to) instantly.

components/asana/completed-task.js
```diff
@@ -0,0 +1,78 @@
...
+      return;
+
+    this.http.respond({
+      status: 200,
+      headers: {
+        "x-hook-secret": event.headers["x-hook-secret"],
+      },
+    });
+
+    const body = get(event, "body");
```
dylburger, line 59: Usually I only use Lodash's `get` function in the cases where an intermediate property of a nested object might not exist. But in these simple cases I'd suggest just doing:

```javascript
const { body } = event
```

components/asana/completed-task.js
```diff
@@ -0,0 +1,78 @@
...
+    if (!body || !body.events) {
+      return;
+    }
+
+    let tasks = [];
```
dylburger, line 64: Small nit but you can still use `const tasks = []` and push new elements to that array. `const` will then just prevent a redeclaration of the entire variable, which is ideal to prevent here.

components/asana/completed-task.js
```diff
@@ -0,0 +1,78 @@
...
+
+    for (const e of body.events) {
+      tasks.push(await this.asana.getTask(e.resource.gid));
+    }
+
+    for (const task of tasks) {
```
dylburger, line 70: Technically since you just push the tasks to the `tasks` array, and then iterate over the same, new array here, you can just combine these two loops to iterate over `body.events`, get the task info, and immediately emit that all in the same loop.

components/asana/completed-task.js
dylburger: Also if that's the only use of `get` here, I'd also suggest removing the `require` to save space in the deployed component (we have internal limits we need to be aware of wrt/ the deployment size, so every little bit helps).

PR #461, opened 2020-09-19
Gitlab new milestone

jverce:

* Implement method to retrieve milestones from the Gitlab API.
* Implement polling logic and emit events for any unprocessed
  milestone.
* Fix typo in method name.
* Fix bug leading to infinite loops.

### DIFF
--- a/components/gitlab/new-milestone.js
+++ b/components/gitlab/new-milestone.js
@@ -11,7 +11,7 @@ module.exports = {
     timer: {
       type: "$.interface.timer",
       default: {
-        intervalSeconds: 60,
+        intervalSeconds: 15 * 60, // 15 minutes
       },
     },
     http: "$.interface.http",
@@ -40,11 +40,10 @@ module.exports = {
         created_at,
         title,
       } = data;
-      const summary = `New milestone created: ${title}`;
       const ts = +new Date(created_at);
       return {
         id,
-        summary,
+        summary: title,
         ts,
       };
     },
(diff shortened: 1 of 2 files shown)


components/gitlab/new-milestone.js
```diff
@@ -0,0 +1,91 @@
...
+  description: "Triggers when a new milestone is created in a project",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    gitlab,
+    projectId: { propDefinition: [gitlab, "projectId"] },
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60,
```
dylburger, line 14: Since free users have limited quota, I like to default sources to `15 * 60` `intervalSeconds`. Then since the user can change the default on component creation, they can always set a lower time.

components/gitlab/new-milestone.js
```diff
@@ -0,0 +1,91 @@
...
+    },
+  },
+  methods: {
+    generateMeta(data) {
+      const {
+        id,
+        created_at,
+        title,
+      } = data;
+      const summary = `New milestone created: ${title}`;
```
dylburger, line 43: Since there's limited visible space in the event summary I normally like to emit just the name of the relevant thing (in this case just the `title`, since the fact that it's a new milestone should be obvious from context.

components/gitlab/new-milestone.js
```diff
@@ -0,0 +1,91 @@
...
+    // don't emit events for them (i.e. we only want to emit events
+    // for new milestones).
+    let lastProcessedMilestoneId = this.db.get("lastProcessedMilestoneId");
+    const opts = {
+      projectId: this.projectId,
+      lastProcessedMilestoneId,
+    };
+    const unprocessedMilestones = await this.collectMilestones(opts);
+
+    if (unprocessedMilestones.length > 0) {
```
dylburger, line 71: Very small nit but I personally love the golden path in these scenarios:

```javascript
if (unprocessedMilestones.length === 0) {
  console.log("No new Gitlab milestones detected");
  return
}

// Actually process milestones here
```

dylburger (commented):
This looks fine to me, added just a few suggestions

components/gitlab/new-milestone.js
jverce: Done

components/gitlab/new-milestone.js
jverce: Done

components/gitlab/new-milestone.js
jverce: Done

PR #465, opened 2020-09-20
Gitlab new merge request

jverce:

* Add event source logic for New Merge Request

### DIFF
--- a/components/gitlab/new-merge-request.js
+++ b/components/gitlab/new-merge-request.js
@@ -1,0 +1,90 @@
+const gitlab = require("https://github.com/PipedreamHQ/pipedream/components/gitlab/gitlab.app.js");
+
+module.exports = {
+  name: "New Merge Request (Instant)",
+  description: "Emits an event when a new merge request is created",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    gitlab,
+    projectId: { propDefinition: [gitlab, "projectId"] },
+    http: "$.interface.http",
+    db: "$.service.db",
+  },
+  hooks: {
+    async activate() {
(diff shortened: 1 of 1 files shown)


components/gitlab/new-merge-request.js
```diff
@@ -0,0 +1,90 @@
...
+      const { hookId, token } = await this.gitlab.createHook(opts);
+      console.log(
+        `Created "merge request events" webhook for project ID ${this.projectId}.
+        (Hook ID: ${hookId}, endpoint: ${hookParams.url})`
+      );
+      this.db.set("hookId", hookId);
+      this.db.set("token", token);
+    },
+    async deactivate() {
+      const hookId = this.db.get("hookId");
```
tjk, line 33: Clearly lots of room for improvement with "app files" or some of form of mixins in the future, but in the meantime maybe the deactivate could be DRY'ed up easily? (the helper can even do the logging)

```
await this.gitlab.deleteHook(this.projectId, this.db.get("hookId"));
```

tjk (approved):

components/gitlab/new-merge-request.js
jverce: Yes and no.

Yes: definitely room for improvement. I've even discussed with @dylburger and @todsac about certain patterns that present themselves frequently.

No: It's hard to do now, since the only way to achive this is to dump every repeated pattern into the main `.app.js` component, which is not semantically correct, and it ends up bloating it with long and complex functions. So in general for the `gitlab` I just went with the pattern used in the `github` component so keep things consistent (I may create more intermediate variables just to add clarity to the code, which can give the impression of duplicated logic).

PR #467, opened 2020-09-21
Gitlab new project

jverce:

* Add event source logic for New Project


components/gitlab/new-project.js
```diff
@@ -0,0 +1,89 @@
...
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 15 * 60, // 15 minutes
+      },
+    },
+    http: "$.interface.http",
+    db: "$.service.db",
+  },
+  hooks: {
+    async activate() {
```
tjk, line 20: Hm, I think I like this pattern of using the activate hook for this. We should backport it to older components, although in older components we typically do emit a recent page (eg. for test event reasons). We should tackle that issue differently though -- this is good.

components/gitlab/new-project.js
```diff
@@ -0,0 +1,89 @@
...
+      const opts = {
+        pageSize: 1,
+      };
+      const projects = await this.collectProjects(opts);
```
tjk, line 24: Am I misreading? We only need the latest project here but it looks like you'd collect all the user's projects but with 1 per_page? Would be nice to just make call for the first page.

tjk (changes requested):
https://github.com/PipedreamHQ/pipedream/pull/467/files#r496457562

components/gitlab/new-project.js
jverce: +1
We should definitely strive for consistent building blocks across the project.

components/gitlab/new-project.js
jverce: I think you're right, good catch!! I'll change this to use what `getCommits` already does (limit the amount of queried items).

jverce: Indeed you were right. Just pushed a fix.

tjk (approved):

PR #471, opened 2020-09-22
Gitlab new review request

jverce:
* Add event source logic for New Review Request

### DIFF
--- a/components/gitlab/new-review-request.js
+++ b/components/gitlab/new-review-request.js
@@ -82,7 +82,7 @@ module.exports = {
       } = data.object_attributes;
       const summary = `New reviewer for "${title}": ${reviewer}`;
       const ts = +new Date(updated_at);
-      const compositeId = `${id}-${ts}`;
+      const compositeId = `${id}-${ts}-${reviewer}`;
       return {
         id: compositeId,
         summary,
--- a/components/gitlab/new-review-request.js
+++ b/components/gitlab/new-review-request.js
@@ -51,7 +51,7 @@ module.exports = {
       // in it are interpreted as new review requests.
       if (action === "open" || action === "reopen") {
         const { assignees = [] } = body;
-        return assignees.map(a => a.username);
+        return assignees;
       }
 
       // Gitlab API provides any merge request update diff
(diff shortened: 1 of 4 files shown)


components/gitlab/new-review-request.js
```diff
@@ -0,0 +1,119 @@
...
+        ts,
+      };
+    },
+  },
+  async run(event) {
+    const { headers, body } = event;
+
+    // Reject any calls not made by the proper Gitlab webhook.
+    if (!this.gitlab.isValidSource(headers, this.db)) {
+      this.http.respond({
```
tjk, line 97: With the new `customResponse` stuff, it might just be better to keep the prop defined the way it is and remove both of these `this.http.respond` calls. Probably fine to respond with a basic (canned) 200 in all cases.

components/gitlab/new-review-request.js
```diff
@@ -0,0 +1,119 @@
...
+    getNewReviewers(body) {
+      const { action, title } = body.object_attributes;
+
+      // When a merge request is first created, any assignees
+      // in it are interpreted as new review requests.
+      if (action === "open" || action === "reopen") {
+        const { assignees = [] } = body;
+        return assignees;
+      }
+
```
tjk, line 56: Should you return unless action is `updated`? At the very least if it is `closed` we should not emit events?

tjk: Oh nevermind I see you are looking at `body.changes`

components/gitlab/new-review-request.js
```diff
@@ -0,0 +1,119 @@
...
+        console.log(`No new assignees in merge request "${title}"`);
+        return [];
+      }
+
+      // If the assignees of the merge request changed, we need to compute
+      // the difference in order to extract the new reviewers.
+      const previousAssigneesUsernames = new Set(assignees.previous.map(a => a.username));
+      const newAssignees = assignees.current.filter(a => !previousAssigneesUsernames.has(a.username));
+      console.log(
+        `Assignees in merge request "${title}" changed to: ${newAssignees.map(a => a.username).join(', ')}`
```
tjk, line 72: Nitpicky: "Assignees added to merged request..." might be more accurate? Since the old ones could still be there?

tjk (changes requested):
One nitpicky thing + note about removing the custom responses, then feel free to merge.

components/gitlab/new-review-request.js
jverce: Forgot about the new `customResponse` field.
I'll add it to the prop and keep the existing response logic since we don't want to reply `200 OK` to unknown requestors.

components/gitlab/new-review-request.js
jverce: Noted, will rephrase.

jverce:
Just double checked on the [events related](https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#merge-request-events) to this type of webhook and we are good.

I discovered a few minor details while testing that I'm fixing in the next commit, as well as all the usages of `$.interface.http` that were not backfilled with the `customResponse` attribute.

PR #472, opened 2020-09-23
adding app metadata to http and rss sources

psavkar:


jverce:
Changes already merged in a separate PR (#1003)

PR #477, opened 2020-09-23
Ghost

michelle0927:


components/ghost/ghost-admin.app.js
```diff
@@ -0,0 +1,52 @@
...
+    },
+    async createHook(token, data) {
+      const config = {
+        method: "post",
+        url: `${await this._getBaseURL()}/webhooks`,
+        headers: await this._getHeader(token),
+        data,
+      };
+      try {
+        return await axios(config);
```
tjk, line 34: We should let this throw.

components/ghost/ghost-admin.app.js
```diff
@@ -0,0 +1,52 @@
...
+      }
+    },
+    async deleteHook(hookId, token) {
+      const config = {
+        method: "delete",
+        url: `${await this._getBaseURL()}/webhooks/${hookId}`,
+        headers: await this._getHeader(token),
+      };
+      try {
+        await axios(config);
```
tjk, line 46: Similar thing here -- or check the error we are catching (ie. if we fail to delete because it's already deleted, that would be okay)

components/ghost/ghost.app.js
```diff
@@ -0,0 +1,17 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "ghost_org_content_api",
+  methods: {
+    async _makeRequest(endpoint, page) {
+      limit = 15;
```
tjk, line 8: Declare `limit` as non-global, please (`const` or `let`)

components/ghost/new-author.js
```diff
@@ -0,0 +1,36 @@
...
+    db: "$.service.db",
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
```
tjk, line 19: It is maybe okay to just rely on the deduper, but it would be nice to use `this.db` to checkpoint and minimize api calls if the api allows?

components/ghost/ghost-admin.app.js
michelle0927: Updated to throw error.

components/ghost/ghost-admin.app.js
michelle0927: Updated to throw error

components/ghost/new-author.js
michelle0927: I updated the api call to use a filter to minimize results to authors with IDs greater than the highest ID we've previously received.

components/ghost/ghost.app.js
michelle0927: Updated to declare limit as const.

components/ghost/ghost-admin.app.js
tjk:
```
try {
  return await axios(config);
} catch (err) {
  throw err;
}
```

should just be written as:

```
return await axios(config);
```

components/ghost/ghost.app.js
```diff
@@ -0,0 +1,14 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "ghost_org_content_api",
+  methods: {
+    async getAuthors(page, latestId) {
+      const limit = 15;
+      return (results = await axios.get(
```
tjk, line 9: Can you use axios request config instead of the unsafe url building?

https://github.com/axios/axios#request-config

components/ghost/ghost-admin.app.js
```diff
@@ -0,0 +1,52 @@
+const axios = require("axios");
+const jwt = require("jsonwebtoken");
+
+module.exports = {
+  type: "app",
+  app: "ghost_org_admin_api",
+  methods: {
+    async _getBaseURL() {
```
tjk, line 8: same comment about being more careful about which functions you tag as `async`

components/ghost/ghost-admin.app.js
```diff
@@ -0,0 +1,52 @@
...
+        keyid: id,
+        algorithm: "HS256",
+        expiresIn: "5m",
+        audience: `/v3/admin/`,
+      });
+    },
+    async createHook(token, data) {
+      const config = {
+        method: "post",
+        url: `${await this._getBaseURL()}/webhooks`,
```
tjk, line 29: (and no need to await if not async)

components/ghost/member-created.js
```diff
@@ -0,0 +1,42 @@
...
+          },
+        ],
+      };
+      const token = await this.ghost._getToken();
+      const resp = await this.ghost.createHook(token, data);
+      this.db.set("hookId", resp.data.webhooks[0].id);
+      this.db.set("token", token);
+    },
+    async deactivate() {
+      console.log(this.db.get("hookId"));
```
tjk, line 30: No need to log this as there should already be a `db.get` observation.

components/ghost/new-author.js
```diff
@@ -0,0 +1,42 @@
...
+    while (page <= total) {
+      let results = await this.ghost.getAuthors(page, latestId);
+      total = results.data.meta.pagination.pages;
+      for (const result of results.data.authors) {
+        this.$emit(result, {
+          id: result.id,
+          summary: result.name,
+          ts: Date.now(),
+        });
+        if (Number(`0x${result.id.substr(1)}`) > Number(`0x${maxId}`)) maxId = result.id;
```
tjk, line 34: Are these UUIDs? I'm not sure this will work...

michelle0927: I wasn't totally sure about using this method either, but it was the only way I could figure out to minimize api calls. Maybe it's better in this case to rely on the deduper?

components/ghost/ghost.app.js
```diff
@@ -0,0 +1,20 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "ghost_org_content_api",
+  methods: {
+    async getAuthors(page) {
+      const config = {
```
tjk, line 8: Much better, thanks!

tjk (approved):

PR #480, opened 2020-09-24
Calendly

michelle0927:


components/calendly/calendly.app.js
```diff
@@ -0,0 +1,55 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "calendly",
+  methods: {
+    async _getHeader() {
```
tjk, line 7: Please don't declare functions async if not using `await` inside (marking functions returning promises is nice too but unnecessary)

components/calendly/calendly.app.js
```diff
@@ -0,0 +1,55 @@
...
+        "X-TOKEN": this.$auth.api_key,
+      };
+    },
+    async _getBaseURL() {
+      return "https://calendly.com/api/v1"
+    },
+    async getEventTypes() {
+      return (
+        await axios.get(`${this._getBaseURL()}/users/me/event_types`, {
+          headers: await this._getHeader(),
```
tjk, line 18: No need to await here once you remove the unnecessary async(s).

components/calendly/calendly.app.js
```diff
@@ -0,0 +1,55 @@
...
+      ).data.data;
+    },
+    async createHook(data) {
+      const config = {
+        method: "post",
+        url: `${await this._getBaseURL()}/hooks`,
+        headers: await this._getHeader(),
+        data,
+      };
+      try {
```
tjk, line 36: Don't try/catch these so the errors propagate

components/calendly/invitee-created.js
```diff
@@ -0,0 +1,39 @@
...
+      };
+      const resp = await this.calendly.createHook(body);
+      this.db.set("hookId", resp.data.id);
+    },
+    async deactivate() {
+      await this.calendly.deleteHook(this.db.get("hookId"));
+    },
+  },
+
+  async run(event) {
```
tjk, line 28: Should make sure we respond to the http request, like so:

```
if (event.headers["x-calendly-hook-id"] != this.db.get("hookId")) {
  this.http.respond({status: 404})
  return;
}

this.http.respond({status: 200})

// ... rest of code
```

components/calendly/calendly.app.js
```diff
@@ -0,0 +1,47 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "calendly",
+  methods: {
+    _getHeader() {
```
dylburger, line 7: Consider renaming this to `_getAuthHeader()` to be explicit about the type of header

components/calendly/invitee-canceled.js
```diff
@@ -0,0 +1,41 @@
...
+
+module.exports = {
+  name: "Invitee Cancelled (Instant)",
+  description: "Emits an event when an invitee cancels a scheduled event",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    calendly,
+    db: "$.service.db",
+    http: "$.interface.http",
```
dylburger, line 11: We made some recent changes to the HTTP interface. When you need a custom response like you do below, you'll need to set 

```javascript
http: {
  type: "$.interface.http",
  customResponse: true,
}
```

components/calendly/invitee-created.js
```diff
@@ -0,0 +1,41 @@
...
+
+module.exports = {
+  name: "Invitee Created (Instant)",
+  description: "Emits an event when an invitee schedules an event",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    calendly,
+    db: "$.service.db",
+    http: "$.interface.http",
```
dylburger, line 11: See http interface comment above

components/calendly/new-event.js
```diff
@@ -0,0 +1,41 @@
...
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
+    const now = new Date();
```
dylburger, line 20: Since you use this code to retrieve the timestamp a month ago in a couple of places, consider abstracting to a method in the app file?

Issue #412, opened 2020-09-10
[SOURCE] BitBucket


todsac:
**Describe the event source. What app, what event?**
BitBucket

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
New Commit
New Issue
New Event
New Repo
Watch team
Watch user
Watch repository

dylburger:
From #411:

* New branch
* New PR
* Pipeline events

jverce:
From BitBucket API docs, teams are deprecated in favor of workspaces:
https://developer.atlassian.com/bitbucket/api/2/reference/resource/teams

Will implement **Watch workspace** instead of **Watch team**

jverce:
Same for **Watch user**, see https://developer.atlassian.com/cloud/bitbucket/bitbucket-api-teams-deprecation/

PR #485, opened 2020-09-27
Vercel event sources

jverce:
* Add Vercel component 
* Add event source logic for New Deployment
* Add event source logic for Deployment Ready


jverce:
Hey @dannyroosevelt, sorry for the late reply.
The reason why this was blocked was because webhook calls could not be [secured as per Vercel's procedure](https://vercel.com/docs/integrations#webhooks/securing-webhooks), since the verification process requires access to the OAuth2 client secret that Pipedream uses to connect to Vercel, which is not accessible to the event source.

PR #486, opened 2020-09-29
Bitbucket new commit comment

jverce:
This PR is in support of issue #412
* Add event source logic for New Commit Comment

(Contains the changes from PR #481, so review that one first and then come back to this one)

PR #487, opened 2020-09-29
Bitbucket new commit

jverce:
* Add event source logic for New Commit

(Contains the changes from PR #486, so review that one first and then come back to this one)


components/bitbucket/new-commit.js
```diff
@@ -0,0 +1,152 @@
...
+    http: "$.interface.http",
+    workspaceId: { propDefinition: [bitbucket, "workspaceId"] },
+    repositoryId: {
+      propDefinition: [
+        bitbucket,
+        "repositoryId",
+        c => ({ workspaceId: c.workspaceId }),
+      ],
+    },
+    branchName: {
```
tjk, line 20: Can we make this prop optional?

tjk: (Maybe should even be `branchNames` but that might be not as convenient as it should punt on supporting 0-n branches for now -- but 0 or 1 might be a nice improvement for now?)

tjk (changes requested):
Just one improvement suggestion if not too tricky, but looks good otherwise.

components/bitbucket/new-commit.js
jverce: The approach here is the same as with Gitlab and Github, so in the spirit of consistency we should do the same here.
(We can always make a required argument optional, but not the other way around without breaking backward compatibility).

However, I agree that it would be nicer to have it as an optional parameter, and I don't think it'd require that much effort other than testing multiple cases.
I can refactor all 3 event sources (i.e. Github, Gitlab and BitBucket) in a separate PR if that's OK.

jverce: Added a task for this: https://github.com/PipedreamHQ/pipedream/issues/508

PR #493, opened 2020-09-30
Add meeting filter to recording completed source

psavkar:


components/zoom/recording-completed.js
```diff
@@ -7,14 +9,29 @@ module.exports = {
...
-  version: "0.0.2",
+  version: "0.0.3",
   props: {
     zoom,
     zoomApphook: {
       type: "$.interface.apphook",
       appProp: "zoom",
       eventNames: ["recording.completed"],
     },
+    meetingFilter: {
```
tjk, line 20: Can you name the variable `meetingIds` to be more clear.

tjk: (you can keep the label as is)

tjk (approved):
Just please rename prop.

PR #499, opened 2020-10-01
improve docs

ABHI12cmd:


Nihlus:
@todsac You should probably mark this as "invalid" as well, to prevent it from counting towards Hacktoberfest's old ruleset.

Issue #470, opened 2020-09-21
[SOURCE] StackOverflow: new question for specific keyword(s)


dylburger:
I'd like a source that emits a new event when a new question is added for a specific set of keywords.

todsac:
Event sources:
- New answer
- New question

PR #541, opened 2020-10-10
StackExchange new question for keywords

jverce:
* StackExchange New Question with keywords

The approach for this solution is the following:
1. Trigger the event source with a timer.
2. Perform an [advanced search](https://api.stackexchange.com/docs/advanced-search) against the StackExchange API for a specific time frame, from the last time the search was performed until now (for a [sample click here](https://api.stackexchange.com/docs/advanced-search#pagesize=10&fromdate=1602262702&todate=1602296890&order=asc&sort=creation&q=python&accepted=True&closed=False&filter=!9_bDE.BDp&site=stackoverflow&run=true))
3. Emit an event for each of the questions in the search result
4. Record the time of the search in the DB, to be used in the next execution

### DIFF
--- a/components/stack-exchange/new-question-for-keywords.js
+++ b/components/stack-exchange/new-question-for-keywords.js
@@ -52,7 +52,6 @@ module.exports = {
       toDate,
       order: 'asc',
       sort: 'creation',
-      accepted: true,
       closed: false,
       site: this.siteId,
       q: keywordsQuery,
--- a/components/stack-exchange/new-question-for-keywords.js
+++ b/components/stack-exchange/new-question-for-keywords.js
@@ -1,0 +1,69 @@
+const stack_exchange = require('./stack-exchange.app');
+
+module.exports = {
+  name: "New Question for Specific Keywords",
+  description: "Emits an event when a new question is posted and related to a set of specific keywords",
+  version: "0.0.1",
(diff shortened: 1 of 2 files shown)


components/stack-exchange/new-question-for-keywords.js
```diff
@@ -0,0 +1,69 @@
...
+      toDate,
+      order: 'asc',
+      sort: 'creation',
+      accepted: true,
+      closed: false,
+      site: this.siteId,
+      q: keywordsQuery,
+    };
+
+    const items = this.stack_exchange.advancedSearch(searchParams);
```
dylburger, line 61: Do you not need to await this?

jverce: It's not needed actually. If you take a look at the `for` loop below, I'm actually doing the `await` there with the `for await` construction (as [explained here](https://javascript.info/async-iterators-generators#async-generators-finally)).

I tried adding an `await` there to see if the error message for the unresolved promises went away, but it didn't :disappointed: 

As a reference, the message I'm getting is the following:
```
Pipedream detected that component was still running code. Make sure you await all promises. (Reason: Socket)
```
As discussed offline, this could be a false positive caused by some issues with the platform itself.

PR #547, opened 2020-10-14
BitBucket event sources refactoring

jverce:
* Extract repeating logic into `common` component
* Cleanup repetitive interfaces in the main `.app.js` component
* Fix bugs in the **New Commit** event source:
  * "Last processed hash" being set incorrectly
  * Metadata incorrectly displaying full commit message


components/bitbucket/new-commit.js
```diff
@@ -82,74 +59,75 @@ module.exports = {
...
-      });
+      // Push events can be for different branches, tags and
+      // causes. We need to make sure that we're only processing events
+      // that are related to new commits in the particular branch
+      // that the user indicated.
+      const eventsForThisBranch = changes.filter(this.isEventForThisBranch);
+      const isEventRelevant = eventsForThisBranch
+        .filter(this.doesEventContainNewCommits)
+        .length > 0;
+      if (isEventRelevant) {
```
dylburger, line 98: Since we use golden-path style elsewhere, consider negating this condition and returning early if the event isn't relevant.

dylburger (approved):

components/bitbucket/new-commit.js
jverce: Noted, will do.

PR #550, opened 2020-10-14
Intercom

michelle0927:


components/intercom/conversation-closed.js
```diff
@@ -0,0 +1,54 @@
+const intercom = require("https://github.com/PipedreamHQ/pipedream/components/intercom/intercom.app.js");
+
+module.exports = {
+  name: "New Closed Conversation",
+  description: "Emits an event each time a conversation is closed.",
```
dylburger, line 5: Will need to add a `key` here and on other components, as we talked about in Slack

components/intercom/conversation-closed.js
```diff
@@ -0,0 +1,54 @@
...
+    intercom,
+    db: "$.service.db",
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
```
dylburger, line 18: Remove newline

components/intercom/conversation-closed.js
```diff
@@ -0,0 +1,54 @@
+const intercom = require("https://github.com/PipedreamHQ/pipedream/components/intercom/intercom.app.js");
```
dylburger, line 1: See note on local requires in Slack

components/intercom/conversation-closed.js
```diff
@@ -0,0 +1,54 @@
...
+        value: lastConversationClosedAt,
+      },
+    };
+
+    let results = null;
+    let starting_after = null;
+
+    while (
+      !results ||
+      (results.data.pages.next !== null &&
```
dylburger, line 36: Do you care about the case where `results.data.pages.next === ""` (empty string)? Otherwise you can just check for `results.data.pages.next` being truthy:

```
while (results || results.data.pages.next) {
```
```

File: components/intercom/new-admin-reply.js [hunk 5]
```diff
@@ -0,0 +1,74 @@
...
+          lastAdminReplyAt = conversation.statistics.last_admin_reply_at;
+        let conversationData = (
+          await this.intercom.getConversation(conversation.id)
+        ).data;
+        let total_count = conversationData.conversation_parts.total_count;
+        if (
+          total_count > 0 &&
+          conversationData.conversation_parts.conversation_parts[
+            total_count - 1
+          ].body != null
```
dylburger, line 54: This might be a good use case for Lodash's `get` method, since in theory `conversationData.conversation_parts.conversation_parts[total_count - 1]` might not exist and you might get an error at this step. I'd suggest doing something like:

```
const get = require("lodash.get")
const conversationBody = get(conversationData, "conversation_parts.conversation_parts[total_count - 1].body")
if (total_count > 0 && conversationBody) { ...
```

components/intercom/conversation-closed.js
```diff
@@ -0,0 +1,54 @@
...
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
+    const monthAgo = this.intercom.monthAgo();
+    let lastConversationClosedAt =
+      this.db.get("lastConversationClosedAt") || Math.floor(monthAgo / 1000);
+    const data = {
```
dylburger, line 23: In this and each of the components below, I'd recommend including the `data` object, as well as the code for filtering on objects since the last item (e.g. `lastConversationClosedAt`), within the app file method itself - here, `this.intercom.searchConversations`.

That is, it'd be great if the component itself just contained this code in the run method:

```javascript
const monthAgo = this.intercom.monthAgo();
let lastConversationClosedAt = this.db.get("lastConversationClosedAt") || Math.floor(monthAgo / 1000);
const conversations = await this.intercom.searchConversations(data, starting_after, lastConversationClosedAt);
for (const conversation of conversations) {
  this.$emit(conversation, {
    id: `${conversation.id}${conversation.last_close_at}`,
    summary: conversation.source.body,
    ts: conversation.last_close_at,
  });
}
```

By passing `lastConversationClosedAt` to the app file method, I think you can handle all of the other logic in the app file, and just return the conversations that meet your criteria.

I think you can do the same in each of the components below - just a little refactor.

components/intercom/new-lead.js
```diff
@@ -0,0 +1,64 @@
...
+        results.data.pages.next !== undefined)
+    ) {
+      if (results) starting_after = results.data.pages.next.starting_after;
+      results = await this.intercom.searchContacts(data, starting_after);
+      for (const lead of results.data.data) {
+        if (lead.created_at > lastLeadCreatedAt)
+          lastLeadCreatedAt = lead.created_at;
+        this.$emit(lead, {
+          id: lead.id,
+          summary: lead.name ? lead.name : lead.id,
```
dylburger, line 56:
```
lead.name || lead.id
```

dylburger (changes requested):

PR #558, opened 2020-10-16
StackExchange new answers for questions

jverce:
* Logic to poll the StackExchange API for new questions
* Refactored some common code to reutilize the API querying/pagination logic.


components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
```diff
@@ -0,0 +1,74 @@
...
+    async activate() {
+      const fromDate = this._getCurrentEpoch();
+      this.db.set("fromDate", fromDate);
+    },
+  },
+  methods: {
+    _getCurrentEpoch() {
+      // The StackExchange API works with Unix epochs in seconds.
+      const epochInMs = +new Date();
+      return Math.floor(epochInMs / 1000);
```
dylburger, line 37: You could shorten this by one line if you wanted:

```javascript
return Math.floor(Date.now() / 1000)
```

components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
```diff
@@ -0,0 +1,74 @@
...
+        id,
+        summary,
+        ts,
+      };
+    },
+  },
+  async run() {
+    const fromDate = this.db.get("fromDate");
+    const toDate = this._getCurrentEpoch();
+    const filter = '!SWKA(ozr4ec2cHE9JK'; // See https://api.stackexchange.com/docs/filters
```
dylburger, line 56: Interesting, so you created this filter and are just referencing its string representation here?

dylburger (approved):

components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
jverce: Yes and no :slightly_smiling_face: 
StackExchange offers this "particular" API to create/predefine response filters so that they can be reused. They are also public, so we can reuse filters created by others. In my case, I used their UI to set up a particular filter, and the UI went ahead and handled the creation (or retrieval, probably) of a filter with the specified characteristics:

stackexchange_filter

components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
jverce: Will do.

PR #559, opened 2020-10-16
Stackexchange new answers from users

jverce:
* Add event source logic for New Answers from Users
* Refactor prop options retrieval to reuse code
* Fix prop description


components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
```diff
@@ -0,0 +1,74 @@
...
+  hooks: {
+    async activate() {
+      const fromDate = this._getCurrentEpoch();
+      this.db.set("fromDate", fromDate);
+    },
+  },
+  methods: {
+    _getCurrentEpoch() {
+      // The StackExchange API works with Unix epochs in seconds.
+      const epochInMs = +new Date();
```
dylburger, line 36: Same small feedback from other PR

components/stack_exchange/stack_exchange.app.js
```diff
@@ -35,6 +36,61 @@ module.exports = {
       label: "Keywords",
       description: "Keywords to search for in questions",
     },
+    questionIds: {
+      type: "string[]",
+      label: "Questions",
+      description: "Questions to monitor (max: 100)",
+      useQuery: true,
+      async options(context) {
+        const q = context.query || '';
```
dylburger, line 45: Any specific reason you need to default to the empty string here, or will it work by just letting `context.query` be undefined when it's not present?

components/stack_exchange/stack_exchange.app.js
```diff
@@ -61,35 +156,51 @@ module.exports = {
...
-          ...baseParams,
-          page,
-        };
-        const requestConfig = {
-          ...baseRequestConfig,
-          params,
-        };
-        const { data } = await axios.get(url, requestConfig);
+        const data = await this.getItemsForPage(url, baseParams, page);
         if (data.items.length === 0) {
```
dylburger, line 176: Will `data.items` always be defined?

dylburger (approved):

components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
jverce: Will do.

components/stack_exchange/stack_exchange.app.js
jverce: Yeah, the StackExchange API doesn't work well when we set `q` to `null`, which is the value that `context.query` has when it's not set. I spend a couple of hours scratching my head because of this :sob:

components/stack_exchange/stack_exchange.app.js
jverce: Yes, for the URL's for which we're calling this method. If no items are present, the API returns an empty array.
I will add a check just in case so that we fail gracefull should something change.

components/stack_exchange/sources/new-answers-for-questions/new-answers-for-questions.js
jverce: Yup, will fix all the occurrences within this component.

components/stack_exchange/stack_exchange.app.js
jverce: Answered before, but just for future contributors:
`context.query` is `null` when left unset, which doesn't go well with the StackExchange API.

PR #560, opened 2020-10-17
BitBucket new repository

jverce:
* Add event source logic for New Repository

### DIFF
--- a/components/bitbucket/new-repository.js
+++ b/components/bitbucket/new-repository.js
@@ -1,0 +1,42 @@
+const common = require("./common");
+const watchWorkspace = require("./watch-workspace");
+
+const EVENT_SOURCE_NAME = "New Repository (Instant)";
+
+module.exports = {
+  ...watchWorkspace,
+  name: EVENT_SOURCE_NAME,
+  description: "Emits an event when a repository is created",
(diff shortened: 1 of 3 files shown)


components/bitbucket/new-repository.js
```diff
@@ -0,0 +1,42 @@
...
+
+const EVENT_SOURCE_NAME = "New Repository (Instant)";
+
+module.exports = {
+  ...watchWorkspace,
+  name: EVENT_SOURCE_NAME,
+  description: "Emits an event when a repository is created",
+  version: "0.0.1",
+  props: {
+    ...common.props,
```
tjk, line 12: Might be better to be more explicit about which props / methods are being borrowed from other files... but since we don't have a good pattern or recommendation for this yet, I'm fine with just shipping like this and we learn over time!

tjk (approved):
One comment to consider but feel free to merge

components/bitbucket/new-repository.js
jverce: Yeah, it's not ideal for sure (I personally don't like it that much 😂).
I think we can do better, but like you said I think it's going to be a process.

PR #561, opened 2020-10-17
BitBucket new review request

jverce:
* Add event source logic for New Review Request

### DIFF
--- a/components/bitbucket/new-review-request.js
+++ b/components/bitbucket/new-review-request.js
@@ -1,0 +1,75 @@
+const common = require("./common");
+const { bitbucket } = common.props;
+
+const EVENT_SOURCE_NAME = "New Review Request (Instant)";
+
+module.exports = {
+  ...common,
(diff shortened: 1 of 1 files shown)


components/bitbucket/new-review-request.js
```diff
@@ -0,0 +1,74 @@
...
+      const compositeId = `${pullRequestId}-${reviewerId}`;
+      return {
+        id: compositeId,
+        summary,
+        ts,
+      };
+    },
+    async processEvent(event) {
+      const { headers, body } = event;
+      body.pullrequest.reviewers.forEach(reviewer => {
```
tjk, line 63: Sometimes I wonder whether we should be more defensive about the expected event shapes... but I think I prefer this approach of letting stuff throw

tjk (approved):

components/bitbucket/new-review-request.js
jverce: Aren't these built internally by the platform?

PR #564, opened 2020-10-20
Bitbucket pipeline events

jverce:
* Cosmetic changes to remove trailing periods
* Add event source logic for New Pipeline Event

### DIFF
--- a/components/bitbucket/new-pipeline-event.js
+++ b/components/bitbucket/new-pipeline-event.js
@@ -1,4 +1,4 @@
-const _ = require("lodash");
+const isEmpty = require("lodash/isEmpty");
 const common = require("./common");
 const { bitbucket } = common.props;
 
@@ -52,7 +52,7 @@ module.exports = {
     isEventRelevant(event) {
       const { state: eventType } = event.body.commit_status;
       return (
-        _.isEmpty(this.eventTypes) ||
+        isEmpty(this.eventTypes) ||
         this.eventTypes.some(i => i === eventType)
       );
     },
--- a/components/bitbucket/new-pipeline-event.js
+++ b/components/bitbucket/new-pipeline-event.js
@@ -1,0 +1,83 @@
+const _ = require("lodash");
+const common = require("./common");
+const { bitbucket } = common.props;
+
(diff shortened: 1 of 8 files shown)


components/bitbucket/new-pipeline-event.js
```diff
@@ -0,0 +1,83 @@
+const _ = require("lodash");
```
tjk, line 1: lodash is [kinda big](https://bundlephobia.com/result?p=lodash@4.17.20) -- if you are just using `_.isEmpty` maybe use [lodash.isempty](https://bundlephobia.com/result?p=lodash.isempty@4.4.0)? This will speed up async option configuring + deploys so it's a good pattern to try to adhere to whenever it's easy.

components/bitbucket/new-pipeline-event.js
```diff
@@ -0,0 +1,83 @@
...
+      const ts = +new Date(eventDate);
+      return {
+        id,
+        summary,
+        ts,
+      };
+    },
+    processEvent(event) {
+      if (this.isEventRelevant(event)) {
+        const parent = common.methods.processEvent.bind(this);
```
tjk, line 78: interesting pattern with spreading common and this... we've discussed mixins / inheritance in the past (clearly code reuse has not been a priority) -- will be nice to take a look if certain strategies emerge

tjk (approved):
One suggestion but then looks good!

components/bitbucket/new-pipeline-event.js
jverce: There are some rough edges, like when trying to call a "parent" method (e.g. those defined in the `.app.js` file) without having the parent instance available.
So to sum up, I like that we ran into the same issues :slightly_smiling_face:

components/bitbucket/new-pipeline-event.js
jverce: I will do some refactoring. I avoided using per-method packages since it's [discouraged by Lodash](https://lodash.com/per-method-packages), but I can definitely do something to avoid requiring the whole package.

PR #574, opened 2020-10-26
Hubspot

michelle0927:


components/hubspot/sources/contact-property-change/contact-property-change.js
```diff
@@ -0,0 +1,82 @@
...
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
+    const lastRun = this.db.get("updatedAfter") || this.hubspot.monthAgo();
+    const updatedAfter = new Date(lastRun);
+
+    const properties = [];
```
dylburger, line 41: You might be able to re-write to something shorter here:

```javascript
const properties = this.properties.map(p => JSON.parse(p).value)
```

components/hubspot/sources/contact-property-change/contact-property-change.js
```diff
@@ -0,0 +1,82 @@
...
+
+    while (hasMore) {
+      let contacts = await this.hubspot.getRecentlyUpdatedContacts(params);
+      hasMore = contacts["has-more"];
+      if (hasMore) params.vidOffset = contacts["vid-offset"];
+      for (const contact of contacts.contacts) {
+        for (let property of this.properties) {
+          property = JSON.parse(property);
+          let updatedAt = new Date(
+            contact.properties[property.value].versions[0].timestamp
```
dylburger, line 65: Are there conditions where intermediate properties might not be present in this object?

components/hubspot/sources/contact-property-change/contact-property-change.js
```diff
@@ -0,0 +1,82 @@
...
+          );
+          if (updatedAt > updatedAfter) {
+            this.$emit(contact, {
```
dylburger, line 68: You might consider a pattern we've been using in other sources when the emit metadata is this complex. We've been keeping a `generateMeta()` method in the component's `methods` section, just to separate the logic here. That method just returns the object of { id, ts, summary } metadata.

This way you can simply run:

```this.$emit(contact, generateMeta(contact, property))```

File: components/hubspot/sources/deal-property-change/deal-property-change.js [hunk 4]
```diff
@@ -0,0 +1,83 @@
...
+      },
+    },
+  },
+
+  async run(event) {
+    const lastRun = this.db.get("updatedAfter") || this.hubspot.monthAgo();
+    const updatedAfter = new Date(lastRun);
+
+    const property = [];
+    for (let prop of this.properties) {
```
dylburger, line 42: Similar feedback here:

```javascript
const property = this.properties.map(p => JSON.parse(p).value)
```

components/hubspot/sources/deal-property-change/deal-property-change.js
```diff
@@ -0,0 +1,83 @@
...
+    while (hasMore) {
+      let results = await this.hubspot.getRecentlyUpdatedDeals(params);
+      hasMore = results["has-more"];
+      if (hasMore) params.vidOffset = results["vid-offset"];
+      for (const deal of results.results) {
+        for (let prop of this.properties) {
+          prop = JSON.parse(prop);
+          if (deal.properties[prop.value]) {
+            let updatedAt = new Date(
+              deal.properties[prop.value].versions[0].timestamp
```
dylburger, line 65: Consider a Lodash `get`.

components/hubspot/sources/new-blog-article/new-blog-article.js
```diff
@@ -0,0 +1,48 @@
...
+        this.$emit(blogpost, {
+          id: blogpost.id,
+          summary: blogpost.name,
+          ts: createdAt.getTime(),
+        });
+        count++;
+      }
+    }
+
+    this.db.set("createdAfter", null);
```
dylburger, line 46: Is this `null` intentional?

components/hubspot/sources/new-contact-in-list/new-contact-in-list.js
```diff
@@ -0,0 +1,67 @@
...
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
+    for (let list of this.lists) {
+      list = JSON.parse(list);
+      const params = {
+        count: 20,
```
dylburger, line 48: Is there any way to filter this API call by time like you do for the others?

components/hubspot/sources/new-contact/new-contact.js
```diff
@@ -0,0 +1,59 @@
...
+  },
+
+  async run(event) {
+    const lastRun = this.db.get("createdAfter") || this.hubspot.monthAgo();
+    const createdAfter = new Date(lastRun);
+    const data = {
+      limit: 10,
+      sorts: [
+        {
+          propertyName: "createdate",
```
dylburger, line 27: Similar question - can this be filtered by created date or another time-based param to reduce the result set?

components/hubspot/sources/contact-updated/contact-updated.js
```diff
@@ -0,0 +1,58 @@
...
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+
+  async run(event) {
+    const lastRun = this.db.get("updatedAfter") || this.hubspot.monthAgo();
+    const updatedAfter = new Date(lastRun);
+    const data = {
+      limit: 10,
```
dylburger, line 23: Here and in other places - are these limits maximums, or can you fetch more than 10 results at a time? Just thinking of ways to reduce API calls as you paginate.

components/hubspot/sources/new-engagement/new-engagement.js
```diff
@@ -0,0 +1,46 @@
...
+
+  async run(event) {
+    const lastRun = this.db.get("createdAfter") || this.hubspot.monthAgo();
+    const createdAfter = new Date(lastRun);
+    const params = {
+      limit: 100,
+    };
+    let results = null;
+
+    while (!results || params.offset != undefined) {
```
dylburger, line 28: Will `while(!results || params.offset)` work?

components/hubspot/sources/new-engagement/new-engagement.js
```diff
@@ -0,0 +1,46 @@
...
+        let createdAt = new Date(result.engagement.createdAt);
+        if (createdAt.getTime() > createdAfter.getTime()) {
+          this.$emit(result, {
+            id: result.engagement.id,
+            summary: result.engagement.type,
+            ts: result.engagement.createdAt,
+          });
+        }
+      }
+      if (results.hasMore == true) params.offset = results.offset;
```
dylburger, line 40: Will `if(results.hasMore)` work?

components/hubspot/sources/new-contact-in-list/new-contact-in-list.js
michelle0927: I don't think this API call allows for filtering by time. https://legacydocs.hubspot.com/docs/methods/lists/get_list_contacts

components/hubspot/sources/new-contact/new-contact.js
michelle0927: For this API call, createdate is not one of the searchable properties, but it does allow you to sort by that property.

components/hubspot/sources/contact-updated/contact-updated.js
michelle0927: These limits can be whatever we want. I just set them as the default values, but let me know if I should increase these.

dylburger: I think you should increase to at least `100` to reduce the need to make multiple API calls for larger customers. That'll reduce the total component runtime.

Issue #532, opened 2020-10-08
[SOURCE] Salesforce


todsac:
**Describe the event source. What app, what event?**
Salesforce

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**

Two potential paths and I prefer the first:

A generic set of triggers:

- New Record | Triggers when a record of the specified Salesforce object (ie. Contact, Lead, Opportunity, etc.) is created.
- Updated Field on Record | Triggers when a field of your choosing (ie. email address, status) is updated on any record of a specified Salesforce object (ie. Contact, Lead, Opportunity, etc.).
- Updated Record | Triggers when any record of the specified Salesforce object (ie. Contact, Lead, Opportunity, etc.) is updated.

Or more specific triggers:

- New Opportunity
- Opportunity updated
- Opportunity closed
- New Email (added to list)
- New Campaign
- New Case
- New Contact
- New Custom Object
- Custom Object updated
- New Event
- New Folder
- New Note
- New Task
- Account updated
- Contact updated
- New Lead
- Lead updated

PR #579, opened 2020-10-28
Salesforce new account

jverce:
This PR is in support of issue #532 
* Add Salesforce component
* Add event source logic for New Account

### DIFF
--- a/components/salesforce/common.js
+++ b/components/salesforce/common.js
@@ -1,0 +1,151 @@
+const slugify = require("slugify");
+const { v4: uuidv4 } = require("uuid");
+
+const salesforce = require("./salesforce.app");
+
+module.exports = {
+  dedupe: "unique",
+  props: {
+    db: "$.service.db",
+    http: {
+      type: "$.interface.http",
+      customResponse: true,
+    },
+    salesforce,
+  },
+  hooks: {
+    async activate() {
+      // Retrieve base API URL's for this account
+      const apiUrls = await this.salesforce.getApiUrls();
+      this.db.set("apiUrls", apiUrls);
+
+      // Enable this Salesforce account to make HTTP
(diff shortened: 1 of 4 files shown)


components/salesforce/common.js
```diff
@@ -0,0 +1,151 @@
...
+      await this.salesforce.deleteApexClass(webhookId);
+
+      const { metadata: metadataApiUrl } = this.db.get("apiUrls");
+      const remoteSiteName = this.db.get("remoteSiteName");
+      await this.salesforce.deleteRemoteSite(metadataApiUrl, remoteSiteName);
+    },
+  },
+  methods: {
+    _getRemoteSiteName() {
+      const eventSourceName = this.getEventSourceName();
```
dylburger, line 64: Won't this throw as written?

components/salesforce/common.js
```diff
@@ -0,0 +1,151 @@
...
+        statusCode: 404,
+      });
+      return;
+    }
+
+    this.http.respond({
+      statusCode: 200,
+    });
+
+    return this.processEvent(event);
```
dylburger, line 149: Do you need to `return` this or can you just call it?

components/salesforce/common.js
dylburger: Oh I see you're overriding in the actual component below

dylburger (approved):

components/salesforce/common.js
jverce: From the component's point of view, I could just do this and it would be the same:
```js
await this.processEvent(event);
```

I'll change to that for clarity.

components/salesforce/common.js
jverce: Yeah, it's a dirty version of classes with virtual methods :joy:

PR #591, opened 2020-10-30
SuperSaaS component

guiprav:
For supersaas.com.

### DIFF
--- a/components/supersaas/sources/changed-appointments.js
+++ b/components/supersaas/sources/changed-appointments.js
@@ -1,0 +1,39 @@
+const supersaas = require('../supersaas.app.js');
+
+module.exports = {
+  key: 'supersaas-changed-appointments',
+  name: 'New or changed appointments',
+  description: `This event source tracks new and changed appointment events from the selected schedules and emits them.`,
+  version: '0.0.1',
+
+  props: {
+    supersaas,
+    schedules: { propDefinition: [supersaas, 'schedules'] },
+    db: "$.service.db",
+    http: '$.interface.http',
+  },
+
+  hooks: {
+    async activate() {
+      const { http, schedules } = this;
+
+      this.db.set('activeHooks', await this.supersaas.createHooks(schedules.map(x => ({
+        event: 'C', // change_appointment
+        parent_id: Number(x),
+        target_url: http.endpoint,
+      }))));
+    },
+
+    async deactivate() {
+      await this.supersaas.destroyHooks(this.db.get('activeHooks') || []);
+      this.db.set('activeHooks', []);
+    },
+  },
+
+  async run(ev) {
+    // TODO: Authenticate requests? (see GitHub's HMAC signature approach)
+
+    console.log('Emitting:', ev.body);
+    this.$emit(ev.body);
+  },
+};
--- a/components/supersaas/sources/changed-appointments.js
+++ b/components/supersaas/sources/changed-appointments.js
@@ -3,7 +3,7 @@ const supersaas = require('../supersaas.app.js');
 module.exports = {
   key: 'supersaas-changed-appointments',
   name: 'New or changed appointments',
-  description: `This event source tracks new and changed appointment events from the selected schedules and emits them.`,
+  description: `Emits an event for every changed appointments from the selected schedules.`,
   version: '0.0.1',
 
   props: {
--- a/components/supersaas/sources/changed-appointments.js
+++ b/components/supersaas/sources/changed-appointments.js
@@ -1,3 +1,5 @@
+const dayjs = require('dayjs');
+const makeEventSummary = require('../utils/makeEventSummary.js');
 const supersaas = require('../supersaas.app.js');
 
 module.exports = {
@@ -27,7 +29,15 @@ module.exports = {
     },
   },
   async run(ev) {
-    console.log('Emitting:', ev.body);
-    this.$emit(ev.body);
+    const outEv = {
+      meta: {
+        summary: makeEventSummary(ev),
+        ts: dayjs(ev.body.created_on).valueOf(),
+      },
+      body: ev.body,
+    };
+
+    console.log('Emitting:', outEv);
+    this.$emit(outEv, outEv.meta);
   },
 };
--- a/components/supersaas/sources/changed-appointments.js
+++ b/components/supersaas/sources/changed-appointments.js
@@ -5,14 +5,12 @@ module.exports = {
   name: 'New or changed appointments',
   description: `Emits an event for every changed appointments from the selected schedules.`,
   version: '0.0.1',
-
   props: {
     supersaas,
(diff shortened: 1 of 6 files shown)


components/supersaas/sources/changed-appointments.js
```diff
@@ -0,0 +1,39 @@
+const supersaas = require('../supersaas.app.js');
+
+module.exports = {
+  key: 'supersaas-changed-appointments',
+  name: 'New or changed appointments',
+  description: `This event source tracks new and changed appointment events from the selected schedules and emits them.`,
```
dylburger, line 6: In other components we normally use the pattern "Emits an event for every..." or "Emits an event for each new..." to make the language the user sees consistent in the UI. Could you try to re-word the description to fit that pattern?

components/supersaas/sources/changed-appointments.js
```diff
@@ -0,0 +1,39 @@
...
+  version: '0.0.1',
+
```
dylburger, line 8: Can you remove the empty newlines between sections here and in the rest of the files? We need to develop ESLint / Prettier rules that automate this, so that should catch this in the future.

components/supersaas/sources/changed-appointments.js
```diff
@@ -0,0 +1,39 @@
...
+      await this.supersaas.destroyHooks(this.db.get('activeHooks') || []);
+      this.db.set('activeHooks', []);
+    },
+  },
+
+  async run(ev) {
+    // TODO: Authenticate requests? (see GitHub's HMAC signature approach)
+
+    console.log('Emitting:', ev.body);
+    this.$emit(ev.body);
```
dylburger, line 37: Did you happen to see the metadata you can pass to the emitted events? https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-API.md#emit

At a minimum `summary` is normally nice to include, since that's what appears in the UI for the emitted event for the user and can make it easier to identify the event. In this case if you have access to the appointment name, that might be nice to include as the summary.

`id` is nice to include if you want to take advantage of the [built-in deduper](https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-API.md#dedupe-strategies). If you include a unique event ID (fixed for a particular webhook request) in the payload, for example, you can pass that as the `id`, then define a `dedupe: unique` strategy in the component, and we'll make sure duplicate events don't get emitted. Optional but a nice feature.

components/supersaas/sources/changed-appointments.js
```diff
@@ -0,0 +1,39 @@
...
+    },
+
+    async deactivate() {
+      await this.supersaas.destroyHooks(this.db.get('activeHooks') || []);
+      this.db.set('activeHooks', []);
+    },
+  },
+
+  async run(ev) {
+    // TODO: Authenticate requests? (see GitHub's HMAC signature approach)
```
dylburger, line 34: Do you pass a signature or other secret on webhook requests? If not, the validation isn't necessary from our perspective - we always include it if the source service allows, but not required.

components/supersaas/sources/changed-users.js
```diff
@@ -0,0 +1,39 @@
...
+  },
+
+  hooks: {
+    async activate() {
+      const { $auth } = this.supersaas;
+      const { http } = this;
+
+      this.db.set('activeHooks', await this.supersaas.createHooks([{
+        event: 'M', // change_appointment
+        parent_id: $auth.account,
```
dylburger, line 22: Can you avoid passing `parent_id` (and `$auth`) here and simply reference the parent ID within the `createHooks` method?

dylburger: Oh I see in the component above, this won't always be `$auth.account`

components/supersaas/sources/changed-appointments.js
guiprav: Nope, that's what I was thinking of adding. I'm only worried of endpoint name conflicts causing servers from other Pipedream users to send payload I'm not interested in and causing errors in our own endpoints. Something to prevent endpoint guess attacks might also be nice, but if you think it may be overkill, I'll trust you anytime :P

components/supersaas/supersaas.app.js
```diff
@@ -0,0 +1,66 @@
...
+      const { axios } = this;
+
+      console.log('Creating hooks:', hookParams);
+
+      return await Promise.all(hookParams.map(
+        x => axios('/api/hooks', { method: 'POST', params: x }),
+      ));
+    },
+
+    // TODO: Better error handling. Dylan suggested retries with a backoff
```
dylburger, line 51: I'm not too worried about this for now, but we'll consider ways to improve this kind of error handling in the future as we talked about in Slack.

dylburger (commented):
Overall looks great, added just a few comments

components/supersaas/sources/changed-users.js
guiprav: I don't think so, because the `parent_id` is only `$auth.account` in this particular invocation. See for example in `changed-appointments.js` how `parent_id` is something else entirely.

components/supersaas/sources/changed-appointments.js
dylburger: Gotcha. Based on our algorithm for generating endpoints the probability of collisions is infinitesimal, and we should also block that from happening at the API level.

components/supersaas/sources/changed-users.js
dylburger: Yep makes sense, thanks

components/supersaas/sources/changed-appointments.js
guiprav: Resource names are sometimes available but mostly not. Our webhooks seem to be sending slightly inconsistent event structures in this regard, but we don't want to make changes to our codebase to support this for now because we don't know yet how worthwhile is this whole effort (only very few users want it in principle) and definitely don't want to accidentally cause trouble to many other customers.

But there is enough information in standard format to write a summary, yes, so I'll do that.

This will probably mean moving some of the code I wrote for the logging action step into the event source code for `credit-balance-changes`.

components/supersaas/sources/changed-appointments.js
guiprav: Are empty lines *always* undesirable, or only inside JSON and object literals? Should the line after the `TODO` comment be kept? What about the other ones from other function bodies?

dylburger: In this case I'd only suggest removing the empty lines between properties of the Node module (e.g. between `version` and `props`, and between `props` and `hooks`. Other empty newlines within methods, for example, are totally fine.

components/supersaas/sources/changed-appointments.js
dylburger: 👍

PR #596, opened 2020-11-03
Generalize Salesforce Event Sources

jverce:
* Add event source logic for:
  * New objects
  * Updated objects
  * Deleted objects
* Add user input prop to allow users to select the type of sObject for
  which new events should be monitored

This PR is in support of issue #532

### DIFF
--- a/components/salesforce/sources/new-object/new-object.js
+++ b/components/salesforce/sources/new-object/new-object.js
@@ -1,0 +1,60 @@
+const startCase = require("lodash/startCase");
+
+const common = require("../../common");
+
+const EVENT_SOURCE_NAME = "New Object (Instant)";
+
+module.exports = {
+  ...common,
+  name: EVENT_SOURCE_NAME,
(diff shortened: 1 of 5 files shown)


components/salesforce/sources/new-object/new-object.js
```diff
@@ -0,0 +1,60 @@
...
+
+const common = require("../../common");
+
+const EVENT_SOURCE_NAME = "New Object (Instant)";
+
+module.exports = {
+  ...common,
+  name: EVENT_SOURCE_NAME,
+  key: "salesforce-new-object",
+  description: "Triggers when a new object is created",
```
dylburger, line 11: Can you rephrase to the "Emits an event..." format?

dylburger (approved):

components/salesforce/sources/new-object/new-object.js
jverce: Done!

PR #614, opened 2020-11-12
Ac/yotpo

adolfo-pd:


components/yotpo/sources/new-reviews/new-reviews.js
```diff
@@ -0,0 +1,34 @@
+const yotpo = require('../../yotpo.app.js')
+module.exports = {
+  name: 'yotpo-new-reviews',
+  version: '0.0.1',
```
dylburger, line 4: You'll need a `key` of the format `[app slug]-[source name]`, and a `description`, which typically reads "Emits an event on each..."

components/yotpo/sources/new-reviews/new-reviews.js
```diff
@@ -0,0 +1,34 @@
...
+        await this.yotpo.sendTestEvents()
+      }
+    },
+    async deactivate() {
+      await this.yotpo.deleteWebhook("review_create", this.http.endpoint)
+      await this.yotpo.deleteWebhook("review_updated", this.http.endpoint)
+    },
+  },
+  async run(event) {
+    this.$emit(event)
```
dylburger, line 32: I'd recommend adding a `summary` to the emit metadata. Are there also conditions where Yotpo might send the same event twice (on retry or other cases)? Normally I'll add a `dedupe` strategy at the top of the component, then pass an `id` in the emit metadata that the deduper uses (see Component API docs).

components/yotpo/yotpo.app.js
```diff
@@ -0,0 +1,74 @@
...
+          params: {
+            access_token: this.$auth.oauth_access_token,
+          },
+        }),
+      }
+    },
+    async sendTestEvents() {
+      const api = this.api().core
+      const response = await api.post("/reviews/dynamic_create", {
+        "appkey": this.$auth.app_key,
```
dylburger, line 28: Just to confirm, is this a secret or can it be made public to users?

dylburger (approved):
lgtm once comments are addressed

dylburger (approved):

Issue #74, opened 2020-05-10
[TRIGGER] Webflow


dylburger:
See https://developers.webflow.com/#create-new-webhook

todsac:
New Event (generic event source)
New Form Submission
New Order
Updated order

PR #625, opened 2020-11-17
Webflow Event Sources

jverce:
This PR is in support of issue #74 

* Add Webflow component and basic common code
* Add event source logic for New Form Submission
* Add event source logic for collection events
* Add event source logic for e-commerce events
* Add event source logic for Site Published

### DIFF
--- a/components/webflow/sources/common.js
+++ b/components/webflow/sources/common.js
@@ -10,8 +10,8 @@ module.exports = {
       type: "string",
       label: "Site",
       description: "The site from which to listen events",
-      async options(actions) {
-        const { page } = actions;
+      async options(context) {
+        const { page } = context;
         if (page !== 0) {
           return {
             options: []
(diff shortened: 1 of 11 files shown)


components/webflow/sources/common.js
```diff
@@ -0,0 +1,77 @@
...
+module.exports = {
+  props: {
+    db: "$.service.db",
+    http: "$.interface.http",
+    webflow,
+    siteId: {
+      type: "string",
+      label: "Site",
+      description: "The site from which to listen events",
+      async options(actions) {
```
tjk, line 13: actions -> context ?

tjk (approved):
nit but looks good

components/webflow/sources/common.js
jverce: Huh, you're right, I think I made a typo here. Will fix it.

PR #648, opened 2020-11-28
Google Sheets

michelle0927:


components/google_sheets/google_drive.app.js
```diff
@@ -0,0 +1,235 @@
+const axios = require("axios");
```
jsendo, line 1: this looks like a copy paste of https://github.com/PipedreamHQ/pipedream/blob/master/components/google_drive/google_drive.app.js
and it's out of date now.  Remove this file and just require the one in the google_drive directory.

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -0,0 +1,172 @@
+const google_sheets = require("../../google_sheets.app.js");
+const google_drive = require("../../google_drive.app.js");
```
jsendo, line 2: `const google_drive = require("../../../google_drive/google_drive.app.js");` should work here

components/google_sheets/sources/new-updates/new-updates.js
```diff
@@ -0,0 +1,232 @@
+const google_sheets = require("../../google_sheets.app.js");
+const google_drive = require("../../google_drive.app.js");
```
jsendo, line 2: same as the file above

components/google_sheets/sources/new-updates/new-updates.js
```diff
@@ -0,0 +1,232 @@
...
+      return;
+    }
+
+    const { files, newPageToken } = await this.google_drive.getModifiedSheets(
+      pageToken,
+      this.drive === "myDrive" ? null : this.drive,
+      sheetIDs
+    );
+    for (const file of files) {
+      const spreadsheet = await this.google_sheets.getSpreadsheet(file.id);
```
jsendo, line 156: maybe refactor this into functions?

jsendo (changes requested):

components/google_sheets/google_drive.app.js
michelle0927: Much of it is copied from the one in the google_drive directory, but the last 3 methods are not. Maybe I could require the one in the google_drive directory in this app file, remove the copied methods, and reference the original methods (from the google_drive directory app file) in this app file?

PR #650, opened 2020-12-02
Formstack

michelle0927:


components/formstack/sources/new-form-submission/new-form-submission.js
```diff
@@ -0,0 +1,48 @@
...
+    formId: { propDefinition: [formstack, "formId"] },
+  },
+  hooks: {
+    async activate() {
+      const { id } = await this.formstack.createHook({
+        id: this.formId,
+        url: this.http.endpoint,
+      })
+      this.db.set("hookId", id);  
+      this.db.set("handshake_key", this.formstack.$auth.oauth_refresh_token);
```
jsendo, line 22: you shouldn't have to save this.

components/formstack/sources/new-form-submission/new-form-submission.js
```diff
@@ -0,0 +1,48 @@
...
+    },  
+  },
+  async run(event) {
+    const body = get(event, "body");
+    if (!body) {
+      return;
+    }
+
+    // verify incoming request
+    if (this.db.get("handshake_key") !== get(body, "HandshakeKey"))
```
jsendo, line 37: related to the node above, you can just do:
 `this.formstack.$auth.oath_refresh_token !== get(body, "HandshakeKey")`

jsendo: you could also log that the tokens mismatched

components/formstack/sources/new-form-submission/new-form-submission.js
```diff
@@ -0,0 +1,48 @@
+const formstack = require("../../formstack.app.js");
+const get = require("lodash.get");
+
+module.exports = {
+  key: "formstack-new-form-submission",
+  name: "New Form Submission (Instant)",
+  description: "Emits an event for each new form submission.",
```
jsendo, line 7: dedupe

components/formstack/sources/new-form/new-form.js
```diff
@@ -0,0 +1,34 @@
...
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+  async run() {
+    const formIds = this.db.get("formIds") || [];
+    const forms = await this.formstack.getForms();
```
jsendo, line 21: getting the entire list of forms for the account every 15 mins seems really inefficient.  We'll end up with a bunch of duplicates that could be avoided. you could save the pages you've iterated through and start from that.  Things could get missed if a form in a previous page has been deleted though.

Assuming that the fromstack api returns everything in id order, you could page through everything and only emit ids > the largest id emitted from the previous run.  Then save the greatest id emitted in the current run for the next run.

components/formstack/formstack.app.js
```diff
@@ -0,0 +1,83 @@
...
+        method: "GET",
+        url: `${this._getBaseURL()}/form.json`,
+        headers: this._getAuthHeaders(),
+        params: {
+          page: 1,
+          per_page: 100,
+          folders: false,
+        }
+      }
+      while (count < total) {
```
jsendo, line 71: don't think count is necessary.  You just need to loop while the total returned from the api is == the page size:

```
while (pageSize === total) {

}
```

for the last page of 0 to 99 forms, the loop will bail.

components/formstack/formstack.app.js
```diff
@@ -0,0 +1,83 @@
...
+        const results = (await axios(config)).data;
+        total = results.total;
+        for (const form of results.forms) {
```
jsendo, line 74: instead of looping through `results.forms` you can just concatenate it to `forms`

components/formstack/formstack.app.js
```diff
@@ -0,0 +1,83 @@
...
+    },
+    async deleteHook({ hookId }) {
+      const config = {
+        method: "DELETE",
+        url: `${this._getBaseURL()}/webhook/${hookId}.json`,
+        headers: this._getAuthHeaders(),
+      }
+      return await axios(config);
+    },
+    async getForms() {
```
jsendo, line 57: this might be cleaner as a wrapper around just the api call, so it takes the folders, page and per_page params.  Then you handle the paging logic in the component code.

components/formstack/formstack.app.js
jsendo: per comment above this loop would move to the component `run()` code.

jsendo (changes requested):

components/formstack/formstack.app.js
michelle0927: Since this function is used both for props and for components, moving the pagination from the function to both of those places would create some duplication. Other reviewers have sometimes recommended consolidating code to reduce cross-component duplication, but I suppose it's just a matter of preference. If your preference is to leave pagination out of app file methods, I'll extrapolate it to both places.

Issue #626, opened 2020-11-17
[BUG] Salesforce event source returns 400 error on creation


dylburger:
Salesforce doesn't allow the programmatic creation of Apex trigger objects directly on a production instance. When creating a Salesforce event source on my production instance, I see the following error:

```json
{
      "message": "Can not create Apex Class on an active organization.",
      "errorCode": "ENTITY_IS_LOCKED",
      "fields": []
}
```

jverce:
Some reference links:
* https://trailblazers.salesforce.com/answers?id=9064S000000D5UzQAK
* https://help.salesforce.com/articleView?id=000324396&type=1&mode=1
* https://salesforce.stackexchange.com/questions/255000/issue-while-creating-apex-class-using-api

jverce:
### Progress so far

* The process and requirements for deploying code in Prod organizations is more involved, and requires for example having tests that provide sufficient coverage. Hence, the changes needed for this bug fix are the following:
  * The Apex code that this component was already sending (i.e. classes and trigger)
  * Tests for the trigger
  * Utility classes for to actually run those tests, like mocks and object factories
* Since most of the content is this component resulted to be very specific to the process of deploying code to Salesforce, I extracted everything Apex-related to a separate [NPM package](https://www.npmjs.com/package/salesforce-webhooks). The strongest reason behind this is that trying to tackle this without the use of templates for the mentioned Apex code became impractical.
* The package mentioned above offers a simple interface to create and delete webhooks in Salesforce, and this component will just focus on handling these webhook events, and getting config options from the user.

### Current focus

The biggest problem here is the fact that in order to test the webhook Apex code we need to create, update and/or delete object instances (e.g. an `Account`, a `Lead`, etc.) but every one of those object types has a specific set of required fields which are not provided via any API. There are some independent efforts to provide something useful for this purpose:
1. https://github.com/mbotos/SmartFactory-for-Force.com: I've tried this options with relative success after doing some fixes. I am doing an exhaustive check right now to verify that this works with all SObject types (e.g. currently it fails to create `Lead` objects for example)
2. https://github.com/apexlarson/SObjectFactory: this is another alternative that I haven't tried but could be considered if the option above cannot be amended

jverce:
A couple of fixes have been merged:
[...]

PR #670, opened 2020-12-09
Salesforce timer-based event sources

jverce:
## Description
This PR provides an alternative to the previous Salesforce event sources, and does not require Apex code deployment (as described in #626)

## Changelog
* Refactor Salesforce event sources for every SObject types to be
  timer-based
* Remove everything related to the SOAP API and Apex, and rely solely
  on the REST API `getUpdated` and `getDeleted` endpoints


components/salesforce/common.js
```diff
@@ -1,152 +1,77 @@
...
-      return this.standardizeName("webhook");
-    },
-    _isValidSource(event) {
-      const {
-        "x-webhook-token": webhookToken,
-      } = event.headers;
-      const secretToken = this.db.get("secretToken");
-      return webhookToken === secretToken;
+    isValidSObject(sobject) {
+      return sobject.replicateable;
```
dylburger, line 53: Could you describe what the function of the `replicateable` field is briefly in a comment in this method?

components/salesforce/sources/new-object/new-object.js
```diff
@@ -2,26 +2,31 @@ const startCase = require("lodash/startCase");
 
 const common = require("../../common");
 
-const EVENT_SOURCE_NAME = "New Object (Instant)";
-
 module.exports = {
   ...common,
-  name: EVENT_SOURCE_NAME,
+  name: "New Object",
```
dylburger, line 7: Is this the name Zapier assigns? Maybe we can edit the description to make the fact that you choose the object type clear

components/salesforce/sources/new-object/new-object.js
```diff
@@ -2,26 +2,31 @@ const startCase = require("lodash/startCase");
...
   key: "salesforce-new-object",
   description: "Emit an event when an object is created",
```
dylburger, line 9: Re: the above comment, maybe add something here re: the fact that this source lets you listen for new objects of any type, and that you choose the type when creating the source.

components/salesforce/sources/new-object/new-object.js
```diff
@@ -30,31 +35,33 @@ module.exports = {
...
+        startTimestamp,
+        endTimestamp,
+      } = eventData;
+      const {
+        ids,
+        latestDateCovered,
+      } = await this.salesforce.getUpdatedForObjectType(this.objectType, startTimestamp, endTimestamp);
+
+      // By the time we try to retrieve an item, it might've been deleted. This
+      // will cause `getSObject` to throw a 404 exception, which will reject it's
```
dylburger, line 49: "reject its promise"

dylburger (approved):

components/salesforce/common.js
jverce: Will do

components/salesforce/sources/new-object/new-object.js
jverce: They actually use `New Record in Salesforce`. We could rename it to that?

components/salesforce/sources/new-object/new-object.js
jverce: Oops, fixed.

components/salesforce/sources/new-object/new-object.js
jverce: Will do. How long are these descriptions supposed to be? Do we have a limit on their length?

jverce: OK, I tried with a multiline string and it's handled fine by the UI so I'll go with that!

PR #671, opened 2020-12-09
Salesforce webhook-based event sources

jverce:
## Description
This PR provides a fix to the previous Salesforce event sources, in which Apex code is deployed through [another API](https://developer.salesforce.com/docs/atlas.en-us.apexcode.meta/apexcode/sforce_api_calls_compileandtest.htm), required for production organizations  (as described in #626). The event source delegates the responsibility of putting together the Apex code, tests, SOAP API request payloads, etc to the [`salesforce-webhooks`](https://www.npmjs.com/package/salesforce-webhooks) package.

## Changelog
* Use the `salesforce-webhooks` NPM package which deploys Apex code
  and tests using the SOAP API (which works with production orgs as
  well)
* Cleanup unused code


components/salesforce/sources/new-object-instant/new-object-instant.js
```diff
@@ -0,0 +1,36 @@
+const startCase = require("lodash/startCase");
+
+const common = require("../../common-instant");
+
+module.exports = {
+  ...common,
+  name: "New Object (Instant)",
+  key: "salesforce-new-object-instant",
+  description: "Emit an event when an object is created",
```
dylburger, line 9: Same feedback as on the other PR re: clarifying that the source allows you to select your target object.

For context, we're planning to display the description more prominently, so you see it alongside the event source's title.

dylburger: It would be good to clarify why users should choose this vs. the non-instant trigger in the description.

components/salesforce/sources/object-deleted-instant/object-deleted-instant.js
```diff
@@ -0,0 +1,43 @@
...
+
+module.exports = {
+  ...common,
+  name: "Object Deleted (Instant)",
+  key: "salesforce-object-deleted-instant",
+  description: "Emit an event when an object is deleted",
+  version: "0.0.1",
+  methods: {
+    ...common.methods,
+    isValidSObject(sobject) {
```
dylburger, line 13: This logic isn't relevant for new objects, correct?

dylburger (approved):

components/salesforce/sources/object-deleted-instant/object-deleted-instant.js
jverce: You are correct, this is something I forgot to remove.

components/salesforce/sources/new-object-instant/new-object-instant.js
jverce: Will do. I was also thinking of adding a `README` with more details, would that make sense?

Issue #557, opened 2020-10-15
[SOURCE] Sendgrid


dylburger:
I'd like event sources for [all delivery and engagement events](https://sendgrid.com/docs/for-developers/tracking-events/event/#delivery-events).

PR #673, opened 2020-12-10
SendGrid Delivery Events

jverce:
This PR is in support of issue #557 

* Refactor common code to be used in multiple components
* Split common code for timer-based vs http-based event sources
* Add event source logic for Delivery Events

### DIFF
--- a/components/sendgrid/sendgrid.app.js
+++ b/components/sendgrid/sendgrid.app.js
@@ -39,7 +39,10 @@ module.exports = {
       const { data } = await axios.post(url, requestData, requestConfig);
       return data;
     },
-    async _withRetries(apiCall, retriableStatusCodes) {
+    _isRetriableStatusCode(statusCode) {
+      [408, 429, 500].includes(statusCode);
+    },
+    async _withRetries(apiCall) {
       const retryOpts = {
         retries: 5,
         factor: 2,
@@ -49,7 +52,7 @@ module.exports = {
           return await apiCall();
         } catch (err) {
           const statusCode = get(err, ["response", "status"]);
-          if (!retriableStatusCodes.has(statusCode)) {
+          if (!this._isRetriableStatusCode(statusCode)) {
             bail(`
               Unexpected error (status code: ${statusCode}):
               ${JSON.stringify(err.response, null, 2)}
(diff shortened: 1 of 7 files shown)


components/sendgrid/sendgrid.app.js
```diff
@@ -68,5 +72,23 @@ module.exports = {
         retriableStatusCodes,
       );
     },
+    async getWebhookSettings() {
+      const url = this._webhookSettingsUrl();
+      const requestConfig = this._makeRequestConfig();
+      const retriableStatusCodes = new Set([408, 429]);
```
tjk, line 78: nit but if these are always the same maybe just default in `withRetries`?

components/sendgrid/sources/common/http-based.js
```diff
@@ -0,0 +1,75 @@
...
+
+module.exports = {
+  ...base,
+  props: {
+    ...base.props,
+    http: "$.interface.http",
+  },
+  hooks: {
+    ...base.hooks,
+    async activate() {
```
tjk, line 11: I think deactive/activate get called anytime props are updated for example... can we be a bit more forgiving here if we are the owners of the webhook?

components/sendgrid/sources/common/http-based.js
```diff
@@ -0,0 +1,75 @@
...
+      throw new Error('webhookEventFlags is not implemented');
+    },
+    generateMeta(data) {
+      const {
+        sg_event_id: id,
+        timestamp: ts,
+      } = data;
+      return {
+        id,
+        summary: "New event",
```
tjk, line 63: can we do a better summary?

tjk: nevermind this is http-base

components/sendgrid/sources/delivery-events/delivery-events.js
```diff
@@ -0,0 +1,84 @@
...
+  name: "Delivery Events (Instant)",
+  description: "Emit an event when a delivery event is detected in SendGrid",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    ...common.props,
+    processed: {
+      type: "boolean",
+      label: "Processed",
+      description: `
```
tjk, line 15: hm i suppose these descriptions with newlines are probably okay for when embedded in html but maybe we shouldn't do them like this? (i suppose we could strip out in other clients maybe.. hm..)

tjk (approved):
mostly minor comments, the activate one is my main question

components/sendgrid/sources/delivery-events/delivery-events.js
jverce: That is actually I question I raised in another PR. If descriptions are a free-form text, and can be arbitrarily long, then at some point someone will add a very long line to the code.
🤷🏻‍♂️

components/sendgrid/sources/common/http-based.js
jverce: That's a good idea. We can check if the URL's match for example. Does that make sense?

components/sendgrid/sendgrid.app.js
jverce: That's a good idea, looks like all API endpoints share these status codes, so we can reuse them.
Will fix!

PR #675, opened 2020-12-11
Fix API URL to use new authentication

jverce:
This PR adds some fixes to use the new Salesforce authentication flow.

### DIFF
--- a/components/salesforce/salesforce.app.js
+++ b/components/salesforce/salesforce.app.js
@@ -11,12 +11,25 @@ module.exports = {
     _instance() {
       return this.$auth.yourinstance;
     },
+    _instanceUrl() {
+      return this.$auth.instance_url;
+    },
+    _subdomain() {
+      return (
+        this._instance() ||
+        this._instanceUrl()
+          .replace('https://', '')
+          .replace('.salesforce.com', '')
+      );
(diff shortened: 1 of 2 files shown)


components/salesforce/salesforce.app.js
```diff
@@ -11,12 +11,25 @@ module.exports = {
...
     },
+    _instanceUrl() {
+      return this.$auth.instance_url;
+    },
+    _subdomain() {
+      return (
+        this._instance() ||
+        this._instanceUrl()
+          .replace('https://', '')
+          .replace('.salesforce.com', '')
```
dylburger, line 22: just to confirm, based on what you know, we should include the full subdomain in all cases, e.g. `toptal3-dev-ed.my` when you have a "my" portion of the domain, and just `na174` for `na174.salesforce.com`?

dylburger (approved):

components/salesforce/salesforce.app.js
jverce: Yup, that's correct. The "instance" ID is just a generic subdomain that users can use to call the API, and in this context it's no different than any other `custom-subdomain.with-multiple-hierarchies.salesforce.com`

PR #677, opened 2020-12-11
SendGrid Events

jverce:
This PR is in support of issue #557 

* Generalise the event source for both Delivery and Engagement events
* Add webhook signature verification


components/sendgrid/sources/common/http-based.js
```diff
@@ -19,60 +26,52 @@ module.exports = {
...
+      const webhookPublicKey = this.db.get("webhookPublicKey");
+
+      const webhookHelper = new EventWebhook();
+      const ecdsaPublicKey = webhookHelper.convertPublicKeyToECDSA(webhookPublicKey);
+      return webhookHelper.verifySignature(ecdsaPublicKey, payload, signature, timestamp);
     },
     processEvent(event) {
+      if (!this._isValidSource(event)) {
+        this.http.respond({
+          status: 404,
```
tjk, line 66: maybe status: 400 with body: "signature check failed" (forget if we support string body) or something just to make the random debugging a bit more explicit

tjk (approved):
one minor common but looks good to me! clean as always, nice work :)

components/sendgrid/sources/common/http-based.js
jverce: Done!

PR #679, opened 2020-12-11
Google Sheets updates

michelle0927:


components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -87,12 +102,12 @@ module.exports = {
...
   },
   methods: {
-    getMeta(spreadsheet, sheet, diff) {
+    getMeta(spreadsheet, sheet) {
       return {
         id: `${spreadsheet.spreadsheetId}${
           sheet.properties.sheetId
         }${Date.now()}`,
-        summary: `${diff} rows added to ${spreadsheet.properties.title} - ${sheet.properties.title}`,
+        summary: `New row added to ${spreadsheet.properties.title} - ${sheet.properties.title}`,
```
dylburger, line 110: The summary will be the same for every row now, so maybe we do something like:

```
summary: data.join(", ")
```

where `data` is the array of new row data?

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -121,19 +136,16 @@ module.exports = {
...
-      if (sheetIDs.length === 0) {
-        const sheets = (
-          await this.google_drive.listSheets(
-            this.drive === "myDrive" ? null : this.drive
-          )
-        ).options;
-        for (const s of sheets) sheetIDs.push(s.value);
-      }
-      const rowCounts = await this.google_sheets.getRowCounts(sheetIDs);
+      const rowCounts = await this.google_sheets.getRowCounts(this.sheetID);
```
dylburger, line 141: Maybe rename the method to `getWorksheetRowCounts`? I was initially confused at why there would be multiple counts for a given spreadsheet.

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -144,21 +156,36 @@ module.exports = {
...
     }
-    const { files, newPageToken } = await this.google_drive.getModifiedSheets(
+    const { file, newPageToken } = await this.google_drive.getModifiedSheet(
       pageToken,
       this.drive === "myDrive" ? null : this.drive,
-      sheetIDs
+      this.sheetID
     );
-    for (const file of files) {
+    if (file) {
```
dylburger, line 164: I'd move the setting of the `newPageToken` above this section so it's colocated with the code that retrieves it above. Then:

```javascript
if (!file) return;
```

so you can avoid the indentation of the rest of the code below.

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -144,21 +156,36 @@ module.exports = {
...
+        ) {
+          continue;
+        }
         let oldRowCount = this.db.get(
           `${spreadsheet.spreadsheetId}${sheet.properties.sheetId}`
         );
-        let rowCount = sheet.properties.gridProperties.rowCount;
-        if (oldRowCount && rowCount > oldRowCount) {
+        let rowCount = sheet.data[0].rowData ? sheet.data[0].rowData.length : 0;
+        if (rowCount > oldRowCount) {
```
dylburger, line 177:
```javascript
if (rowCount <= oldRowCount) return;
```

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -144,21 +156,36 @@ module.exports = {
...
-          this.$emit(sheet, this.getMeta(spreadsheet, sheet, diff));
+          let range = `${sheet.properties.title}!${
+            rowCount - (diff - 1)
+          }:${rowCount}`;
+          let newRowValues = await this.google_sheets.getSpreadsheetValues(
+            spreadsheet.spreadsheetId,
+            range
+          );
+          for (const newRow of newRowValues.values) {
+            this.$emit({ newRow, sheet }, this.getMeta(spreadsheet, sheet));
```
dylburger, line 187: I noticed the new emit format didn't contain the `range`. Is there a reason you left it out here? That could still be useful to emit - I think some consumers will need it.

components/google_sheets/sources/new-updates/new-updates.js
```diff
@@ -210,17 +215,27 @@ module.exports = {
...
+      this.sheetID
     );
-    for (const file of files) {
+    if (file) {
       const spreadsheet = await this.google_sheets.getSpreadsheet(file.id);
       for (const sheet of spreadsheet.sheets) {
-        const { oldValues, currentValues } = await this.getValues(spreadsheet, sheet, file);
+        if (
+          this.worksheetIDs.length > 0 &&
+          !this.worksheetIDs.includes(sheet.properties.sheetId)
```
dylburger, line 228: should read `sheet.properties.sheetId.toString()`. In my tests we were `continue`ing every time because `sheetId` was an int but `this.worksheetIDs` contains strings.

dylburger (changes requested):
Just a few changes. Only major one is in the new updates PR (the `toString()` comment).

PR #680, opened 2020-12-11
Todoist

michelle0927:
Added incomplete-task.js to this set of components created by @psavkar. Todoist requires webhooks to be set up in their App Management Console, but sync-resources.js can poll for events/changes to any resources.


components/todoist/sources/completed-task/completed-task.js
```diff
@@ -0,0 +1,46 @@
...
+    for (const property in result) {
+      if(Array.isArray(result[property])) {
+        for (const element of result[property]) {
+          let matchingProject = await this.todoist.isProjectInList(element.project_id, this.selectProjects)
+          if(element.checked === 1 && matchingProject) {
+            let dedupeId = `${element.id}-${(new Date(element.date_completed)).getTime()}`
+            this.$emit(element, {
+              summary: element.content,
+              id: dedupeId, 
+            })
```
jverce, line 38: We should extract the metadata elements filtering and the "generation" into separate methods (e.g. `isElementRelevant`, `generateMeta`, etc.) so that we can further abstract this piece of code.

For example:
```js
// sources/common.js
methods: {
  generateMeta(element) {
    const {
      id: elementId,
      content: summary,
      date_completed: dateCompleted,
    } = element;
    const ts = new Date(dateCompleted).getTime();
    const id = `${elementId}-${ts}`;
    return {
      id,
      summary,
      ts,
    };
  }
},
async run() {
  const syncResult = await this.getSyncResult();
  Object.values(syncResult)
    .filter(Array.isArray)
    .flat()
    .filter(this.isElementRelevant)
    .filter(element => this.todoist.isProjectInList(element.project_id, this.selectProjects))
    .forEach(element => {
      const meta = this.generateMeta(element);
      this.$emit(element, meta);
    });
}

// sources/completed-task/completed-task.js
const common = require("../../common");
module.exports = {
  ...common,
  // Key, description, etc.
  methods: {
    isElementRelevant(element) {
      return element.checked === 1;
    },
  },
};
```

This way, we can limit each event source to only contain logic that is particular to its use case, and have the rest of the boilerplate in common modules.
I suggest having:
- `sources/common-task.js`
- `sources/common-project.js`
- `sources/common.js` (parent of the above)

components/todoist/todoist.app.js
```diff
@@ -0,0 +1,112 @@
...
+      if (selectedProjectIds.length > 0) {
+        if(selectedProjectIds.includes(project_id)) {
+          return true
+        } else {
+          return false
+        }
+      } else {
+        return true
+      }  
+    },
```
jverce, line 85: Can we rewrite it like this, to make it cleaner?
```js
isProjectInList(projectId, selectedProjectIds) {
  return (
    selectedProjectIds.length === 0 ||
    selectProjectIds.includes(projectId)
  );
}
```

Also, this is not `async` right?

components/todoist/todoist.app.js
```diff
@@ -0,0 +1,112 @@
+const axios = require('axios')
+const querystring = require('querystring')
+const resourceTypes = ['labels','projects','items','notes','sections','filters','reminders','locations','user','live_notifications','collaborators','user_settings','notification_settings'].sort()
```
jverce, line 3: Given that this is static, can we extract it into a JSON file?
You should be able to do something like this:
```js
// resource-types.json
[
  "labels",
  "projects",
  // ...
]

// todoist.app.js
const resourceTypes = require("./resource-types")
```

You can even sort it when creating the JSON file itself, so that the JSON file contains a list of resource types that's already sorted.

components/todoist/sources/incomplete-task/incomplete-task.js
```diff
@@ -0,0 +1,46 @@
...
+    })
+  
+    for (const property in result) {
+      if(Array.isArray(result[property])) {
+        for (const element of result[property]) {
+          let matchingProject = await this.todoist.isProjectInList(element.project_id, this.selectProjects)
+          if(element.checked === 0 && matchingProject) {
+            let dedupeId = `${element.id}-${(new Date(element.date_completed)).getTime()}`
+            this.$emit(element, {
+              summary: element.content,
```
jverce, line 36: Isn't `element.content` too long to be used as a summary?

jverce (changes requested):

components/todoist/sources/incomplete-task/incomplete-task.js
michelle0927: Since incomplete-task.js was the only source I created in this branch, I chose to make it consistent with the other Todoist sources. What do you recommend using as the summary?

components/todoist/todoist.app.js
michelle0927: Pipedream does not support this, but I was able to extract the resource types into a separate .js module.

components/todoist/sources/common-task.js
```diff
@@ -0,0 +1,32 @@
...
+  ...common,
+  props: {
+    todoist,
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 5,
+      },
+    },
+    db: "$.service.db",
```
jverce, line 14: These props are also very common and would be a good idea to have them on `sources/common.js`.

The we can do:
```js
module.exports = {
  ...common,
  props: {
    ...common.props,
    selectProjects: { propDefinition: [todoist, "selectProjects"] },
  },
};
```

components/todoist/sources/common-task.js
```diff
@@ -0,0 +1,32 @@
+const todoist = require("../todoist.app.js");
+const common = require("./common.js");
+
+module.exports = {
```
jverce, line 4: We should add a base implementation of `isElementRelevant` in this module just so that future event sources don't throw exceptions due to the fact that the method is missing (if a developer forgets to implement it):

I suggest this:
```js
module.exports = {
  methods: {
    isElementRelevant() {
      return true;
    },
  },
};
```

components/todoist/todoist.app.js
jverce: Yeah, I verified and you're right, that's a shame 🙁 
Nice workaround though 🙂

components/todoist/sources/incomplete-task/incomplete-task.js
jverce: For the summaries I usually do something very short and descriptive, like `New incomplete task: ${id}`

jverce (approved):
Added some minor comments, feel free to address them and merge!

Issue #649, opened 2020-11-30
[SOURCE] Unfollow source for twitter.


DAVE4547:
**Describe the event source. What app, what event?**
A source for when you or another user looses a follower on twitter.
**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
You probably already have this but this is it: https://developer.twitter.com/en/docs

jverce:
A couple of event sources have been pushed to `master` to address this. Their structure is similar to the existing `new-follower-of-me` and `new-follower-of-user` event sources.

A few considerations:
- Twitter does not have an API to retrieve recent unfollows, so the way to compute this is to keep track of the whole list of followers at different points in time, and compute the difference.
- This will put extra pressure on the API limits, especially for accounts with a high amount of followers. Hence, the timer configuration should be set such that it doesn't go above a ratio of **5000 followers / min**.
- For example, if an account has 15000 followers, the event source should be triggered with a frequency of at most 3 minutes. A  shorter time will result in Twitter throttling the event source.

PR #688, opened 2020-12-15
Add event source logic for New Unfollower

jverce:
* Add method to the Twitter app file to retrieve the entire list of
  followers of a Twitter account
* Extract common logic for new follower/unfollower event
  sources (refactor pending for previous event sources)
* Add logic for New Unfollower (of current user and of another user)
* Move languages data to separate file

Solves #649

### DIFF
--- a/components/twitter/twitter.app.js
+++ b/components/twitter/twitter.app.js
@@ -182,7 +182,7 @@ module.exports = {
         }
       }
       config.headers.authorization = authorization
-      return await axios(config)
+      return axios(config)
     },
     async getFollowers(screen_name) {
       return (await this._makeRequest({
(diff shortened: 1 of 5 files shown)


components/twitter/twitter.app.js
```diff
@@ -917,6 +193,47 @@ module.exports = {
...
+        url,
+        params: baseParams,
+      };
+      const { data } = await this._makeRequest(config);
+      let {
+        ids,
+        next_cursor: nextCursor,
+        previous_cursor: prevCursor,
+      } = data;
+      while (nextCursor !== 0 || prevCursor === 0) {
```
tjk, line 212: why the prevCursor check here?

tjk: or for first iteration...

tjk (approved):
looks great!

PR #705, opened 2020-12-18
Fix missing parenthesis in code example

bell-steven:
Copying this code block will lead to a syntax error because of the missing parenthesis.


dylburger:
Good catch, and we very much appreciate the contribution @bell-steven .

PR #708, opened 2020-12-18
Pipefy

michelle0927:
Created event sources for the triggers listed, and published Create Card and Update Card Field actions in Pipedream.

### DIFF
--- a/components/pipefy/pipefy.app.js
+++ b/components/pipefy/pipefy.app.js
@@ -1,0 +1,63 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "pipefy",
+  methods: {
+    _getBaseUrl() {
+      return "https://app.pipefy.com";
+    },
+    _getHeaders() {
+      return {
+        Accept: "application/json",
+        Authorization: `Bearer ${this.$auth.token}`,
+      };
+    },
+    async _makeRequest(data, endpoint) {
+      const config = {
+        method: "POST",
+        url: `${this._getBaseUrl()}/${endpoint}`,
+        headers: this._getHeaders(),
+        data,
+      };
+      return (await axios(config)).data;
+    },
+    async _makeQueriesRequest(data = null) {
+      return await this._makeRequest(data, "queries");
+    },
+    async _makeGraphQlRequest(data = null) {
+      return await this._makeRequest(data, "graphql");
+    },
+    async createHook({ pipe_id, name, email, url, actions }) {
+      const data = {
+        query: `mutation { createWebhook(input: { pipe_id: ${pipe_id} name: "${name}" email: "${email}" url: "${url}" actions: ["${actions}"] headers: \"{\\\"token\\\": \\\"${this.$auth.token}\\\"}\" }) { webhook { id name } } }`,
+      };
+      return (await this._makeQueriesRequest(data)).data.createWebhook;
+    },
+    async deleteHook(hookId) {
+      const data = {
+        query: `mutation { deleteWebhook(input: { id: ${hookId} }) { success } }`,
+      };
+      await this._makeQueriesRequest(data);
+    },
+    async getMe() {
+      const data = {
+        query:
+          "{ me { id name username email avatar_url created_at locale time_zone } }",
+      };
+      return (await this._makeQueriesRequest(data)).data.me;
+    },
+    async getCard(id) {
+      const data = {
+        query: `{ card(id: ${id}) { id title url comments{text} createdAt createdBy{name} due_date } }`,
+      };
+      return await this._makeGraphQlRequest(data);
+    },
+    async listCards(pipe_id) {
+      const data = {
+        query: `{ cards(pipe_id: ${pipe_id}, first: 100) { edges { node { id title assignees { id } comments { text } comments_count current_phase { name id } done late expired due_date fields { name value } labels { name } phases_history { phase { name } firstTimeIn lastTimeOut } url } } } }`,
+      };
+      return (await this._makeQueriesRequest(data)).data.cards;
+    },
+  },
+};
--- a/components/pipefy/pipefy.app.js
+++ b/components/pipefy/pipefy.app.js
@@ -11,6 +11,7 @@ module.exports = {
       return {
         Accept: "application/json",
         Authorization: `Bearer ${this.$auth.token}`,
+        "User-Agent": "@PipedreamHQ/pipedream v0.1",
       };
     },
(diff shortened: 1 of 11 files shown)


components/pipefy/pipefy.app.js
```diff
@@ -0,0 +1,63 @@
...
+    },
+    async _makeQueriesRequest(data = null) {
+      return await this._makeRequest(data, "queries");
+    },
+    async _makeGraphQlRequest(data = null) {
+      return await this._makeRequest(data, "graphql");
+    },
+    async createHook({ pipe_id, name, email, url, actions }) {
+      const data = {
+        query: `mutation { createWebhook(input: { pipe_id: ${pipe_id} name: "${name}" email: "${email}" url: "${url}" actions: ["${actions}"] headers: \"{\\\"token\\\": \\\"${this.$auth.token}\\\"}\" }) { webhook { id name } } }`,
```
jverce, line 33: Can we re-indent this so that it can be easily modified in the future if necessary?

```js
const query = `
  mutation {
    createWebhook(input: {
      pipe_id: ${pipe_id}
      name: "${name}"
      email: "${email}"
      url: "${url}"
      actions: [${actions.map(i => `"${i}"`).join(", ")}]
      headers: "{\\\"token\\\": \\\"${this.$auth.token}\\\"}\"
    }
    {
      webhook {
        id
        name
      }
    }
  }
`;
```

jverce: Same for all the GraphQL queries below

jverce: Also, looks like the `email` field is not necessary, and we should remove it so that:
- We handle as little private data as necessary
- The user doesn't get bombarded with e-mails that they didn't request

jverce (changes requested):
As I mentioned before in previous PR's, it's probably a good idea to encapsulate common code into separate modules so that it can be reused and easily maintained. In this case, it looks like there's some common code related to:
- Some props (the `.app.js` module, `db`, `http` for instant events, `timer` for timer-based, etc)
- The hooks (except the `actions` property which can be abstracted into a `getActions` method)
- The webhook call validation in the `run()` method

There's a [fairly short](https://github.com/PipedreamHQ/pipedream/tree/8650984b1a5576231501eb71106dd4c9283b8140/components/webflow) example of what I'm talking about, and we can keep discussing if needed.

components/pipefy/sources/common-polling.js
```diff
@@ -0,0 +1,28 @@
...
+    const cards = await this.pipefy.listCards(this.pipeId);
+    for (const edge of cards.edges) {
+      const { node } = edge;
+      const { due_date } = node;
+      const due = due_date ? new Date(due_date) : new Date();
+      if (!this.isCardRelevant(node, due)) continue;
+      this.$emit(node, {
+        id: this.getEmitId(node),
+        summary: node.title,
+        ts: due.getTime(),
```
jverce, line 24: I don't think this is correct. `ts` is supposed to be the timestamp of the event itself, but in this case the timestamp of the event (i.e. the time at which an event was emitted) could have a different meaning depending on the content of the card and use case.

I think in this case the logic from `card-overdue` leaked itself out of its own module and into this common module (which is supposed to only contain logic related to polling the API for new records and processing them in an abstract way).

If we do something like this instead, then we can refine the behaviour for each use case:
```js
// common-polling.js
async run() {
  const cards = await this.pipefy.listCards(this.pipeId);
  for (const edge of cards.edges) {
    const { node } = edge;
    if (!this.isCardRelevant({ node, event })) continue;

    const meta = this.getMeta(node);
    this.$emit(node, meta);
  }
}
```

Then each event source would implement the specific logic. In the case of `card-overdue`, the due date would be in fact the timestamp (assuming that's the expected behaviour):
```js
// card-overdue.js
module.exports = {
  methods: {
    isCardRelevant({ node, event }) {
      // We take `event` as input since this is the
      // actual monotonic timestamp that should be used as reference.
      // This will avoid calling Date.now(), which will make the
      // event source deterministic.
      const { timestamp: eventTimestamp } = event;

      // To match Javascript's timestamp, since in Javascript epochs
      // are expressed in milliseconds instead of seconds
      const eventDate = eventTimestamp * 1000;
      
      const { due_date: dueDateIso } = node;
      const dueDate = Date.parse(dueDateIso);
      
      return (
        dueDate < eventDate &&
        !node.done
      );
    },
    getMeta(node) {
      const {
        id,
        title: summary,
        due_date: dueDate,
      } = node;
      const ts = Date.parse(dueDate) || Date.now();
      return {
        id,
        summary,
        ts,
      };
    },
  },
};
```

The other cases are much simpler I think, for example:
```js
// card-expired.js
module.exports = {
  methods: {
    isCardRelevant({ node }) {
      return (
        node.expired &&
        !node.done
      );
    },
    getMeta(node) {
      const {
        id: nodeId,
        title: summary,
        current_phase: { id: currentPhaseId },
      } = node;
      const id = `${nodeId}${currentPhaseId}`;
      const ts = Date.now();
      return {
        id,
        summary,
        ts,
      };
    },
  },
}; 
```

components/pipefy/sources/common-webhook.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!this.verifyEvent(event)) return;
+
+    const cardData = event.body.data;
+    const cardId = cardData.card.id;
+    const card = (await this.pipefy.getCard(cardId)).data.card;
+    const { body, id, summary } = this.getMeta(card, cardData);
+    this.$emit(body, {
+      id,
+      summary,
+      ts: Date.now(),
```
jverce, line 47: **Minor/question**
Is it always the case with every event source that the timestamp is not provided and we'd have to use `Date.now()`?

components/pipefy/sources/card-created/card-created.js
```diff
@@ -0,0 +1,22 @@
...
+  description: "Emits an event for each new card created in a Pipe.",
+  version: "0.0.1",
+  methods: {
+    ...common.methods,
+    getActions() {
+      return ["card.create"];
+    },
+    getMeta(card, cardData) {
+      return {
+        body: card,
```
jverce, line 16: Question: is `body` supposed to be equal to `cardData` and there was a typo here? Or is `cardData` never actually used?

jverce (changes requested):
Overall LGTM.
There is one comment in `common-polling.js` that I think we should address since it can cause incorrect/unexpected behaviour.

jverce (approved):

Issue #636, opened 2020-11-18
[SOURCE] Snowflake


todsac:
**Describe the event source. What app, what event?**
Snowflake

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
New Row | Triggers when a new row is added in a table in Snowflake (i.e., query snowflake and emit results as a single event or each row as an individual event).

dylburger:
Additionally, I'd like a source that accepts a SQL query, runs on a schedule, and emits the results of the query. For this source, I'd like the option to emit the results as a single event, or as individual records, similar to https://github.com/PipedreamHQ/pipedream/blob/master/components/pipedream/sources/new-records-from-sql-query/sql.js

PR #721, opened 2020-12-22
Add event source logic for Snowflake New Row

jverce:
This PR is in support of issue #636

### DIFF
--- a/components/snowflake/snowflake.app.js
+++ b/components/snowflake/snowflake.app.js
@@ -44,9 +44,13 @@ module.exports = {
       return this.collectRows({ sqlText });
     },
     async listFieldsForTable(tableName) {
-      const sqlText = `DESCRIBE TABLE ${tableName}`;
+      const sqlText = "DESCRIBE TABLE IDENTIFIER(:1)";
+      const binds = [
+        tableName,
+      ];
       const statement = {
         sqlText,
+        binds,
       };
       return this.collectRows(statement);
     },
--- a/components/snowflake/snowflake.app.js
+++ b/components/snowflake/snowflake.app.js
@@ -1,0 +1,54 @@
+const { promisify } = require("util");
+const snowflake = require("snowflake-sdk");
+
+module.exports = {
+  app: "snowflake",
+  type: "app",
+  methods: {
+    async _getConnection() {
+      if (this.connection) {
+        return this.connection;
+      }
+
+      this.connection = snowflake.createConnection(this.$auth);
+      await promisify(this.connection.connect)();
+      return this.connection;
+    },
+    async getRows(statement) {
+      const connection = await this._getConnection();
+      const executedStatement = connection.execute(statement);
+      return executedStatement.streamRows();
+    },
+    async collectRows(statement) {
+      const rowStream = await this.getRows(statement);
+      const rows = [];
+      for await (const row of rowStream) {
+        rows.push(row);
+      }
+      return rows;
+    },
+    async *collectRowsPaginated(statement, pageSize = 1) {
+      const rowStream = await this.getRows(statement);
+      let rows = [];
+      for await (const row of rowStream) {
+        rows.push(row);
+        if (rows.length === pageSize) {
+          yield rows;
+          rows = [];
+        }
+      }
+      yield rows;
+    },
+    async listTables() {
+      const sqlText = "SHOW TABLES";
+      return this.collectRows({ sqlText });
+    },
+    async listFieldsForTable(tableName) {
+      const sqlText = `DESCRIBE TABLE ${tableName}`;
+      const statement = {
+        sqlText,
+      };
+      return this.collectRows(statement);
+    },
+  },
+};
--- a/components/snowflake/snowflake.app.js
+++ b/components/snowflake/snowflake.app.js
@@ -11,7 +11,7 @@ module.exports = {
       }
 
       this.connection = snowflake.createConnection(this.$auth);
-      await promisify(this.connection.connect)();
+      await promisify(this.connection.connect).bind(this.connection)();
       return this.connection;
     },
     async getRows(statement) {
--- a/components/snowflake/sources/common-table-scan.js
+++ b/components/snowflake/sources/common-table-scan.js
@@ -1,3 +1,4 @@
+const isString = require("lodash/isString");
 const snowflake = require("../snowflake.app");
 
 module.exports = {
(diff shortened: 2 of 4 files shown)


components/snowflake/sources/new-row/README.md
```diff
@@ -0,0 +1,43 @@
...
+higher index than previous records, hence it is necessary for the target table
+to be indexed by a monotonically increasing column, or at least have a column
+with such a property so that this component can detect new rows.
+
+## Prerequisites
+
+To set up your Snowflake credentials, go to the [Pipedream
+Apps](https://pipedream.com/apps) page, and click on the **CONNECT AN APP**
+button. Look for the **Snowflake** app, and click on it. Fill in every field
+required in the, and click on **SAVE**.
```
tjk, line 16: word missing here

components/snowflake/sources/new-row/README.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Usage
+
+1. Visit the [Pipedream Sources](https://pipedream.com/sources) page and click
```
tjk, line 20: Can just link to `https://pipedream.com/sources/new?key=snowflake-new-row` ?

components/snowflake/sources/new-row/new-row.js
```diff
@@ -0,0 +1,54 @@
...
+        ts,
+      };
+    },
+    generateMetaForCollection(data) {
+      const {
+        lastResultId: id,
+        rowCount,
+        timestamp: ts,
+      } = data;
+      const summary = `${rowCount} new rows`;
```
tjk, line 30:
```javascript
const summary = `${rowCount} new row${rowCount === 1 ? "" : "s"}`;
```

components/snowflake/sources/new-row/new-row.js
```diff
@@ -0,0 +1,54 @@
...
+        id,
+        summary,
+        ts,
+      };
+    },
+    async processEvent(lastResultId, event) {
+      const { timestamp } = event;
+      const sqlText = `
+        SELECT *
+        FROM ${this.tableName}
```
tjk, line 41: this is quite spooky without prepared statements... is there no lightweight node lib that does it? (if snowflake sdk is causing issues with this)

components/snowflake/sources/common-table-scan.js
```diff
@@ -0,0 +1,142 @@
...
+      default: {
+        intervalSeconds: 60 * 15, // 15 minutes
+      },
+    },
+    tableName: {
+      type: "string",
+      label: "Table Name",
+      description: "The name of the table to watch for new rows",
+      async options(context) {
+        const { page } = context;
```
tjk, line 19: I wonder if it's worth trying to be clever here... like making `context` a proxy object and seeing if various fields are accessed and in those cases disable pagination, etc. it's maybe a little too much magic but i also hate to see this repeated so often!

components/snowflake/sources/common-table-scan.js
```diff
@@ -0,0 +1,142 @@
...
+      label: "Event Size",
+      description: "The number of rows to include in a single event (by default, emits 1 event per row)",
+      default: 1,
+      min: 1,
+    },
+  },
+  hooks: {
+    async activate() {
+      let lastResultId = this.db.get("lastResultId");
+      if (isNil(lastResultId)) {
```
tjk, line 54: this is just `lastResultId == null` right?

tjk: we should try to minimize deps especially in cases like these

components/snowflake/snowflake.app.js
```diff
@@ -0,0 +1,54 @@
...
+  app: "snowflake",
+  type: "app",
+  methods: {
+    async _getConnection() {
+      if (this.connection) {
+        return this.connection;
+      }
+
+      this.connection = snowflake.createConnection(this.$auth);
+      await promisify(this.connection.connect)();
```
tjk, line 14: I'm sure you've tested this but kind of surprised connect doesn't use `this`... (as in no bind required)

tjk (approved):
A couple very minor comments. I think it would be good to use something to do prepared statements... or at the very least aggressive validation on `tableName`, `uniqueKey` props in `activate`? Otherwise looks great!

components/snowflake/sources/new-row/README.md
jverce: Oops... fixed

components/snowflake/sources/new-row/README.md
jverce: Nice, fixed

components/snowflake/sources/new-row/new-row.js
jverce: Done

components/snowflake/sources/common-table-scan.js
jverce: `null` or `undefined`, but nevertheless I think this was kept by error.
I'll change it to a comparison against `undefined` and remove the dependency.

components/snowflake/snowflake.app.js
jverce: Hmm that's actually a good point. I looked at the code and everything indicates that this shouldn't work 😆 
I'll bind it to be safe.

components/snowflake/sources/common-table-scan.js
jverce: I'm on you on that. Maybe an extra/optional parameter on the prop itself?
```js
props: {
  someProp: {
    type: 'string',
    label: 'Some Prop',
    paginateOptions: false,  // Set to `true` by default to be backwards compatible
    async options(context) {
      const { page } = context;
      assert(page === 0);  // Always on the first page
    },
  },
},
```

components/snowflake/sources/new-row/new-row.js
jverce: It's very tricky to implement on the client side, especially when dealing with custom types for example.
I don't like this either TBH, as it doesn't handle all data types (`VARCHAR` for example will break this).
I will spend some more time looking into possible alternatives.

jverce: OK, I pushed a change that makes me feel more comfortable since it uses prepared statements.
To make it work I additionally do this:
- Use the [`IDENTIFIER`](https://docs.snowflake.com/en/sql-reference/identifier-literal.html) function to refer to the table using a prepared statement
- Validate a column name against the existing columns in a table (requires an extra API call in each execution)

PR #732, opened 2020-12-26
Async code

dylburger:


docs/docs/workflows/steps/code/async/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## The problem
+
+**Any asynchronous code within a Node.js code step must complete before the next step runs**. This ensures future steps have access to its data. If Pipedream detects that code is still running by the time the step completes, you'll see the following warning below the code step:
+
+> **This step was still trying to run code when the step ended. Make sure you await all Promises, or promisify callback functions.**
+
+As the warning notes, this often arises from one of two issues:
+
+- You forgot to `await` a Promise. [All Promises must be awaited](#await-all-promises) so they run synchronously.
```
tjk, line 15: In both these cases I would say it's not about them running synchronously (because they are still async) but more about just ensuring the wrapping step function properly awaits their resolution.

docs/docs/workflows/steps/code/async/README.md
```diff
@@ -0,0 +1,116 @@
...
+```javascript
+file.on("finish", () => {
+  console.log(fs.statSync(this.fileName));
+});
+```
+
+run:
+
+```javascript
+await new Promise((resolve, reject) => {
```
tjk, line 95: (This is a bit more parallel so that what happens in the callback now happens after awaiting the promise.)

```javascript
// would be good to listen for errors inside promise callback and reject(err)
await new Promise(resolve => file.on("finish", resolve))
const stats = fs.statSync(this.filePath);
```

PR #746, opened 2020-12-31
ActiveCampaign

michelle0927:


components/activecampaign/sources/contact-added-to-list/contact-added-to-list.js
```diff
@@ -0,0 +1,71 @@
...
+        const list = (await this.activecampaign.getList(contactList.list)).list;
+        this.$emit(list, {
+          id: `${contact.id}${list.id}`,
+          summary: `${contact.email} added to ${list.name}`,
+          ts: Date.now(),
+        });
+      }
+      this.db.set(contact.id, contactListIds);
+    }
+  },
```
jverce, line 70: I believe the `subscribe` webhook fits this exact same purpose?
At least that's what I observed by creating the same trigger in Zapier.
This will simplify this a lot, since we don't need to scan the entire contact database, their contact lists, and so on.

components/activecampaign/sources/new-contact-task/new-contact-task.js
```diff
@@ -0,0 +1,48 @@
...
+          id: task.id,
+          summary: task.title,
+          ts: Date.now(),
+        });
+      }
+      total = tasks.length;
+      offset += limit;
+    }
+    this.db.set("largestPreviousId", largestId);
+  },
```
jverce, line 47: Doesn't the `contact_task_add` event webhook do this already?

If not, is there a way we could avoid scanning the entire list of tasks? From the docs, I see that the tasks are ordered by ID in ascending order.
I'd imagine that grows over time which will make each execution slower and slower. Maybe caching the `offset`?

jverce (changes requested):
Some general comments/questions/suggestions:
1. Why do we have event sources for specific events if `sources/new-event` can already achieve the same thing? Is it so that we users can provide an additional prop to filter the events?
2. Any chance we could abstract repeated logic into a common module? `activate`/`deactivate` hooks and event processing follow a pattern that I think we can abstract away.

components/activecampaign/sources/new-event/new-event.js
```diff
@@ -0,0 +1,40 @@
...
+module.exports = {
+  name: "New Event (Instant)",
+  key: "activecampaign-new-event",
+  description:
+    "Emits an event for the specified event type from ActiveCampaign.",
+  version: "0.0.1",
+  props: {
+    activecampaign,
+    db: "$.service.db",
+    hhttp: "$.interface.http",
```
jverce, line 12: `s/hhttp/http/`

components/activecampaign/activecampaign.app.js
```diff
@@ -0,0 +1,181 @@
...
+      return { "Api-Token": this.$auth.api_key };
+    },
+    async createHook(events, url, sources) {
+      const config = {
+        method: "POST",
+        url: `${this.$auth.base_url}/api/3/webhooks`,
+        headers: this._getHeaders(),
+        data: {
+          webhook: {
+            name: "Pipedream Hook",
```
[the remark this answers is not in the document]
michelle0927: Updated. However, the ActiveCampagin dashboard does not display webhooks created via API. It only displays webhooks created within the dashboard.

components/activecampaign/sources/campaign-bounce/campaign-bounce.js
```diff
@@ -0,0 +1,47 @@
...
+      );
+      this.db.set("hookId", hookData.webhook.id);
+    },
+    async deactivate() {
+      await this.activecampaign.deleteHook(this.db.get("hookId"));
+    },
+  },
+  async run(event) {
+    const { body } = event;
+    if (!body) {
```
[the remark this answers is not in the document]
michelle0927: Probably not if the request is coming from ActiveCampaign. If it's coming from somewhere else, I think it could. I believe that's why this is often included in Pipedream's webhook event sources.

components/activecampaign/sources/common-webhook.js
```diff
@@ -0,0 +1,49 @@
...
+
+    if (!this.isRelevant(body)) return;
+
+    const dateTime = new Date(body.date_time);
+    const { id, summary } = await this.getMeta(body);
+    this.$emit(body, {
+      id,
+      summary,
+      ts: dateTime.getTime(),
+    });
```
jverce, line 47: Minor: why not have the `getMeta` method return the `ts` as well, since it's part of the metadata?

In the event source...
```js
module.exports = {
  methods: {
    getMeta(body) {
      // Extract ID and summary...
      const { date_time: dateTimeIso } = body;
      const ts = Date.parse(dateTimeIso);
      return {
        id,
        summary,
        ts,
      };
    };
  },
};
```

Then in `common-webhook`:
```js
module.exports = {
  async run(event) {
    // Initial checks...
    const meta = await this.getMeta(body);
    this.$emit(body, meta);
  },
};
```

I see the potentially duplicated code in the different `getMeta` implementations, but that's only accidental, and removing that duplicate logic introduces other downsides, like having different responsibilities of `getMeta` depending on the event source we're looking at (an example is `new-automation-webhook`, which in fact returns the complete metadata information of the event).

Again, this is a **minor** comment and I'm not blocking the PR because of this, just food for though 🙂

jverce (approved):

PR #773, opened 2021-01-08
Fixed object destructuring syntax

bloycey:
Syntax should be `const { data } = resp;` rather than `const { data } = resp.data`


dylburger:
@bloycey thank you very much for the patch!

PR #785, opened 2021-01-12
Google Sheets component refactor

jverce:
* Consolidate Google Drive specific functionality in the original
  `google_drive.app.js` file
* Extract commonly-used functionality specific to Google Sheets in
  `sources/common.js` file to be easily reused by other components
* Fix bugs with event source:
  * ~For new rows, detect when rows are deleted in a spreadsheet so that we don't miss new rows added in such places~
  * For updated rows, make sure we're not pulling the contents of the whole sheet file, but rather do it for the relevant worksheets
* Rename some methods to avoid name clashing
* Remove unused/duplicated code and make minor cosmetic changes


components/google_drive/google_drive.app.js
```diff
@@ -95,9 +95,62 @@ module.exports = {
...
+      if (incomingChannelID !== channelID) {
+        console.log(
+          `Channel ID of ${incomingChannelID} not equal to deployed component channel of ${channelID}`
+        );
+        return false;
+      }
+
+      if (headers["x-goog-resource-id"] !== subscription.resourceId) {
+        console.log(
+          `Resource ID of ${resourceId} not currently being tracked. Exiting`
```
dylburger, line 128: `subscription.resourceId`? I don't think `resourceId` is in scope here.

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -165,54 +96,55 @@ module.exports = {
...
-      );
-      let rowCount = worksheet.data[0].rowData
+
+      const oldRowCount = this.db.get(`${sheetId}${worksheetId}`);
+      const rowCount = worksheet.data[0].rowData
         ? worksheet.data[0].rowData.length
         : 0;
+      this.db.set(`${sheetId}${worksheetId}`, rowCount);
+
       if (rowCount <= oldRowCount) continue;
```
dylburger, line 131: Re: "For new rows, detect when rows are deleted in a spreadsheet so that we don't miss new rows added in such places", I believe this will cause the first new row after a deletion to fail if the changes come as a pair (deletion and addition together), since the new row count will be less than or equal to the old. I tested this and saw this behavior:

- Delete two existing rows and add a new row (didn't emit since `rowCount` was less than `oldRowCount`)
- Added another row (correctly emitted)

I could be reading incorrectly but lmk if you're seeing the same behavior.

components/google_drive/google_drive.app.js
jverce: Good catch

components/google_sheets/sources/new-row-added/new-row-added.js
jverce: Yes, that is definitely a possibility. Worst-case scenario, we have the same behaviour as before: we won't detect new rows if they were added in replacement of previous ones. Best-case scenario, we get notified right away by the subscription when the rows were deleted, the event source "prepares" itself and then is able to detect the newly added rows.

The way the notifications work I don't think it's possible to overcome this without keeping a complete copy of the spreadsheet within the event source storage, just like we do with `google_sheets-new-updates`.

dylburger: Sounds good, let's punt.

michelle0927: If the data in a row is removed, Google Sheets considers the row changed, not deleted. Rows are only ever added to a sheet, not deleted. So you're right that if adding data to an existing row that previously had no data is considered adding a new row, we'll have to save and compare a complete copy of each sheet being watched. We might want to consider what a "new row" should be. For instance, if data is added to an empty row in the middle or top of a sheet, is that a new row? Or should it only be when adding data to a row below the last line that has data in it? "New row" could mean a variety of things here. I think it's just a matter of preference.

jverce: That's a good point. If we add a row in the middle of the spreadsheet, we would detect that a new row was added, would not emit the proper event.

I think this is a big enough caveat for us to work on a `README.md` file. WDYT?

dylburger: How does Zapier handle these cases?

jverce: Based on tests on my side, it handles them like we currently do in production.
I'm reverting the behaviour back to that.

michelle0927 (approved):
Looks good!

Issue #768, opened 2021-01-07
[BUG] Google Sheets New Updates source times out on creation for certain sheets


dylburger:
The **New Updates** Google Sheets event source times out when the source is created under certain conditions. We need to simplify what happens in the `activated()` hook and parallelize the HTTP requests we make to fetch the initial state of the sheet.

jverce:
PR submitted for review: #787

PR #787, opened 2021-01-12
Reduce activation time of Google Sheets New Updates event source

jverce:
# Description
This makes a few optimisations to address [this bug](https://github.com/PipedreamHQ/pipedream/issues/768).
~It depends on PR #785~

# Changelog
* When retrieving spreadsheet data from Google Sheets, make all calls
  in parallel
* Only retrieve those worksheets that were selected by the user
* Use a set instead of a list to check if a worksheet was selected by
  the user or not
* Fix bug that shows every worksheet from a spreadsheet, even if they
  are data sources


components/google_sheets/sources/common.js
```diff
@@ -0,0 +1,104 @@
...
+      // We use and increment the pageToken as new changes arrive, in run()
+      this.db.set("pageToken", startPageToken);
+
+      // Save metadata on the subscription so we can stop / renew later
+      // Subscriptions are tied to Google's resourceID, "an opaque value that
+      // identifies the watched resource". This value is included in request headers
+      this.db.set("subscription", { resourceId, expiration });
+      this.db.set("channelID", channelID);
+
+      await this.takeSheetSnapshot();
```
dylburger, line 45: Do you think this is worth moving to the `run` method (i.e. the first time the component runs)? I noticed the new updates source still took some time to run the `activate()` hook and I'm concerned even with the modified code, it might time out for the user.

jverce: Yes, it's worth doing since the improvements don't really make any difference if the user selects every worksheet in a spreadsheet.

jverce:
A high percentage of the lines in the diff are due to moving around some code, so I recommend hiding the whitespaces when reviewing the code:

hide_whitespaces

components/google_sheets/google_sheets.app.js
```diff
@@ -18,16 +18,13 @@ module.exports = {
...
-            sheetId: value,
-          } = sheet.properties;
-          return {
-            label,
-            value,
-          };
-        });
+        return sheets
+          .map(({ properties }) => properties)
+          .filter(({ sheetType }) => sheetType === 'GRID')
```
dylburger, line 23: Probably worth a comment here on what the "GRID" sheets actually mean (is there a link to docs?)

components/google_sheets/sources/common.js
jverce: Done!

components/google_sheets/google_sheets.app.js
jverce: Good idea. Done.

components/google_sheets/sources/common.js
```diff
@@ -42,7 +42,7 @@ module.exports = {
       this.db.set("subscription", { resourceId, expiration });
       this.db.set("channelID", channelID);
 
-      await this.takeSheetSnapshot();
+      this.db.set("isInitialized", false);
```
michelle0927, line 45: On the first run, how do we know what changes have been made to a file if the content hasn't been initialized?

components/google_sheets/sources/common.js
```diff
@@ -97,8 +99,26 @@ module.exports = {
     getWorksheetIds() {
```
michelle0927, line 99: Since getWorksheetIds and getSheetId are the same in both event sources, couldn't they be removed and just use the methods in common.js if we changed the content of getWorksheetIds to "return this.worksheetIDs;" and getSheetId to "return this.sheetID;" Also, the props sheetId and worksheetId are the same in both components, so couldn't these be moved to common.js?

components/google_sheets/sources/new-row-added/new-row-added.js
```diff
@@ -63,90 +63,90 @@ module.exports = {
         );
       }
     },
-  },
-  async run(event) {
-    let subscription = this.db.get("subscription");
-    let channelID = this.db.get("channelID");
-    let pageToken = this.db.get("pageToken");
+    async processEvent(event) {
```
michelle0927, line 66: Much of the processEvent method looks the same in both components. Couldn't the matching code be put in common.js?

michelle0927 (commented):
Looks good. Just a couple questions/comments.

components/google_sheets/sources/common.js
```diff
@@ -97,8 +99,26 @@ module.exports = {
...
   },
+  async run(event) {
+    const isInitialized = this.db.get("isInitialized");
+    if (!isInitialized) {
+      await this.takeSheetSnapshot();
+      this.db.set("isInitialized", true);
+    }
+
+    const { headers } = event;
+    if (headers["x-goog-resource-state"] === "sync") {
```
dylburger, line 117: If the component is invoked by a timer, `headers` won't be defined and this will throw.

components/google_sheets/sources/common.js
jverce: On the first execution of `run()`, the flag will be false and it will kick in the initialization.
That's actually one of the fixes to avoid timing out during `activate()`. We basically put that burden on the `run()` method instead, which is OK if it takes a long time.

components/google_sheets/sources/common.js
jverce: The idea behind that is because I want to use `common.js` in another event source (i.e. Google Forms) which has a different way of providing this data.

In that component (ideally, it's still being discussed) we wouldn't have those props (i.e. `sheetID` and `worksheetIDs`), but rather a `formId` prop through which we would figure out the other ones. For that reason, we can't assume that `sheetID` and `worksheetIDs` are props that will always be available.

components/google_sheets/sources/new-row-added/new-row-added.js
jverce: Done. Wanted to keep the scope to the bug only, but it's a good opportunity to do it while we're in that mental context given all the refactoring we've been doing 🙂

components/google_sheets/sources/common.js
jverce: Fixed.

michelle0927 (approved):

PR #831, opened 2021-01-21
Twist

michelle0927:


components/twist/twist.app.js
```diff
@@ -0,0 +1,128 @@
...
+      },
+    },
+    channel: {
+      type: "string",
+      label: "Channel",
+      description: "The channel to watch for new events.",
+      optional: true,
+      async options(opts) {
+        if (!opts.workspace)
+          return [{ label: "No Channels Found", value: "none" }];
```
jverce, line 30: This is an interesting edge case, where a prop that depends on another one and the prop it depends on is missing or empty.

@dylburger food for thought

jverce (approved):

Issue #803, opened 2021-01-15
[Sources and Actions] TextLocal


todsac:
**Describe the event source. What app, what event?**
TextLocal

### Triggers
- New Contact | Triggers when a new contact is added.
- New Inbox Message | Triggers when a message is received into an inbox.
- New Sent API Message | Triggers when a message is sent via Textlocal's API.

### Actions
- Create Contact | Creates a new contact.
- Send SMS | Sends a SMS message.

jverce:
- ~New Contact | Triggers when a new contact is added.~ Done
- New Inbox Message | Triggers when a message is received into an inbox.:
  - In order to exercise this, we need an inbox (which TextLocal does not sell to non-UK end customers)
- ~New Sent API Message | Triggers when a message is sent via Textlocal's API.~ Done

dannyroosevelt:
@lcaresia are you working on this one?

dannyroosevelt:
This is ready for release!

PR #839, opened 2021-01-23
TextLocal New Message Sent from API

jverce:
* Add app logic for TextLocal event sources
* Add event source logic for New Sent API Message

Tracking issue: #803


jverce:
Moved to #852

PR #852, opened 2021-01-27
TextLocal Event Sources

jverce:
* New Message Sent from API
* Add event source logic for New Sent API Message
* Add event source logic for New Contact

Issue: #803

### DIFF
--- a/components/textlocal/sources/new-contact/new-contact.js
+++ b/components/textlocal/sources/new-contact/new-contact.js
@@ -1,0 +1,99 @@
+const common = require("../common/timer-based");
+
+module.exports = {
+  ...common,
+  key: "textlocal-new-contact",
+  name: "New Contact",
+  version: "0.0.1",
+  dedupe: "unique",
+  props: {
+    ...common.props,
+    groupId: {
+      type: "string",
+      label: "Contact Group",
+      description: "The contact group to monitor for new contacts",
+      async options(context) {
+        const { page } = context;
+        if (page !== 0) {
+          return [];
+        }
+
+        const { groups } = await this.textlocal.getContactGroups();
+        const options = groups.map((group) => ({
+          label: group.name,
+          value: group.id,
+        }));
+        return {
+          options,
+        };
+      },
+    },
+  },
+  methods: {
+    ...common.methods,
+    _isContactProcessed(contact) {
+      const { number } = contact;
+      return this.db.get(number) || false;
+    },
+    _markContactAsProcessed(contact) {
+      const { number } = contact;
+      this.db.set(number, true);
+    },
+    async takeContactGroupSnapshot() {
+      const contactScan = await this.textlocal.scanContactGroup({
+        groupId: this.groupId,
+      });
+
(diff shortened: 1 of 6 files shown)


components/textlocal/sources/new-contact/new-contact.js
```diff
@@ -0,0 +1,99 @@
...
+        summary,
+        ts,
+      };
+    },
+    async processEvent(event) {
+      const isInitialized = this.db.get("isInitialized");
+      if (!isInitialized) {
+        await this.takeContactGroupSnapshot();
+        this.db.set("isInitialized", true);
+        return;
```
dylburger, line 77: If this is the first time the source runs, and it isn't initialized, is there a reason why you `return` early? That is, on the first run, is there still real data that should be processed and emitted?

components/textlocal/sources/new-sent-api-message/new-sent-api-message.js
```diff
@@ -0,0 +1,66 @@
+const common = require("../common/timer-based");
+
+module.exports = {
+...common,
+  key: "textlocal-new-sent-api-message",
+  name: "New Sent API Message",
+  version: "0.0.2",
```
dylburger, line 7: Has this component been published publicly before?

dylburger (approved):
Logic looks great. Generally, I'd love to see more links to API docs / other interesting things you learned in the comments of these sources. You're digesting so much information and I'd love to see some of the "why" behind these methods.

For example, since I'm unfamiliar with TextLocal, I'm unsure what kind of data is expected by the `getMaskedName` or `getMaskedNumber` methods. What is a masked name / number? Those kinds of comments would be helpful for the eng team here and users at large.

components/textlocal/sources/new-contact/new-contact.js
jverce: I wanted to take the same approach we took with Google Sheets to avoid timing out during `activate()`.
I could take the snapshot during the activation hook but that could cause a timeout in case of big contact lists.

So the first time this event is triggered, we'd take the intial snapshot. At that point, we don't know which contacts were created between during the time the event source was activated and the time it was triggered for the first time (since the TextLocal API does not sort the results or provide timestamps for us to tell which contacts are new vs old). The only way to tell whether a contact is new or not is to check if it's currently present in the contact list but is not in the DB.

This means that the first execution of the event source must be "sacrificed" to collect information about the complete contact list.

I will add a `README.md` to this component so that users are aware of this.

components/textlocal/sources/new-sent-api-message/new-sent-api-message.js
jverce: No, this is my bad. Since I created a PR for this particular component and this was a follow-up PR, I bumped the version. But that PR was never merged to `master` so this should be `0.0.1`. I will fix it.

jverce:
I agree with adding comments/documentation. One approach I can take moving forward is to add some docstring to utility methods that would make their purpose clearer.

PR #869, opened 2021-01-30
Mercury new-transaction

michelle0927:


components/mercury/mercury.app.js
```diff
@@ -0,0 +1,40 @@
...
+    _getBaseURL() {
+      return "https://backend.mercury.com/api/v1";
+    },
+    _getHeaders() {
+      return {
+        Authorization: `Bearer ${this.$auth.api_key}`,
+        "Content-Type": "application/json",
+      };
+    },
+    async _makeGetRequest(endpoint, params = null) {
```
dylburger, line 16: Can you abstract this method to accept the HTTP method and rename `_makeRequest`? That way it'll support `POST` requests, etc. naturally as we evolve the app file.

components/mercury/mercury.app.js
```diff
@@ -0,0 +1,40 @@
...
+      const config = {
+        method: "GET",
+        url: `${this._getBaseURL()}${endpoint}`,
+        headers: this._getHeaders(),
+        params,
+      };
+      return (await axios(config)).data;
+    },
+    dayAgo() {
```
dylburger, line 25: Similar feedback: maybe rename to `daysAgo` and pass the the number of days as an arg? Or rename to `oneDayAgo()` to make the return value explicit?

components/mercury/mercury.app.js
```diff
@@ -0,0 +1,40 @@
...
+      const dayAgo = new Date();
+      dayAgo.setDate(dayAgo.getDate() - 1);
+      return dayAgo;
+    },
+    async getAccounts() {
+      return (await this._makeGetRequest("/accounts")).accounts;
+    },
+    async getTransactions(accountId, params, limit, offset) {
```
dylburger, line 33: `limit` and `offset` aren't used - did you mean to pass them?

components/mercury/sources/new-transaction/new-transaction.js
```diff
@@ -0,0 +1,61 @@
...
+  props: {
+    mercury,
+    db: "$.service.db",
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+    account: {
```
dylburger, line 18: Does it make sense to move this to the app file to support other sources in the future?

components/mercury/sources/new-transaction/new-transaction.js
```diff
@@ -0,0 +1,61 @@
...
+      type: "string",
+      label: "Account",
+      optional: false,
```
dylburger, line 21: This is the default, so you can remove

components/mercury/sources/new-transaction/new-transaction.js
```diff
@@ -0,0 +1,61 @@
...
+      async options() {
```
dylburger, line 22: Does this API support pagination?

components/mercury/sources/new-transaction/new-transaction.js
```diff
@@ -0,0 +1,61 @@
...
+      },
+    },
+  },
+  methods: {
+    getMeta(transaction) {
+      const { id, counterpartyName, postedAt } = transaction;
+      const postedAtDate = new Date(postedAt);
+      const ts = postedAtDate.getTime();
+      const summary = counterpartyName;
+      return { id, summary, ts };
```
dylburger, line 38: You can drop a couple of lines by doing something like:

```javascript
const { id, counterpartyName: summary, postedAt } = transaction;
const ts = (new Date(postedAt)).getTime();
return { id, summary, ts };
```

dylburger (changes requested):
Mostly small / style changes

components/mercury/sources/new-transaction/new-transaction.js
michelle0927: It does for listing transactions, but not for listing accounts. 
https://docs.mercury.com/reference#accounts

dylburger (approved):
lgtm!

PR #890, opened 2021-02-05
Procore

michelle0927:


components/procore/sources/common.js
```diff
@@ -0,0 +1,70 @@
...
+    async activate() {
+      const hook = await this.procore.createHook(
+        this.http.endpoint,
+        this.company,
+        this.project
+      );
+      this.db.set("hookId", hook.id);
+      // create hook triggers
+      eventTypes = this.getEventTypes();
+      resourceName = this.getResourceName();
```
jverce, line 29: Minor: we might want to add a default implementation of `getResourceName`, even if it's dummy or if child classes must implement it:
```js
module.exports = {
  methods: {
    getResourceName() {
      throw new Error("getResourceName is not implemented");
    },
  },
};
```

This mainly helps with readability, so it's a minor request at this point.

components/procore/sources/common.js
```diff
@@ -0,0 +1,70 @@
...
+    http: "$.interface.http",
+    company: { propDefinition: [procore, "company"] },
+    project: {
+      propDefinition: [procore, "project", (c) => ({ company: c.company })],
+    },
+  },
+  methods: {
+    getEventTypes() {
+      return ["create", "update", "delete"];
+    },
```
jverce, line 17: Minor: even better would be to use the `eventTypes` array in `procore.app.js`:
```js
module.exports = {
  methods: {
    getEventTypes() {
      return procore.eventTypes.map(({ value }) => value);
    };
  },
};
```

This way, any change we do in `procore.eventTypes` will be automatically reflected by this component.

components/procore/sources/budget-snapshot/budget-snapshot.js
```diff
@@ -0,0 +1,50 @@
+const common = require("../common.js");
+
+module.exports = {
+  ...common,
+  name: "New or Updated Budget Snapshot (Instant)",
```
jverce, line 5: The name should be just "Budget Snapshot Event (Instant)" instead, since it includes new, updated or deleted snapshots, right?

(This is repeated in multiple places so I'm making the comment here, but it applies to other components)

components/procore/sources/budget-snapshot/budget-snapshot.js
```diff
@@ -0,0 +1,50 @@
...
+          this.project,
+          resourceId,
+          limit,
+          offset
+        );
+        snapshotRows = snapshotRows.concat(resource);
+        total = resource.length;
+        offset += limit;
+      }
+      return { body, snapshotRows };
```
jverce, line 33: I wouldn't expect the event to query the whole object before emitting the event downstream.
Sending a resource ID is fine IMHO. cc @dylburger WDYT?

jverce (approved):

Issue #528, opened 2020-10-06
[BUG] Twitter event sources fail to honor include { retweets, replies } parameters


dylburger:
Due to a change made to Twitter event sources, the **Retweets** and **Replies** parameters for all Twitter event sources are currently not working. We're investigating the issue and will update this ticket as soon as we have a fix.

PR #892, opened 2021-02-06
Bug fixes for tweet search

jverce:
**Note: the change contains a lot of removed trailing whitespaces, so I suggest hiding those when doing your review**

Fixes issue #528

* Properly handle `includeRetweets` and `includeReplies` as strings
  instead of booleans
* Add new language option (**Undefined**) to allow for tweet searches
  where the tweet language was not identified
* Manage the `lang` property as a string array to allow users to fetch
  tweets of multiple languages (including "undefined"). This required
  setting the `lang` parameter as part of the search query instead of
  a URL query param.

### DIFF
--- a/components/twitter/sources/language-codes.js
+++ b/components/twitter/sources/language-codes.js
@@ -1,5 +1,9 @@
 module.exports = [
   {
+    English: "Undefined",
+    alpha2: "und",
+  },
+  {
     English: "Abkhazian",
     alpha2: "ab",
   },
--- a/components/twitter/sources/language-codes.js
+++ b/components/twitter/sources/language-codes.js
@@ -1,6 +1,6 @@
 module.exports = [
   {
-    English: "Undefined",
+    English: "Undetected",
     alpha2: "und",
   },
   {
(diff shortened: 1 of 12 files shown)


components/twitter/sources/language-codes.js
```diff
@@ -1,4 +1,8 @@
 module.exports = [
+  {
+    English: "Undefined",
```
tjk, line 3: Would it make sense to use something more user friendly as the english label for `Undefined`?

tjk (approved):
one minor comment. looks good!

components/twitter/sources/language-codes.js
jverce: It definitely does make sense, I just don't know what to use here 😂 
The [Twitter API docs](https://developer.twitter.com/en/docs/twitter-api/enterprise/search-api/guides/operators) use the term **undefined** so I just used it here as well, but we can change it to **Undetected** maybe?

tjk: Does `Any` make sense?

jverce: Well... that would imply that this option captures tweets in any language, when in fact what it does is capture those tweets which language couldn't be properly identified by Twitter. WDYT?

PR #893, opened 2021-02-06
Fix bug related to initial empty records being captured

### LINKED ISSUE #695: [BUG] Airtable event sources should not emit empty records
Airtable adds empty records to new tables by default. When I create a **New Records** source, I see these records emitted, but they're empty (this makes sense). I'd like the source to filter out these empty records, and not emit them.

jverce:
Fixes #695

* Manage the "starting" timestamp through the components hooks in
  addition to doing it during `run()`, so that we only capture records
  created/updated/deleted after the activation of the event source.
* Refactored some common logic into a single common component for
  reusability

### DIFF
--- a/components/airtable/sources/common.js
+++ b/components/airtable/sources/common.js
@@ -10,6 +10,7 @@ module.exports = {
         intervalSeconds: 60 * 5,
       },
     },
+    baseId: { type: "$.airtable.baseId", appProp: "airtable" },
   },
   hooks: {
     activate() {
--- a/components/airtable/sources/common.js
+++ b/components/airtable/sources/common.js
@@ -1,0 +1,23 @@
+const airtable = require("../airtable.app");
+
+module.exports = {
+  props: {
+    airtable,
+    db: "$.service.db",
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 5,
+      },
+    },
+  },
+  hooks: {
+    activate() {
+      const startTimestamp = new Date().toISOString();
+      this.db.set("lastTimestamp", startTimestamp);
+    },
+    deactivate() {
+      this.db.set("lastTimestamp", null);
+    },
+  },
+};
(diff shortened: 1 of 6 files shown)


components/airtable/sources/common.js
```diff
@@ -0,0 +1,23 @@
...
+    db: "$.service.db",
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 5,
+      },
+    },
+  },
+  hooks: {
+    activate() {
```
dylburger, line 15: btw, we've recently shipped a `deploy` hook that runs only at deploy time (instead of on every component update). [See this source as an example](https://github.com/PipedreamHQ/pipedream/blob/446b92a5173b550cc89145b0e6285c6f1b8a637b/components/uservoice/sources/new-nps-ratings/new-nps-ratings.js#L24).

components/airtable/sources/new-or-modified-records-in-view/new-or-modified-records-in-view.js
```diff
@@ -1,24 +1,19 @@
...
   name: "New or Modified Records in View",
   description: "Emit an event for each new or modified record in a view",
   key: 'airtable-new-or-modified-records-in-view',
-  version: "0.0.2",
+  version: "0.0.3",
   props: {
-    db: "$.service.db",
-    airtable,
+    ...common.props,
     baseId: { type: "$.airtable.baseId", appProp: "airtable" },
```
dylburger, line 14: Does it make sense to include `baseId` in the common props or was there a reason you left it out?

components/airtable/sources/common.js
jverce: Yeah I noticed that one, but I don't think it applies here.
If a user decided to disable this event source, and then later re-enables it, we want to make sure that we update the start timestamp so that we don't capture any data generated during the period in which this event source was disabled.

dylburger (approved):

components/airtable/sources/common.js
dylburger: yeah this comment was more of an aside. I just wasn't sure if you knew about the `deploy` hook.

dylburger:
lgtm

components/airtable/sources/new-or-modified-records-in-view/new-or-modified-records-in-view.js
jverce: It does make sense actually, looks like Airtable uses this "base" concept for every product, so it makes sense to have it in `common`. I'll move it there.

Issue #634, opened 2020-11-18
[SOURCE] RingCentral


todsac:
**Describe the event source. What app, what event?**
RingCentral

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
Custom webhook events | Triggers on new events
New Calls | Triggers when there is a new incoming or outgoing call.
New Call Recordings | Returns new call recordings since the last check.
New Faxes | Triggers when a new fax has been received.
New Incoming SMS Messages | Triggers when a new SMS message is received.
New Messages | Watches for new messages.
New Missed Calls | Triggers when there is a new missed call.
New Voice Message | Triggers when a new voice message has been received.

jverce:
This is blocked until I get a response from RingCentral's customer support for assistance with the creation of a sandbox account.

jverce:
Was able to activate the sandbox account and I am able to call RingCentral's API.
I am still unable to authenticate via Pipedream but that's something I can address later on.

todsac:
We have done a lot with RingCentral and have a good product contact there so happy to help.  When we were working with the app we were able to create a test user within our developer account.  Happy to walk you through the setup we used.

todsac:
Also, why can't you authenticate with PD?

jverce:
I was also able to create a test/sandbox user in my developer account as well, and that allowed me to call the API.
However, when authenticating through PD (or through the RingCentral login page as well) I am unable to use that test user's credentials:
![image](https://user-images.githubusercontent.com/5479513/106078418-3c84e480-60c8-11eb-8422-7b4b2c73d48e.png)

When using my developer account credentials in PD, I get the following error:
![image](https://user-images.githubusercontent.com/5479513/106078355-21b27000-60c8-11eb-83b3-404021be082b.png)

I guess the latter issue is related to the account plan? According to their [pricing page](https://www.ringcentral.com/office/plansandpricing.html), not all plans have access to their developer platform:
![image](https://user-images.githubusercontent.com/5479513/106078611-8bcb1500-60c8-11eb-8574-17fd7a7bc2c2.png)

PR #902, opened 2021-02-10
RingCentral Event Sources

jverce:
Solves #634 

* Add common logic for the main `.app.js` file, for HTTP-based and for
  timer-based event sources
* Add event sources for:
  * Custom webhook events
  * New Calls
  * New Call Recordings
  * New Faxes
  * New Incoming SMS Messages
  * New Messages
  * New Missed Calls
  * New Voice Message

### DIFF
--- a/components/ringcentral/sources/common/http-based.js
+++ b/components/ringcentral/sources/common/http-based.js
@@ -1,0 +1,137 @@
+const template = require("lodash/template");
+const base = require("./base");
+const notificationTypes = require("./notification-types");
+
+module.exports = {
+  ...base,
+  dedupe: "unique",
+  props: {
+    ...base.props,
+    http: {
+      type: "$.interface.http",
+      customResponse: true,
+    },
+  },
+  hooks: {
+    ...base.hooks,
+    async activate() {
+      const opts = {
+        address: this.http.endpoint,
+        eventFilters: this._getEventFilters(),
+      };
+      const {
+        id: webhookId,
+        verificationToken,
+      } = await this.ringcentral.createHook(opts);
+
+      this.db.set("webhookId", webhookId);
+      this.db.set("verificationToken", verificationToken);
+    },
+    async deactivate() {
+      const webhookId = this.db.get("webhookId");
+      await this.ringcentral.deleteHook(webhookId);
+
+      this.db.set("verificationToken", null);
+    },
+  },
+  methods: {
+    ...base.methods,
+    _getPropValues() {
+      return Object.entries(this)
+        .filter(([_, value]) => value != null)
+        .reduce((accum, [prop, value]) => ({
+          ...accum,
+          [prop]: value,
+        }), {});
+    },
+    _getEventFilters() {
+      const eventKeys = this.getSupportedNotificationTypes();
+      return notificationTypes
+        .filter(({ key }) => eventKeys.has(key))
+        .map(({ filter }) => template(filter))
+        .map((templateFn) => templateFn(this._getPropValues()));
+    },
+    /**
+     * Provides the set of notification types to which an HTTP-based event
+     * source subscribes. This should be a subset of the `key` properties
+     * available in the `notification-types` module.
+     *
+     * @return {Set}  The set of supported notification type keys
+     */
+    getSupportedNotificationTypes() {
+      throw new Error("getSupportedNotificationTypes is not implemented");
+    },
+    /**
+     * Validate that the incoming HTTP event comes from the expected source, and
+     * reply with a `200` status code and the proper validation token header, as
+     * described here:
+     * https://community.ringcentral.com/questions/1306/validation-token-is-not-returned-when-creating-a-s.html
+     *
+     * In case the event comes from an unrecognized source, reply with a `404`
+     * status code.
+     *
+     * The result of this method indicates whether the incoming event was valid
+     * or not.
+     *
+     * @param {object}  event The HTTP event that triggers this event source
+     * @return {boolean}  The outcome of the validation check (`true` for valid
+     * events, `false` otherwise)
+     */
+    validateEvent(event) {
+      const {
+        "validation-token": validationToken,
+        "verification-token": verificationToken,
+      } = event.headers;
+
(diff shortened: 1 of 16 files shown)


components/ringcentral/sources/common/http-based.js
```diff
@@ -0,0 +1,137 @@
...
+     * @return {boolean}  The outcome of the validation check (`true` for valid
+     * events, `false` otherwise)
+     */
+    validateEvent(event) {
+      const {
+        "validation-token": validationToken,
+        "verification-token": verificationToken,
+      } = event.headers;
+
+      const expectedVerificationToken = this.db.get("verificationToken") || verificationToken;
```
tjk, line 86: why are we being lenient about the `verificationToken` not being in db?

components/ringcentral/sources/new-call-recording/new-call-recording.js
```diff
@@ -0,0 +1,46 @@
...
+  },
+  methods: {
+    ...common.methods,
+    generateMeta(data) {
+      const {
+        id,
+        startTime: timestamp,
+      } = data;
+
+      const summary = `New call recording`;
```
tjk, line 20: nothing else we can include in summary here?

components/ringcentral/sources/new-event/new-event.js
```diff
@@ -0,0 +1,71 @@
...
+        }));
+      },
+    },
+  },
+  methods: {
+    ...common.methods,
+    _getEventTypeFromFilter(eventFilter) {
+      return eventFilter
+        .replace(/\/restapi\/v\d+\.\d+\//, "")
+        .replace(/account\/.*?\//, "")
```
tjk, line 47: why the `?`

components/ringcentral/sources/common/http-based.js
```diff
@@ -0,0 +1,137 @@
...
+    async deactivate() {
+      const webhookId = this.db.get("webhookId");
+      await this.ringcentral.deleteHook(webhookId);
+
+      this.db.set("verificationToken", null);
+    },
+  },
+  methods: {
+    ...base.methods,
+    _getPropValues() {
```
tjk, line 39: could you remove this and `lodash/template` in favor of just setting the `filter` attribute to function and use `this` knowing function will be called in the component method? not sure if you think that is too hacky.. not a huge fan of this `_getPropValues`

components/ringcentral/sources/common/http-based.js
jverce: Whenever we create a hook, RingCentral makes an initial `POST` call to verify that the provided endpoint is valid and healthy. Once it verifies that, the hook creation succeeds. This means that this event source is triggered by an HTTP call in parallel to the execution of the `activate()` method (where the hook is created). The problem is that when the event source is handling that initial HTTP call, the DB is not up-to-date and so we always get a null value when we do `this.db.get("verificationToken")`.

components/ringcentral/sources/new-event/new-event.js
jverce: It's just the "non-greedy" regex command, otherwise the `.*` will match every character up until the **last** `/`. Using `.*?` instead of `.*` will match every character up until the **first** `/`.

components/ringcentral/sources/common/http-based.js
jverce: You mean something like this?
```js
// notification-types.js
module.exports = [
  {
    key: "message-event-outbound",
    label: "Message Event (Outbound)",
    filter: ({ extensionId, messageType }) => `/restapi/v1.0/account/~/extension/${extensionId}/message-store?direction=Outbound&type=${messageType}`),
  },
};
```

I'm up to it, with the only caveat that I wanted to keep the `notification-types` as a "data" module without any logic itself (it'd had been a JSON originally if it weren't for the fact that we don't support loading JSON files yet). If you don't see a problem with adding logic to this module then I'm fine with moving the templating logic into that module. To me it's as hacky as the approach I took, so I'm comfortable both ways 🙂

components/ringcentral/sources/new-call-recording/new-call-recording.js
jverce: We can add the following, whichever makes more sense to you:
- Session ID (i.e. what identifies the call that was recorded)
- Phone number of the client (the "from" phone number if it was an inbound call, or "to" phone number if it was an outbound call)

components/ringcentral/sources/common/http-based.js
tjk: i was thinking to use unbound functions so the `this` would be the component when called but I like the concept of keeping that more of a simple data module without functions so let's keep it as is.

components/ringcentral/sources/common/http-based.js
tjk: Oh that sounds a bit brittle! We might not be great about this use case of webhooks making that initial call an excepting us to be activated already... hmm... There is a chance that initial POST could even fail, right? Is that handled correctly?

jverce: If that initial POST fails, then the hook creation fails as well and the activation would fail. However, my understanding is that the HTTP endpoint is already set up by the platform before running `activate()`, correct? In which cases could this fail, other than due to a bug on the event source itself?

components/ringcentral/sources/new-call-recording/new-call-recording.js
tjk: Either of those is fine to me. Just something to differentiate the events from summary perspective. Maybe the phone number is more useful? But a user of the source would know best what's useful so let's start with whichever you think is better for now.

components/ringcentral/sources/new-event/new-event.js
tjk: Oh interesting! (I'd just write `/account\/[^/]*\//` which does the same thing right?)

tjk: (I don't mean to change it, just sharing 😛)

PR #928, opened 2021-02-16
PR for initial feedback

sergioeliot2039:
This is a first PR for initial feedback.
There's an issue with $auth variable being null. Plus I have pending to add the subreddit as a prop, for now hardcoded while troubleshooting.

### DIFF
--- a/components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
+++ b/components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
@@ -1,0 +1,64 @@
+const uservoice = require("../../reddit.app.js");
+
+module.exports = {
+    key: "new-link-on-a-subreddit",
+    name: "New Link on a subreddit",
+    description: "Emmits a new link submitted to a subreddit",
+    version: "0.0.1",	
+    props: {
+        reddit,
+        timer: {
+            label: "Polling schedule",
+            description: "Pipedream poll for new links and emit accordingly with this schedule.",
+            type: "$.interface.timer",
+            default: {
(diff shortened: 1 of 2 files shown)


components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
```diff
@@ -0,0 +1,64 @@
+const uservoice = require("../../reddit.app.js");
```
dylburger, line 1: You need to change this from `uservoice` to `reddit`.

components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
```diff
@@ -0,0 +1,64 @@
...
+                summary: reddit_link.data.name,
+            });
+
+        },
+    },
+    async run() {
+
+        let after = this.db.get("after");
+        const reddit_things = await this.getNewSubredditLinks(after);
+        var after = reddit_things.data.data.after;
```
dylburger, line 54: The source failed for me here, since you already declared `let after` two lines above. Try removing the `var` here.

sergioeliot2039: I renamed to `current_after`, and `new_after`

components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
sergioeliot2039: Changed.

PR #930, opened 2021-02-17
Twitch

michelle0927:


components/twitch/sources/common.js
```diff
@@ -0,0 +1,43 @@
...
+            items.push(this.getTopicString(item));
+          }
+          // if called from polling component, emit data
+          else if (type == "polling") {
+            this.$emit(item, this.getMeta(item));
+          }
+          // if type is list, return list of items
+          else if (type == "list") {
+            items.push(item);
+          }
```
jverce, line 28: If we transfor `paginate` into a generator, we can move the responsibility of figuring out what to do with the response of the API requests to the consumers of this function. Otherwise this function has to know the details of each consumer so that it can handle the API responses in the particular manner that the consumers wish.

Instead I propose the following:
```js
async function* paginate(resourceFn, params, max = null) {
  let done = false;
  let count = 0;
  do {
    const {
      data,
      pagination,
    } = await resourceFn(params);

    for (const item of data) {
      count++;
      if (max && count > max) {
        return;
      }

      yield item;
    }

    const { cursor } = pagination;
    params.after = cursor;
    done = !cursor;
  } while (!done);
}
```

Then each consumer can do this:
```js
for (const followed of followedUsers) {
  const params = {
    user_id: followed.to_id,
    // Period during which the video was created. Valid values: "all",
    // "day", "week", "month".
    period: "day",
  };
  const videosScan = await this.paginate(
    this.twitch.getVideos.bind(this),
    params,
    this.max,
  );
  for await (const video of videosScan) {
    const meta = this.getMeta(video);
    this.$emit(video, meta);
  }
}
```

components/twitch/sources/common-webhook.js
```diff
@@ -0,0 +1,77 @@
...
+          "unsubscribe",
+          this.http.endpoint,
+          topic,
+          subscriptionExpiration
+        );
+    },
+  },
+  methods: {
+    ...common.methods,
+    async subscribeToHooks(topics) {
```
jverce, line 38: Can we change this a bit so that we can use it in `deactivate()` as well?
```js
module.exports = {
  hooks: {
    async activate() {
      const topics = await this.getTopics();
      this.db.set("topics", topics);
      await this.manageHookForTopics("subscribe", topics);
    },
    async deactivate() {
      const topics = this.db.get("topics");
      await this.manageHookForTopics("unsubscribe", topics);
    },
  },
  methods: {
    manageHookForTopics(mode, topics) {
      for (const topic of topics) {
        await this.twitch.manageHook(
          mode,
          this.http.endpoint,
          topic,
          subscriptionExpiration,
        );
      }
    },
  },
};
```

We should also add a `try/catch` block around the `manageHook` call so that we don't leave active subscriptions in case one of the calls fails.

components/twitch/twitch.app.js
```diff
@@ -0,0 +1,107 @@
...
+      label: "Stream Language",
+      description:
+        'Watch for games streamed in the specified language. A language value must be either the ISO 639-1 two-letter code for a supported stream language or "other".',
+    },
+    max: {
+      type: "integer",
+      label: "Max Items",
+      description:
+        "The maximum number of items to return at one time. Streams are returned sorted by number of viewers, in descending order. Videos and Clips are returned sorted by publish date.",
+    },
```
jverce, line 30: We can add an additional `min` parameter so that `min: 1`.

jverce (changes requested):
I like how you structured the whole event source.
My comments are mostly to refactor some pieces that could hurt future maintenance, and to address some failure modes (e.g. throttling/retries).

components/twitch/sources/common-webhook.js
```diff
@@ -0,0 +1,77 @@
+const common = require("./common.js");
+const subscriptionExpiration = 864000; // seconds until webhook subscription expires, maximum 10 days (864000 seconds)
```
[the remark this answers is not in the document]
michelle0927: Updated to cut the timer interval in half.

components/twitch/twitch.app.js
```diff
@@ -0,0 +1,107 @@
...
+    async _makeRequest(method, endpoint, params = null) {
+      const config = {
+        method,
+        url: `${this._getBaseUrl()}/${endpoint}`,
+        headers: this._getHeaders(),
+      };
+      if (params) config.params = params;
+      return await axios(config);
+    },
```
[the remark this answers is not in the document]
michelle0927: Updated to use paramsSerializer.

components/twitch/twitch.app.js
```diff
@@ -0,0 +1,107 @@
...
+        "hub.callback": url,
+        "hub.mode": mode,
+        "hub.topic": `${this._getBaseUrl()}/${topic}`,
+        "hub.lease_seconds": leaseSeconds,
+        "hub.secret": this.$auth.oauth_access_token,
+      };
+      return await this._makeRequest("POST", "webhooks/hub", data);
+    },
+    verifyWebhookRequest(headers, body) {
+      var hubSecret = headers["x-hub-signature"].split("=");
```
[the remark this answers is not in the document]
michelle0927: Updated.

components/twitch/twitch.app.js
```diff
@@ -0,0 +1,107 @@
...
+      const hash = crypto
+        .createHmac(hubSecret[0], this.$auth.oauth_access_token)
+        .update(JSON.stringify(body))
```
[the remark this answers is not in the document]
michelle0927: Updated to use bodyRaw.

components/twitch/twitch.app.js
```diff
@@ -0,0 +1,107 @@
...
+    },
+    // Uses Twitch API to create or delete webhook subscriptions.
+    // Set mode to "subscribe" to create a webhook and "unsubscribe" to delete it.
+    async manageHook(mode, url, topic, leaseSeconds) {
+      const data = {
+        "hub.callback": url,
+        "hub.mode": mode,
+        "hub.topic": `${this._getBaseUrl()}/${topic}`,
+        "hub.lease_seconds": leaseSeconds,
+        "hub.secret": this.$auth.oauth_access_token,
```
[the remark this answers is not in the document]
michelle0927: Updated to generate the secret this way: 
const secretToken = uuidv4();

components/twitch/sources/common.js
michelle0927: Updated the paginate method to be a generator function.

components/twitch/twitch.app.js
michelle0927: Done.

components/twitch/sources/common-webhook.js
michelle0927: Updated to include a try/catch block in manageHookForTopics method.

components/twitch/sources/common.js
```diff
@@ -0,0 +1,43 @@
...
+    twitch,
+    db: "$.service.db",
+  },
+  methods: {
+    async paginate(resourceFn, params, type, max = null) {
+      const items = [];
+      let done = false;
+      let count = 0;
+      while (!done) {
+        const { data, pagination } = await resourceFn(params);
```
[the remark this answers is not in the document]
michelle0927: Let me know if I've implemented this correctly or if I need to make changes since this is my first time including retry and backoff logic in a component. Should I be using more of the header info to calculate the backoff?

components/twitch/sources/common.js
```diff
@@ -0,0 +1,48 @@
...
+      try {
+        const response = await resourceFn(params);
+        if (response.status == 200) {
+          return response.data;
+        }
+        if (retries > 0) {
+          if (!backOff) backOff = response.headers["ratelimit-limit"];
+          setTimeout(() => {
+            return retryFn(resourceFn, params, retries - 1, backOff * 2);
+          }, backoff);
```
jverce, line 39: This will set a timer for an asynchronous execution of `retryFn`, and return right away, with a `void` value.
I think the correct approach would be something like this:
```js
const { promisify } = require("util");
const pause = promisify((delay, fn) => setTimeout(fn, delay));

async function retryFn(resourceFn, params) {
  const { "ratelimit-limit": delay } = response.headers;
  // Delay must be in milliseconds
  await pause(delay);
  if (retries <= 0) {
    throw new Error('some error');
  }
  return retryFn(resourceFn, params);  
}
```

I might be wrong though, and if you tested this and worked well then let's go with your solution.

jverce (changes requested):
Minor changes, but overall looks great 🙂

jverce (approved):

PR #959, opened 2021-02-22
Code step exports `return` example clarification

NReilingh:
This clarifies how to use $return_value in the example given


dylburger:
@NReilingh thanks for the fix, much appreciated!

PR #960, opened 2021-02-22
Shopify bug fix

michelle0927:


components/shopify/shopify.app.js
```diff
@@ -1,4 +1,8 @@
 const axios = require("axios");
+const Bottleneck = require('bottleneck');
+const limiter = new Bottleneck({
```
dylburger, line 3: I might just add a small comment here linking to the Spotify API docs on their rate limits, noting that the `minTime` was set as a result of their rate limit

components/shopify/sources/new-event/new-event.js
```diff
@@ -25,27 +29,42 @@ module.exports = {
...
+      if (results[results.length - 1])
+        this.db.set(eventType, results[results.length - 1].id);
     },
+    async getEvents(eventType, sinceId) {
+      const results = await this.shopify.getEvents(
+        sinceId,
+        JSON.parse(eventType).filter,
+        JSON.parse(eventType).verb
+      );
+      return { results, eventType };
```
dylburger, line 42: Any reason you need to return `eventType` since the caller should have access to it (i.e. since it's passed as a param)?

dylburger: I see what you're doing below now, that makes sense and that's fine.

PR #969, opened 2021-02-24
Fix new follower Twitter event sources

### LINKED ISSUE #966: [BUG] Twitter new followers event source emits duplicate events intermittently
**Describe the bug**

As per @psavkar:
> Specifically, can you review (and either update or wholesale re-write) the Twitter "new follower" sources (i.e., `new-follower-of-me` &` new-follower-of-user`)?
>
> We were running a workflow to DM users when they follow @pipedream on Twitter, but we just found that we've been sending multiple DMs to some users. based on my initial investigation, i found the source is intermittently emitting duplicate events (every few hours with no clear pattern) for at least two specific users (but there may be others as well). I wrote the first version of this source but I can't figure out what's wrong (I also don't have logs for the executions where dupe events were emitted, so I don't understand what data was returned by Twitter and what data was cached for the source).
>
> One of the challenges with this source is that Twitter doesn't provide any metadata about the follow event that we can use for deduping (i.e., there's no unique ID or timestamp for the follow event -- they only provide an ordered list of user IDs). 
Additionally, since users can unfollow at any time, our default deduping strategies aren't sufficient (i.e., we can't rely on the last or greatest user ID or the last 100 unique IDs for deduping).
>
> The use cases these sources should handle include:
> 1. Emit the most recent followers on first execution
> 2. Emit new followers on subsequent executions (note: any number of historical followers may be lost, including the most recent one)
> 3. If a user unfollows and follows again, we should emit as a new follower (as long as we're polling, there probably has to be a source execution between the unfollow/follow event)
> 4. Do not emit historical followers as new (e.g., if we cache the last 100 followers and some followers are lost, some old followers could be interpreted as being based on a simple comparison)
>
> I don't think we can use the Twitter account activity webhook because our response latency is too high, but feel free to look into that approach (at least for new-follower-of-me).

**Screenshots**
image

**Additional context**
- The source runs every minute but the duplicate events seem to be emitted at random times (i.e., not on every source execution) - @psavkar 
- While working on the `new-unfollower-of-me` I noticed the same thing, that we need to keep track of historical data explicitly, so I have a solution in mind that aligns to what you're saying - @jverce

jverce:
Fixes #966

* Refactor the **New Follower of Me** and **New Follower of User** to
  use the existing logic in their _unfollower_ counterparts (see the
  [original PR](https://github.com/PipedreamHQ/pipedream/pull/688) for
  further context. The logic relies on keeping an up-to-date list of
  all the followers of an account in order to handle deduplication and
  diff detection.
* Add retry logic around Twitter API calls to handle rate limit
  responses (as well as other retriable status codes).


components/twitter/sources/common-followers.js
```diff
@@ -60,20 +62,52 @@ module.exports = {
...
+
         result.push(id);
       }
+
+      this.setFollowersCache(result);
       return result;
     },
+    setFollowersCache(followers) {
+      const followersCacheSize = Math.max(this.getFollowersCacheSize(), 0);
+      const trimmedFollowers = followers.slice(0, followersCacheSize);
```
casret, line 83: Won't this trim all the followers in the unlimited case?

components/twitter/sources/new-follower-of-me/new-follower-of-me.js
```diff
@@ -1,68 +1,57 @@
...
-        }
-      }
+  methods: {
+    ...common.methods,
+    getFollowersCacheSize() {
+      return this.followersCacheSize;
+    },
+    async getRelevantIds() {
+      const isFirstExecution = !this.db.get("hasExecuted");
+      const mostRecentFollowers = this.db.get("followers");
```
casret, line 27: Small nit.  Can we stick an accessor for this in common_followers?  I don't like that it's a magic string.

casret (changes requested):

components/twitter/sources/common-followers.js
jverce: Ah yes, nice catch! I need to fix this.

components/twitter/sources/new-follower-of-me/new-follower-of-me.js
jverce: Makes sense, will do!

casret (approved):
Sorry, didn't realize you were waiting on me again.  Looks great.

PR #991, opened 2021-02-26
Spotify bug fix - fetch more than 20 playlists

michelle0927:


components/spotify/sources/common.js
```diff
@@ -0,0 +1,38 @@
...
+      while (!done) {
+        const results = await resourceFn(params);
+        const { items } = results;
+        for (const item of items) {
+          yield item;
+        }
+        if (items.length < params.limit)
+            done = true;
+        params.offset += params.limit;
+        await new Promise(resolve => setTimeout(resolve, 500));
```
dylburger, line 34: Was there a specific reason you set this to half a second? I see that Spotify returns a `Retry-After` header that tells us the number of seconds to wait before sending another request. Could we use that, instead? https://developer.spotify.com/documentation/web-api/

components/spotify/spotify.app.js
```diff
@@ -12,34 +33,41 @@ module.exports = {
...
-      );
+        // if rate limit is exceeded, Retry-After will contain the # of seconds to wait before retrying
+        delay = (response && response.status == 429) ? (response.headers['Retry-After']*1000) : 500;
+        await pause(delay);
+        return await this.retry(config, retries - 1);
+      }
+    },
+    async getPlaylistItems(params) {
+      const { playlistId } = params;
+      return await this._makeRequest("GET", `playlists/${playlistId}/tracks`, params);
```
dylburger, line 63: Small nit: in other components we typically include the `/` in the path passed to `_makeRequest`. i.e. you should remove the path separator in `_makeRequest` and pass:

```javascript
`/playlists/${playlistId}/tracks`
```

Even though it's an extra character in each call to `_makeRequest`, it's a common convention, so I'd recommend sticking with that

components/spotify/spotify.app.js
```diff
@@ -12,34 +33,41 @@ module.exports = {
...
+      try {
+        response = await axios(config);
+        return response.data;
+      } catch(err) {
+        if (retries <= 1) {
+          throw new Error(err);
         }
-      );
+        // if rate limit is exceeded, Retry-After will contain the # of seconds to wait before retrying
+        delay = (response && response.status == 429) ? (response.headers['Retry-After']*1000) : 500;
```
dylburger, line 56: `const delay`? Or is `delay` declared somewhere else above?

dylburger (approved):
looks good, just a couple of small comments

components/spotify/sources/common.js
michelle0927: The Retry-After header is only sent if the response status is 429 (too many requests). I couldn't find what their usual rate limit is, but figured 2 requests per second would be a decent default to use for regular pagination. Would it be better to leave this part out and just let the retry method in the app file handle it if the rate limit is exceeded?

components/spotify/spotify.app.js
michelle0927: Updated.

components/spotify/spotify.app.js
michelle0927: Updated.

components/spotify/sources/common.js
dylburger: I'd recommend checking for the presence of the `Retry-After` header here, and if it's absent / undefined, continue.  When a `Retry-After` mechanism is supported for an API like it is here, I'd recommend we use it, since a user can create multiple event sources that would trigger the rate limit even at a 2/sec rate for an individual source.

PR #993, opened 2021-02-27
ActiveCampaign - New Contact Added to List event source - support filtering on specific lists

michelle0927:


components/activecampaign/sources/contact-added-to-list/contact-added-to-list.js
```diff
@@ -6,9 +6,57 @@ module.exports = {
...
     ...common.methods,
+    async getHookData(events, endpoint, sources, list) {
+      const { webhook } = await this.activecampaign.createHook(
+        events,
+        endpoint,
+        sources,
+        list
+      );
+      return webhook;
+    },
```
jverce, line 59: We should remove this method, as it can be misleading.
`getHookData` implies that it's retrieving some value, when in fact what it does is create a webhook.

components/activecampaign/sources/contact-added-to-list/contact-added-to-list.js
```diff
@@ -6,9 +6,57 @@ module.exports = {
...
+      if (this.lists.length > 0) {
+        for (const list of this.lists) {
+          const { id } = await this.getHookData(
+            events,
+            this.http.endpoint,
+            sources,
+            list
+          );
+          hookIds.push(id);
+        }
```
jverce, line 31: In case we're unable to create a new webhook, we should delete the webhooks that we created so far.
Otherwise the event source will leave the user's account in an inconsistent state.

jverce (approved):
A couple of minor comments, but overall looks good 🙂

PR #1000, opened 2021-02-28
Update README.md to match Pipedream's UI

lirdis:
Changed "New Source" to "New +" to match the current UI.


dylburger:
Great catch!

PR #1003, opened 2021-03-01
Actions

psavkar:


components/airtable/actions/create-multiple-records/create-multiple-records.js
```diff
@@ -0,0 +1,44 @@
...
+    airtable,
+    baseId: {type: "$.airtable.baseId", appProp: 'airtable'},
+    tableId: { type: '$.airtable.tableId', baseIdProp: 'baseId' },
+    records: { propDefinition: [airtable, "records"] },
+  },
+  methods: {
+    async addRecords(records) {
+      const base = new Airtable({apiKey: this.airtable.$auth.api_key}).base(this.baseId);
+      return await base(this.tableId).create(records)
+    },
```
jverce, line 22: I would put this in the `.app.js` module, and keep any logic specific to the interaction against the Airtable API within that module, or at least add a method there that does this:

**airtable.app.js**
```js
const 
module.exports = {
  methods: {
    _apiKey() {
      return this.$auth.api_key;
    },
    getNewClient(baseId, tableId) {
      const clientOpts = {
        apiKey: this._apiKey(),
      },
      const airtableClient = new Airtable(clientOpts);
      return airtableClient.base(baseId)(tableId);
    },
  },
}; 
```

components/airtable/actions/create-multiple-records/create-multiple-records.js
```diff
@@ -0,0 +1,44 @@
+const airtable=require('../../airtable.app.js')
+const axios = require('axios')
```
jverce, line 2: This Axios import can be removed, right?

components/discord_webhook/discord_webhook.app.js
```diff
@@ -0,0 +1,26 @@
...
+  type: 'app',
+  app: 'discord_webhook',
+  propDefinitions: {
+    message: {
+      type: "string",
+      description: "Enter a simple message up to 2000 characters. This is the most commonly used field. However, it's optional if you pass embed content.",
+    },
+    embeds: {
+      type: "any",
+      description: 'Optionally pass an [array of embed objects](https://birdie0.github.io/discord-webhooks-guide/discord_webhook.html). E.g., ``[{"description":"Use markdown including *Italic* **bold** __underline__ ~~strikeout~~ [hyperlink](https://google.com) `code`"}]``. To pass data from another step, enter a reference using double curly brackets (e.g., `{{steps.mydata.$return_value}}`).\nTip: Construct the `embeds` array in a Node.js code step, return it, and then pass the return value to this step.',
```
jverce, line 11: I've dealt with long descriptions myself.
Question: is there some guideline we should follow in those cases?

jverce (commented):
I have 2 main observations:
- Extracting common logic into a separate module simplifies the creation of new components, and we could apply this pattern here as well.
- I see that the `.app.js` modules just contain the prop definitions, and the logic to make the API calls is contained within each component. I would do the opposite: the `.app.js` encapsulates all the logic related to making the API calls, and have the props defined by each component depending on what they need. If some props are shared between components, those can be defined in a common module.

WDYT?

components/slack/slack.app.js
```diff
@@ -4,10 +4,102 @@ module.exports = {
...
+      label: "User",
+      //description: "Select one or more channels or DM conversations to monitor for new messages.",
+      async options({ prevContext }) {
+        let { types, cursor, userNames } = prevContext
+        if (types == null) {
+          const scopes = await this.scopes()
+          types = ["im"]
+          // TODO use paging
+          userNames = {}
+          for (const user of (await this.users()).users) {
```
casret, line 62: If you are wondering why your async options are taking so long, it's stuff like this where we are asking for every single user in a workspace before we start to page.  And we do this every time we ask for a new page.

components/slack/actions/send-direct-message/send-direct-message.js
```diff
@@ -0,0 +1,30 @@
+const slack = require('../../slack.app.js')
+const { WebClient } = require('@slack/web-api')
+
+module.exports = {  
+  key: "slack-send-direct-message",
+  name: "Send a Direct Message",
+  description: "Send a direct message to a single user.",
```
casret, line 7: I'm curious, did you test this out?  Casual reading of the slack API makes it sound like you can't directly message between 2 users even with the as_user set.

components/slack/slack.app.js
casret: Actually it seems like this users call is redundant since this.avaliableConversations is already doing paging calls through the users?

components/slack/slack.app.js
```diff
@@ -3,6 +3,243 @@ const { WebClient } = require('@slack/web-api')
...
```
casret, line 62: I think this will cause a huge delay in async options for slack with many users, and I'm not sure that this is useful at all.  The other comments lay out some alternatives.

components/slack/slack.app.js
```diff
@@ -3,6 +3,243 @@ const { WebClient } = require('@slack/web-api')
...
+            userNames[user.id] = user.name
+          }
+        }
+        const resp = await this.availableConversations(types.join(), cursor)
```
casret, line 66: So it seems like this will only get the list of conversations for users you've ever had a conversation with?  I would think a prop named 'user' would return user objects.  In fact I think the action should be using that prop (one that return all the users), and then use postMessage with the user id.  (Though I also feel like that might not work?)

components/slack/slack.app.js
```diff
@@ -3,6 +3,243 @@ const { WebClient } = require('@slack/web-api')
...
+        return {
+          options: resp.conversations.map((c) => {
+            return { label: `@${userNames[c.user]}`, value: c.id }
```
casret, line 69: What does c.name return here BTW? https://api.slack.com/types/conversation doesn't really specify, but I think we may be able to eliminate the lookup table if it returns something useful?

components/slack/slack.app.js
```diff
@@ -3,6 +3,243 @@ const { WebClient } = require('@slack/web-api')
...
+          context: { types, cursor: resp.cursor, userNames },
+        }
+      },
+    },
+    privateChannel: {
+      type: "string",
+      label: "Channel",
+      //description: "Select one or more channels or DM conversations to monitor for new messages.",
+      async options({ prevContext }) {
+        let { types, cursor, userNames } = prevContext
```
casret, line 37: Same here.

components/slack/slack.app.js
```diff
@@ -3,6 +3,243 @@ const { WebClient } = require('@slack/web-api')
...
+          context: { types, cursor: resp.cursor, userNames },
+        }
+      },
+    },
+    group: {
+      type: "string",
+      label: "Group",
+      //description: "Select one or more channels or DM conversations to monitor for new messages.",
+      async options({ prevContext }) {
+        let { types, cursor, userNames } = prevContext
```
casret, line 80: usernames not used, and in this case we are taking the hit of grabbing it for no reason.

components/slack/slack.app.js
```diff
@@ -3,6 +3,243 @@ const { WebClient } = require('@slack/web-api')
...
+            for (const user of (await this.users()).users) {
+              userNames[user.id] = user.name
+            }
+          }
+        }
+        const resp = await this.availableConversations(types.join(), cursor)
+        return {
+          options: resp.conversations.map((c) => {
+            if (c.is_im) {
+              return { label: `Direct messaging with: @${userNames[c.user]}`, value: c.id }
```
casret, line 127: A lookup here may be unavoidable if c.name doesn't return us anything useful.  However I think doing an on demand users.info rather than paging through the entire list probably make more sense for most use cases.

casret (changes requested):
A review on the slack app.  There are some issues around the user lookup table that is probably causing the long async options times that you mentioned.

PR #1055, opened 2021-03-05
Google Sheets Initial Events

jverce:
Emit a small sample of events right after the event source has been created


components/google_sheets/sources/common.js
```diff
@@ -16,6 +16,14 @@ module.exports = {
       },
     },
     watchedDrive: { propDefinition: [google_sheets, "watchedDrive"] },
+    initialEventCount: {
```
dylburger, line 19: We actually don't need to make this a prop or make it optional. You can simply emit the events (assuming the user has records in the specified worksheet). With Zapier and other platforms, this is the default, too, so the user doesn't have to think about it - they just get data.

jverce: I transformed it to a constant instead.

dylburger (approved):

PR #1072, opened 2021-03-09
Discourse sources

dylburger:


components/discourse/discourse.app.js
```diff
@@ -0,0 +1,192 @@
...
+          value: c.id,
+        }));
+        const options = sortBy(rawOptions, ["label"]);
+
+        return {
+          options,
+        };
+      },
+    },
+    eventTypes: {
```
dylburger, line 32: This prop and the two that follow aren't used yet, but I'm planning to create some components where they will be.

components/discourse/common.js
```diff
@@ -0,0 +1,61 @@
...
+        );
+      }
+    },
+    generateMeta() {
+      throw new Error("Generate meta is not implemented");
+    },
+    validateEventAndEmit(event, eventName, key) {
+      const { body, headers } = event;
+      if (isEmpty(headers) || !headers["x-discourse-event-signature"]) {
+        console.log("Discourse signature header not present. Exiting");
```
tjk, line 46: Maybe this should throw? Seems on the same level as the sig check failing vs. the event name filtering.

components/discourse/discourse.app.js
```diff
@@ -0,0 +1,192 @@
...
+  type: "app",
+  app: "discourse",
+  propDefinitions: {
+    categories: {
+      type: "string[]",
+      label: "Categories",
+      description:
+        "The Discourse categories you want to watch for changes. **Leave blank to watch all categories**.",
+      async options() {
+        // Categories endpoint does not implement pagination
```
tjk, line 16: You should do the only first query does work then -- ie. https://github.com/PipedreamHQ/pipedream/blob/master/components/ringcentral/sources/new-event/new-event.js#L32

components/discourse/discourse.app.js
```diff
@@ -0,0 +1,192 @@
...
+        return data;
+      }
+
+      // No way to filter items by category via API, so we filter here
+      return data.filter(
+        (el) => el.category_id && categories.includes(el.category_id.toString())
+      );
+    },
+    async _makeRequest(opts) {
+      if (!opts.headers) {
```
tjk, line 108:
```
if (!opts.headers) opts.headers = {}
// always set Content-Type and Accept like the other headers
```

tjk (approved):
Please at least address the categories async options please, but nice work!

components/discourse/discourse.app.js
dylburger: good call, fixed

components/discourse/discourse.app.js
dylburger: There are no cases we need to override currently. They do allow for URL-encoded form data, as well, but I see no reason why the source needs to use that. Fixed.

PR #1085, opened 2021-03-12
Google Drive: New Event Sources

michelle0927:


components/google_drive/sources/new-or-modified-comments/new-or-modified-comments.js
```diff
@@ -0,0 +1,185 @@
...
+// https://developers.google.com/drive/api/v3/push and here:
+// https://developers.google.com/drive/api/v3/manage-changes
+//
+// This source has two interfaces:
+//
+// 1) The HTTP requests tied to changes in the user's Google Drive
+// 2) A timer that runs on regular intervals, renewing the notification channel as needed
+
+const { uuid } = require("uuidv4");
+const includes = require("lodash.includes");
```
jverce, line 12: The usage of these per-method packages is [discouraged as per the Lodash docs](https://lodash.com/per-method-packages).

This should fix it:
```suggestion
const includes = require("lodash/includes");
```

components/google_drive/sources/new-or-modified-comments/new-or-modified-comments.js
```diff
@@ -0,0 +1,185 @@
...
+// implementing strategy enumerated in the Push Notifications API docs:
+// https://developers.google.com/drive/api/v3/push and here:
+// https://developers.google.com/drive/api/v3/manage-changes
+//
+// This source has two interfaces:
+//
+// 1) The HTTP requests tied to changes in the user's Google Drive
+// 2) A timer that runs on regular intervals, renewing the notification channel as needed
+
+const { uuid } = require("uuidv4");
```
jverce, line 11: The [documentation of this package](https://github.com/thenativeweb/uuidv4#please-note) says the following:
> This module will be deprecated in the future in favour of module [uuid](https://www.npmjs.com/package/uuid). Most of the functionality of this module is already included in `uuid` since version `8.3.0`, so most of the functions of this module have already been marked as deprecated.

This should do the trick:
```suggestion
const { v4: uuid } = require("uuid");
```

components/google_drive/sources/new-or-modified-comments/new-or-modified-comments.js
```diff
@@ -0,0 +1,185 @@
...
+            changed: headers["x-goog-changed"], // "Additional details about the changes. Possible values: content, parents, children, permissions"
+          },
+        };
+
+        this.$emit(eventToEmit, {
+          summary: comment.content,
+          id: headers["x-goog-message-number"],
+        });
+      }
+      this.db.set(file.id, maxModifiedTime);
```
jverce, line 182: If we extract this into a method, we can basically replace the rest of the code in this file by extending the `new-or-modified-files` event source, correct?

jverce (changes requested):
`new-or-modified-comments` is very similar to `new-or-modified-folders`, so the comments made in the former also apply to the latter.

In summary:
1. The imported packages are soon to be deprecated, but refactoring that should be pretty straightforward
2. The 2 event sources mentioned above are almost identical to `new-or-modified-files`, so we should probably extract the common functionality like we did for other event sources.

jverce (approved):

Issue #1013, opened 2021-03-01
[SOURCE] Twitter: New Event Sources


michelle0927:
**Describe the event source. What app, what event?**
Twitter

**New Tweet in List** | Triggers when a new tweet is created in a specific list.
**Watch Retweets** | Triggers when someone retweets a tweet from a user being watched.
**New Message** | Triggers when a new message is received.

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
https://developer.twitter.com/en/docs

PR #1113, opened 2021-03-25
New event sources for Twitter Developer App

jverce:
* Add event source logic for New Message
* Fix the logic of the New Tweet Metrics logic to switch from the
  deprecated API to the new V2 API

Part of #1013


components/twitter_developer_app/sources/new-message/new-message.js
```diff
@@ -0,0 +1,76 @@
...
+      const {
+        id,
+        created_timestamp: ts,
+        message_create: { message_data: { text = "" } },
+      } = data;
+      const maxTextLength = 12;
+      const trimmedText = text.length > maxTextLength
+        ? `${text.slice(0, maxTextLength)}...`
+        : text;
+      const summary = `New message: ${trimmedText}`;
```
dylburger, line 49: Since the source itself emits new messages, I think you can remove the "New message: " prefix and just emit the text.

components/twitter_developer_app/sources/new-message/new-message.js
```diff
@@ -0,0 +1,76 @@
...
+    _getNextCursor() {
+      return this.db.get("nextCursor");
+    },
+    _setNextCursor(nextCursor) {
+      this.db.set("nextCursor", nextCursor);
+    },
+    _isRelevant(directMessage) {
+      return (
+        directMessage.type === "message_create" &&
+        directMessage.message_create.sender_id != this._getAccountId()
```
dylburger, line 32: Will the `message_create` intermediate object ever not exist?

components/twitter_developer_app/twitter_developer_app.app.js
```diff
@@ -1,4 +1,139 @@
+const TwitterV1 = require("twit");
+const TwitterV2 = require("twitter-v2");
+
 module.exports = {
   type: "app",
   app: "twitter_developer_app",
-}
+  propDefinitions: {
+    count: {
```
dylburger, line 8: Is this prop used anywhere?

dylburger (approved):
A few small comments. lgtm otherwise!

components/twitter_developer_app/twitter_developer_app.app.js
jverce: Nope, I'll remove it

components/twitter_developer_app/sources/new-message/new-message.js
jverce: It will if the type of the message is `message_create`, but I'll add a check just to be safe

components/twitter_developer_app/sources/new-message/new-message.js
jverce: Will do!

PR #1140, opened 2021-04-01
Update app for Google Sheets sources

psavkar:


components/google_sheets/google_sheets.app.js
```diff
@@ -2,7 +2,8 @@ const { google } = require("googleapis");
 const google_drive = require("../google_drive/google_drive.app");
 
 module.exports = {
-  ...google_drive,
+  type: "app",
+  app: "google_sheets",
```
jverce, line 6: We still want to inherit other properties, and only override the `app` attribute in this case.
This will achieve that:
```suggestion
  ...google_drive,
  app: "google_sheets",
```

jverce (changes requested):

jverce (approved):

PR #1147, opened 2021-04-05
Youtube: Limit results on first run

michelle0927:


components/youtube/sources/common.js
```diff
@@ -0,0 +1,71 @@
...
+  },
+  async run(event) {
+    let publishedAfter = this._getPublishedAfter();
+    const componentParams = this.getParams();
+    const params = {
+      ...componentParams,
+      part: "snippet",
+      type: "video",
+      order: "date",
+    };
```
jverce, line 65: Be aware that having `...componentParams` at the top will make it susceptible to be overriden, which I guess it's not what you want.

Is this what you actually want maybe?
```js
module.exports = {
  async run() {
    const baseParams = {
      part: "snippet",
      type: "video",
      order: "date",
    };
    const componentParams = this.getParams();
    const params = {
      ...baseParams,
      ...componentParams,
    };
    // ...
  },
};
```

components/youtube/sources/new-videos-by-username/new-videos-by-username.js
```diff
@@ -1,84 +1,57 @@
...
-        intervalSeconds: 60 * 15,
-      },
+  },
+  methods: {
+    ...common.methods,
+    getParams(channelId) {
+      return {
+        part: "snippet",
+        type: "video",
+        order: "date",
```
jverce, line 24: This is not needed, right?

components/youtube/sources/new-videos-by-username/new-videos-by-username.js
```diff
@@ -1,84 +1,57 @@
...
-        results.data.items.forEach(function (video) {
-          videos.push(video);
-          count++;
-        });
-        if (!nextPageToken) break;
-        if (!results.data.items || results.data.items.length < 1) break;
-      }
+      const params = this.getParams(channelId);
+      if (publishedAfter) params.publishedAfter = publishedAfter;
+      else params.maxResults = 10;
```
jverce, line 46: Is this logic in place so that we send the initial events?
Because since we're overriding the `run()` method here, we kind of know that `publishedAfter` will be set, unless it's the very first execution.

I think it might be cleaner to have this logic in the hooks of the `common` module perhaps?
```js
// sources/common.js

module.exports = {
  hooks: {
    async deploy() {
      const params = {
        ...this._getBaseParams(),
        maxResults: 10,
        ...this.getParams(),
      };
      // Paginate through videos...
    },
    async activate() {
      this._setPublishedAfter(Date.now());
    },
  },
  methods: {
    _getBaseParams() {
      return {
        part: "snippet",
        type: "video",
        order: "date",
      };
    },
  },
  async run() {
    const publishedAfter = this._getPublishedAfter();
    const params = {
      ...this._getBaseParams(),
      publishedAfter,
      ...this.getParams(),
    };
    // Do the rest...
  },
};
```

jverce (changes requested):
Nice cleanup!
Made a few comments, let me know what you think

components/youtube/sources/common.js
```diff
@@ -0,0 +1,71 @@
...
+        if (!results.items || results.items.length < 1) break;
+        if (!results.nextPageToken) break;
+        else params.pageToken = results.nextPageToken;
+      }
+      return lastPublished;
+    },
+  },
+  async run(event) {
+    let publishedAfter = this._getPublishedAfter();
+    const componentParams = this.getParams();
```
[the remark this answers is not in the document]
michelle0927: Updated

components/youtube/sources/common.js
michelle0927: Updated

components/youtube/sources/common.js
```diff
@@ -0,0 +1,71 @@
...
+        summary: snippet.title,
+        ts: Date.parse(snippet.publishedAt),
+      };
+    },
+    emitEvent(result) {
+      const meta = this.generateMeta(result);
+      this.$emit(result, meta);
+    },
+    // paginate through results from getVideos(), emit each video, and return the
+    // most recent date that a video was published
```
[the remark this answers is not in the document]
michelle0927: Updated

components/youtube/sources/new-videos-by-username/new-videos-by-username.js
michelle0927: Moved to _getBaseParams()

components/youtube/sources/new-videos-by-username/new-videos-by-username.js
michelle0927: The purpose of this PR is to limit the videos pulled on the first run.

> On the initial run of the trigger, we attempt to fetch a number of sample videos for the specified user, and the user you specified had so many videos that we hit the YouTube quota just from trying to list those sample videos.
> 
> We're optimizing how we fetch these initial videos and should have an approach that keeps you under limits. I'll let you know when we have a new version of the trigger for you to test.

The idea here is to add in maxResults on the first run in order to limit the videos pulled and try to avoid going over Youtube's quota limit. I moved that logic to a deploy hook in common.js and overwrote it for components that don't need it or need different logic in the deploy.

jverce (approved):

PR #1164, opened 2021-04-12
Twilio: New Event Sources

michelle0927:


components/twilio/sources/new-call/new-call.js
```diff
@@ -0,0 +1,29 @@
...
+  methods: {
+    ...common.methods,
+    getWebhookFn() {
+      return this.twilio.setIncomingCallWebhookURL.bind(this);
+    },
+    generateMeta(body, headers) {
+      return {
+        /** if Twilio retries a message, but we've already emitted, dedupe */
+        id: headers["i-twilio-idempotency-token"],
+        summary: `New call from ${body.From}`,
```
jverce, line 20: **-Minor-**

This will expose phone numbers, which can present compliance issues for users.
I wrote [this method](https://github.com/PipedreamHQ/pipedream/blob/master/components/ringcentral/sources/common/base.js#L21) for a similar event source for RingCentral that might be of interest to you.

components/twilio/sources/common-webhook.js
```diff
@@ -0,0 +1,94 @@
...
+          this.authToken,
+          twilioSignature,
+          /** This must match the incoming URL exactly, which contains a / */
+          `${this.http.endpoint}/`,
+          body
+        )
+      ) {
+        throw new Error(
+          "Computed Twilio signature doesn't match signature received in header"
+        );
```
jverce, line 64: Shouldn't this return `false`?
Or actually, just forward the call to the Twilio client:
```js
module.exports = {
  methods: {
    validateRequest(body, headers) {
      const twilioSignature = headers["x-twilio-signature"];
      if (!twilioSignature) {
        console.log("No x-twilio-signature header in request. Exiting.");
        return false;
      }

      return twilioClient.validateRequest(
        this.authToken,
        twilioSignature,
        `${this.http.endpoint}/`,
        body,
      );
    },
  },
};
```

components/twilio/sources/common-webhook.js
```diff
@@ -0,0 +1,94 @@
...
+        body: responseBody,
+      });
+    }
+
+    if (typeof body !== "object")
+      body = Object.fromEntries(new URLSearchParams(body));
+
+    if (!this.isRelevant(body)) return;
+
+    if (!this.validateRequest(body, headers)) return;
```
jverce, line 90: **-Minor-**

Some logs here? Something like _Event could not be validated. Skipping..._

jverce (changes requested):
Some comments, but otherwise LGTM

components/twilio/sources/common-polling.js
```diff
@@ -0,0 +1,42 @@
...
+    emitEvent(result) {
+      const meta = this.generateMeta(result);
+      this.$emit(result, meta);
+    },
+  },
+  async run(event) {
+    const resourceFn = this.getResourceFn();
+    let createdAfter = this._getCreatedAfter();
+    const params = {};
+    if (createdAfter) params.dateCreatedAfter = createdAfter;
```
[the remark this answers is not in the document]
michelle0927: Updated

components/twilio/sources/new-call/new-call.js
```diff
@@ -0,0 +1,29 @@
...
+  name: "New Call (Instant)",
+  description:
+    "Configures a webhook in Twilio, tied to a phone number, and emits an event each time a call to that number is completed",
+  version: "0.0.1",
+  dedupe: "unique",
+  methods: {
+    ...common.methods,
+    getWebhookFn() {
+      return this.twilio.setIncomingCallWebhookURL.bind(this);
+    },
```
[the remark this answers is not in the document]
michelle0927: Updated

components/twilio/sources/new-call/new-call.js
michelle0927: Thanks! Updated.

components/twilio/sources/common-webhook.js
michelle0927: Updated to forward the call to the Twilio client

components/twilio/sources/common-webhook.js
michelle0927: Added logs

jverce (approved):
LGTM!

Issue #1024, opened 2021-03-01
[SOURCE] Mailgun: New Event Sources


michelle0927:
**Describe the event source. What app, what event?**
Mailgun

- [x] **New Unsubscribe Event** | Triggers when an Unsubscribe event occurs.
- [x] **New Complaint Event** | Triggers when the Spam Complaints event occurs.
- [x] **New List** | Triggers when a new mailing list is added to your account.
- [x] **New Open Event** | Triggers when an Open event occurs.
- [x] **New Click Event** | Triggers when a Click event occurs.
- [x] **New Bounce Event** | Triggers when a Bounce event occurs.
- [x] **New Log Data** | Triggers for most actions within your account.
- [x] **New Failed Delivery Event** | Triggers when a Failed Delivery Attempts event occurs.
- [x] **New Delivery Event** | Triggers when a Successful Delivery event occurs.
- [x] **New Temporary Failure Event** | Triggers when a Temporary Failure event occurs.
- [x] **New Permanent Failure Event** | Triggers when a Permanent Failure event occurs.

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
https://documentation.mailgun.com/en/latest/api_reference.html

PR #1172, opened 2021-04-15
[SOURCE] Mailgun: New Event Sources #1024

sergioeliot2039:
Request for pulling mailgun assets. Main app file mailgun.app.js and common webhook logic file has an initial implementation, new-delivery as first implemented source. thanks


components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,153 @@
...
+      });
+      const opts = {
+        method: "put",
+        headers: {
+          "Content-Type": `multipart/form-data; boundary=${data._boundary}`,
+        },
+        path: `v3/domains/${mailgunDomain}/webhooks/${webhookName}`,
+        data,
+      };
+      return await this._withRetries(() => this._makeRequest(opts));
```
jverce, line 142: Any reason why we're not using the [official Mailgun SDK](https://github.com/mailgun/mailgun-js)?
Using an official SDK should is preferrable unless there's a good reason not to (e.g. flexibility).

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,92 @@
...
+      const crypto = require("crypto");
+      const encodedToken = crypto
+        .createHmac("sha256", signingKey)
+        .update(timestamp.concat(token))
+        .digest("hex");
+      return encodedToken === signature;
+    },
+    emitEvent(eventWorkload) {
+      const eventPayload = eventWorkload["event-data"];
+      if (eventTypes.indexOf(eventPayload.event) > -1) {
```
jverce, line 31: This is not defined anywhere.
Did you see any error logs while testing this?

BTW, there's a specific method for this case:
```suggestion
      if (eventTypes.includes(eventPayload.event)) {
```

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,92 @@
...
+        webhookDetails.urls &&
+        webhookDetails.urls.length > 1
+      ) {
+        const currentWebhookUrls = webhookDetails.urls.slice();
+        const newWebhookUrls = [];
+        currentWebhookUrls.forEach((url) => {
+          if (!(url === this.http.endpoint)) {
+            newWebhookUrls.push(url);
+          }
+        });
```
jverce, line 77: **-Minor-**

This can be simplified:
```suggestion
        const newWebhookUrls = currentWebhookUrls
          .filter(url => url !== this.http.endpoint);
```

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,92 @@
...
+      } else {
+        await this.mailgun.deleteWebhook(this.domain, webhookName);
+      }
+    },
+  },
+  async run(eventWorkload) {
+    const { timestamp, token, signature } = eventWorkload.signature;
+    const eventPayload = eventWorkload["event-data"];
+    if (!verify(this.webhookSigningKey, timestamp, token, signature)) {
+      throw new Error("signature mismatch");
```
jverce, line 88: The main reason why a verification would fail is because some other source is calling this endpoint.
The usual way to handle this is to just skip the event and return a 404 error.

An exception here can cause unintended issues, like triggering an alarm on the customer's side, so we should be careful.

components/reddit/reddit.app.js
```diff
@@ -252,3 +252,17 @@ module.exports = {
...
+		    params: {
+		    	after:after_link
+		    },
+          });
+         
+  		return newSubredditLinks;
+
+        },
+    },
+  };
```
jverce, line 268: This is legacy code from the rebase, right?

components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
```diff
@@ -0,0 +1,68 @@
...
+        if(new_after){
+            if(reddit_things.data.data.children.length>0){
+                reddit_things.data.data.children.forEach(reddit_link => {
+                    this.emitMeInfo(reddit_link);
+                });
+                this.db.set("after",new_after);
+            }                        
+        }                		
+    },
+};
```
jverce, line 68: Same here

jverce (changes requested):
It'd probably be a good idea to squash all these commits into a single one, because if someone accidentally merges all of them it will pollute the history in the main repo/branch.

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,92 @@
...
+    webhookSigningKey: {
+      type: "string",
+      secret: true,
+      label: "Mailgun webhook signing key",
+      description:
+<<<<<<< HEAD
+        "Your Mailgun webhook signing key, found [in your Mailgun dashboard](https://app.mailgun.com/app/dashboard), located under Settings on the left-hand nav and then in API Keys look for webhook signing key. Required to compute the authentication signature of events. ",
+=======
+        "Your Mailgun webhook signing key, found [in your Mailgun dashboard](https://app.mailgun.com/app/dashboard), located under Settings on the left-hand nav and then in API Keys look for webhook signing key. Required to compute the authentication signature of events.",
+>>>>>>> 3894ceaeca041938f51947f077049a80a007ebaf
```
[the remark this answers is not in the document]
sergioeliot2039: Yep, weird. Looks like the conflict resolution wasn't saved and didn't make it in the commit.

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,153 @@
...
+      default: "US",
+      options: ["US", "EU"],
+    },
+  },
+  methods: {
+    _apiKey() {
+      return this.$auth.api_key;
+    },
+    _apiUrl() {
+      return `https://api.mailgun.net`; //TODO: How to handle base  URL for both Mailgun regions US, and EU.
```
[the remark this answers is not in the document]
sergioeliot2039: Yes, I was thinking in deduce the subdomain as well.

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,153 @@
...
+    },
+    async _makeRequest(opts) {
+      const credentials = "api" + ":" + this._apiKey();
+      const credentialsBuffer = Buffer.from(credentials);
+      const base64Credentials = credentialsBuffer.toString("base64");
```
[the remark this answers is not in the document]
sergioeliot2039: Right.

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,153 @@
...
+      opts.headers.Authorization = `Basic ${base64Credentials}`;
+      opts.headers["user-agent"] = "@PipedreamHQ/pipedream v0.1";
+      const { path } = opts;
+      delete opts.path;
+      opts.url = `${this._apiUrl()}${path[0] === "/" ? "" : "/"}${path}`;
+      return (await axios(opts)).data;
+    },
+    _isRetriableStatusCode(statusCode) {
+      [402, 429, 500, 502, 503, 504].includes(statusCode);
+      //TODO: Remove this comment. For code review purposes, I am linking to Mailgun's HTTP status code returned: https://documentation.mailgun.com/en/latest/api-intro.html?highlight=502#status-codes Errro 408, common in other Pipedream retries documentedly not returned by Mailgun, so it was not included.
```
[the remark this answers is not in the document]
sergioeliot2039: No, I didn't forget but added it for discussion here on the review. Ok, I will modify and leave the docs link.

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,153 @@
...
+      const results = [];
+      let skip = 0;
+      do {
+        domains_set = await this.getDomains(skip);
+        domains_set.items.forEach((domain) => {
+          results.push({
+            name: domain.name,
+          });
+        });
+        skip = skip + 100;
```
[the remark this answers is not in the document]
sergioeliot2039: Right, not sure why I just used the max length of the batch, instead of that in the response.

components/mailgun/mailgun.app.js
sergioeliot2039: Ah, sorry. I see there are a couple of actions already using mailgun SDK, but the Pipedream code to create an action `"Run Node.js code with Mailgun"` is using axios and hitting the API with HTTP request, and some how I based off from that. Should I switch to the SDK then?

components/mailgun/sources/common-webhook.js
sergioeliot2039: Ok

components/reddit/sources/new-link-on-a-subreddit/new-link-on-a-subreddit.js
sergioeliot2039: Yes, legacy from rabse.

components/reddit/reddit.app.js
sergioeliot2039: Right

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,155 @@
...
+      async options(context) {
+        const q = context.query;
+        const options = [];
+        const results = await this.getAllDomains();
+        for (const domains of results) {
+          options.push({
+            label: domains.name,
+            value: domains.name,
+          });
+        }
```
michelle0927, line 22: FYI, if label and value are the same, you can just create an array of the domain names instead of an array of label/value pairs.
```
for (const domains of results) {
  options.push(domains.name);
}
```

components/mailgun/mailgun.app.js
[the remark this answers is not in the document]
sergioeliot2039: Left the docs link.

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,192 @@
...
+        return options;
+      },
+    },
+    baseRegion: {
+      type: "string",
+      label: "Base region",
+      description:
+        'The region that Mailgun will use to send and receive email, which determines the URL to use when requesting Mailgun API. This must be the same region where "domain" was created.',
+      default: "US",
+      options: ["US", "EU"],
```
dylburger, line 30: Are these the only regions Mailgun supports?

components/mailgun/sources/new-log-data/new-log-data.js
```diff
@@ -0,0 +1,81 @@
+const common = require("../common-webhook");
```
dylburger, line 1: Since this is a timer-based source, do you want to require everything from `common-webhook`?

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,192 @@
...
+      return require("mailgun-js")({ apiKey, domain, host });
+    },
+    _apiKey() {
+      return this.$auth.api_key;
+    },
+    _apiHost() {
+      const subdomain = this.baseRegion == "EU" ? "api.eu" : "api";
+      return `${subdomain}.mailgun.net`;
+    },
+    async _makeRequest(opts) {
```
dylburger, line 52: To reduce code, I'd recommend destructuring `opts` here:

```javascript
const { domain, method, path, params } = opts
```

so that you can reference the vars directly.

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,192 @@
...
+      const mailgun = this.mailgun(
+        this._apiKey(),
+        opts.domain,
+        this._apiHost()
+      );
+      switch (opts.method) {
+        case "get":
```
dylburger, line 59: Can you replace all instances of lowercase `get` with `GET`, `post` with `POST`, etc.? Small style thing but even though `axios` allows for the lowercase versions, we like to use the capitalized versions to align with the HTTP spec

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,192 @@
...
+      );
+      return get(webhooks, `webhooks.${webhookName}`);
+    },
+    async getAllDomains() {
+      let domains_set;
+      const results = [];
+      let skip = 0;
+      do {
+        domains_set = await this.getDomains(skip);
+        domains_set.items.forEach((domain) => {
```
dylburger, line 122: You can shorten this code to something like:

```javascript
results.push(...domains_set.items.map(domain => domain.name))
```

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,192 @@
...
+    timer: {
+      label: "Polling schedule",
+      description: "Pipedream polls Reddit for events on this schedule.",
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15, // by default, run every 15 minutes.
+      },
+    },
+  },
+  methods: {
```
dylburger, line 41: For methods in the app file, we're trying to document these well for future devs. IMO there's no need to document private methods like `_apiKey()`, but it would be useful for others. Could you take a pass at adding [JS Docs](https://github.com/PipedreamHQ/pipedream/blob/master/COMPONENT-GUIDELINES.md#methods) to the core, public methods?

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+    verify(signingKey, timestamp, token, signature) {
+      const crypto = require("crypto");
+      const encodedToken = crypto
+        .createHmac("sha256", signingKey)
+        .update(timestamp.concat(token))
+        .digest("hex");
+      return encodedToken === signature;
+    },
+    emitEvent(eventPayload) {
+      const eventTypes = this.getEventType();
```
dylburger, line 26: Can you add a `getEventType` method stub here that throws a "not implemented" error? e.g. see https://github.com/PipedreamHQ/pipedream/blob/7f03b49859616f3576d0c3d24d24816560962770/components/snowflake/sources/common.js#L67

components/mailgun/sources/new-bounce/new-bounce.js
```diff
@@ -0,0 +1,50 @@
...
+  props: {
+    ...common.props,
+    domain: { propDefinition: [mailgun, "domain"] },
+  },
+  methods: {
+    ...common.methods,
+    getEventName() {
+      return ["bounce"];
+    },
+    getEventType() {
```
dylburger, line 21: These need to return arrays in each of the components, like `["bounced"]`. The way you use it in `emitEvent`:

```javascript
const eventTypes = this.getEventType();
if (eventTypes.includes(eventPayload.event)) {
```

input like "bounce" or "bou" will also return true, since that's the way the `includes` method works for strings. If you return an array, `includes` demands that the argument be an element of the array, which means only "bounces" will match.

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+      if (eventTypes.includes(eventPayload.event)) {
+        const meta = this.generateMeta(eventPayload);
+        this.$emit(eventPayload, meta);
+      }
+    },
+  },
+  hooks: {
+    async activate() {
+      const webhookNames = this.getEventName();
```
dylburger, line 35: Same comment re: the method stub as for `getEventType` above

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+      for (let i = 0; i < webhookNames.length; i++) {
```
dylburger, line 36: a `for ... of` loop will save you one line and I personally prefer them in almost all cases where I'm iterating over an array now:

```javascript
for (const webhookName of webhookNames) {
  // remaining code
}

File: components/mailgun/sources/common-webhook.js [hunk 23]
```diff
@@ -0,0 +1,97 @@
...
+        const webhookName = webhookNames[i];
+        const webhookDetails = await this.mailgun.getWebhookDetails(
+          this.domain,
+          webhookName
+        );
+        if (webhookDetails && webhookDetails.urls) {
```
dylburger, line 42: If you use `lodash/get` and run:

```javascript
const urls = get(webhookDetails, 'urls', [])
```

Then you can avoid the `slice` below, and simply push to `urls`:

```javascript
urls.push(this.http.endpoint)
```

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+            this.domain,
+            webhookName,
+            this.http.endpoint
+          );
+        }
+      }
+    },
+    async deactivate() {
+      const webhookNames = this.getEventName();
+      for (let i = 0; i < webhookNames.length; i++) {
```
dylburger, line 61: Same comment re: `for ... of` loop above

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+        const webhookName = webhookNames[i];
+        const webhookDetails = await this.mailgun.getWebhookDetails(
+          this.domain,
+          webhookName
+        );
+        if (
```
dylburger, line 67: Prefer `lodash/get` here, as above

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+            webhookName,
+            newWebhookUrls
+          );
+        } else {
+          await this.mailgun.deleteWebhook(this.domain, webhookName);
+        }
+      }
+    },
+  },
+  async run(eventWorkload) {
```
dylburger, line 87: This is a small nit, but by convention, the parameter passed to `run()` should be called `event`

components/mailgun/sources/common-webhook.js
```diff
@@ -0,0 +1,97 @@
...
+    const { timestamp, token, signature } = eventWorkload.body.signature;
```
dylburger, line 88: You should check that `eventWorkload.body.signature` is set before destructuring, and bail early if not set:

```
const sig = get(event, ['body', 'signature'])
if (!sig) {
  console.log("No signature present in event")
  return
}

File: components/mailgun/sources/new-bounce/new-bounce.js [hunk 28]
```diff
@@ -0,0 +1,50 @@
+const common = require("../common-webhook");
+const { mailgun } = common.props;
```
dylburger, line 2: Is this used? It looks like you destructure `...common.props` within `props` already and I don't see a reference to it elsewhere.

components/mailgun/sources/new-bounce/new-bounce.js
```diff
@@ -0,0 +1,50 @@
...
+
+module.exports = {
+  ...common,
+  key: "new-bounce",
```
dylburger, line 6: These `keys` should read `mailgun-new-bounce` (keys should always be preceded by the app name)

components/mailgun/sources/new-bounce/new-bounce.js
```diff
@@ -0,0 +1,50 @@
...
+  name: "New bounce",
+  description:
+    "Emit an event when the email recipient who could not be reached.",
```
dylburger, line 9: remove the "who" in this phrase

components/mailgun/sources/new-bounce/new-bounce.js
```diff
@@ -0,0 +1,50 @@
+const common = require("../common-webhook");
+const { mailgun } = common.props;
+
+module.exports = {
+  ...common,
+  key: "new-bounce",
+  name: "New bounce",
```
dylburger, line 7: Capitalize "bounce"

components/mailgun/sources/new-bounce/new-bounce.js
```diff
@@ -0,0 +1,50 @@
...
+      return ["bounce"];
+    },
+    getEventType() {
+      return "bounced";
+    },
+    generateMeta(eventPayload) {
+      const ts = eventPayload.timestamp;
+      return {
+        id: `${eventPayload["X-Mailgun-Sid"]}${ts}`,
+        summary: `New Bounce on message id: ${eventPayload["Message-Id"]} by ${eventPayload.recipient}`,
```
dylburger, line 28: In general I like a `summary` that's short and doesn't duplicate any information. The user created a New Bounce source, so we can leave out "New Bounce on message ID". I think the most useful information here is the `recipient`, so I would actually only include that in this case and also remove the `Message-Id`.

components/mailgun/sources/new-bounce/new-bounce.js
```diff
@@ -0,0 +1,50 @@
...
+    emitEvent(eventPayload) {
+      const eventTypes = this.getEventType();
+      if (eventTypes.includes(eventPayload.event)) {
+        const meta = this.generateMeta(eventPayload);
+        this.$emit(eventPayload, meta);
+      }
+    },
+  },
+  async run(eventRawData) {
+    const eventWorkload = eventRawData.body;
```
dylburger, line 41: This looks almost the same as the `run` method from the `common-webhook` file. Is there a reason it's unique?

components/mailgun/sources/new-click/new-click.js
```diff
@@ -0,0 +1,33 @@
+const common = require("../common-webhook");
+const { mailgun } = common.props;
+
+module.exports = {
+  ...common,
+  key: "new-click",
```
dylburger, line 6: Same comment re: preceding with `mailgun-`

components/mailgun/sources/new-click/new-click.js
```diff
@@ -0,0 +1,33 @@
...
+  name: "New click",
```
dylburger, line 7: Same comment re: capitalizing "click"

components/mailgun/sources/new-click/new-click.js
```diff
@@ -0,0 +1,33 @@
...
+    ...common.props,
+    domain: { propDefinition: [mailgun, "domain"] },
+  },
+  methods: {
+    ...common.methods,
+    getEventName() {
+      return ["clicked"];
+    },
+    getEventType() {
+      return "CLICKED";
```
dylburger, line 22: Do you know if there's a reason Mailgun capitalizes only this event name, and not the others?

components/mailgun/sources/new-click/new-click.js
```diff
@@ -0,0 +1,33 @@
...
+    },
+    generateMeta(eventPayload) {
+      const ts = eventPayload.timestamp;
+      return {
+        id: `${eventPayload.id}${ts}`,
+        summary: `New Click on message id: ${eventPayload.message.headers["message-id"]} by ${eventPayload.recipient}`,
```
dylburger, line 28: Same feedback here re: shortening the summary to include only the most important, human-readable information. Here, I'd also only include the recipient

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,192 @@
...
+      description: "Domain used to send and receive email.",
+      async options(context) {
+        const q = context.query;
+        const options = [];
+        const results = await this.getAllDomains();
+        for (const domains of results) {
+          options.push(domains.name);
+        }
+        return options;
+      },
```
michelle0927, line 22: Is q used here? I think you can shorten this:
```
async options() {
  const domains = await this.getAllDomains();
  return domains.map((domain) => domain.name);
}
```

components/mailgun/sources/new-mailing-list/new-mailing-list.js
```diff
@@ -0,0 +1,67 @@
...
+      mailgunListItems.forEach(this.emitEvent);
+      this.db.set("next", mailgunLists.paging.next);
+    },
+  },
+  methods: {
+    ...common.methods,
+    generateMeta(eventPayload) {
+      const ts = moment(eventPayload.created_at).unix();
+      return {
+        id: `${ts}`,
```
dylburger, line 36: You can just reference `ts`

components/mailgun/sources/new-mailing-list/new-mailing-list.js
```diff
@@ -0,0 +1,67 @@
...
+      }
+      const { items: mailgunListItems = [] } = mailgunLists;
+      mailgunListItems.forEach(this.emitEvent);
+      this.db.set("next", mailgunLists.paging.next);
+    },
+  },
+  methods: {
+    ...common.methods,
+    generateMeta(eventPayload) {
+      const ts = moment(eventPayload.created_at).unix();
```
dylburger, line 34: Is `eventPayload.created_at` an ISO 8601 timestamp? You can actually avoid `moment` and use just:

```
+new Date(eventPayload.created_at);
```

components/reddit/reddit.app.js
```diff
@@ -252,3 +252,17 @@ module.exports = {
     },
   },
 };
+  	  async getNewSubredditLinks(after_link, subreddit){
```
dylburger, line 255: Did you mean to push these changes to the Reddit branch?

components/mailgun/sources/new-log-data/new-log-data.js
```diff
@@ -0,0 +1,81 @@
...
+      }
+      mailgunEvents = await this.mailgun.getMailgunEvents(
+        this.domain,
+        nextUrlPart,
+        10
+      );
+      if (mailgunEvents.items.length === 0) {
+        console.log(`No data available, skipping iteration`);
+        break;
+      }
```
michelle0927, line 75: Is checking that there are no items the right way to do this? I was looking at this part of the docs:
https://documentation.mailgun.com/en/latest/api-events.html#event-polling
It recommends checking the timestamp of the last event on the result page to see if the results are legit. Then if they're not, wait 15 seconds and retry.

components/mailgun/mailgun.app.js
sergioeliot2039: Yes, they only support two regions. Each region has its own API host, which is reflected in the code.

components/mailgun/mailgun.app.js
sergioeliot2039: Good idea, I made the change. I am still getting to use to argument destructuing\

components/mailgun/mailgun.app.js
sergioeliot2039: Yep, this was just replaced.

components/mailgun/mailgun.app.js
sergioeliot2039: All right, I made the change.

components/mailgun/mailgun.app.js
sergioeliot2039: Interesting. I have created documentation for public methods under the mailgun app file.

components/mailgun/sources/common-webhook.js
sergioeliot2039: Yes, it was added.

components/mailgun/sources/new-bounce/new-bounce.js
sergioeliot2039: ah, I see. In the examples that I saw, I thought of using the arrays as in `["bounced"]`  and thought it was because the related service being used to create the component for had more than 1 type. I think it was Github the sample I saw. 

The change was made to return arrays.

components/mailgun/sources/common-webhook.js
sergioeliot2039: This method stub was added to for `getEventName`

components/mailgun/sources/common-webhook.js
sergioeliot2039: This was interesting. I initially tried and I was getting 0-based array indexes instead of the actual values or each element. I went ahead and this time it worked. I don't know why it didn't work when I implemented it initially.

components/mailgun/sources/common-webhook.js
sergioeliot2039: I looked upon the web examples of using `slice` without arguments as to make the copy of the array and found this statement: 

> No competent programmer would ever make code more complicated than it needs to be. 

Yeah it is much clearer using `lodash/get`. 👍

components/mailgun/sources/common-webhook.js
sergioeliot2039: Yep, this was changed already.

components/mailgun/mailgun.app.js
sergioeliot2039: Sorry I need to take that back. I made a concept test with two array on this [workflow ](https://pipedream.com/@sergio/javascript-playground-p_5VCL7by/edit)and it works there. I don't know why it didn't work here. I tried a couple of variations such as declaring an anonymous function and pass it as argument to the map function.

components/mailgun/sources/common-webhook.js
sergioeliot2039: Ok, noted, will be using in upcoming implementations.
I always keep in my that workload/payload analogy where a cargo trailer carries certain load - the workload, but gets pay only for a part of all this load, the payload. Somewhere I saw this being applied to data in the request and hence was using it.

components/mailgun/sources/common-webhook.js
sergioeliot2039: Ok made the change. Renamed the flag to `hasSignature`

components/mailgun/sources/new-bounce/new-bounce.js
sergioeliot2039: ah! similarly, changes to reddit will need to be made too.

components/mailgun/sources/new-bounce/new-bounce.js
sergioeliot2039: Removed.

components/mailgun/sources/new-bounce/new-bounce.js
sergioeliot2039: Capitalized

components/mailgun/sources/new-bounce/new-bounce.js
sergioeliot2039: ok noted. i changed to `summary: eventPayload.recipient,`

components/mailgun/sources/new-bounce/new-bounce.js
sergioeliot2039: The reason why it is unique is that `maigun-new-bounce` is using a deprecated endpoint from Mailgun API that ultimately sends a different data structure when posting to the subscribed webhooks. In new Mailgun API endpoints `signature` and `event-data` are wrapped inside separated object, for `maigun-new-bounce` the signature and event data are inside the same object.

As an alternative we could have only one `run` method from `common-webhook` that asks for the `this.eventName` to see where to look for the signature and body.

In the Github issue for Maigun sources `maigun-new-bounce` this is included regardless of the deprecation status, and is implemented as well by the other services in the no-code database.

components/mailgun/sources/new-click/new-click.js
sergioeliot2039: I just inspected each webhook payload and noticed only `CLICKED` was sent capitalized.

components/mailgun/sources/new-click/new-click.js
sergioeliot2039: Summary was set to `eventPayload.recipient`

components/mailgun/mailgun.app.js
sergioeliot2039: No, `q` is not used. You are right. This was shorten.

components/mailgun/sources/new-mailing-list/new-mailing-list.js
sergioeliot2039: Right

components/mailgun/sources/new-mailing-list/new-mailing-list.js
sergioeliot2039: I checked and it was more of a RFC 2822 Date, but that instruction worked so we don't new moment.

components/reddit/reddit.app.js
sergioeliot2039: No, sorry.  But when I rebased that file had conflict so I had to resolved, and I committed. Was the best to do here? Stash them and not commit/push them?

components/mailgun/sources/new-log-data/new-log-data.js
sergioeliot2039: I don't think checking the timestamp is necessary because that algorithm applies to polling events taking into account a time range. So my thinking is that the timestamp check is to make sure results are trustworthy in that they have data older than the most recent time of the range.

In the documentation, right before the link you provide, it says that when the `last` page is reached and no items are returned, eventually the `next` that was included in said `last` page will do have items, once new events arrive.

> If the range is ascending then events will continue to be recorded but will not show in the current request time range pages that are provided. So after the most recent events have been retrieved and an empty result page has been reached, then requesting next page URL returned with the last page some time later will return events that occurred since then. And this can go on indefinitely.

Initially, when the source is activated, we query Mailgun from 1 days old to current time events, we then skip all the results and go to the last page, so my assumption is that we didn't have to check for any previous events and simply wait for new events to appear in the `next` page that we saved in the source's `db` prop.

Not sure if there are other thoughts on this? @michelle0927 @dylburger @jverce

components/mailgun/sources/new-click/new-click.js
sergioeliot2039: Others sources where prefixed with `mailgun-`

components/mailgun/sources/new-bounce/new-bounce.js
dylburger: Thanks for the explanation, having the separate method makes sense. Can you add this information (essentially the first paragraph of your explanation) as a comment in the code so that's clear?

sergioeliot2039: Ok, this information was added.

components/mailgun/sources/new-log-data/new-log-data.js
sergioeliot2039: Thoughts?

michelle0927: If what you have now is working fine, that's good with me!

sergioeliot2039: Ok, I left it as is.

Issue #1103, opened 2021-03-18
[SOURCE] Microsoft OneDrive: New File


dylburger:
**Describe the event source. What app, what event?**

I'd like to trigger a workflow when a new file is added to my OneDrive drive. Optionally, I want to watch a specific folder. If the folder is blank, I'd like to watch for any new file in my drive.

Otherwise, I'd like the source to operate like the Dropbox - New File source (e.g. I'd like to implement the **Recursive** option, and other props like download URL- in general we should attempt to map these props to their equivalent functions in OneDrive). 

If there's any other OneDrive-specific logic that Zapier, Integromat or other triggers implement that we should here, we should implement that functionality, as well.

dannyroosevelt:
@dylburger do you have any state on this? It doesn't look like it's been reviewed, but do you know if it's in progress?

dannyroosevelt:
@js07 this was originally written by Jay, who is no longer working with us, and Jonathon had done most of the review (he's also not working much w/ us lately), but it should be ready for another review. If there are changes to make here, feel free to just implement them yourself and drop it in "Ready for QA" when you're done.

dannyroosevelt:
@js07 this is ready for release! 

>Added via [Pipedream](https://pipedream.com/@/p_8rCkMkO?o=a&a=github)

dannyroosevelt:
Not sure why this tagged Jacob -- should've tagged @vellames-turing -- this is ready to go

PR #1180, opened 2021-04-21
Microsoft One Drive

jverce:
* Add `.app.js` file for Microsoft One Drive
* Add base logic for event sources based on webhooks and drive items
* Add event source logic for **New File**
* Add event source logic for **New File in Folder**

Partially addresses #1103

### DIFF
--- a/components/microsoft_onedrive/sources/new-file-in-folder/new-file-in-folder.js
+++ b/components/microsoft_onedrive/sources/new-file-in-folder/new-file-in-folder.js
@@ -1,0 +1,46 @@
+const base = require("../new-file/new-file");
+
+module.exports = {
+  ...base,
+  key: "microsoft_onedrive-new-file-in-folder",
+  name: "New File in Folder (Instant)",
+  description: "Emit an event when a new file is added to a specific folder in a OneDrive drive",
+  version: "0.0.1",
+  dedupe: "unique",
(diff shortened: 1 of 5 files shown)


components/microsoft_onedrive/sources/new-file-in-folder/new-file-in-folder.js
```diff
@@ -0,0 +1,46 @@
+const base = require("../new-file/new-file");
```
dylburger, line 1: Can you walk me through the logic of requiring the `new_file` component instead of keeping the common logic in `base` and requiring that?

dylburger (approved):

components/microsoft_onedrive/sources/new-file-in-folder/new-file-in-folder.js
jverce: The `new-file-in-folder` component is a refinement over `new-file`:
- It also scans the drive for new files. This logic is implemented entirely in the `new-file` component and it can be completely reused by the `new-file-in-folder` component
- It adds some logic on top of the "new file" event source logic:
  - Accept an additional user prop (i.e. the "Folder" prop)
  - When a change happens in the drive, only scan for files under the specified folder

So the hierarchy would be something like this:
image

PR #1181, opened 2021-04-21
New Event Sources for Shopify

### LINKED ISSUE #1013: [SOURCE] Twitter: New Event Sources

### LINKED ISSUE #1027: [SOURCE] Shopify: New Event Sources
Shopify

**Updated Customer** | Triggers when a customer's information is updated.
**Updated Order** | Triggers when an order is updated.
**New Paid Order** | Triggers whenever a new purchase is paid.
**New Cancelled Order** | Triggers whenever an order is cancelled.

https://shopify.dev/docs/admin-api/rest/reference

js07:
Add event sources:

- Updated Customer
- Updated Order
- New Paid Order
- New Cancelled Order

Comments:
- Polling is used to get data for events instead of a webhook because [Shopify webhooks](https://shopify.dev/tutorials/manage-webhooks#verifying-webhooks) include an HMAC header generated using the request data and the app's (pipedream's) shared secret. I don't believe the shared secret is accessible in event sources, so webhook requests can't be verified(?).
- Shopify Admin GraphQL API requests (used in only New Paid Order) returns orders with some different fields than the REST API, and [field names are camelCase](https://shopify.dev/docs/admin-api/graphql/reference/orders/order#fields-2021-04) rather than [snake_case](https://shopify.dev/docs/admin-api/rest/reference/orders/order?api%5Bversion%5D=2021-04#properties-2021-04). Should field names from GraphQL API requests be renamed and snake_cased in the emitted event to be consistent with the REST API?

Resolves #1027


components/shopify/sources/new-cancelled-order/new-cancelled-order.js
```diff
@@ -0,0 +1,40 @@
...
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+    shopify,
+  },
+  async run() {
+    const lastUpdatedAt = this.db.get("last_updated_at") || null;
```
jverce, line 20: We're trying to move towards a getter/setter approach when dealing with `$.service.db` to avoid potential typos and leverage encapsulation:

```js
module.exports = {
  methods: {
    _getLastUpdatedDate() {
      return this.db.get("last_updated_at") || null;
    },
    _setLastUpdatedDate(date) {
      this.db.set("last_updated_at", date);
    },
  },
  async run() {
    const lastUpdatedAt = this._getLastUpdatedDate();

    // Do some stuff...

    this._setLastUpdatedDate(results[results.length - 1].updated_at);
  },
};
```

components/shopify/sources/new-paid-order/new-paid-order.js
```diff
@@ -0,0 +1,47 @@
...
+        }
+      }
+      if (!lastTransactionDate) {
+        return false;
+      }
+      const timeFromTransactToOrderUpdate =
+        Date.parse(order.updatedAt) - lastTransactionDate;
+      // If the order was updated long after the last transaction, assume becoming 'paid' was not the cause of update
+      // allow 5 minutes between transaction and order update
+      if (timeFromTransactToOrderUpdate > 1000 * 60 * 5) {
```
jverce, line 34: Where do these 5 minutes come from? Are there any docs that users can refer to?

components/shopify/sources/updated-customer/updated-customer.js
```diff
@@ -0,0 +1,36 @@
...
+    shopify,
+  },
+  async run() {
+    const lastUpdatedAt = this.db.get("last_updated_at") || null;
+    let results = await this.shopify.getCustomers(null, lastUpdatedAt);
+
+    for (const customer of results) {
+      const dedupeId = `${customer.id}-${customer.updated_at}`;
+      this.$emit(customer, {
+        id: dedupeId,
```
jverce, line 26: **-Minor-**
```suggestion
      const id = `${customer.id}-${customer.updated_at}`;
      this.$emit(customer, {
        id,
```

components/shopify/sources/updated-customer/updated-customer.js
```diff
@@ -0,0 +1,36 @@
...
+        summary: `${customer.first_name} ${customer.last_name}`,
```
jverce, line 27: `<opinion>`
First and last names are personal information that might cause issues with users that require to be compliant with something like GDPR. I'm not an expert in this field so take my opinion with a grain of salt, but perhaps we should log something like `Customer updated: 1234` where `1234` is the customer ID?
`</opinion>`

components/shopify/sources/updated-customer/updated-customer.js
```diff
@@ -0,0 +1,36 @@
...
+        ts: Date.now(),
```
jverce, line 28: For this field, `customer.updated_at` would be a more appropiate value.
We use `Date.now()` as a last resort for a timestamp since it's the less likely to reflect the time at which an event happened in the original source.

In order of relevance, these are the values that you can use for this field:
- `customer.updated_at` (use `Date.parse` so that you get a JS timestamp)
- `event.timestamp` (x1000 since this value is expressed in seconds, not milliseconds like JS timestamps)
- `Date.now()`

jverce (changes requested):
Some of my comments apply to multiple event sources but I only made them once to avoid repetition and noise.

Thanks for this PR BTW!! 🙂🤍

components/shopify/sources/new-paid-order/new-paid-order.js
```diff
@@ -29,19 +73,61 @@ module.exports = {
...
+      // (because time from Order Transaction `created_at` to Order `updated_at` would be too large)
+      // The larger interval could cause Orders updated multiple times within the interval to be considered 'paid' twice,
+      // but those Order events would be deduped if there are fewer than 100 paid Orders in 2x timer inverval
+      if (
+        timeFromTransactToOrderUpdate >
+        this._getAllowedTransactToOrderUpdateMs()
+      ) {
         return false;
       }
       return true;
```
jverce: **-Minor-**

```suggestion
      return (
        timeFromTransactToOrderUpdate < this._getAllowedTransactToOrderUpdateMs()
      );
```

components/shopify/sources/new-paid-order/new-paid-order.js
```diff
@@ -1,23 +1,67 @@
...
+      return [
+        "updatedAt",
+        "name",
+        "legacyResourceId", // The ID of the corresponding resource (Order) in the REST Admin API.
+        "transactions.createdAt",
+      ];
+    },
+    getQueryFilter({ updatedAt = null }) {
+      return `financial_status:paid updated_at:>${updatedAt}`;
+    },
```
jverce, line 60: Can you document these 2 methods?
It's likely that they'll be moved to a parent/common component as more and more GraphQL-based components are developed.

Thanks!

jverce (changes requested):
Looks good!

I just made a few **minor** comments, and a request for documentation on some parts.

Also, the linter is failing (we recently put it in place), but it should be fairly straightforward to fix. At the root of the project, run the following commands:

```shell
npm ci
npx eslint --fix components/shopify
```

The linter will check and fix every JS and JSON file in the `components/shopify` directory. The linter checks in the PR are only checking for the modified files, so you only need to run ESLint against the files that you touched, not the entire `components/shopify` directory.

Thanks again! 🙂

jverce (approved):

components/shopify/sources/new-cancelled-order/new-cancelled-order.js
```diff
@@ -0,0 +1,49 @@
...
+    },
+  },
+  async run() {
+    const lastUpdatedAt = this._getLastUpdatedDate();
+    let results = await this.shopify.getOrders(
+      "any",
+      true,
+      null,
+      lastUpdatedAt,
+      "cancelled"
```
jverce, line 34:
```suggestion
      "cancelled",
```

components/shopify/sources/new-paid-order/new-paid-order.js
```diff
@@ -0,0 +1,195 @@
...
+     * Returns the allowed time interval between Order Transaction and Order
+     * update where the Order update can still be considered the result of
+     * becoming 'paid'.
+     *
+     * @returns {number} The number of milliseconds after a Transaction is created
+     */
+    _getAllowedTransactToOrderUpdateMs() {
+      return Math.max(
+        MIN_ALLOWED_TRANSACT_TO_ORDER_UPDATE_MS,
+        2 * DEFAULT_TIMER_INTERVAL_SECONDS * 1000
```
jverce, line 65:
```suggestion
        2 * DEFAULT_TIMER_INTERVAL_SECONDS * 1000,
```

components/shopify/sources/new-paid-order/new-paid-order.js
```diff
@@ -0,0 +1,195 @@
...
+        updatedAt: prevCursor
+          ? null
+          : dateToISOStringWithoutMs(this.shopify.dayAgo()),
+      }),
+      prevCursor,
+    };
+    const orderStream = this.shopify.queryOrders(queryOrderOpts);
+    const relevantOrderIds = [];
+    for await (const orderInfo of orderStream) {
+      const { order, cursor } = orderInfo;
```
jverce, line 179:
```suggestion
      const {
        order,
        cursor,
      } = orderInfo;
```

components/shopify/sources/updated-order/updated-order.js
```diff
@@ -0,0 +1,49 @@
...
+      this.db.set("last_updated_at", date);
+    },
+  },
+  async run() {
+    const lastUpdatedAt = this._getLastUpdatedDate();
+    let results = await this.shopify.getOrders(
+      "any",
+      true,
+      null,
+      lastUpdatedAt
```
jverce, line 33:
```suggestion
      lastUpdatedAt,
```

jverce (commented):
Adding some suggestions to solve the linting issues.

PR #1187, opened 2021-04-22
Calendly: emit Pipedream events for each invitee on events that support multiple invitees

michelle0927:
Updated the deduped meta ID to concatenate both the event ID and invitee ID, and also cleaned up the code.

### DIFF
--- a/components/calendly/sources/common-webhook.js
+++ b/components/calendly/sources/common-webhook.js
@@ -25,7 +25,10 @@ module.exports = {
     },
   },
   methods: {
-    generateMeta(body) {
+    generateMeta() {
+      throw new Error("generateMeta is not implemented");
+    },
+    generateInviteeMeta(body) {
       const eventId = get(body, "payload.event.uuid");
       const inviteeId = get(body, "payload.invitee.uuid");
       const summary = get(body, "payload.event_type.name");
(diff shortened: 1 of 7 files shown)


components/calendly/sources/common-webhook.js
```diff
@@ -0,0 +1,52 @@
...
+      const resp = await this.calendly.createHook(body);
+      this.db.set("hookId", resp.data.id);
+    },
+    async deactivate() {
+      await this.calendly.deleteHook(this.db.get("hookId"));
+    },
+  },
+  methods: {
+    generateMeta(body) {
+      const eventId = get(body, "payload.event.uuid");
```
dylburger, line 29: Are these properties guaranteed to exist for all events? I see the intent behind this change, so that you can use this as a default for the invitee sources and override `generateMeta` only as necessary, but this might yield an empty string for the `id` if another component _doesn't_ override it and `payload.invitee.uuid` (for example) isn't defined.

Even though it's a little more code, I'd recommend having the default `generateMeta` throw a similar error we do elsewhere to warn authors if they fail to implement it:

```javascript
throw new Error("generateMeta is not implemented");
```

Then, you can define an additional method (e.g. `generateInviteeMeta`) here in the common file, and have `generateMeta` in the invitee-related components reference that:

```javascript
methods: {
  ...common.methods,
  generateMeta(body) {
    return this.generateInviteeMeta(body)
  }
}
```

Would that work?

dylburger (approved):
Very small change re: `generateMeta`, but otherwise lgtm. I appreciate you taking the time to clean this up and abstract the common logic. I tested the multiple invitee case and it looks great.

PR #1191, opened 2021-04-24
OneDrive new file event sources

jverce:
* Add **New File in Folder** event source
* Add **New Folder in Folder** event source
* Add **New File of Types in Folder** event source

### DIFF
--- a/components/microsoft_onedrive/sources/common/base.js
+++ b/components/microsoft_onedrive/sources/common/base.js
@@ -3,13 +3,7 @@ const {
   MAX_INITIAL_EVENT_COUNT,
   WEBHOOK_SUBSCRIPTION_RENEWAL_SECONDS,
 } = require("./constants");
-
-function toSingleLineString(multiLineString) {
-  return multiLineString
-    .trim()
-    .replace(/\n/g, " ")
-    .replace(/\s{2,}/g, "");
-}
+const { toSingleLineString } = require("./utils");
 
 module.exports = {
   props: {
@@ -42,9 +36,10 @@ module.exports = {
 
       let eventsToProcess = Math.max(MAX_INITIAL_EVENT_COUNT, 1);
       for await (const driveItem of itemsStream) {
-        if (driveItem.deleted) {
-          // We don't want to process items that were deleted from the drive
-          // since they no longer exist
+        if (!this.isItemTypeRelevant(driveItem)) {
+          // If the type of the item being processed is not relevant to the
+          // event source we want to skip it in order to avoid confusion in
+          // terms of the actual payload of the sample events
           continue;
         }
 
@@ -134,7 +129,11 @@ module.exports = {
           break;
         }
 
-        if (!this.isItemRelevant(value)) {
+        const shouldSkipItem = (
+          !this.isItemTypeRelevant(value) ||
+          !this.isItemRelevant(value)
+        );
+        if (shouldSkipItem) {
           // If the retrieved item is not relevant to the event source, we skip it
           continue;
         }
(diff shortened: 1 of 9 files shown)


components/microsoft_onedrive/sources/common/base.js
```diff
@@ -1,19 +1,13 @@
...
-    .trim()
-    .replace(/\n/g, " ")
-    .replace(/\s{2,}/g, "");
-}
+const { toSingleLineString } = require("./utils");
 
 module.exports = {
   props: {
-    microsoft_onedrive,
+    onedrive,
```
compwright, line 10: The convention has been to keep this name the same as the app file, if it was renamed to deal with eslint rules I would rather override eslint with a comment and keep the consistent naming.

components/microsoft_onedrive/sources/common/utils.js
```diff
@@ -0,0 +1,24 @@
...
+   *   This is a much
+   *   cleaner string
+   * `);
+   *
+   * @param {string}  multiLineString the input string to reformat
+   * @returns a formatted string based on the content of the input argument,
+   * without newlines and multiple spaces
+   */
+  toSingleLineString(multiLineString) {
+    return multiLineString
```
compwright, line 19: Would be a good idea to wrap in String() so it doesn't break if `null` or `undefined` is passed.

components/microsoft_onedrive/sources/new-file-of-types-in-folder/mime-types.js
```diff
@@ -0,0 +1,304 @@
+// A list of the most common MIME types defined in the IANA registry. The list
```
compwright, line 1: Consider using [mime-db](https://www.npmjs.com/package/mime-db) instead (h/t @michelle0927 for pointing this package out to me)

components/microsoft_onedrive/sources/new-folder-in-folder/new-folder-in-folder.js
```diff
@@ -0,0 +1,27 @@
...
+  dedupe: "unique",
+  props: {
+    ...baseNewFolder.props,
+    ...baseNewFileInFolder.props,
+  },
+  methods: {
+    ...baseNewFolder.methods,
+    ...baseNewFileInFolder.methods,
+    getDeltaLinkParams(...args) {
+      return baseNewFileInFolder.methods.getDeltaLinkParams.bind(this)(...args);
```
compwright, line 24: I find this to be slightly less awkward in this situation:

```javascript
return baseNewFileInFolder.methods.getDeltaLinkParams.call(this, ...args);
```

compwright (approved):
Thanks @jverce!

I have flagged several small things that could be improved.

Also, we need to switch to the official SDK, either here or in a new PR. [@microsoft/microsoft-graph-client](https://www.npmjs.com/package/@microsoft/microsoft-graph-client)

components/microsoft_onedrive/sources/new-folder/new-folder.js
```diff
@@ -0,0 +1,19 @@
+const base = require("../new-file/new-file");
+
+module.exports = {
+  ...base,
+  key: "microsoft_onedrive-new-folder-in-folder",
```
compwright, line 5: Name conflicts with previous action, fails to publish

compwright (changes requested):
One of the new sources fails to publish (`new-folder`), will need to be fixed.

compwright:
@jverce also got this error: https://www.dropbox.com/s/ugjgv2zwlhhi9z7/Screen%20Shot%202021-09-17%20at%209.47.34%20PM.png?dl=0

PR #1199, opened 2021-04-29
Provide root path info to the Dropbox client

### LINKED ISSUE #1179: [BUG] Dropbox event sources fail to list correct root folders for a Business account
I want to be able to watch a folder in the root folder, but none of the folders at the root level come up as a choice in the folder picker.

Reported [here](https://pipedream-users.slack.com/archives/CPTJYRY5A/p1618954351073500) in Slack.

jverce:
Fixes #1179

### DIFF
--- a/components/dropbox/dropbox.app.js
+++ b/components/dropbox/dropbox.app.js
@@ -24,8 +24,28 @@ module.exports = {
     },
   },
   methods: {
-    sdk() {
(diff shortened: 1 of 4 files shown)


dylburger (approved):
Tested with my Business account. Code looks fine. Thanks for the other improvements

PR #1203, opened 2021-04-30
Troubleshooting mailgun new log data source

sergioeliot2039:
There is an issue we are facing when implementing Mailgun's new-log-data source.
This is due to behavior when paginating results from Mailgun's Get Event endpoint that consist in that the last page of results always has items, and traversing to the next page won't bring empty results list. One consequence of this is that the source execution time is exceeded. Furthermore, after the execution time is exceeded, the next source of execution won't bring new results, as if the results were persisted by the API, and thus it seems a new request from the "first" page of results would be needed.
One possible fix could be that the current results page is compared with the previous one if they are the same we could halt the execution.
There's a discussion about this on the Slack channel:
https://pipedream-users.slack.com/archives/C01QQFC9TV0/p1619664687073000


umang-eng:
Ìmã ki ßd FF xêùözçvm ïwkï...

PR #1208, opened 2021-05-04
Eventbrite: New Event Sources

michelle0927:


components/eventbrite/eventbrite.app.js
```diff
@@ -0,0 +1,96 @@
...
+  propDefinitions: {
+    organization: {
+      type: "string",
+      label: "Organization",
+      async options({ prevContext }) {
+        const {
+          prevHasMore: hasMore,
+          prevContinuation: continuation,
+        } = prevContext;
+        const params = prevContext && hasMore ? { continuation } : null;
```
jverce, line 15: **-Minor-**
Simplifying it a bit...

```suggestion
        const {
          prevHasMore: hasMore = false,
          prevContinuation: continuation,
        } = prevContext;
        const params = hasMore ? { continuation } : null;
```

components/eventbrite/eventbrite.app.js
```diff
@@ -0,0 +1,96 @@
...
+    async _makeRequest(
+      method,
+      endpoint,
+      params = null,
+      data = null,
+      url = null
+    ) {
+      const config = {
+        method,
+        url: url || `${this._getBaseUrl()}${endpoint}`,
```
jverce, line 55: **-Minor-**
Let the language take care of it 🙂 

```suggestion
      url = `${this._getBaseUrl()}${endpoint}`,
    ) {
      const config = {
        method,
        url,
```

components/eventbrite/sources/common-webhook.js
```diff
@@ -0,0 +1,40 @@
+const common = require("./common.js");
+const get = require("lodash/get");
+
+module.exports = {
+  ...common,
+  props: {
+    ...common.props,
+    http: "$.interface.http",
+  },
+  hooks: {
```
jverce, line 10: Make sure you extend the `hooks` from the parent module to avoid future headaches:

```suggestion
  hooks: {
    ...common.hooks,
```

components/eventbrite/sources/common-webhook.js
```diff
@@ -0,0 +1,40 @@
...
+    async activate() {
+      const data = {
+        actions: this.getActions(),
+        endpoint_url: this.http.endpoint,
+      };
+      const { id } = await this.eventbrite.createHook(this.organization, data);
+      this.db.set("hookId", id);
```
jverce, line 17: Use setter/getter methods for these cases.

components/eventbrite/sources/common.js
```diff
@@ -0,0 +1,36 @@
...
+      let results = [];
+      let hasMoreItems;
+      do {
+        const result = await resourceFn(params);
+        results = results.concat(result[resource]);
+        hasMoreItems = result.pagination.has_more_items;
+        params.continuation = result.pagination.continuation;
+      } while (hasMoreItems);
+      return results;
+    },
```
jverce, line 34: This seems like a good case for a generator, which would:
- Avoid accumulating all the items in a collection before we start processing them (which could put pressure on the memory)
- If an issue happens midway then we'd lose all progress

```suggestion
    async *resourceStream(resourceFn, resource, params = null) {
      let hasMoreItems;
      do {
        const {
          [resource]: items,
          pagination = {},
        } = await resourceFn(params);        
        for (const item in items) {
          yield item;
        }
        
        hasMoreItems = !!pagination.has_more_items;
        params.continuation = pagination.continuation;
      } while (hasMoreItems);
    },
```

components/eventbrite/sources/new-event-ended/new-event-ended.js
```diff
@@ -0,0 +1,59 @@
...
+  props: {
+    ...common.props,
+    timer: {
+      type: "$.interface.timer",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+  },
+  hooks: {
```
jverce, line 19: Remember to include the parent's hooks:

```suggestion
  hooks: {
    ...common.hooks,
```

components/eventbrite/sources/common.js
jverce: If you in fact want to retrieve all the items (e.g. to emit an event containing the whole collection), you can accumulate those results where they are actually needed.

For example, you can have something like this in `new-attendee-registered`:
```js
const base = require("./common-webhook");
module.exports = {
  ...base,
  methods: {
    ...base.methods,
    async getData(order) {
      const { id: orderId } = order;
      const attendeesStream = this.resourceStream(
        this.eventbrite.getOrderAttendees.bind(this),
        "attendees",
        orderId,
      );
      const attendees = [];
      for await (const attendee of attendeesStream) {
        attendees.push(attendee);
      }
      // Do the rest...
    },
  },
};
```

For the rest of the cases:

```js
const base = require("../common");
module.exports = {
  ...base,
  hooks: {
    ...base.hooks,
    async deploy() {
      const params = {
        orgId: this.organization,
        params: {
          status: "ended,completed",
        },
      };
      const eventStream = await this.resourceStream(
        this.eventbrite.listEvents.bind(this),
        "events",
        params,
      );
      for await (const event of eventStream) {
        this.emitEvent(event);
      }
    },
  },
};
```

components/eventbrite/sources/new-event-ended/new-event-ended.js
```diff
@@ -0,0 +1,59 @@
...
+      );
+      for (const event of events) {
+        this.emitEvent(event);
+      }
+    },
+  },
+  methods: {
+    ...common.methods,
+    generateMeta(event) {
+      return this.generateEventMeta(event);
```
jverce, line 40: We could have another parent method specific for "events" event sources, with a hierarchy like this:

`common/base <- common/event <- new-event-ended`

What do you think?

jverce (changes requested):

components/eventbrite/eventbrite.app.js
michelle0927: Updated

components/eventbrite/sources/common-webhook.js
michelle0927: Updated

components/eventbrite/sources/common.js
michelle0927: Updated to use generator

components/eventbrite/sources/new-event-ended/new-event-ended.js
michelle0927: Updated

components/eventbrite/sources/new-event-ended/new-event-ended.js
michelle0927: I added events.js, but I wasn't sure the best way to implement this with new-event-ended using polling and new-event using a webhook. The way I have it set up now works, but seems a bit awkward to me. Any suggestions?

jverce: Yeah, it seems a bit awkward to me too, especially since the timer-based event source creates an HTTP endpoint that it doesn't use.
The straightforward alternative here is to extend both base components (e.g. `common/event.js` and `common/webhook.js`) in the event source:

```js
const timerBase = require("../common/timer");
const eventBase = require("../common/event");

module.exports = {
  ...timerBase,
  ...eventBase,
  hooks: {
    ...eventBase.hooks,
    async deploy() {
      // Do something specific here...
    },
  },
  methods: {
    ...timerBase.methods,
    ...eventBase.methods,
  },
  async run(event) {
    // Do something specific here...
  },
};
```

Also, `common/event.js` should not extend either `common/webhook.js` or `common/timer.js` since event sources for these events can be either webhook-based or timer-based.

As you can see, it gets complicated really quickly, especially considering that this resembles a multiple inheritance architecture where you might have multiple definitions of the same method/prop/etc. So if your current approach works then let's keep it as is and we can discuss it further and revisit it as part of a future refactor.

jverce (approved):

PR #1212, opened 2021-05-07
Shopify: Update to remove unresponsive event types

michelle0927:


jverce:
@michelle0927 what's the motivation behind the removal of these event types?

michelle0927:
> @michelle0927 what's the motivation behind the removal of these event types?

Simply put, they don't work. Apparently, this has been an existing Shopify issue for at least a couple of years. Here's one thread about it:
https://community.shopify.com/c/Shopify-APIs-SDKs/Event-API-not-firing-cancelled-orders-events/td-p/470264

jverce:
> > @michelle0927 what's the motivation behind the removal of these event types?
> 
> Simply put, they don't work. Apparently, this has been an existing Shopify issue for at least a couple of years. Here's one thread about it:
> https://community.shopify.com/c/Shopify-APIs-SDKs/Event-API-not-firing-cancelled-orders-events/td-p/470264

That's a shame 😞 
Yeah, Shopify's API has those WTF's every now and then 😂

jverce (approved):

jverce (changes requested):
Sorry, just forgot you'll need to bump the sources versions.

jverce (approved):

PR #1238, opened 2021-05-18
Typeform new-submission source typo

NReilingh:
Corrected "timezome" to "timezone"


dylburger:
Thanks for this fix, as well, @NReilingh

Issue #1026, opened 2021-03-01
[SOURCE] Mailchimp: New Event Sources


michelle0927:
**Describe the event source. What app, what event?**
Mailchimp

**New Audience** | Triggers when a new audience is added to your Mailchimp accounts.
**New Campaign** | Triggers when a new campaign is created or sent.
**New Subscriber** | Triggers when a new subscriber is added to an audience.
**New Subscriber in Segment or Tag** | Triggers when a subscriber is added to a segment or tag within an audience.
**New Unsubscriber** | Triggers when any current subscriber unsubscribes from an audience.
**New or Updated Subscriber** | Triggers when a subscriber is added or updated in an audience.
**New Customer** | Triggers when a new customer is added to a selected store.
**Email Opened** | Triggers when a recipient opens an email in a specific campaign.
**Link Clicked** | Triggers when a recipient clicks a pre-specified link in your campaign.
**New Order** | Triggers when a new order is created.
**New List Event** | Triggers when an event occurs on a list.
**New List** | Triggers when a new list is created.
**New or Updated List Segment** | Triggers when a segment is created or updated.
**New File** | Triggers when a new file is added.

**Please provide a link to the relevant API docs for the specific service / operation this trigger is tied to.**
https://mailchimp.com/developer/marketing/api/

dannyroosevelt:
@sergioeliot2039 is there a PR for this? I see it's "in review" but I think maybe that's a mistake?

js07:
PR for this issue: https://github.com/PipedreamHQ/pipedream/pull/1289

dannyroosevelt:
@michelle0927 this is ready for release! 

>Added via [Pipedream](https://pipedream.com/@/p_8rCkMkO?o=a&a=github)

PR #1239, opened 2021-05-18
[SOURCE] Mailchimp: New Event Sources #1026

sergioeliot2039:
Hi, this is a new PR for Mailchimp sources.

Here are some notes:

* New Order, and New Ecommerce have not been tested, they depend on Ecommerce Stores module which Mailchimp has in Beta, can't get to create a store and reached out to Mailchimp support.
* New Segment Subscribers, don't have a `since` and `before` parameters to narrow down pagination, so currently they paginate across all items.
* New List from the related Pipedream Issue for Mailchimp sources is the same as New Audience (there was a duplicated source in the issue)
* New or Updated Subscriber was implemented as New Subscriber, and Email Address change or a Profile change (3 different events to Mailchimp)
* New List Events is a superset to New or Updated Subscriber because it exposes all available events related to Audience Lists. However, I thought having New or Updated Subscriber as a different source could be relevant to Mailchimp users.


sergioeliot2039:
I received feedback from Mailchimp, they suggested connecting an Ecommerce Store to Mailchimp.
So I spun a Big Cartel instance and connected to Mailchimp.
I was able to test New Customer, New order with it. Lastest commits have the relevant changes (only New Order and the Mailchimp app file needed changes).

PR #1241, opened 2021-05-18
attempt at fixing a 404 on this link

NReilingh:
Going to see what happens in the vercel preview


NReilingh:
That worked! Fix appears to be valid

PR #1261, opened 2021-05-21
[ACTIONS] Mailgun

compwright:


components/mailgun/actions/send-email/send-email.js
```diff
@@ -0,0 +1,104 @@
...
+      description: "Enables or disables DKIM signatures",
+      optional: true,
+    },
+    tracking: {
+      type: "boolean",
+      default: true,
+      description: "Enables or disables tracking",
+      optional: true,
+    },
+    haltOnError: {
```
dylburger, line 70: As an aside, this led to a good discussion internally. I think we're likely to make this a step-level primitive (such that you can control the error-handling logic for any step native to the platform). For now this pattern is fine.

components/mailgun/actions/send-email/send-email.js
```diff
@@ -0,0 +1,104 @@
...
+      type: "string",
+      label: "HTML",
+      description: "Message body (HTML)",
+    },
+    testMode: {
+      type: "boolean",
+      default: true,
+      description: "Enables sending in test mode",
+    },
+    dkim: {
```
dylburger, line 57: For more complex props like this, does it make sense to link to the Mailgun docs in the description for an explanation of how the functionality works? The `description` should support Markdown links.

components/mailgun/actions/send-email/send-email.js
```diff
@@ -0,0 +1,104 @@
...
+      type: "string",
+      label: "From Email",
+      description: "Sender email address",
+    },
+    replyTo: {
+      type: "string",
+      label: "Reply Email",
+      description: "Sender Reply-To email address",
+    },
+    to: {
```
dylburger, line 32: Should `cc` and `bcc` be supported as optional string arrays, as well?

components/mailgun/actions/suppress-email/suppress-email.js
```diff
@@ -0,0 +1,84 @@
...
+      ],
+    },
+  },
+  async run () {
+    try {
+      const suppression = {
+        address: this.email,
+      };
+
+      switch (this.category) {
```
dylburger, line 62: Is a `default` case necessary to capture any type of code / error / tag? Or does the `suppression` object above capture the correct default case?

components/mailgun/mailgun.app.js
```diff
@@ -0,0 +1,58 @@
...
+    api (api) {
+      const mailgun = new Mailgun(formData);
+      const mg = mailgun.client({
+        username: "api",
+        key: this.$auth.api_key,
+        public_key: this.$auth.api_key,
+      });
+      return mg[api];
+    },
+    async suppress (domain, type, suppression) {
```
dylburger, line 18: Is `suppress` not available via the JS client lib?

components/mailgun/actions/send-email/send-email.js
compwright: Good catch!

components/mailgun/mailgun.app.js
compwright: @dylburger heh, it is but no matter what I tried I couldn't get it to work. I think it must have a bug, or Mailgun changed their API behavior and their official client library hasn't caught up. Whatever it is, I noticed that it is particularly touchy about the Content-Type header and the exact format of the request body.

components/mailgun/actions/suppress-email/suppress-email.js
compwright: @dylburger no default case needed, bounces and unsubscribes are the only cases that accept additional information, and all cases require address.

compwright:
@dylburger thanks for the review! Should be good to go.

PR #1263, opened 2021-05-21
[ACTION] Add missing Airtable typecast setting

compwright:
* Adds typecast setting to components/airtable/actions/create-single-record
* Adds typecast setting to components/airtable/actions/create-multiple-records
* Unfolds the code logic in all actions for readability
* Conforms all files to established code style

### DIFF
--- a/components/airtable/actions/create-multiple-records/create-multiple-records.js
+++ b/components/airtable/actions/create-multiple-records/create-multiple-records.js
@@ -44,9 +44,9 @@ module.exports = {
       inputRecords = JSON.parse(inputRecords);
     }
 
-    inputRecords.forEach(record => { records.push({
-      fields: record,
-    }); });
+    const records = inputRecords.map((fields) => ({
+      fields,
+    }));
 
     const recordsSets = chunk(records, BATCH_SIZE);
     for (const recordsSet of recordsSets) {
--- a/components/airtable/actions/create-multiple-records/create-multiple-records.js
+++ b/components/airtable/actions/create-multiple-records/create-multiple-records.js
@@ -7,7 +7,7 @@ module.exports = {
   key: "airtable-create-multiple-records",
   name: "Create Multiple Records",
   description: "Create one or more records in a table by passing an array of objects containing field names and values as key/value pairs.",
-  version: "0.0.29",
+  version: "0.0.30",
   type: "action",
   props: {
     airtable,
--- a/components/airtable/actions/create-multiple-records/create-multiple-records.js
+++ b/components/airtable/actions/create-multiple-records/create-multiple-records.js
@@ -1,6 +1,7 @@
 const airtable = require("../../airtable.app.js");
 const chunk = require("lodash.chunk");
-const Airtable = require("airtable");
+
+const BATCH_SIZE = 10; // Airtable API allows to update up to 10 rows per request.
 
 module.exports = {
   key: "airtable-create-multiple-records",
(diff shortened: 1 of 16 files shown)


components/airtable/actions/create-multiple-records/create-multiple-records.js
```diff
@@ -10,34 +10,49 @@ module.exports = {
...
     }
 
-    inputRecords.forEach(record => { records.push({ fields: record }) })
-    
-    const records_sets = chunk(records, BATCH_SIZE)
-    for (const records_set of records_sets) {        
-      response_records = response_records.concat((await this.addRecords(records_set)))
+    inputRecords.forEach(record => { records.push({
+      fields: record,
+    }); });
```
jverce, line 49: I know it's out of scope but I'm going to leave a quick suggestion here to piggyback on this PR:

```suggestion
    const records = inputRecords.map((fields) => ({
      fields,
    }));
```

components/airtable/actions/create-single-record/create-single-record.js
```diff
@@ -8,15 +8,41 @@ module.exports = {
...
+      [
+        {
+          fields: this.record,
+        },
+      ],
+      {
+        typecast: this.typecast,
+      },
+    ];
+    return (await table.create(...args))[0];
```
jverce, line 46: Since `table.create` accepts only 2 arguments, I suggest removing the varargs here:

```suggestion
    const recordsData = [
      {
        fields: this.record,
      },
    ];
    const params = {
      typecast: this.typecast,
    };
    const [
      createdRecord,
    ] = await table.create(recordsData, params);
    return createdRecord;
```

components/airtable/actions/list-records-in-view/list-records-in-view.js
```diff
@@ -1,48 +1,85 @@
...
-         records.forEach(function(record) {
-             data.push(record._rawJson)
-         });
+      ...config,
+    })
+      .eachPage(function page(records, fetchNextPage) {
+      // This function (`page`) will get called for each page of records.
+        records.forEach(function(record) {
+          data.push(record._rawJson);
+        });
```
jverce, line 75:
```suggestion
        data.push(...records.map((record) => record._rawJson));
```

components/airtable/actions/list-records/list-records.js
```diff
@@ -1,46 +1,80 @@
...
-         records.forEach(function(record) {
-             data.push(record._rawJson)
-         });
+      ...config,
+    })
+      .eachPage(function page(records, fetchNextPage) {
+      // This function (`page`) will get called for each page of records.
+        records.forEach(function(record) {
+          data.push(record._rawJson);
+        });
```
jverce, line 70:
```suggestion
        data.push(...records.map((record) => record._rawJson));
```

components/airtable/actions/create-single-record/create-single-record.js
```diff
@@ -8,15 +8,41 @@ module.exports = {
   type: "action",
```
jverce, line 8: Make sure you bump the version of this component.

```suggestion
  version: "0.0.12",
  type: "action",
```

jverce: (Github does not allow me to comment on a line that's far from a changed section, which is the case of line no. 7 here)

jverce (changes requested):
A few minor details (except for the version bump), other than that LGTM!
One thing to keep in mind is that the CI/CD pipeline will attempt to deploy all the affected components in the PR, and if they don't have their versions bumped then the pipeline will fail.

Thanks for reformatting the code BTW!

jverce:
Thanks for the changes @compwright !
Looks like there was a recent refactoring of the Airtable components and a bunch of conflicts arose.
Sorry for the back and forth 🙁

Issue #1248, opened 2021-05-19
[ACTION] Support SendGrid Actions


sergioeliot2039:
**Is there a specific app this action is for?**
Sendgrid

Add Emails to Global Suppression | Adds recipient email address to the global suppression group.
Add or Update a Contact |	Creates new contact or updates existing based on the email address.
Add or Update a Recipient (Legacy) | Adds a new recipient or updates an existing one based on the email address. 
Add Recipients to a List (Legacy) |	Adds existing recipients to a list. 
Create a List | Creates a new list.
Delete a Global Suppression | Removes an email address from the global suppressions group.
Delete a List |	Deletes a list.
Delete all Bounces | Deletes all bounces.
Delete Blocks |	Deletes either all or specific emails on your block list.
Delete Bounces | Deletes specified bounces.
Delete Contacts | Deletes one or more contacts.
Delete Recipient from a List (Legacy)	| Deletes recipient from a specific list.
Delete Recipients (Legacy) | Deletes one or more recipients.
Get a Block	| Retrieves a specific email address from your block list.
Get a Global Suppression	| Retrieves a specific global suppression email.
Make an API Call | Performs an arbitrary authorized API call.
Remove Contacts from a List	| Removes contacts from the given list. The contacts will not be deleted. 
Send an Email	| Sends an email.
Validate an Email	| Validates an email address.
Get all Bounces	| Retrieves all of your bounces.
Get all Contacts | Retrieves all your contacts.
Get all Lists	| Retrieves all your contact lists.
Get all Recipients | Retrieves all of your recipients or recipients under a specified list.
List Blocks	| Lists all email addresses that are currently on your block list.
List Global Suppressions | Lists all email addresses that are globally suppressed.
Search Contacts | Searches for contacts. 

**Please provide a link to the relevant API docs for the specific service / operation.**
https://sendgrid.com/docs/api-reference/
For Legacy endpoints: https://www.twilio.com/docs/sendgrid/api/v2/using_the_web_api

dylburger:
Relevant: https://github.com/PipedreamHQ/pipedream/issues/1129

dannyroosevelt:
Moving this back to "To do" -- to whoever picks this up next, please refer to the most recent comments on [the PR](https://github.com/PipedreamHQ/pipedream/pull/1288)

dannyroosevelt:
This is ready for release!

PR #1264, opened 2021-05-22
[ACTIONS] Support SendGrid Actions #1248

sergioeliot2039:
Hi, 

Here is a new PR for SendGrid actions.

Currently, it only has SendMail implemented.

I think I found a couple of issues when working with actions and the component model.
* Props can't be made optional, regardless of the `optional: true,` directive. So all props will appear as parameters when the action is used on a workflow.
* The actions keep running and then time-outs. I tried with and without a `return` instruction inside the `run` method. Attached is a screenshot of how it looks. The component is not using `timer` nor `db` interface props, so can't think of I reason for this behavior.

I prepared a Workflow using a modified version of the action where nonrequired
parameters were out commented, to show the behavior of the action being executed repeatedly to then time out.
[https://pipedream.com/@sergio/sendgrid-send-mail-timing-out-p_8rCMlBd](https://pipedream.com/@sergio/sendgrid-send-mail-timing-out-p_8rCMlBd)

In the attachment, there's the workflow timing out where console.logs added for debugging appear repeatedly. Had the action execute once there would have been only one console.log 
TinyTake21-05-2021-06-40-06

I will be adding the remaining SendGrid actions as I finish them.


sergioeliot2039:
I will create a new PR, doing a new rebase, and hoping it won't pick files from other non related features.

PR #1268, opened 2021-05-24
Fix: Google Calendar to honor 'New Events Only' flag

### LINKED ISSUE #1267: [BUG] Google Calendar - New and Updated Event (Instant) fails to honor New Events only? flag
Setting **New Events Only?** to `true` still yields update events (e.g. invitee accepts invite, title changes, etc). It appears that the check we run within the component's `run` method depends on the event's sequence number, which doesn't appear to increment on changes.

We should fix the component to honor this flag, so that I can filter on only new events.

jverce:
* Fix the logic around the detection of new vs updated calendar events
* Apply some reformatting to make the linter happy
* Enforce parenthesis around arrow function args
* Enforce spaces around arrow function's arrows

Fixes #1267

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -36,6 +36,8 @@
   ],
   "root": true,
   "rules": {
+    "arrow-parens": "error",
+    "arrow-spacing": "error",
     "array-bracket-newline": [
       "error",
       {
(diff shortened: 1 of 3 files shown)


dylburger (approved):
Tested and approved

PR #1274, opened 2021-05-27
AWS Event Sources Refactoring

jverce:
* Refactor existing AWS event sources to leverage common logic around
  AWS SNS

* Update documentation to include architectural approach and internal
  logic around SNS notifications

* Improve documentation around required IAM permissions for every
  SNS-based event source

* Add utility functions in `aws.app.js` and extract service-specific
  logic into common, scoped modules

* Bump minor versions


components/aws/aws.app.js
```diff
@@ -1,27 +1,202 @@
+const awsRegions = require("./regions");
```
dylburger, line 1: I believe [this API endpoint](https://docs.aws.amazon.com/AWSJavaScriptSDK/latest/AWS/EC2.html#describeRegions-property) lists the regions enabled for your account (it's under the EC2 class, but it appears like it's not unique to EC2). Can we reference that so we don't have to keep the list static as new regions are added?

components/aws/aws.app.js
```diff
@@ -1,27 +1,202 @@
...
+     */
+    async createServiceRole(region, service, roleDescription) {
+      const assumeRolePolicyDocument = JSON.stringify(
+        this._getAssumeRolePolicyForService(service),
+      );
+      const params = {
+        AssumeRolePolicyDocument: assumeRolePolicyDocument,
+        RoleName: generateRandomUniqueName(),
+        Description: roleDescription,
+        Path: "/pipedream/",
```
dylburger, line 88: Does the `path` appear with the name of the role in the IAM console? Should we prefix the `name` with "pipedream-" or some string to clarify that the role was created by PD?

components/aws/sources/README.md
```diff
@@ -0,0 +1,200 @@
+# AWS Event Sources <!-- omit in toc -->
+
+This folder contains several event sources either directly related to or
+supported by AWS services:
+
+- **Directly related**: event sources that emit events whenever an AWS service
+  experiences changes that the user is interested in watching. For example, an
+  e-mail received by AWS SES (Simple Email Service).
+- **Supported**: event sources that are not dependant on watching events that
```
dylburger, line 9: "dependant" -> "dependent"

components/aws/sources/README.md
```diff
@@ -0,0 +1,200 @@
+# AWS Event Sources <!-- omit in toc -->
```
dylburger, line 1: Great job on this `README`

components/aws/aws.app.js
jverce: Done. I'm keeping both options since calling this API requires additional permissions that the account might not have.
What would be the best place to document these permissions? (we don't have a `components/aws/README.md` file yet).

components/aws/aws.app.js
jverce: It doesn't appear in the console unless you go into the role itself.
The output of the `generateRandomUniqueName` function is already prefixed with `pd-`. Should I change that to `pipedream-` instead?

components/aws/sources/README.md
jverce: Done

components/aws/aws.app.js
dylburger: I see, nah `pd-` is totally fine, thanks

components/aws/aws.app.js
dylburger: You make a good point re: the IAM permissions necessary to make this call. I agree that's hard to communicate given the current state. Sorry to go back and forth on this, but I think we should keep the static list you had until we address that.

dylburger: Ignore the last comment. I looked at the code and I see what you did re: the fallback. Works great.

dylburger (approved):

PR #1281, opened 2021-05-29
Test Pull Request

sergioeliot2039:
TEST. You can Ignore. After creating a test Pull REquest (after rebasing, resolved conflicts, cherry picking commits, etc. ) I modified files that still showed differences in with changes already in upstream/master on Reddit and Mailgun features. They should be the same now. Test PR before squashing the several commits.


sergioeliot2039:
Looks like there are differences in new line character at EOF in reddit related files.

PR #1291, opened 2021-06-01
Mailgun event sources and actions

### LINKED ISSUE #1024: [SOURCE] Mailgun: New Event Sources

sergioeliot2039:
Hi,

This is an initial merge of the Mailgun actions that were marge to PipedreamHQ/master while we had Mailgun sources on review.

I believe there are chances of further merging, for example. the method that creates a Mailgun SDK object could be shared both between actions and sources, however, doing so would require further modifications on either action and/or sources. So looking forward to discussing on this as needed.

The previous PR for Mailgun sources is [here](https://github.com/PipedreamHQ/pipedream/pull/1172). Furthermore, this PR doesn't touch files from reddit, as the previously unintendedly had.

Resolves #1024 and #1471


compwright (changes requested):
@sergioeliot2039 thank you so much for untangling this, and for all the new sources! 🎉

There are a few things which need to be done, which I'm actually going to work on, and then have you review what I've done, if you'd be so kind:

- [x] Rebase against master to get up-to-date
- [x] Fix all outstanding eslint issues
- [x] Replace the abandoned `mailgun-js` library with the currently maintained official `mailgun.js` library
- [x] Move the base region option into the Mailgun app authentication settings
- [x] Reconcile as much as possible any code duplication between the sources and the current actions

If there's anything I've missed, let me know.

sergioeliot2039:
Hi @compwright , looks good to me.
When it is ready for me to review please just ping me. Thanks!

compwright:
@sergioeliot2039 haven't tested yet, but you can give this a review now.

compwright (approved):

dannyroosevelt:
Not sure if I'm missing something but I can't get any of the actions I tried testing working:

-Add a Mailing List Member
Screenshot 9-20-21 at 4 13 02 PM

-List Domains 
Screenshot 9-20-21 at 4 13 28 PM

-List a Mailing List Member
Screenshot 9-20-21 at 4 12 21 PM

sergioeliot2039:
Most actions were failing because they leverage from a method `withErrorHandler` in the common file to avoid duplicating boiler plate code which had this issues: the app object was undefined, props values weren't being passed properly to the code segment actually hitting mailgun api. once those fixed there was an underlying issue that even if the code didn't break nor returned an error, it would return null and shown a workflow output message saying that return value is not serializable.

the fixups for the issues described above included: sending the mailgun app object from `withErrorHandler`  in the common file to the delegate function hitting the api, the props were passed as additinal parameters to `withErrorHandler` which are then relayed to the delegate function hitting the api. the output message about return value not being serializable was fixed by importing `withErrorHandler` to the action file via the methods component section instead of doing a direct import at the top of the file as a declaration.

while some extra boilerplate code was added for passing the variables between the coupled `withErrorHandler`  in the common file with the deleage function hitting the api in the related component file, or doing imports via the methods component section, i think the ethos of using  `withErrorHandler` for handling errors instead of using repetitives try/catch in the component files is preserved.

finally, each some actions had other logic errors on its own, i tried to fix all but here are the outstanding:
`verify-email` -- Validations are not enabled.
`create-route`-- Routes quota (0) is exceeded for these two i assumed it's due to my using a mailgun free tier account or just missing configuration. Since those didn't look like a problem with the request config itself i stopped there but i can look further if required.
`send-email` -- is not including the `cc` nor `bcc` parameters. this issue is extremely interesting because with the mailgun sdk it won't accept parameters from the UI for `cc` nor `bcc` because of the in memory handling of the arrays. if i hardcoded an array and did a direct assignment to these parameters with double quote they would work (i.e. `data.cc = ["test@mail.com]`), i consoled out the parameters received from the UI and the console out as  `['test@mail.com']`, i tried to JSON.stringify the params, do a replace from ' to " and JSON.parse back but this didn't work. I had to remove  `cc` nor `bcc`. Odd because assigning an array from the UI on the `to` prop works.

yes -- there are ESLint issues in this PR but those are in the sources. I didn't touch sources files just this time, but let me know if should tackle them.

sergioeliot2039:
forgot to mention, i stripped off pagination on action `list-mailinglist-members` because the related sdk method is buggy -- it always get the first page and doesn't honor the input page/skip parameters, so it goes on a infinite loop until the workflows times out. i set the code to return the first page only.

dannyroosevelt:
I was able to successfully QA everything except Create a Route (not because that action is broken, just because I didn't go deep in setting up the account / use case properly).

sergioeliot2039:
@js07 this is ready for another review.
I made adjustments to the sources, so that they work properly with the sdk.
I also refactored the common-webhooks file into a base, http-based, and timer-based files. this is to take off the webhook signature off from timer based sources, which don't use this prop.
i followed your suggestions of doing a check on event type/subtype in the common file (now http-based), as of the suggestion for checking "opened" event types, i left both spellings "OPENED" and "opened" because originally malign was sending it on uppercase, so just in case.
looking forward to your reviews.

js07 (commented):
@sergioeliot2039 thanks for making those changes!
I made some additional comments.

sergioeliot2039:
> Is there a reason to stringify `vars` and add it to `data` separately from the other `opts`?

sergioeliot2039:
> Do you know if you could instead use [this.mailgun.api("webhooks").destroy(domain, webhook);](https://github.com/mailgun/mailgun.js#destroy-1), so the Mailgun client handles building the endpoint URL?
this worked. i applied this change.

sergioeliot2039:
> It looks like the events emitted in deploy() are the earliest 5 events after 1 day ago. Instead, I think we want to emit the latest (most recent) 5 events on deploy.
> 
> To fetch the 5 most recent events, you could use these [query options](https://documentation.mailgun.com/en/latest/api-events.html?highlight=begin#query-options):
> 
> this.mailgun.api("events").get(this.domain,{
>   begin: Math.floor(Date.now() / 1000),
>   ascending: "no",
>   limit: 5,
> });

Hey, I had to use this logic to make ascending: "no" work. Essentially, the descending records had to be emitted from the current date and earlier.
On the other hand, the ascending: "yes" records, are emitted using the current day minus one date.
If I didn't use the current date for ascending: "no", then no record are returned by Mailgun. THis is the relevant code:
```
      const date = new Date();
      if ([
        "yes",
      ].includes(ascending)) {
        date.setDate(date.getDate() - 1);
      } else {
        date.setDate(date.getDate());
      }
      const config = {
        begin: Math.floor(date.valueOf() / 1000),
        ascending,
        limit,
      };
```

sergioeliot2039:
> Since this.mailgun.api("lists").list() returns mailing lists for all domains, I think events for incorrect/unselected domains (domain !== this.domain) could also be emitted?
> 
> To fix this issue, I'd suggest filtering sortedList the way mailing lists are sorted in the run method.

I confirmed the issue. The source was emitting new mailing lists for all domains on deploy.
I committed your changes, tested and it works like a charm. Great catch!

sergioeliot2039:
> It looks like we're fetching all events created after 1 day ago each time this source runs. If there are more than 100 events in 1 day, this source could emit the same event multiple times (since it's using the unique [deduping strategy](https://pipedream.com/docs/components/api/#dedupe-strategies))?
> 
> To avoid refetching events (beginning 1 day ago) on each run, could we store the timestamp of the last seen event using [$.service.db](https://pipedream.com/docs/components/api/#db) and request only events after that timestamp? E.g., Google Drive [New or Modified Comments](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_drive/sources/new-or-modified-comments/new-or-modified-comments.mjs), Twilio [common-polling](https://github.com/PipedreamHQ/pipedream/blob/master/components/twilio/sources/common-polling.mjs)

I implemented your suggestion to avoid refetching events as described by checking for the timestamp of the last seen event and leveraging from `$.service.db` to do so.

sergioeliot2039:
@js07 @dannyroosevelt this is ready for another review!

components/mailgun/sources/new-mailing-list/new-mailing-list.js
```diff
@@ -0,0 +1,88 @@
...
+  ...common
+} = require("../common/timer-based");
+
+module.exports = {
+  ...common,
+  key: "mailgun-new-mailing-list",
+  name: "New Mailing List",
+  type: "source",
+  description: "Emit new event when a new mailing list is added to the associated Mailgun account.",
+  version: "0.0.2",
```
js07, line 13:
```suggestion
  version: "0.0.1",
```

components/mailgun/actions/send-email/send-email.js
```diff
@@ -15,108 +19,118 @@ module.exports = {
...
-      return err;
+  async run() {
+    const msg = {
+      "from": `${this.fromName} <${this.from}>`,
+      "to": this.to,
+      "subject": this.subject,
+      "text": this.text,
+      "html": this.html,
+      "o:testmode": this.testMode,
+    };
```
js07, line 116: When I run this action with the `testMode` prop set to `true`, I get the error:
```
ErrorTypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer or Uint8Array. Received type boolean (true)
```

According to the [mailgun API docs](https://documentation.mailgun.com/en/latest/api-sending.html?highlight=testmode#sending), the `o:testmode` prop should be set to `yes` to enable sending in test mode.
Conditionally setting `msg["o:testmode"] = "yes"` fixed the error for me:

```suggestion
    };
    if (this.testMode) {
      msg["o:testmode"] = "yes";
    }
```

components/mailgun/sources/new-log-data/new-log-data.js
```diff
@@ -0,0 +1,90 @@
...
+      pages: {
+        next: {
+          id: null,
+        },
+      },
+    };
+    while (true) {
+      result = await this.getLatestEvents(result.pages.next.number, 300);
+      if (result.items.length === 0) {
+        console.log("No data available, skipping iteration");
```
js07, line 77: Remove this `console.log("No data available, skipping iteration")`, since it's called on every run of this event source, even if events were emitted?

components/mailgun/sources/new-log-data/new-log-data.js
```diff
@@ -0,0 +1,90 @@
...
+  dedupe: "unique",
+  props: {
+    ...props,
+  },
+  methods: {
+    ...methods,
+    generateMeta(payload) {
+      return {
+        id: payload.id,
+        summary: `${payload.timestamp} [${payload["log-level"]}] ${payload.event}`,
```
js07, line 24: **Suggestion**
Remove `payload.timestamp` from this `summary` so more of the `event` is visible in the UI.
log-data
E.g.,
```js
        summary: `[${payload["log-level"]}] ${payload.event}`,
```

js07 (changes requested):
@sergioeliot2039 thanks for making those changes.

In addition to the comments in this review, there are some unresolved, hidden conversations in my previous review above: https://github.com/PipedreamHQ/pipedream/pull/1291#pullrequestreview-774425788

hidden-load-more

Note: When testing the sources and actions, the logs repeatedly contained the message `ky UMD !!!!`. This seems to be an issue in the [mailgun.js](https://github.com/mailgun/mailgun.js/issues/238) sdk.
ky-log

sergioeliot2039:
@js07 this is ready for a new review!
I addressed the older review comments that were on hidden conversations and the lastest from newest review round.
Thanks

jcortes (approved):
> @js07 this is ready for a new review! I addressed the older review comments that were on hidden conversations and the lastest from newest review round. Thanks

All the changes were addressed, it is good to go!

PR #1294, opened 2021-06-01
Docusign: New Action - create-signature-request

michelle0927:


components/docusign/docusign.app.js
```diff
@@ -0,0 +1,129 @@
+const axios = require("axios");
+
+module.exports = {
+  type: "app",
+  app: "docusign",
+  propDefinitions: {
+    account: {
+      type: "string",
+      label: "Account",
+      description: "",
```
dylburger, line 10: Does Zapier have any description of the account? If this prop is obvious to the user, feel free to also just remove the description altogether (vs. keeping the empty string)

components/docusign/docusign.app.js
```diff
@@ -0,0 +1,129 @@
...
+        headers: this._getHeaders(),
+        data,
+        params,
+      };
+      return (await axios(config)).data;
+    },
+    async getUserInfo() {
+      return await this._makeRequest(
+        "GET",
+        "https://account-d.docusign.com/oauth/userinfo",
```
dylburger, line 100: Is this URI distinct from the base URI retrieved for the user's account below?

components/docusign/docusign.app.js
```diff
@@ -0,0 +1,129 @@
...
+    async listTemplates(baseUri, accountId) {
+      return await this._makeRequest(
+        "GET",
+        `${baseUri}/restapi/v2.1/accounts/${accountId}/templates`,
+      );
+    },
+    async listTemplateRecipients(baseUri, accountId, templateId) {
+      return await this._makeRequest(
+        "GET",
+        `${baseUri}/restapi/v2.1/accounts/${accountId}/templates/${templateId}/recipients`,
```
dylburger, line 118: Will all REST API requests be sent to paths under

```javascript
/restapi/v2.1/accounts/${accountId}/
```

i.e. does it make sense to include that in the base URI?

components/docusign/docusign.app.js
michelle0927: Yes. For developer accounts, the Base URI is https://demo.docusign.net.

components/docusign/actions/create-signature-request/create-signature-request.js
```diff
@@ -0,0 +1,78 @@
+const docusign = require("../../docusign.app.js");
+
+module.exports = {
+  key: "docusign-create-signature-request",
+  name: "Create Signature Request",
+  description: "Creates a signature request from a template",
+  version: "0.0.56",
```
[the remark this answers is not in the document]
michelle0927: Updated

components/docusign/docusign.app.js
```diff
@@ -0,0 +1,129 @@
...
+        const { accounts } = await this.getUserInfo();
+        return accounts.map((account) => {
+          return {
+            label: account.account_name,
+            value: account.account_id,
+          };
+        });
+      },
+    },
+    template: {
```
dylburger, line 21: Based on Zapier's action, I believe only account and template should be required. Everything else can inherit from the template settings, so I think all other props should default to `optional: true`

components/docusign/actions/create-signature-request/create-signature-request.js
```diff
@@ -0,0 +1,78 @@
+const docusign = require("../../docusign.app.js");
+
+module.exports = {
+  key: "docusign-create-signature-request",
+  name: "Create Signature Request",
+  description: "Creates a signature request from a template",
+  version: "0.0.01",
```
dylburger, line 7: You have an extra "0" in front of the last "1" :)

components/docusign/docusign.app.js
```diff
@@ -0,0 +1,129 @@
...
+            label: template.name,
+            value: template.templateId,
+          };
+        });
+      },
+    },
+    role: {
+      type: "string",
+      label: "Recipient Role",
+      description: "",
```
dylburger, line 42: Can you add the same description Zapier has for their action?

dylburger: I'd also move this last prop to show up last in the list (also like Zapier)

components/docusign/docusign.app.js
michelle0927: Updated

components/docusign/docusign.app.js
michelle0927: Updated

components/docusign/actions/create-signature-request/create-signature-request.js
michelle0927: Updated

components/docusign/docusign.app.js
michelle0927: Updated

PR #1314, opened 2021-06-10
Fix spellcheck workflow to always check a file

jverce:
The spellcheck PR workflow currently scans the entire repository if the PR does not modify any markdown files, which at the moment makes them fail due to the fact that some markdown files in the repo have spelling mistakes.

Example of a failed check:
https://github.com/PipedreamHQ/pipedream/runs/2788759237

The cause seems to be that when the corresponding Github action receives an empty string as the "input source files" it doesn't skip the checks but actually assumes the default behaviour (which is to scan the entire repository).

This change makes sure that the action is fed with at least one file to scan (`README.md` by default if the PR does not modify any markdown files).

### DIFF
--- a/.github/workflows/pull-request-checks.yaml
+++ b/.github/workflows/pull-request-checks.yaml
@@ -11,7 +11,8 @@ on:
       - master
 
 jobs:
-  build:
+
+  spellcheck-markdown-files:
     name: Spellcheck
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


.github/workflows/pull-request-checks.yaml
```diff
@@ -19,14 +19,21 @@ jobs:
...
             files="${f} ${files}"
-          fi;
+          fi
         done
-        echo "files=${files}" >> $GITHUB_ENV
+
+        # Setting the list of files to include at least one entry, since
+        # otherwise the spellcheck action might not work properly if no markdown
+        # files are part of the changed files.
+        echo "files=${files:-README.md}" >> $GITHUB_ENV
```
jverce, line 36:
```suggestion
        # Exiting early if no markdown files are part of the PR
        [ -z "$files" ] && exit 0
        echo "files=${files}" >> $GITHUB_ENV
```

Issue #1229, opened 2021-05-13
[ACTION]


todsac:
**Is there a specific app this action is for?**
Notion

**Please provide a link to the relevant API docs for the specific service / operation.**
https://developers.notion.com/

Actions:
 - Create database item
 - Update database item
 - Find database item

vellames-turing:
@dylburger Looks like this is not ready for review, right? The last interaction here was some months ago, how to proceed? I think we should move this issue to another column

sergioeliot2039:
that's right @vellames-turing , I moved it back to "doing" for now.

dmathukorala:
Verification completed for sources and actions including regression testing for bug fixes. This is ready to merge.

PR #931, opened 2021-02-17
Reddit event sources

sergioeliot2039:

PR #1317, opened 2021-06-10
Notion components

### LINKED ISSUE #1228: [SOURCE] Notion
Notion just launched their API.

https://developers.notion.com/

Triggers:
- new database item (triggered when new item is created)
(optional) - new user (triggered when a new user is added to a workspace)

sergioeliot2039:
Actions:

- [ ] add content to to page or block
- [ ] add page
- [ ] get all users
- [ ] get block children
- [ ] get database
- [ ] get page
- [ ] get user
- [ ] query database pages
- [ ] search items
- [ ] update page properties

Sources:

- [ ] new checklist item
- [ ] new page
- [ ] new user
- [ ] updated database item

Notes: from the list for notion in the no code database, I didn't implement New Comment, for which I didn't find a matching endpoint in the API. didn't implement checked item list because thought it wasn't optimal given the current notion API and PD implementation. new user has to iterate over all users in order to pick up new ones, paginating to the last page on deploy doesn't help forgetting new users after the deploy. there were a couple of actions that were implemented because the related API calls were needed for source's props, so I created an action around it to, i.e. get block children

Resolves #1228, #1229


components/notion/actions/get-all-users/get-all-users.js
```diff
@@ -0,0 +1,31 @@
...
+      label: "Page Size",
+      description:
+        "The number of items from the full list desired in the response. Maximum: 100",
+      min: 1,
+      max: 100,
+      optional: true,
+    },
+  },
+  async run() {
+    return await this.notion.getAllUsers(this.startCursor, this.pageSize);
```
compwright, line 29: Let's have this auto paginate and return the entire result set instead of putting that on the workflow.

components/notion/actions/get-block-children/get-block-children.js
```diff
@@ -0,0 +1,46 @@
...
+module.exports = {
+  key: "notion-get-block-children",
+  name: "Get Block Children",
+  description:
+    "Gets a list of child block objects contained in a block given its identifier.",
+  version: "0.0.3",
+  type: "action",
+  props: {
+    notion,
+    blockId: {
```
compwright, line 12: Is an options method here to provide a list possible? https://pipedream.com/docs/components/api/#async-options-example

If not, how does one get a blockId?

components/notion/actions/get-block-children/get-block-children.js
```diff
@@ -0,0 +1,46 @@
...
+      min: 1,
+      max: 100,
+      optional: true,
+    },
+  },
+  async run() {
+    if (!this.blockId) {
+      throw new Error("Must provide blockId parameter.");
+    }
+    return await this.notion.getBlockChildren(
```
compwright, line 40: Let's have this auto paginate and return the entire result set instead of putting that on the workflow.

components/notion/actions/get-database/get-database.js
```diff
@@ -0,0 +1,24 @@
...
+
+module.exports = {
+  key: "notion-get-database",
+  name: "Get Database",
+  description: "Gets details of a database given its identifier.",
+  version: "0.0.2",
+  type: "action",
+  props: {
+    notion,
+    databaseId: {
```
compwright, line 11: Needs an options method here: https://pipedream.com/docs/components/api/#async-options-example

compwright (changes requested):
@sergioeliot2039 Thanks for doing this! A lot of people are looking forward to this integration.

Needs some changes, and then I will re-review:

- [x] Rebase off latest master branch
- [ ] Fix all eslint issues
- [ ] Refactor all props that are used more than once into notion.app.js
- [ ] Add missing options() methods wherever they are supported by the Notion API
- [ ] Make all pagination props optional with no defaults, and auto paginate and return the entire result set if the are not set

LMK when you're ready for me to take another look.

sergioeliot2039:
> @sergioeliot2039 Thanks for doing this! A lot of people are looking forward to this integration.
> 
> Needs some changes, and then I will re-review:
> 
> * [ ]  Rebase off latest master branch
> * [ ]  Fix all eslint issues
> * [ ]  Refactor all props that are used more than once into notion.app.js
> * [ ]  Add missing options() methods wherever they are supported by the Notion API
> * [ ]  Make all pagination props optional with no defaults, and auto paginate and return the entire result set if the are not set
> 
> LMK when you're ready for me to take another look.

@compwright I don't think returning the entire result set is ideal, because the component can incur in timeout on execution. Here's a[ previous discussion](https://github.com/PipedreamHQ/pipedream/pull/931#discussion_r578814323) with Dylan about it:

> Good job on this one - we do like to emit sample events on the first run. I'd recommend only emitting 10 events on the first run, instead of all historical links. There's a timeout of 30 seconds on the deploy hook that users are likely to hit if you pull the whole history, and users typically won't need to see that many events - just a sample is appropriate

While that comment refers to the deploy hook, there is still a timeout for both actions and sources when executing the `run` method. @dylburger not sure if you have further guidance on this?

compwright:
Fair point, but I expect in these cases the list will be short enough not to use up the invocation time limit, and if that assumption is reasonable then we shouldn't require the user of the action components in question to incorporate pagination logic into their workflow to get all the records.

dylburger:
@sergioeliot2039 it is a valid point and we definitely need to think more about how we approach potentially-large pages. For now, I would paginate within the method and return all results.

We had to implement that logic for sources because they're fixed at a timeout of 30 seconds. For workflows / actions, users can raise their execution timeout to 5 minutes, so it's less of an issue here.

I agree with @compwright that the default user experience should be as simple as possible. Most people probably want to fetch all records, and don't want to think about pagination / max records. I _do_ think some users will want to fix the number of records they pull, and we could consider optional props for actions in the future to provide this option.

I'll take a look at how Integromat / Zapier solve this, since I think we can learn a lot from them. If you see another issue in this specific case fetching all records, let me know.

sergioeliot2039:
> @sergioeliot2039 Thanks for doing this! A lot of people are looking forward to this integration.
> 
> Needs some changes, and then I will re-review:
> 
> * [x]  Rebase off latest master branch
> * [x]  Fix all eslint issues
> * [ ]  Refactor all props that are used more than once into notion.app.js
> * [ ]  Add missing options() methods wherever they are supported by the Notion API
> * [ ]  Make all pagination props optional with no defaults, and auto paginate and return the entire result set if the are not set
> 
> LMK when you're ready for me to take another look.

@compwright here with an interesting case about refactoring props that appear more than once.
there's a `databaseId` that appears in two actions, and in two sources, they refer to the same data but the one for actions is optional, and for sources is mandatory.
wondering what would be the best approach here?
1. create two diferrent props in the notion.app file, one for actions, one for sources, with different names say `databaseIdActions` and `databaseSources`.
2. create the props as `databaseId` but instead of using the notion.app file, create them in the common file. in this case, we would need to create common file for actions and have that extended in the actions needing the props defined there. sources already have a common file.
thoughts?

dylburger:
@sergioeliot2039 you can override the optionality (and anything else about a prop) when you include it in a component. For example, see this: https://github.com/PipedreamHQ/pipedream/blob/5814ac0529db5d7969fb50b5f5b5e507b21dcd25/components/slack/actions/add_star/add_star.js#L11

components/notion/actions/add-page/add-page.js
```diff
@@ -0,0 +1,43 @@
...
+module.exports = {
+  key: "notion-add-page",
+  name: "Add Page",
+  description:
+    "Adds a new page to the specified parent object, a database or an existing page.",
+  version: "0.0.18",
+  type: "action",
+  props: {
+    notion,
+    parent: {
```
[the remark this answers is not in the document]
sergioeliot2039: I checked about pages having a workspace parent, but they are not in included in the results sets available to pull pages. I checked the /Search endpoint, /pages breaks when you used a page id with a workspace parent.

components/notion/actions/add-content-to-page-or-block/add-content-to-page-or-block.js
```diff
@@ -0,0 +1,31 @@
...
+  key: "notion-add-content-to-page-or-block",
+  name: "Add Content To Page Or Block",
+  description:
+    "Adds content to a specified page or block. In Notion, content is added via [blocks](https://developers.notion.com/reference/block), which can be headings, paragraphs, lists to dos, etc.",
+  version: "0.0.4",
+  type: "action",
+  props: {
+    notion,
+    blockId: {
+      type: "string",
```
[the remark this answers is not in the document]
sergioeliot2039: Perhaps the user could get the blockId from a previous steps in the workflow.

There is no direct way to retrieve the full set of blocks for Notion. This is what can be done.

1. Retrieve all databases
2. For each database, retrieve all pages.
3. For each page, recursively retrieve all blocks. 
Note from the Notion API: _In order to receive a complete representation of a block, you may need to recursively retrieve the block children of child blocks._

I am not convinced on going this way. 
I would discard the option to have an additional prop for pageID that would be parent to blocks, because in PD atm we can't dynamically load options of a prop given the values selected in the previous one.
Thoughts?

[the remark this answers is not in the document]
compwright: @sergioeliot2039 nah, I think that's too intense and maybe not even useful? I guess lets not do an options() for blockId and see how (and if) it gets used in real life.

components/notion/actions/get-all-users/get-all-users.js
sergioeliot2039: OK, let this autopaginate. Returning entire result set now.

components/notion/actions/get-block-children/get-block-children.js
sergioeliot2039: Ok, added auto pagination and entire result set is being returned now.

components/notion/actions/get-database/get-database.js
sergioeliot2039: Done.

components/notion/actions/get-block-children/get-block-children.js
sergioeliot2039: See my comment about pulling all blocks ids.

sergioeliot2039:
@compwright completed changes as per your PR review.
Please take another look.

Here are some notes:
There was a prop named `parent` but they were different, one expected a parentId the other expected a parent object. I added the `options` method to `parentId`, however for  a `parent` object it would be complex because given the Id we would have to query the `getPage` and/or `getDatabase` to see if they are a page or a database (we would do this if the endpoint returned an 404), then we need to contract an object based on the structur expected by the API.

WHere validation was needed, I started using `validate.js` as it was recommneded in another PR.

Some components had to be modified as a result of returning the entire result set in action, specially sources.

Looking forward to your coments.

sergioeliot2039:
I received input from @todsac about Notion having sent a developer update with changes in the Notion API,

- You can now use the Notion API to create a database

- Select values can now be dynamically created via Create and Update Page endpoints

etc.

I evaluated the impact and we are not affectedon existng components.  Because of the new `create database` endpoint, I created an action "create database" and a new source "new database". I also took the change to refactored code to getting a database and page titles to methods in the app file.

components/notion/actions/search-items/search-items.js
```diff
@@ -0,0 +1,130 @@
...
+        const options = [];
+        const notionItems = [];
+        const notionDatabases = await this.notion.getAllItems("database");
+        const notionPages = await this.notion.getAllItems("page");
+        notionDatabases.forEach((notionDatabase) =>
+          notionItems.push(notionDatabase));
+        notionPages.forEach((notionPage) => notionItems.push(notionPage));
+        for (const notionItem of notionItems) {
+          //Populating options with Notion databases
+          if ([
```
compwright, line 29: Please simplify, here and elsewhere:

```javascript
if (notionItem.object === "database") {
```

components/notion/actions/search-items/search-items.js
```diff
@@ -0,0 +1,130 @@
...
+      optional: true,
+      async options() {
+        const options = [];
+        const notionItems = [];
+        const notionDatabases = await this.notion.getAllItems("database");
+        const notionPages = await this.notion.getAllItems("page");
+        notionDatabases.forEach((notionDatabase) =>
+          notionItems.push(notionDatabase));
+        notionPages.forEach((notionPage) => notionItems.push(notionPage));
+        for (const notionItem of notionItems) {
```
compwright, line 27: Use [Array.prototype.map()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/map)

components/notion/notion.app.js
```diff
@@ -0,0 +1,586 @@
...
+    },
+  },
+  methods: {
+    _authToken() {
+      return this.$auth.oauth_access_token;
+    },
+    _apiUrl() {
+      return "https://api.notion.com";
+    },
+    async _makeRequest(opts) {
```
compwright, line 86: We need to use the [Notion SDK](https://github.com/makenotion/notion-sdk-js) instead.

components/notion/sources/new-checklist-item/new-checklist-item.js
```diff
@@ -0,0 +1,139 @@
...
+const get = require("lodash/get");
+
+module.exports = {
+  key: "notion-new-checklist-item",
+  name: "New Checklist Item",
+  description:
+    "Emits an event when a new checklist item is added to a database or page.",
+  version: "0.0.1",
+  dedupe: "unique",
+  type: "action",
```
compwright, line 12: Change to `source` here and in the other source components

compwright (changes requested):
Thanks @sergioeliot2039 for your work on this. There are a number of changes needed to address usability issues and to comply with the Pipedream [contribution guidelines](https://pipedream.com/docs/components/guidelines/#guidelines-patterns), in particular:

* Required Metadata
* Using APIs vs Client Libraries
* Reuse existing Prop Definitions
* Reuse methods defined in the app file

Happy to discuss any thoughts.

andrewjschuang:
@dannyroosevelt I think we can close this in favor of #2256?

dannyroosevelt:
Closing in favor of https://github.com/PipedreamHQ/pipedream/pull/2256

Issue #1349, opened 2021-06-23
[BUG] "Google Sheets - New Row Added" stops processing events after a while


jverce:
**Describe the bug**

Quoting the [Slack thread](https://pipedream-users.slack.com/archives/CPTJYRY5A/p1623080544183200) where the bug was reported:

**Trever F** -
> If I run for the next pageToken received
> it shows up

@dylburger - 
> Trever and I were chatting over DM and he asked about pagination for this API call. I looked into the docs and it looks like the [REST API](https://developers.google.com/drive/api/v3/reference/changes/list) for `changes.list` defaults to returning the last 100 changes. It's not clear whether the Node SDK will handle pagination on our behalf from the docs, but they do also take a `maxResults` param (and return a `nextLink`), so I wonder if that could be it. If we accrue more than 100 changes for the user's drive since the last time the spreadsheet was edited, I can see how we could miss the change and get into this state
>
> [drive.changes.list docs here](https://googleapis.dev/nodejs/googleapis/latest/drive/classes/Resource$Changes.html#list)

**Additional context**
- Slack thread: https://pipedream-users.slack.com/archives/CPTJYRY5A/p1623080544183200

PR #1350, opened 2021-06-24
Fix GSheets event sources during high GDrive traffic

jverce:
# Changelog

* Fix typo when destructuring result from the
  `getModifiedSheet` method in `google_sheets/common`
* Fix the logic that detects whether the incoming event is driven by
  the timer or by the Google Drive subscription

# Additional Context

This fix takes care of a [reported issue](https://github.com/PipedreamHQ/pipedream/issues/1349)
with some Google Sheets event sources that after some time they stop emitting
events event though they should.

The root cause was the fact the the Google Drive API reports the
changed files via the `changes.list` endpoint, which is paginated. In
order to fully retrieve the changes from Google Drive, the result of
such API call must properly take into account the pagination token
that the endpoint returns. The source code **contained a typo** that
prevented the event source from properly storing such token to use it
in subsequent API calls.


dylburger (commented):
This change looks good re: setting the next page token correctly, but don't we also need to make sure we paginate through all changes? Can't we still encounter a case where we have 200 changes, and the change to our spreadsheet is on page 2 on the `drive.changes.list()` response, and thus won't be retrieved?

dylburger:
Or does the Node SDK paginate for us?

jverce:
> This change looks good re: setting the next page token correctly, but don't we also need to make sure we paginate through all changes? Can't we still encounter a case where we have 200 changes, and the change to our spreadsheet is on page 2 on the `drive.changes.list()` response, and thus won't be retrieved?

Ahh yes, I forgot about that. This would still work because every execution will use a newer page token (which was not the case before) but it will at most process 100 file events, which could make the event source lag behind if the amount of traffic is high. I'll add the pagination logic to prevent this.

components/google_drive/sources/changes-to-specific-files/changes-to-specific-files.js
```diff
@@ -29,31 +31,35 @@ module.exports = {
...
-
-        return await this.googleDrive.listFiles({
-          pageToken: nextPageToken,
-          corpora: "drive",
-          driveId: this.drive,
-          includeItemsFromAllDrives: true,
-          supportsAllDrives: true,
-        });
+        const baseOpts = {};
+        const opts = this.drive === "myDrive"
```
dylburger, line 38: Does it make sense to have a method like `isMyDrive` to capture this logic? Or perhaps a `MY_DRIVE_LABEL` constant to capture the "myDrive" string?

sirloinofbeef:
Thank you for your amazing work on this guys!

dylburger:
@jverce did do an amazing job on this. Lots of improvements. Thanks for the note @sirloinofbeef .

PR #1373, opened 2021-06-29
Google Drive: New Action - create-file

michelle0927:


components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+      description: `The type of upload request to the /upload URI. If you are uploading data (using an /upload URI), this field is required. If you are creating a metadata-only file, this field is not required. 
+        media - Simple upload. Upload the media only, without any metadata.
+        multipart - Multipart upload. Upload both the media and its metadata, in a single request.
+        resumable - Resumable upload. Upload the file in a resumable fashion, using a series of at least two requests where the first request includes the metadata.
+      `,
+      options: [
+        "media",
+        "multipart",
+        "resumable",
+      ],
```
jverce, line 31: **-Minor-**
We can probably make this description shorter so that the UI has no trouble rendering it.
Maybe something like this?
```suggestion
      `,
      options: [
        {
          label: "Media (upload the media only, without any metadata)
          value: "media"
        },
        {
          label: "Multipart (upload both the media and its metadata, in a single request)
          value: "multipart"
        },
        {
          label: "Resumable (Upload the file in a resumable fashion, using a series of requests)
          value: "resumable"
        },
      ],
```

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+        "The path to the file, e.g. /tmp/myFile.csv . Must specify either File URL or File Path.",
+      optional: true,
+    },
+    ignoreDefaultVisibility: {
+      type: "boolean",
+      label: "Ignore Default Visibility",
+      description: `Whether to ignore the domain's default visibility settings for the created file. Domain administrators can choose to make all uploaded files visible to the domain by default; this parameter bypasses that behavior for the request. Permissions are still inherited from parent folders. 
+        (Default: false)
+      `,
+      optional: true,
```
jverce, line 53: **-Minor-**
Explicitly setting the default value via the prop definition improves the UX (in my opinion), since it doesn't require the user to read the whole description and it also auto-populates the field in the UI, so the user will be inclined to accept this fact and proceed to the next prop without too much friction.

```suggestion
      `,
      default: false,
```

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+    },
+    includePermissionsForView: {
+      type: "string",
+      label: "Include Permissions For View",
+      description:
+        "Specifies which additional view's permissions to include in the response. Only 'published' is supported.",
+      optional: true,
```
jverce, line 60:
```suggestion
      optional: true,
      options: [
        "published",
      ],
```

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+        "The ID of the parent folder which contain the file. If not specified as part of a create request, the file will be placed directly in the user's My Drive folder.",
+      optional: true,
+      async options({ prevContext }) {
+        const { nextPageToken } = prevContext;
+        let results;
+        if (this.drive === "myDrive") {
+          results = await this.googleDrive.listFiles({
+            pageToken: nextPageToken,
+            q: "mimeType = 'application/vnd.google-apps.folder'",
+          });
```
jverce, line 165: Just a question: since we do this same call in other places, would it be a good idea to have a new `listFolders` method in the app module?
```js
module.exports = {
  methods: {
    listFolderOptions(pageToken, opts = {}) {
      return this.listFilesOptions(pageToken, {
        ...opts,
        q: "mimeType = 'application/vnd.google-apps.folder'",
      });
    },
  },
};
```

By the way, we recently pushed a few breaking changes to the app module, so make sure you merge the latest changes in `master` onto this branch.

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+      optional: true,
+    },
+    originalFilename: {
+      type: "string",
+      label: "Original Filename",
+      description:
+        "The original filename of the uploaded content if available, or else the original value of the name field. This is only available for files with binary content in Google Drive.",
+      optional: true,
+    },
+    parents: {
```
jverce, line 152: Just for the sake of UX, I'd move this options to the top since they are more relevant to the user setting up the action: we want users to get going as fast as possible, and each additional prop that they have to configure gets them one step closer to quitting the whole thing in anger 😂 

As a reference, Zapier only provides a handful of very important props for users to configure and actually hides the rest of the options from users:
image

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+      description:
+        "Whether users with only writer permission can modify the file's permissions. Not populated for items in shared drives.",
+      optional: true,
+    },
+  },
+  async run() {
+    const drive = this.googleDrive.drive();
+    const body = this.fileUrl
+      ? await got.stream(this.fileUrl)
+      : fs.createReadStream(`${this.filePath}`);
```
jverce, line 203:
```suggestion
      : fs.createReadStream(this.filePath);
```

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,244 @@
...
+          copyRequiresWriterPermission: this.copyRequiresWriterPermission,
+        },
+        media: {
+          mimeType: this.mimeType,
+          uploadType: this.uploadType,
+          body,
+        },
+        fields: "*",
+      })
+    ).data;
```
jverce, line 242: This call should live inside a method in the app module, like we do with event sources.

jverce (changes requested):
As I mentioned in one of my comments, make sure you merge the latest changes from `master` since we did some major refactoring recently that affect the methods used for listing files from a drive.

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Added listFolders to the app module & merged master

components/google_drive/actions/create-file/create-file.js
michelle0927: Moved parents prop up

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated to use method in the app module

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,248 @@
...
+      propDefinition: [
+        googleDrive,
+        "watchedDrive",
+      ],
+    },
+    parents: {
+      type: "string",
+      label: "Parent Folder",
+      description:
+        "The ID of the parent folder which contain the file. If not specified as part of a create request, the file will be placed directly in the user's My Drive folder.",
```
compwright, line 23: Typo (`contain` should be `contains`). Also wrap this and any other long lines to 100 chars.

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,248 @@
...
+  type: "action",
+  props: {
+    googleDrive,
+    drive: {
+      propDefinition: [
+        googleDrive,
+        "watchedDrive",
+      ],
+    },
+    parents: {
```
compwright, line 19: Should probably be singular (`parent`), since you aren't accepting multiple options here.

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,248 @@
...
+      type: "string",
+      label: "Upload Type",
+      description: `The type of upload request to the /upload URI. If you are uploading data (using an /upload URI), this field is required. If you are creating a metadata-only file, this field is not required. 
+        media - Simple upload. Upload the media only, without any metadata.
+        multipart - Multipart upload. Upload both the media and its metadata, in a single request.
+        resumable - Resumable upload. Upload the file in a resumable fashion, using a series of at least two requests where the first request includes the metadata.
+      `,
+      options: [
+        {
+          label: "Media (upload the media only, without any metadata)",
```
compwright, line 51: Let's remove the explanations from the labels, since you covered that in the description.

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,248 @@
...
+      type: "boolean",
+      label: "Keep Revision Forever",
+      description:
+        "Whether to set the 'keepForever' field in the new head revision. This is only applicable to files with binary content in Google Drive. Only 200 revisions for the file can be kept forever. If the limit is reached, try deleting pinned revisions.",
+      default: false,
+    },
+    ocrLanguage: {
+      type: "string",
+      label: "OCR Language",
+      description:
```
compwright, line 104: Need options array for this prop

components/google_drive/actions/create-file/create-file.js
```diff
@@ -0,0 +1,248 @@
...
+      type: "string",
+      label: "Folder Color RGB",
+      description:
+        "The color for a folder as an RGB hex string. If an unsupported color is specified, the closest color in the palette will be used instead.",
+      optional: true,
+    },
+    mimeType: {
+      type: "string",
+      label: "Mime Type",
+      description: `The MIME type of the file. Google Drive will attempt to automatically detect an appropriate value from uploaded content if no value is provided. The value cannot be changed unless a new revision is uploaded. If a file is created with a Google Doc MIME type, the uploaded content will be imported if possible.
```
compwright, line 166: Need options array for this prop

compwright (changes requested):
@michelle0927 this is looking really good! There are a few minor UX-related things that should be addressed, and then this will be good to go.

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Updated

components/google_drive/actions/create-file/create-file.js
michelle0927: Created file language-codes.js for prop options

components/google_drive/actions/create-file/create-file.js
michelle0927: Created file google-mime-types.js for Google specific mime types and used require("mime-db") for the rest. It's too much data to load all at once, so I have it loading 10 per "page".

compwright (approved):
Looks great @michelle0927!

farzaneghm:
>

PR #1380, opened 2021-07-01
Clickup: New Actions

michelle0927:
Create Task
Create Subtask
Create List


components/clickup/actions/create-list/create-list.js
```diff
@@ -0,0 +1,72 @@
...
+      ],
+      type: "string",
+      label: "Assignee",
+      description: "Assignee to be added to this list.",
+      optional: true,
+    },
+    status: {
+      type: "string",
+      label: "Status",
+      description:
```
compwright, line 53: Inaccurate description, please remove - the documentation is confusing, and correct me if I'm wrong but I believe the status here is just a status, which can be anything you want because they are user defined. Doesn't have to be a color.

components/clickup/actions/create-subtask/create-subtask.js
```diff
@@ -0,0 +1,26 @@
+const clickup = require("../../clickup.app.js");
+const createTask = require("../create-task/create-task.js");
```
compwright, line 2: Nice job reusing from another very similar task!

This would be cleaner and more readable if you destructure here:

```javascript
const { props, run } = require("../create-task/create-task.js");
```

Then you can explicitly include `props` and `run` in the appropriate places below.

components/clickup/actions/create-task/create-task.js
```diff
@@ -0,0 +1,167 @@
...
+    },
+    dueDateTime: {
+      type: "boolean",
+      label: "Due Date Time",
+      description:
+        "Set to true if you want to enable the due date time for the task",
+      optional: true,
+    },
+    timeEstimate: {
+      type: "string",
```
compwright, line 75: Use `integer` type instead

components/clickup/actions/create-task/create-task.js
```diff
@@ -0,0 +1,167 @@
...
+      label: "Time Estimate",
+      description: "Use milliseconds",
+      optional: true,
+    },
+    startDate: {
+      type: "string",
```
compwright, line 81: Use `integer` type instead

components/clickup/actions/create-task/create-task.js
```diff
@@ -0,0 +1,167 @@
...
+      type: "boolean",
+      label: "Start Date Time",
+      description: "Select true if you want to enable the start date time",
+      optional: true,
+    },
+    notifyAll: {
+      type: "boolean",
+      label: "Notify All",
+      description:
+        "If notify_all is true, creation notifications will be sent to everyone including the creator of the task.",
```
compwright, line 97: In the description, you should refer to this by label, not by name. The user can't see the name.

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+            value: workspace.id,
+          };
+        });
+      },
+    },
+    space: {
+      type: "string",
+      label: "Space",
+      async options({ workspace }) {
+        const spaces = (await this.getSpaces(workspace)).spaces;
```
compwright, line 24: @dylburger isn't it possible to access `this.workspace` here if that prop is provided, rather than explicitly passing it in with a function? If so, is there a preference?

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+module.exports = {
+  type: "app",
+  app: "clickup",
+  propDefinitions: {
+    workspace: {
+      type: "string",
+      label: "Workspace",
+      async options() {
+        const workspaces = (await this.getWorkspaces()).teams;
+        return workspaces.map((workspace) => {
```
compwright, line 12: I generally prefer a simple map like this to not use a function block:

```javascript
return workspaces.map((workspace) => ({
 label: workspace.name,
 value: workspace.id,
}));
```

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+            label: task.name,
+            value: task.id,
+          };
+        });
+      },
+    },
+    priority: {
+      type: "integer",
+      label: "Priority",
+      description: `1 is Urgent
```
compwright, line 122: @dylburger will line breaks be preserved in the description?

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+        2 is High
+        3 is Normal
+        4 is Low`,
+      optional: true,
+    },
+  },
+  methods: {
+    _getBaseUrl() {
```
compwright, line 130: Unnecessary method, refactor into `_makeRequest()`

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+      return "https://api.clickup.com/api/v2/";
+    },
+    _getHeader() {
```
compwright, line 133: Unnecessary method, refactor into `_makeRequest()`

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+    },
+    async _makeRequest(method, endpoint, data = null, params = null) {
+      const config = {
+        headers: this._getHeader(),
+        method,
+        url: `${this._getBaseUrl()}${endpoint}`,
+        data,
+        params,
+      };
+      return (await axios(config)).data;
```
compwright, line 147: Implement retry, perhaps with [retry-axios](https://www.npmjs.com/package/retry-axios)

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+      );
+    },
+    async getFolderlessLists(spaceId) {
+      return await this._makeRequest(
+        "GET",
+        `space/${spaceId}/list?archived=false`,
+      );
+    },
+    async getWorkspaceMembers(workspaceId) {
+      const workspaces = (await this.getWorkspaces()).teams;
```
compwright, line 180: Simpler to destructure:

```javascript
const { teams } = await this.getWorkspaces();
```

components/clickup/clickup.app.js
```diff
@@ -0,0 +1,205 @@
...
+      const workspace = workspaces.filter(
+        (workspace) => workspace.id == workspaceId,
+      );
+      return workspace[0].members;
```
compwright, line 184: This will break if the `workspaceId` is not found.

compwright (changes requested):
@michelle0927 great job on this! I especially like the way you set up the API and prop definitions in the app file.

There are a few things, mostly minor, that I've flagged that need to be fixed or can be improved throughout. 

In addition, please also:

- [ ] Refactor the following common prop definitions into the main app file: `name`, `dueDate`, `dueDateTime`, (task) `parent`
- [ ] Verify that eslint passes
- [ ] Wrap all long lines (max 100 chars per line)

components/clickup/actions/create-list/create-list.js
michelle0927: This is something I copied directly from the documentation (https://jsapi.apiary.io/apis/clickup20/reference/0/lists/create-list.html)
It seems weird, but I think it's correct. When I test with something other than a color in the status field, the API returns a 400 error.

components/clickup/clickup.app.js
michelle0927: Looks like line breaks are preserved in the description in the UI.
Screen Shot 2021-08-16 at 3 54 42 PM

components/clickup/actions/create-task/create-task.js
```diff
@@ -0,0 +1,167 @@
...
+    },
+    checkRequiredCustomFields: {
+      type: "boolean",
+      label: "Check Required Custom Fields",
+      description:
+        "Indicates whether or not your new task will include data for required Custom Fields (true) or not (false). The default is false. If you set this option to true, and do not include information for required Custom Fields, then you will receive an error that 'One or more required fields is missing'.",
+      optional: true,
+    },
+    customFields: {
+      type: "string[]",
```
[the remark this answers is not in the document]
michelle0927: Tried changing this to type object[], and this is the error I get:

> component create api call responded with status: 400, body: {"error":"{\"name\":\"UserError\",\"message\":\"unsupported array element type for prop customFields: object\"}"}

components/clickup/actions/create-list/create-list.js
```diff
@@ -0,0 +1,72 @@
...
+      description: "New list name",
+    },
+    content: {
+      type: "string",
+      label: "Content",
+      description: "New list content",
+      optional: true,
+    },
+    dueDate: {
+      type: "string",
```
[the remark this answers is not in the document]
michelle0927: Updated

components/clickup/actions/create-subtask/create-subtask.js
michelle0927: Updated

components/clickup/actions/create-task/create-task.js
michelle0927: Updated

components/clickup/actions/create-task/create-task.js
michelle0927: Updated

components/clickup/actions/create-task/create-task.js
michelle0927: Updated

components/clickup/clickup.app.js
michelle0927: Updated

components/clickup/clickup.app.js
michelle0927: Updated

components/clickup/clickup.app.js
michelle0927: Updated

components/clickup/clickup.app.js
michelle0927: Updated to use axios-retry

compwright:
@michelle0927 I just deployed these to my account and I'm able to connect my account and select my workspace, but then no spaces load, and without selecting a space I cannot select any other props, and the action is unusable. Are spaces loading for you? I'm guessing there's a bug somewhere.

michelle0927:
@compwright Yes, spaces are loading for me. Is it possible that there are no Spaces in your Clickup account?

Screen Shot 2021-08-28 at 12 43 19 PM

compwright:
@michelle0927 yep that was it. Thanks!

compwright (approved):

components/clickup/clickup.app.js
dylburger: The way we implemented the `options` method, the values of props that have been configured at runtime, and which future props' `options` methods need access to, must be passed in the current manner, e.g.

```javascript
propDefinition: [
  common.props.clickup,
  "status",
  (c) => ({
    list: c.list,
  }),
],
```

I am forgetting the core _reason_ we implemented it this way (so I asked the implementor). 

I realized we also don't have docs in this in the component API, so I added an issue to write those.

dylburger:
@michelle0927 I merged latest `master` and pulled in the new ESLint rules. Would you mind addressing the items in the app file re: adding default descriptions for those props? You can ignore the warnings re: default values for `optional` props (unless any defaults make sense there).

I reviewed the rest of the components and tested them and they're looking good. Once you address the prop `description` items, please merge.

PR #1493, opened 2021-07-29
DX-2122

ajrice6713:


components/bandwidth/sources/new-incoming-sms/new-incoming-sms.js
```diff
@@ -0,0 +1,23 @@
...
+  description: "Emits an event each time a `message-received` event is received at the source url",
+  props: {
+    http: {
+      type: "$.interface.http",
+      customResponse: true,
+    },
+  },
+
+  async run(event) {
+    let messageBody = event.body[0];
```
dylburger, line 12: Small style comment: prefer `const` to `let` if you don't need to re-assignt the value later.

components/bandwidth/sources/new-incoming-sms/new-incoming-sms.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+  name: "New Incoming SMS",
+  description: "Emits an event each time a `message-received` event is received at the source url",
```
dylburger, line 3: Add a `key` (`bandwidth-new-incoming-sms`) and a `version` (0.0.1) here, as well. See https://pipedream.com/docs/components/guidelines/#required-metadata

components/bandwidth/sources/new-incoming-sms/new-incoming-sms.js
dylburger: Also, I'd recommend checking to make sure `messageBody` is defined prior to moving on:

```javascript
if (!messageBody) {
  return
}
```

Does Bandwidth send signatures with its webhook requests that can be validated here?

components/bandwidth/sources/new-incoming-sms/new-incoming-sms.js
```diff
@@ -0,0 +1,23 @@
...
+
+  async run(event) {
+    let messageBody = event.body[0];
+    this.http.respond({
+      status: 204,
+    });
+
+    if (messageBody.message.direction == "in") {
+      this.$emit(messageBody, {
+        summary: "Message Received",
```
dylburger, line 19: Is there a unique event ID Bandwidth passes in the webhooks that you can pass to the `id` property of this metadata? See the `id` section of [the `this.$emit` docs](https://pipedream.com/docs/components/api/#emit). That will help ensure that Pipedream only emits one event if we happen to receive multiple requests for the same event.

dylburger: If an event ID is available and you emit it as a part of the metadata, you'll also want to define [a dedupe strategy](https://pipedream.com/docs/components/api/#dedupe-strategies).

PR #1497, opened 2021-07-29
[REDDIT SOURCES] Fix: Reddit subreddit's posts prop options not shown

sergioeliot2039:
Fixes the issue reported by Slack space user Brian Jordan about posts not shown when setting up "New comments on a post"


dylburger:
@sergioeliot2039 can you increase the version on the Reddit components that use these methods, so we publish new versions to the registry when this is merged? If you change the app file and not the version of the component, the changes won't get published.

sergioeliot2039:
Ok,s I increased the affected components' version numbers. Had to modify to comply with the Lint/JS spell checks.

PR #1519, opened 2021-08-02
[ACTION] Bandwidth

ajrice6713:


components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
...
+    },
+    message: {
+      type: "string",
+      label: "Message",
+      description: "The text message content.",
+    },
+    messagingApplicationId: {
+      type: "string",
+      label: "Messaging Application ID",
+      description: "The ID from the messaging application created in the [Bandwidth Dashboard](https://dashboard.bandwidth.com). \n\nThe application must be associated with the location that the `from` number lives on.",
```
hamermike, line 25:
```suggestion
      description: "The ID from the messaging application created in the [Bandwidth Dashboard](https://dashboard.bandwidth.com).\n\nThe application must be associated with the location that the `from` number lives on.",
```

components/bandwidth/actions/send-sms/send-sms.js
```diff
@@ -0,0 +1,48 @@
+const bandwidth = require("../../bandwidth.app.js");
+
+module.exports = {
+  key: "bandwidth-send-sms",
+  name: "Send SMS",
+  description: "Send an SMS message using Bandwidth's Messaging API.",
+  type: "action",
+  version: "1.0.0",
+
```
dylburger, line 9: Very small comment, but you can remove this newline (I'll see if we can add this into our ESLint rules)

components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
...
+      label: "Message",
+      description: "The text message content.",
+    },
+    messagingApplicationId: {
+      type: "string",
+      label: "Messaging Application ID",
+      description: "The ID from the messaging application created in the [Bandwidth Dashboard](https://dashboard.bandwidth.com).\n\nThe application must be associated with the location that the `from` number lives on.",
+    },
+    mediaUrl: {
+      type: "string[]",
```
dylburger, line 28: Is it expected the user can enter more than one media URL?

components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
+const bandwidthMessaging = require("@bandwidth/messaging");
+
 module.exports = {
   type: "app",
   app: "bandwidth",
+  propDefinitions: {
+    messageTo: {
+      type: "string",
+      label: "To",
+      description: "The number the message will be sent to, in E164 format ex `+19195551234`.",
```
dylburger, line 10: Technically you don't need periods at the end of these descriptions, since they aren't complete sentences.

components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
...
+    },
+    from: {
+      type: "string",
+      label: "From",
+      description: "The number the call or message event will come from, in E164 format ex `+19195551234`.",
```
dylburger, line 15: Will these numbers always be tied to a Bandwidth customer's account, or can the user enter any number here?

components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
...
+    },
+    message: {
+      type: "string",
+      label: "Message",
+      description: "The text message content.",
+    },
+    messagingApplicationId: {
+      type: "string",
+      label: "Messaging Application ID",
+      description: "The ID from the messaging application created in the [Bandwidth Dashboard](https://dashboard.bandwidth.com).\n\nThe application must be associated with the location that the `from` number lives on.",
```
dylburger, line 25: Is there a Bandwidth API for retrieving this value from the user's account?

components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
...
+          to,
+        ],
+        from: from,
+        text: message,
+      };
+      try {
+        const response = await controller.createMessage(this.$auth.accountId, data);
+        return response;
+      } catch (error) {
+        console.log(error.result);
```
dylburger, line 54: When an action returns an error in a step, Pipedream should display the full error, with the stack trace, just below the step. So you can remove the `try / catch` logic and simply let the error bubble up if one occurs.

There are cases where a user may want to wrap a step with try / catch logic on their own (which we're planning to allow on a step-level), but for now, our default behavior has been to let the error bubble up and halt the workflow, which will happen anytime a step throws an error. That way the user will be notified of the error in [the global error workflow](https://pipedream.com/docs/workflows/error-handling/global-error-workflow/#global-error-workflow).

components/bandwidth/bandwidth.app.js
```diff
@@ -1,4 +1,59 @@
...
+      const data = {
+        applicationId: messagingApplicationId,
+        to: [
+          to,
+        ],
+        from: from,
+        text: message,
+      };
+      try {
+        const response = await controller.createMessage(this.$auth.accountId, data);
```
dylburger, line 51: If you're not using `response` anywhere else, I'd recommend doing:

```javascript
return await controller.createMessage(this.$auth.accountId, data);
```

to save a line.

components/bandwidth/bandwidth.app.js
ajrice6713: For messaging - yes

For outbound phone calls - no, users can create outbound calls from any number to give the appearance of an automated call coming from someones main office line, so if the end user calls back they are directed to the office line as opposed to calling a tn on bandwidth's network

components/bandwidth/bandwidth.app.js
ajrice6713: ah - here are those error docs 😅

components/bandwidth/bandwidth.app.js
dylburger: Great thanks. Does it make sense to create one prop that lists numbers in the user's account that you can use for messaging sources / actions, and use this field as a freeform text prop that you can use for phone components?

In general, we always prefer listing information tied to the user's account via API, if possible. This lets the user choose items from a list without having to look them up, and there's no room for mistaken entry. We've found it dramatically improves the usability of components, so that's why I'm asking so many questions about the API functionality.

components/bandwidth/bandwidth.app.js
dylburger: > would you rather us build that logic in as opposed to asking the user for the application ID?

If possible, yes. See my comment above re: the "From" prop. Generally, this dramatically improves the usability of components.

If you can pull this behind the scenes, it probably doesn't even need to be a prop, which would further simplify.

PR #1578, opened 2021-08-10
Upgrade Mailgun Verify action to v4 API

compwright:
The `mailgun.js` API client currently uses the old v3 validation API.

Current API documentation: https://documentation.mailgun.com/en/latest/api-email-validation.html#single-validation

Also adds an option to stop execution if the email is invalid.


compwright:
Merged with #1291

PR #1629, opened 2021-08-25
Adding eslint-plugin-pipedream and ESLint rules

dylburger:


compwright (commented):
@dylburger why is this added to a specific component rather than the project root?

dylburger:
That was a mistake, thanks for catching.

PR #1634, opened 2021-08-25
Google Sheets sources: watch files instead of drive

michelle0927:
Updates most Google Sheets sources (New Row Added, New Updates, and New Worksheet) to only watch for changes to a specified file instead of watching for changes on the drive.


dylburger (approved):
Looks great, thanks for tackling this. I'm merging now.

sirloinofbeef:
As always, thanks for your awesome work everyone.

PR #1672, opened 2021-09-08
Google Sheets - new actions

michelle0927:


compwright:

components/google_sheets/actions/find-row/find-row.js
```diff
@@ -0,0 +1,60 @@
...
+    const sheets = this.googleSheets.sheets();
+
+    const colValues = (await sheets.spreadsheets.values.get({
+      spreadsheetId: this.sheetId,
+      range: `${this.sheetName}!${this.column}:${this.column}`,
+    })).data.values;
+    return colValues.map((value, index) => {
+      if (value == this.value)
+        return index + 1;
+    }).filter((value) => value);
```
compwright, line 58: Would be better (easier to read and reason about) and more efficient to filter first, then map:

```suggestion
    return colValues.filter(value => value == this.value)
      .map((value, index) => index + 1);
```

components/google_sheets/actions/update-row/update-row.js
```diff
@@ -0,0 +1,73 @@
...
+    const cells = this.cells;
+
+    // validate input
+    if (!cells || !cells.length) {
+      throw new Error("Please enter an array of elements in `Cells / Column Values`.");
+    } else if (!Array.isArray(cells)) {
+      throw new Error("Cell / Column data is not an array. Please enter an array of elements in `Cells / Column Values`.");
+    } else if (Array.isArray(cells[0])) {
+      throw new Error("Cell / Column data is a multi-dimensional array. A one-dimensional is expected.");
+    }
```
compwright, line 60: For your consideration: [Else Considered Harmful](https://www.youtube.com/watch?v=XDqZemwVXUk)

```suggestion
    }
    if (!Array.isArray(cells)) {
      throw new Error("Cell / Column data is not an array. Please enter an array of elements in `Cells / Column Values`.");
    }
    if (Array.isArray(cells[0])) {
      throw new Error("Cell / Column data is a multi-dimensional array. A one-dimensional is expected.");
    }
```

compwright (changes requested):
@michelle0927 great job on this! Found one typo that must be fixed before proceeding, and suggested some minor improvements in a few places.

components/google_sheets/actions/find-row/find-row.js
michelle0927: When I test this with filtering first, then mapping, it seems to actually change the index/row values. For instance, if I have a column C with value 'z' in rows 5 and 13, it should return [ 5, 13 ]. When I filter it first, I get [ 1, 2 ]. Is there a better way to write this where we can keep the correct row numbers?

components/google_sheets/actions/create-worksheet/create-worksheet.js
```diff
@@ -0,0 +1,50 @@
...
+      ],
+    },
+    title: {
+      type: "string",
+      label: "Title",
+      description: "The title of the new worksheet",
+    },
+  },
+  async run() {
+    const request = {
```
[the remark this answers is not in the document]
michelle0927: I wasn't able to find a way to create a new worksheet without using batchUpdate. Would you be able to point me in the proper direction?

components/google_sheets/actions/find-row/find-row.js
compwright: Good point. In that case, you can combine both operations with reduce:

```suggestion
    return colValues.reduce((values, value, index) => {
      if (value == this.value) {
        values.push(index + 1);
      }
      return values;
    });
```

components/google_sheets/actions/create-worksheet/create-worksheet.js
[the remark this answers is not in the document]
compwright: That's weird. Evidently there is not a way!

compwright:
@michelle0927 looks like you have conflicts, be sure to resolve those too.

components/google_sheets/actions/find-row/find-row.js
michelle0927: Updated

components/google_sheets/actions/update-row/update-row.js
michelle0927: Updated

compwright (approved):

PR #1687, opened 2021-09-11
Fix typo

ml4den:
Add closing quotes


dylburger:
Thanks so much for the fix @ml4den !

PR #1689, opened 2021-09-13
[ACTIONS]: Discord bot new actions

jcortes:
# Actions
- [x] Rename Channel
- [x] Remove User Role
- [x] Add Role
- [x] Find Channel
- [x] Find User
- [x] Create Guild Channel
- [x] Create Channel Invite
- [x] Modify Channel
- [x] Delete Channel
- [x] List Channel Invites
- [x] List Channels
- [x] List Guild Members
- [x] List Users that Reacted with Emoji
- [x] **Get Message** | Return a specific message in a channel
- [x] **Delete Message** | Delete a message
- [x] **List Channel Messages** | Return the messages for a channel
- [x] **Post Reaction with Emoji** | Post a reaction for a message with an emoji


dylburger:
@jcortes thanks so much for the PR! We'll take a look at this ASAP.

components/discord_bot/actions/add-role/add-role.mjs
```diff
@@ -0,0 +1,57 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Looks like the only thing discord_bot/actions/common.mjs adds on top of discord_bot/common.mjs is the `channelId` prop. Since this action doesn't use `channelId`, I think we could just include discord_bot/common.mjs in this component and any others that don't use `channelId`.

`import common from "../..common.mjs`

components/discord_bot/actions/add-role/add-role.mjs
```diff
@@ -0,0 +1,57 @@
...
+  name: "Add Role",
+  description: "Assign a role to a user. Remember that your bot requires the `MANAGE_ROLES` permission. [See the docs here](https://discord.com/developers/docs/resources/guild#add-guild-member-role)",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    discord,
+    guildId: {
+      ...common.props.guildId,
+    },
+    userId: {
```
michelle0927, line 17: Make sure to include all common props. Then you don't need to list discord or guildId here since they're already in common.props.
```
props: {
  ...common.props,
   userId: {
```

components/discord_bot/actions/common.mjs
```diff
@@ -0,0 +1,20 @@
...
+      propDefinition: [
+        discord,
+        "channelId",
+        ({ guildId }) => ({
+          guildId,
+          notAllowedChannels: constants.NOT_ALLOWED_CHANNELS,
+        }),
+      ],
+    },
+  },
```
michelle0927, line 19: Make sure to include everything from common:
```
export default {
  ...common,
  props: {
    ...common.props,
    channelId: {
```

components/discord_bot/actions/create-channel-invite/create-channel-invite.mjs
```diff
@@ -0,0 +1,96 @@
...
+          {
+            label: "Stream",
+            value: constants.INVITE_TARGET_TYPES.STREAM,
+          },
+          {
+            label: "Embedded Application",
+            value: constants.INVITE_TARGET_TYPES.EMBEDDED_APPLICATION,
+          },
+        ];
+      },
```
michelle0927, line 53: Just noticed that this doesn't need to be a function. You can just return the array:
```
options: [
  {
    label: "Stream",
    value: constants.INVITE_TARGET_TYPES.STREAM,
  },
  {
    label: "Embedded Application",
    value: constants.INVITE_TARGET_TYPES.EMBEDDED_APPLICATION,
  },
]
```

components/discord_bot/actions/create-guild-channel/create-guild-channel.mjs
```diff
@@ -0,0 +1,159 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Since this component doesn't use `channelId`, we should include the common.mjs file in the discord_bot directory.
`import common from "../../common.mjs"`

components/discord_bot/actions/create-guild-channel/create-guild-channel.mjs
```diff
@@ -0,0 +1,159 @@
...
+  name: "Create Guild Channel",
+  description: "Create a new channel for the guild. [See the docs here](https://discord.com/developers/docs/resources/guild#create-guild-channel)",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    discord,
+    guildId: {
+      ...common.props.guildId,
+    },
+    name: {
```
michelle0927, line 18: Include `common.props`, and you can leave out `discord` and `guildId` since they're already included in `common.props`.
```
props: {
  ...common.props,
  name: {
```

components/discord_bot/actions/create-guild-channel/create-guild-channel.mjs
```diff
@@ -0,0 +1,159 @@
...
+          {
+            label: "Store",
+            value: String(constants.CHANNEL_TYPES.GUILD_STORE),
+          },
+          {
+            label: "Stage Voice",
+            value: String(constants.CHANNEL_TYPES.GUILD_STAGE_VOICE),
+          },
+        ];
+      },
```
michelle0927, line 51: You can just return the array here:
```
options: [
  {
    label: "Text",
    value: String(constants.CHANNEL_TYPES.GUILD_TEXT),
  },
  {
    label: "Voice",
    value: String(constants.CHANNEL_TYPES.GUILD_VOICE),
  },
  {
```

components/discord_bot/actions/delete-channel/delete-channel.mjs
```diff
@@ -0,0 +1,19 @@
...
+export default {
+  ...common,
+  key: "discord_bot-delete-channel",
+  name: "Delete Channel",
+  description: "Delete a Channel.",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    ...common.props,
+  },
```
michelle0927, line 12: No need to overwrite `props` here since no additional props are being added. You can leave out the `props` section here.

components/discord_bot/actions/find-channel/find-channel.mjs
```diff
@@ -0,0 +1,41 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Since this component doesn't use `channelId`, you can include common.mjs from the discord_bot directory.
`import common from "../../common.mjs"`;

components/discord_bot/actions/find-channel/find-channel.mjs
```diff
@@ -0,0 +1,41 @@
...
+  name: "Find Channel",
+  description: "Find an existing channel by name. [See the docs here](https://discord.com/developers/docs/resources/guild#get-guild-channels)",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    discord,
+    guildId: {
+      ...common.props.guildId,
+    },
+    channelName: {
```
michelle0927, line 17: Include `common.props` and you can leave out `discord` and `guildId`.
```
props: {
  ...common.props,
  channelName: {
```

components/discord_bot/actions/find-user/find-user.mjs
```diff
@@ -0,0 +1,39 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Since this component doesn't use `channelId`, you can include common.mjs from the discord_bot directory.
`import common from "../../common.mjs"`;

components/discord_bot/actions/find-user/find-user.mjs
```diff
@@ -0,0 +1,39 @@
...
+  name: "Find User",
+  description: "Find an existing user by name. [See the docs here](https://discord.com/developers/docs/resources/guild#search-guild-members)",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    discord,
+    guildId: {
+      ...common.props.guildId,
+    },
+    query: {
```
michelle0927, line 17:
```
props: {
  ...common.props,
  query: {
```

components/discord_bot/actions/list-channel-invites/list-channel-invites.mjs
```diff
@@ -0,0 +1,35 @@
...
+  ...common,
+  key: "discord_bot-list-channel-invites",
+  name: "List Channel Invites",
+  description: "Return a list of invitees for the channel. Only usable for guild channels.",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    ...common.props,
+    channelId: {
+      propDefinition: [
```
michelle0927, line 16: `channelId` is already included in common.props, so you can remove this `props` section.

components/discord_bot/actions/list-channels/list-channels.mjs
```diff
@@ -0,0 +1,24 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Since this component doesn't use `channelId`, you can include common.mjs from the discord_bot directory.
`import common from "../../common.mjs"`;

components/discord_bot/actions/list-channels/list-channels.mjs
```diff
@@ -0,0 +1,24 @@
...
+  name: "List Channels",
+  description: "Return a list of channels. [See the docs here](https://discord.com/developers/docs/resources/guild#get-guild-channels)",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    discord,
+    guildId: {
+      ...common.props.guildId,
+    },
+  },
```
michelle0927, line 17: Since `guildId` and `discord` are already included in common.props, you can remove this `props` section.

components/discord_bot/actions/list-guild-members/list-guild-members.mjs
```diff
@@ -0,0 +1,39 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Since this component doesn't use `channelId`, you can include common.mjs from the discord_bot directory.
`import common from "../../common.mjs"`;

components/discord_bot/actions/modify-channel/modify-channel.mjs
```diff
@@ -0,0 +1,223 @@
...
+      description: "Please select the channel you'd like to change its name",
+      propDefinition: [
+        discord,
+        "channelId",
+        ({ guildId }) => ({
+          guildId,
+          notAllowedChannels: NOT_ALLOWED_CHANNELS,
+        }),
+      ],
+    },
```
michelle0927, line 34: You can leave out `channelId` here since it's already included in common.mjs

components/discord_bot/actions/remove-user-role/remove-user-role.mjs
```diff
@@ -0,0 +1,58 @@
+import common from "../common.mjs";
```
michelle0927, line 1: Since this component doesn't use `channelId`, you can include common.mjs from the discord_bot directory.
`import common from "../../common.mjs"`;

components/discord_bot/actions/remove-user-role/remove-user-role.mjs
```diff
@@ -0,0 +1,58 @@
...
+  name: "Remove User Role",
+  description: "Remove a selected role from the specified user. [See the docs here](https://discord.com/developers/docs/resources/guild#remove-guild-member-role)",
+  type: "action",
+  version: "0.0.1",
+  props: {
+    discord,
+    guildId: {
+      ...common.props.guildId,
+    },
+    userId: {
```
michelle0927, line 17:
```
props: {
  ...common.props,
  userId: {
```

components/discord_bot/sources/common.mjs
```diff
@@ -0,0 +1,16 @@
+import constants from "../constants.mjs";
+import common from "../common.mjs";
+
+export default {
+  props: {
+    ...common.props,
+  },
+  methods: {
+    _getLastMessageIDs() {
```
michelle0927, line 9: Include all of `common` here, and you can leave out the `props` section. Also, don't forget to include common.methods here.
```
export default {
  methods: {
    ...common.methods,
    _getLastMessageIDs() {
```

michelle0927 (changes requested):
Looking good, just a little more refactoring.

components/discord_bot/actions/list-guild-members/list-guild-members.mjs
jcortes: In this case I had to do it that way since the `description` for `guildId` is a bit different

components/discord_bot/actions/list-channel-invites/list-channel-invites.mjs
jcortes: In this case I had to do it a bit different since the property `notAllowedChannels` in `channelId` property is a bit different

components/discord_bot/actions/modify-channel/modify-channel.mjs
jcortes: Yeah same here since `notAllowedChannels` and `description` are a bit different so I need to rewrite

components/discord_bot/actions/list-guild-members/list-guild-members.mjs
michelle0927: Yep, you're right. You should leave in `guildId` so you can overwrite the description. This action doesn't use `channelId` though, right? So you could import `"../../common.mjs"` instead of `"../common.mjs"`.

```
props: {
   ....common.props,
   guildId: {
```

michelle0927 (changes requested):

components/discord_bot/actions/list-guild-members/list-guild-members.mjs
michelle0927: Was the props section updated to include `...common.props` and remove `discord`? I'm not seeing that change.

jcortes: yeah but remember the description here is a bit different

michelle0927: Yes, you want to overwrite guildId to add the new description. However, this component is still inheriting from common, so when you're overwriting the props section, you should add back in any base props from common with the line `...common.props,`. Then, since `discord` is a prop in common, you don't need to list it again in the component props.

jcortes: Yeah I totally missed that, thanks!

michelle0927:
The timer in the source `new-messages-in-channel` should default to 15 minutes instead of 1 minute.
```
    timer: {
      type: "$.interface.timer",
      default: {
        intervalSeconds: 60 * 15,
      },
    },
```

michelle0927:
I noticed the prop `after` being used in actions `list-channel-messages`, `list-guild-members`, & `list-users-with-emoji-reactions`. Can I ask about the purpose of that prop? It seems like a variable that would be used for pagination in combination with `limit`. Instead of having the user specify `after`, I'd create a loop in `async run()` to paginate through all messages/members/users. However, since a channel might have a ton of messages, we might want to have a maxLimit to return. Does the endpoint `/channels/${channelId}/messages` return the messages in reverse chronological order? I'd assume most users would want the most recent messages, so if they want a max of 20 messages returned, it would be the 20 most recent messages.

The loop would be something like:
```
const allTheItems = [];
const params = stuff from props
do {
  const items = make api call
  totalReturned = items.length
  allTheItems.push(items)
  params.after = id of last item in list
} while(totalReturned == limit && allTheItems.length <= max)
```

Once the number of items returned is less than the limit, we know there are no more pages.

jcortes:
> essages/members/use

Yes according to the [docs](https://discord.com/developers/docs/resources/channel#get-channel-messages) the `after` query string parameter is represented by a `messageId` so the system can paginate after that id. It can either be combined or not with `limit` since by default is `50`. 
The loop for me makes more sense in `source` components which is done there, since they are just reading data from the endpoint without having to interact too much with the configuration while as for `action` components doesn't make sense, because the user will lose flexibility of the API. That's why I don't loop over in `action` components. Unless that is a standard for action components. And yes the messages returned are latest first, so you always will have the most recent messages.

components/discord_bot/actions/common.mjs
```diff
@@ -0,0 +1,70 @@
...
+            before,
+            after,
+            around,
+          }
+          : {
+            limit: lastLimit < limit
+              ? lastLimit
+              : limit,
+            before: lastId,
+          };
```
michelle0927, line 35: I think `after` should be included here as well so that we don't accidentally go past the message/item specified by `after` in case it's not on the first page of results.

components/discord_bot/actions/list-guild-members/list-guild-members.mjs
```diff
@@ -0,0 +1,39 @@
...
+      ],
+    },
+  },
+  async run({ $ }) {
+    return await this.discord.getGuildMembers({
+      $,
+      guildId: this.guildId,
+      limit: this.limit,
+      after: this.after,
+    });
```
michelle0927, line 37: Since this uses `after` and `limit`, we should include pagination here as well. You might be able to update the `paginateMessages` method in `common.mjs` to work with resources other than messages. Check out the `paginate` method in `twitch/sources/common.js`. It takes a parameter called `resourceFn` that tells it which method from the app file to get results from.

components/discord_bot/actions/list-users-with-emoji-reactions/list-users-with-emoji-reactions.mjs
```diff
@@ -0,0 +1,80 @@
...
+        .map((messageId) => {
+          return this.discord.getUserReactions({
+            $,
+            channelId,
+            emoji,
+            messageId,
+            after,
+            limit,
+          });
+        });
```
michelle0927, line 65: Since `getUserReactions` uses `after` and `limit`, we should include pagination here as well.

michelle0927 (changes requested):

components/discord_bot/actions/common.mjs
jcortes: I didn't quite get what you meant here. Since `before`, `after` and `around` are mutually exclusive I can't use `before` and `after` at the same time. And for me It would make more sense if I paginate using `before` because I will always get the latest message first so I will be paginating backwards in time.

michelle0927: Oh yeah, you’re right. We do want to check that we’re not going past the after ID though. Can we check for it at the bottom of the loop, or as the messages are added to the array? Once we’ve reached the after ID, we don’t need to continue the pagination.

components/discord_bot/actions/list-users-with-emoji-reactions/list-users-with-emoji-reactions.mjs
jcortes: There is a question here I neeed to answer first, would it be better looping over messages or just asking the user for the exact `messageId` with a `prop`? I would rather go with `messageId`. What do you think about it?

michelle0927: Integromat's version of this action asks for a `messageId`, so I agree that asking for a `messageId` is best.

michelle0927 (approved):

components/discord_bot/discord_bot.app.mjs
```diff
@@ -0,0 +1,530 @@
...
+
+        return {
+          options,
+          context: {
+            after: lastMemberId,
+          },
+        };
+      },
+    },
+    max: {
```
dannyroosevelt, line 241: @jcortes shouldn't `max` be optional?

jcortes: Yeah I agree, let me do the change then!

dannyroosevelt:
@jcortes plz merge away!

PR #1694, opened 2021-09-14
Discord Webhook: New Action send-message-with-file

michelle0927:


components/discord_webhook/package.json
```diff
@@ -0,0 +1,22 @@
...
+    "discord_webhook"
+  ],
+  "homepage": "https://pipedream.com/apps/discord_webhook",
+  "author": "Pipedream <support@pipedream.com> (https://pipedream.com/)",
+  "license": "MIT",
+  "dependencies": {
+    "axios": "^0.21.1",
+    "form-data": "^4.0.0",
+    "request": "^2.88.2",
+    "fs": "^0.0.1-security"
```
dylburger, line 17: Did you add this manually or did this come from somewhere else? I don't believe we'll need it since `fs` is built-in.

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
```diff
@@ -0,0 +1,73 @@
+const discordWebhook = require("../../discord_webhook.app.js");
+var request = require("request");
```
dylburger, line 2: [`request` is deprecated](https://www.npmjs.com/package/request). Could you use `axios`, instead?

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
```diff
@@ -0,0 +1,73 @@
...
+      label: "File URL",
+      description:
+        "The URL of the file to attach. Must specify either File URL or File Path.",
+      optional: true,
+    },
+    filePath: {
+      type: "string",
+      label: "File Path",
+      description:
+        "The path to the file, e.g. /tmp/myFile.csv . Must specify either File URL or File Path.",
```
dylburger, line 48: Can you also add a runtime check for this to make this error clear to the user? We've found that users hit these cases a lot, so adding an error like this can help:

```javascript
if (!fileUrl && !filePath) {
  throw new Error("This action requires either File URL or File Path. Please enter one or the other above.");
}
```

@dannyroosevelt this use case doesn't come up a lot, but occasionally actions require _at least one of N_ props to work correct. For example, you can pass _either_ message text _or_ a "blocks" object (rich text messages) to the Slack Send a message action. Neither are required, but at least one of them is. We have no elegant way to surface that to the user, and I haven't checked to see what Zapier / Integromat do.

dylburger: Separately, can you update the description to read:

```
The path to the file, e.g. `/tmp/myFile.csv`. Must specify either **File URL** or **File Path**.
```

I'll add this to the component guidelines, but I'm norming on using `code` syntax anywhere there's a reference to a filepath / example JSON / other code, and **bold** when referencing the names of props.

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
```diff
@@ -0,0 +1,73 @@
...
+      propDefinition: [
+        discordWebhook,
+        "avatarURL",
+      ],
+    },
+    fileUrl: {
+      type: "string",
+      label: "File URL",
+      description:
+        "The URL of the file to attach. Must specify either File URL or File Path.",
```
dylburger, line 41: Bold File URL and File Path (see note on the File Path prop below)

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
```diff
@@ -0,0 +1,73 @@
...
+        "The path to the file, e.g. /tmp/myFile.csv . Must specify either File URL or File Path.",
+      optional: true,
+    },
+  },
+  async run() {
+    const {
+      message: content,
+      avatarURL,
+      threadID,
+      username,
```
dylburger, line 57: destructure `fileUrl` and `filePath` here, as well

components/discord_webhook/discord_webhook.app.js
```diff
@@ -1,20 +1,24 @@
...
     },
     embeds: {
       type: "any",
+      label: "Embeds",
       description: "Optionally pass an [array of embed objects](https://birdie0.github.io/discord-webhooks-guide/discord_webhook.html). E.g., ``{{ [{\"description\":\"Use markdown including *Italic* **bold** __underline__ ~~strikeout~~ [hyperlink](https://google.com) `code`\"}] }}``. To pass data from another step, enter a reference using double curly brackets (e.g., `{{steps.mydata.$return_value}}`).\nTip: Construct the `embeds` array in a Node.js code step, return it, and then pass the return value to this step.",
       optional: true,
     },
     username: {
       type: "string",
+      label: "Username",
```
dylburger, line 21: Thanks for adding these labels

components/discord_webhook/discord_webhook.app.js
```diff
@@ -62,5 +66,33 @@ module.exports = {
...
+      if (file) data.append("file", file);
+      const resp = await axios({
+        method: "POST",
+        url: this.url(),
+        headers: {
+          "Content-Type": "multipart/form-data; boundary=" + data._boundary,
+        },
+        validateStatus: () => true,
+        params: {
+          thread_id: threadID
```
dylburger, line 85: I would do just:

```javascript
params: {
  thread_id: threadID,
}
```

I see that `sendMessage` does the same as you did here, so if you don't mind, please refactor that method, as well.

dylburger (changes requested):
Very small changes, good work

components/discord_webhook/package.json
michelle0927: Updated to remove `fs` and `request`

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
michelle0927: Updated to use `axios`

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
michelle0927: Added runtime check and updated descriptions

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
michelle0927: Updated

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
michelle0927: Updated

components/discord_webhook/discord_webhook.app.js
michelle0927: Updated these and added `if (!threadID) threadID = undefined;` above to avoid getting this error:
`Error: {"thread_id":["Value \"\" is not snowflake."]}`

components/discord_webhook/actions/send-message-with-file/send-message-with-file.js
```diff
@@ -0,0 +1,83 @@
...
+  description: "Post a message with an attached file",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    discordWebhook,
+    message: {
+      propDefinition: [
+        discordWebhook,
+        "message",
+      ],
```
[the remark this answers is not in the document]
michelle0927: Updated to make `message` optional.

PR #1718, opened 2021-09-21
[ACTIONS] Google Drive new actions

### LINKED ISSUE #1432: [ACTION] Google Drive new actions
Google Drive

https://developers.google.com/drive/api/v3/reference

**Replace File** | Upload a file that replaces an existing file
**Add File Sharing Preference** | Add a sharing scope to the sharing preference of a file and provides a sharing URL
**Find Folder** | Search for a specific folder by name
**Create New File From Text** | Create a new file from plain text
**Get Folder ID for a Path** | Retrieve a folderId for a path
**Create Folder** | Create a new empty folder
**Upload File**  | Copy an existing file to Google Drive
**Copy File** | Create a copy of the specified file
**Move File** | Move a file from one folder to another
**Find File** | Search for a specific file by name
**Download File** | Download a file
**Update File** | Update a file's metadata and/or content
**Move File to Trash** | Move a file or folder to trash
**Delete File** | Permanently delete a file or folder without moving it to the trash
**Search for Shared Drives** | Search for shared drives with query options
**Create Shared Drive** | Create a new shared drive
**Get Shared Drive** | Get a shared drive's metadata by ID
**Update Shared Drive** | Update an existing shared drive
**Delete Shared Drive** | Delete a shared drive without any content

### LINKED ISSUE #1801: [FEATURE] speed up async options for Google Drive app integrations
The async option for selecting "My Drive" or "Shared Drive" is slow when configuring Actions and Sources for Google Sheets or Google Drive. Since most users are typically selecting "My Drive", can we set that as the default Drive, while still allowing users to change to "Shared Drive" if needed?

js07:
Add Google Drive actions:

- Replace File
- Add File Sharing Preference
- Find Folder
- Create New File From Text
- Get Folder ID for a Path
- Create Folder
- Upload File
- Copy File
- Move File
- Find File
- Download File
- Update File
- Move File to Trash
- Delete File
- Search for Shared Drives
- Create Shared Drive
- Get Shared Drive
- Update Shared Drive
- Delete Shared Drive

Sets "My Drive" as the default Drive for Google Drive and Google Sheets actions and sources. Users can still select a shared drive by enabling [structured mode](https://pipedream.com/docs/workflows/steps/params/#structured-params).

Comments:

- API calls using Google API client for Shared Drives are not yet fully tested because `drives` [create](https://developers.google.com/drive/api/v3/reference/drives/create), [update](https://developers.google.com/drive/api/v3/reference/drives/update), [delete](https://developers.google.com/drive/api/v3/reference/drives/delete), and [some](https://developers.google.com/drive/api/v3/ref-search-terms#drive_properties) [list](https://developers.google.com/drive/api/v3/reference/drives/list) parameters require [Domain Admin Privilege](https://support.google.com/a/answer/1219251?hl=en).

Resolves #1432, resolves #1801


dylburger:

js07:
While testing, files listed using the `googleapis` client library stopped filtering according to the query parameter `q`. For example:
```js
drive.files.list({
  q: 'mimeType = 'application/vnd.google-apps.folder''
})
```
A few days ago, this request would return only Google Drive folders (files with Google's folder MIME type). But now, it returns all files.
I've tried using an [older version](https://github.com/googleapis/google-api-nodejs-client/tree/v67.0.0) of the `googleapis` npm package through a user-scoped package, but that didn't seem to fix the issue.
If I'm not mistaken, this issue seems to affect other actions that use a filtered result from `drive.files.list`, like [Google Sheets actions](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_sheets/google_sheets.app.js#L99).

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,96 @@
+const googleDrive = require("../../google_drive.app");
+
+module.exports = {
+  key: "google_drive-add-file-sharing-preference",
+  name: "Add File Sharing Preference",
+  description: "Add File Sharing Preference",
```
dylburger, line 6: Is there a document you can link to here that describes what file sharing preferences are, or otherwise expands on what you have in `name`?

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,96 @@
...
+  version: "0.0.12",
```
dylburger, line 7: We need to figure out a better way to handle this, but the registry version should start at `0.0.1` for all of these actions. It's hard when you're constantly running `pd publish` and have to increment the version in your own account. We plan to ship a `pd dev` for actions at some point, which should facilitate testing.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,96 @@
...
+  type: "action",
+  props: {
+    googleDrive,
+    drive: {
+      propDefinition: [
+        googleDrive,
+        "watchedDrive",
+      ],
+      description: "The drive you want to find a file in",
+      optional: true,
```
dylburger, line 17: What happens if the user doesn't provide a drive? Shouldn't this be required, since the `fileId` depends on this prop being set?

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,96 @@
...
+      default: "",
```
dylburger, line 18: I see, so if we default to `""`, that means we end up fetching data for the user's authenticated drive - is that right?

Would you mind taking a look at [Zapier](https://zapier.com/) to see how they implement? They have a solid user experience, so I'd like to see if they make that prop optional, as well.

@dannyroosevelt when we get to the QA stage with this we should talk about whether to make the drive optional or required. If it's not set, it defaults to your current drive. I think this behavior could be OK, it just looks a little odd in the current UI because of how we handle optional props.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,96 @@
...
+        "fileId",
+        (c) => ({
+          drive: c.drive,
+          baseOpts: {
+            q: "'me' in owners",
+          },
+        }),
+      ],
+      optional: false,
+      description: "The file to add a file sharing preference to.",
```
dylburger, line 32: You can remove the period at the end of this description and others since it's not a complete sentence.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,96 @@
...
+        "anyone",
+      ],
+    },
+    domain: {
+      type: "string",
+      label: "Domain",
+      description:
+        "The domain to which this permission refers if type is 'domain'.",
+      optional: true,
+      default: "",
```
dylburger, line 68: Just to confirm, is `""` the correct default for this param, or `undefined`?

components/google_drive/actions/create-file-from-text/create-file-from-text.js
```diff
@@ -0,0 +1,66 @@
+const googleDrive = require("../../google_drive.app");
+const { Readable } = require("stream");
+
+module.exports = {
+  key: "google_drive-create-file-from-text",
+  name: "Create New File From Text ",
```
dylburger, line 6: Trailing space

components/google_drive/actions/create-folder/create-folder.js
```diff
@@ -0,0 +1,51 @@
...
+      label: "Name",
+      description: "The name of the new folder.",
+      optional: true,
+      default: "",
+    },
+  },
+  async run() {
+    const {
+      parentId,
+      name = undefined,
```
dylburger, line 44: `undefined` should already be the default if `name` is not set

components/google_drive/actions/download-file/download-file.js
```diff
@@ -0,0 +1,84 @@
...
+    let file;
+    if (mimeType.includes(GOOGLE_DRIVE_MIME_TYPE_PREFIX)) {
+      // Google MIME type
+      // See https://developers.google.com/drive/api/v3/mime-types for a list of
+      // Google MIME types.
+      file = (
+        await drive.files.export(
+          {
+            fileId: this.fileId,
+            mimeType: "application/pdf",
```
dylburger, line 60: Is `application/pdf` intentional?

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
js07: If the user doesn't provide a drive, `files.list` searches files in the `user` [corpus](https://developers.google.com/drive/api/v3/search-files#search_the_corpora) by default, which [includes](https://developers.google.com/drive/api/v3/about-files#org) all files in "My Drive" and "Shared with me."

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
js07: Yes, if we default to `""` or undefined, we fetch data for the user's authenticated drive, including files in the "Shared with me" category.

js07: I set a default of `""` because of the ESLint warning, and because `""` currently works the same as `undefined` because the `drive` prop is used in a loose equality check. But I'm guessing there's a better solution. I meant to leave a comment about defaults for optional props. 😅

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
js07: `undefined` is the correct default for this param.

components/google_drive/actions/download-file/download-file.js
js07: Yes, but other functionality might be preferable. Besides Apps Scripts Google Workspace Files that have a [Google Drive specific MIME type](https://developers.google.com/drive/api/v3/mime-types) can be be converted to `application/pdf` when using [export](https://developers.google.com/drive/api/v3/reference/files/export). I should add a comment about that.

js07: Maybe we could add a `File Type` prop to choose the type of the downloaded file, and the options can be determined by the type of the selected file. The [About resource](https://developers.google.com/drive/api/v3/reference/about#resource) in the Google Drive API contains an `exportFormats` object, which is "a map of source MIME type to possible targets for all supported exports".

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
dylburger: Thanks. I'll defer to @dannyroosevelt on the question of whether the prop should be optional (from a product perspective). I think the default listing is probably what most users want. The only issue from my perspective is that it's listed as an optional prop first, which our current UI doesn't handle well. But long-term keeping it optional is probably best.

Can you remove the `default: ""` here and anywhere where `undefined` will work as a default? I think I'll probably just remove the ESLint warning. In general I'll only add `default` when there's a non-empty value.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
dylburger: Thanks, you can remove completely then (see prior comment)

components/google_drive/actions/download-file/download-file.js
dylburger: @dannyroosevelt can you sync with Jacob on this one? We should probably think through the ideal experience for this action / see what Zapier does.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,97 @@
...
+module.exports = {
+  key: "google_drive-add-file-sharing-preference",
+  name: "Add File Sharing Preference",
+  description:
+    "Add a sharing permission to the sharing preferences of a file and provide a sharing URL",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    googleDrive,
+    /* eslint-disable pipedream/default-value-required-for-optional-props */
```
js07, line 12: I used an ESLint configuration comment to disable the `pipedream/default-value-required-for-optional-prop` (and silence the warning). `/* eslint-disable */` disables the rule for the entire file. Should we use `/* eslint-disable-next-line */` (which disables the rule for the next line) instead before each optional prop without a default value?

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
js07: Zapier makes the `drive` prop optional, and it's positioned as the first prop to configure. On Zapier, its description includes "If nothing is selected, then your personal Google Drive will be used."

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,97 @@
...
+    drive: {
+      propDefinition: [
+        googleDrive,
+        "watchedDrive",
+      ],
+      description:
+        "The drive to use. If not specified, your personal Google Drive will be used. If you are connected with any [Google Shared Drives](https://support.google.com/a/users/answer/9310351), you can select it here.",
```
js07, line 19: This (updated) description for the `drive` prop is repeated for many of the actions. The `drive` prop (including its base `description`) is used in other components. Should we move the `drive` prop to a `common` action file, is repeating the description OK, or should the descriptions be changed to be more action-specific?

components/google_drive/actions/download-file/download-file.js
js07: Zapier doesn't seem to have a **Download File** action.
If "Show advanced settings" is enabled, Integromat's **Download File** Action shows a prop to choose a file conversion format for each Google Workspace file type:
downloadprops3

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
js07: Updated the [description](https://github.com/PipedreamHQ/pipedream/pull/1718/files#diff-285e469c2f1ee33054a35177fab40ba02a5bfb76cd9663ed31efd25c100be8beR12-R13) to be more descriptive.

vellames-turing:
@dylburger Getting this PR for my review

vellames-turing (changes requested):
@js07 
I am starting to do a more detailed review, meanwhile, could you please convert all files to ESModules and add the `summary` for each action?

vellames-turing (changes requested):
I realized that the descriptions are without the document reference. Could you please add it?

components/google_drive/google_drive.app.js
```diff
@@ -26,11 +31,51 @@ module.exports = {
...
+      description: "The file in the drive",
+      options({
+        prevContext,
+        drive,
+        baseOpts = {
+          q: `mimeType != '${GOOGLE_DRIVE_FOLDER_MIME_TYPE}'`,
+        },
+      }) {
+        const { nextPageToken } = prevContext;
+        return this.listDriveFilesOptions(drive, nextPageToken, baseOpts);
```
vellames-turing, line 61: Can you add the path file? I think in some cases, the user may have the same file name in different folders. Show the file may help it

components/google_drive/google_drive.app.js
```diff
@@ -26,11 +31,51 @@ module.exports = {
...
+      description: "The folder in the drive",
+      options({
+        prevContext,
+        drive,
+        baseOpts = {
+          q: `mimeType = '${GOOGLE_DRIVE_FOLDER_MIME_TYPE}'`,
+        },
+      }) {
+        const { nextPageToken } = prevContext;
+        return this.listDriveFilesOptions(drive, nextPageToken, baseOpts);
```
vellames-turing, line 46: There is some way to show the entire folder path? I think we may have a scenario where the user has two folders with the same name but in different paths. Does it make sense?

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,104 @@
...
+          drive: c.drive,
+          baseOpts: {
+            q: "'me' in owners",
+          },
+        }),
+      ],
+      optional: false,
+      description: "The file to share",
+    },
+    role: {
```
vellames-turing, line 41: Maybe you can move this to the app? Looks like a variable that can be used in future

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
```diff
@@ -0,0 +1,104 @@
...
+      options: [
+        "owner",
+        "organizer",
+        "fileOrganizer",
+        "writer",
+        "commenter",
+        "reader",
+      ],
+    },
+    type: {
```
vellames-turing, line 56: Maybe you can move this to the app? Looks like a variable that can be used in future

components/google_drive/actions/download-file/download-file.js
```diff
@@ -0,0 +1,91 @@
...
+      ],
+      description: "The file to download",
+    },
+    filePath: {
+      type: "string",
+      label: "Destination File Path",
+      description:
+        "The destination path for the file in /tmp (e.g., `/tmp/myFile.csv`).",
+    },
+    mimeType: {
```
vellames-turing, line 49: Do you have any reason to use the `mimeType` from `app`? Looks like you are changing everything. I am wondering if you need to use this `propDefinition` here

components/google_drive/actions/find-file/find-file.js
```diff
@@ -0,0 +1,34 @@
...
+        googleDrive,
+        "fileNameSearchTerm",
+      ],
+    },
+  },
+  async run() {
+    const opts = getListFilesOpts(this.drive || undefined, {
+      q: `name contains '${this.nameSearchTerm}'`,
+    });
+    return (await this.googleDrive.listFilesInPage(null, opts)).files;
```
vellames-turing, line 32: Maybe you can use `lodash.get` or `?` before trying to access `.files`?

components/google_drive/actions/find-folder/find-folder.js
```diff
@@ -0,0 +1,37 @@
...
+        "fileNameSearchTerm",
+      ],
+      description: "The name of the folder to search for",
+    },
+  },
+  async run() {
+    const opts = getListFilesOpts(this.drive || undefined, {
+      q: `mimeType = '${GOOGLE_DRIVE_FOLDER_MIME_TYPE}' and name contains '${this.nameSearchTerm}'`,
+    });
+    return (await this.googleDrive.listFilesInPage(null, opts)).files;
```
vellames-turing, line 35: Maybe you can use `lodash.get` or `?` before trying to access `.files`?

components/google_drive/actions/search-shared-drives/search-shared-drives.js
```diff
@@ -0,0 +1,32 @@
...
+      ],
+    },
+  },
+  async run() {
+    return (
+      await this.googleDrive.searchDrives({
+        q: this.q,
+        useDomainAdminAccess: this.useDomainAdminAccess,
+      })
+    ).drives;
```
vellames-turing, line 30: Maybe you can use `lodash.get` or `?` before trying to access `.drives`?

components/google_drive/actions/upload-file/upload-file.js
```diff
@@ -0,0 +1,89 @@
...
+    if (!fileUrl && !filePath) {
+      throw new Error("One of File URL and File Path is required.");
+    }
+    const file = await getFileStream({
+      fileUrl,
+      filePath,
+    });
+    return await this.googleDrive.createFileFromOpts({
+      file,
+      mimeType: mimeType || undefined,
```
vellames-turing, line 84: Do you really need this `|| undefined` here?

components/google_drive/google_drive.app.js
```diff
@@ -26,11 +31,51 @@ module.exports = {
...
+    },
+    fileOrFolderId: {
+      type: "string",
+      label: "File or Folder",
+      description: "The file or folder in the drive",
+      options({
+        prevContext, drive, baseOpts = {},
+      }) {
+        const { nextPageToken } = prevContext;
+        return this.listDriveFilesOptions(drive, nextPageToken, baseOpts);
```
vellames-turing, line 72: If possible, it would be nice to show the path here as well

components/google_drive/google_drive.app.js
```diff
@@ -599,5 +730,431 @@ module.exports = {
...
+      if (parentId) {
+        q += ` and '${parentId}' in parents`;
+      }
+      if (excludeTrashed) {
+        q += " and trashed != true";
+      }
+      const listOpts = getListFilesOpts(driveProp, {
+        q,
+      });
+      return (await drive.files.list(listOpts)).data.files;
```
vellames-turing, line 768: I'd like to suggest you to use `?` or `lodash.get` to get the `data.files`

components/google_drive/actions/update-shared-drive/update-shared-drive.js
```diff
@@ -0,0 +1,73 @@
...
+      propDefinition: [
+        googleDrive,
+        "useDomainAdminAccess",
+      ],
+    },
+    backgroundImageLink: {
+      type: "string",
+      label: "Background Image Link",
+      description:
+        "A link to the newbackround image for the shared drive. Cannot be set if `Theme ID` is set in the same request.",
```
vellames-turing, line 29:
```suggestion
        "A link to the new background image for the shared drive. Cannot be set if `Theme ID` is set in the same request.",
```

components/google_drive/google_drive.app.js
```diff
@@ -343,12 +437,30 @@ module.exports = {
...
      */
+
     async listFolderOptions(pageToken, opts = {}) {
       return await this.listFilesOptions(pageToken, {
         ...opts,
         q: "mimeType = 'application/vnd.google-apps.folder'",
       });
     },
+    /**
+     * Gets a list of prop options for a GDrive file in a particalar GDrive
```
vellames-turing, line 448:
```suggestion
     * Gets a list of prop options for a GDrive file in a particular GDrive
```

vellames-turing (changes requested):
I finished my review. I have just minor suggestions. The code looks great!
Another suggestion, completely optional is to change the requests in `actions` to use pipedream `axios` in order to have better error tracking.
It's not in your scope, but I found a spelling error in `create-file.js`. Looks like `keeprevisionForever` is supposed to be `keePrevisionForever`

components/google_drive/actions/find-file/find-file.js
js07: According to the [API docs](https://developers.google.com/drive/api/v3/reference/files/list#response), a successful request to the [Files: list](https://developers.google.com/drive/api/v3/reference/files/list) endpoint should return a response body containing a `files` property. If the response body is not an object, I think we'd prefer that an error is thrown, for debugging, rather than see an `undefined` return value.

components/google_drive/google_drive.app.js
js07: I added file paths to labels for files and folders listed in async options for props used in actions.

For example, if the file path is `MyDrive/folder1/folder2/myFile.csv`, the option label would be `folder1 > folder2 > myFile.csv`, using angle brackets similar to the Google Drive UI. If the file path is `MyDrive/folder1/folder2/folder3/folder4/myFile.csv`, the option label would be `... > folder4 > myFile.csv`, truncating the file path so the file name is displayed within the UI.

components/google_drive/actions/download-file/download-file.js
js07: Added `async options` to the [`mimeType` prop in **Download File**](https://github.com/PipedreamHQ/pipedream/blob/f461bfb3793a727c705007472dbc9927e0a6579c/components/google_drive/actions/download-file/download-file.mjs#L64) which returns the following:
- If the value of the `fileId` prop is the ID of a Google Workspace document, return the [valid conversion format MIME types](https://developers.google.com/drive/api/v3/ref-export-formats) for that file type
- If the value of the `fileId` prop is the ID of a non-Google Doc file, return an empty array (because this prop is not used)
- Otherwise, return all [Google Workspace document export MIME types](https://developers.google.com/drive/api/v3/ref-export-formats)

The `mimeType` prop is optional, but an error is throw in the `run` method when the file being downloaded is a [Google Workspace document](https://developers.google.com/drive/api/v3/ref-export-formats) and `mimeType` is blank.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.js
js07: Updated descriptions for the drive prop in this commit: https://github.com/PipedreamHQ/pipedream/pull/1718/commits/bc5aa50090f97d06f61e1afb757f9b666ab7696c

- Made the `drive` prop optional, with default value `"My Drive"`
- Changed the base description for `drive` prop to a more general component-independent description
- Redefined description for the `drive` prop in Google Drive & Sheets sources
- Updated previously redefined `drive` prop description in Google Drive & Sheets actions

Alternatively, we could make **Drive** optional and additionally treat the value `undefined` as  "My Drive" so structured mode is enabled by default. For example:
opt2-2

js07:
All new Google Drive actions work for me except **Create Shared Drive**,  **Get Shared Drive**, **Update Shared Drive**, and **Delete Shared Drive**, which I haven't tested with a Google Workspace account.

I tested a Google Sheets action, a Google Sheets source, and a Google Docs action affected by changes in this PR. Other affected Google Sheets and Docs components are untested.

dannyroosevelt:
I tested the ones you didn't, @js07, and they're good to go (all the shared drive actions for Google Workspace accts)

- [x] Create Shared Drive
- [x] Get Shared Drive
- [x] Update Shared Drive
- [x] Delete Shared Drive

PR #1747, opened 2021-09-29
Add missing / in the link

SinghCoder:
Due to missing /, the url was coming relative to current URL, because of which instead of pointing to [expected link](https://pipedream.com/docs/components/api/), it pointed to [this link](https://pipedream.com/docs/api/rest/components/api/) under [this section](https://pipedream.com/docs/api/rest/#parameters)


dylburger:
Another good catch! Thanks!

PR #1749, opened 2021-09-29
[ACTIONS] Telegram Bot new actions

### LINKED ISSUE #1443: [ACTION] Telegram Bot new actions
Telegram Bot

https://core.telegram.org/bots/api

**Send a Text Message or Reply** | Sends a text message or a reply to your Telegram Desktop application
**Edit a Text Message** | Edits text of game messages
**Forward a Message** | Forwards messages of any kind
**Delete a Message** - Deletes a message
**Pin a Message** | Pins a message
**Unpin a Message** | Unpins a message
**Send an Audio File** | Sends an audio file to your Telegram Desktop application
**Send a Document/Image** | Sends a document or an image to your Telegram Desktop application
**Send an Album (Media Group)** | Sends a group of photos or videos as an album
**Send a Photo** | Sends a photo to your Telegram Desktop application
**Send a Sticker** | Sends a .webp sticker to you Telegram Desktop application
**Send a Video** | Sends a video file to your Telegram Desktop application
**Send a Video Note** | As of v.4.0, Telegram clients support rounded square mp4 videos of up to 1 minute long. Use this method to send video messages
**Send a Voice Message** | Sends a voice message
**Edit a Media Message** | Edits photo or video messages
**Send Media by URL or ID** | Sends a file (document, photo, video, audio, ...) by HTTP URL or by ID that exists on the Telegram servers
**List Updates** | Retrieves a list of updates from the Telegram server
**List Chats** | List available Telegram chats
**List Administrators In Chat** | Use this module to get a list of administrators in a chat
**Get the Number of Members in a Chat** | Use this module to get the number of members in a chat
**Kick a Chat Member** | Use this method to kick a user from a group, a supergroup or channel
**Promote a Chat Member** | Use this method to promote or demote a user in a supergroup or a channel
**Restrict a Chat Member** | Use this method to restrict a user in a supergroup

js07:
Adds Telegram Bot actions:

- Send a Text Message or Reply
- Edit a Text Message
- Forward a Message
- Delete a Message
- Pin a Message
- Unpin a Message
- Send an Audio File
- Send a Document/Image
- Send an Album (Media Group)
- Send a Photo
- Send a Sticker
- Send a Video
- Send a Video Note
- Send a Voice Message
- Edit a Media Message
- Send Media by URL or ID
- List Updates
- List Chats
- List Administrators In Chat
- Get the Number of Members in a Chat
- Kick a Chat Member
- Promote a Chat Member
- Restrict a Chat Member

Resolves #1443


dylburger:

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -1,4 +1,20 @@
+// If unset, set `process.env.NTBA_FIX_319` to true to enable cancellation of
```
jcortes, line 1: Out of curiosity are we also migrating this to ES6 modules?

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -1,4 +1,20 @@
...
+// Promises. See: https://github.com/yagop/node-telegram-bot-api/issues/319 for
+// more information.
+process.env.NTBA_FIX_319 = (process.env.NTBA_FIX_319 !== undefined)
+  ? process.env.NTBA_FIX_319
+  : true;
```
jcortes, line 6:
```suggestion
process.env.NTBA_FIX_319 = process.env.NTBA_FIX_319 ?? true;
```

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -1,4 +1,20 @@
...
+const TelegramBot = require("node-telegram-bot-api");
 const axios = require("axios");
```
jcortes, line 8: can you use [@pipedream/platform](https://pipedream.com/docs/pipedream-axios/#pipedream-platform-axios) instead.

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -1,4 +1,20 @@
...
+const {
+  TELEGRAM_BOT_API_UI_MEDIA_AUDIO,
+  TELEGRAM_BOT_API_UI_MEDIA_DOCUMENT,
+  TELEGRAM_BOT_API_UI_MEDIA_PHOTO,
+  TELEGRAM_BOT_API_UI_MEDIA_VIDEO,
+  TELEGRAM_BOT_API_UI_MEDIA_VIDEO_NOTE,
+  TELEGRAM_BOT_API_UI_MEDIA_STICKER,
+  TELEGRAM_BOT_API_UI_MEDIA_VOICE,
+} = require("./constants");
 const updateTypes = [
```
jcortes, line 18: My suggesetion here is to set this `const` as a constant variable in `constants` file

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -58,6 +74,161 @@ module.exports = {
         "Only emit events for the selected update types.",
       options: updateTypes,
     },
+    chatId: {
+      type: "string",
+      label: "Chat ID",
+      description: "Enter the unique identifier for the target chat or username of the target channel (in the format `@channelusername` or `@supergroupusername`).",
```
jcortes, line 80: I had a hard time trying to find a Chat ID in this format and I ended up setting a normal id like `1035597319` to test an action component and it worked. So I guess it would be nice to have more context of how the user con get this data or at least to have a link where the user can get a real Chat ID

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -58,6 +74,161 @@ module.exports = {
...
+    text: {
+      type: "string",
+      label: "Text",
+      description: "Enter or map the message text to send.",
+      optional: true,
+    },
+    parse_mode: {
+      type: "string",
+      label: "Parse Mode",
+      description: "Select [MarkdownV2-style](https://core.telegram.org/bots/api#markdownv2-style), [HTML-style](https://core.telegram.org/bots/api#html-style), or [Markdown-style](https://core.telegram.org/bots/api#markdown-style) of the text if you want Telegram apps to show bold, italic, fixed-width text or inline URLs in your bot's message.",
```
jcortes, line 91: I guess we can use [toSingleLineString](https://github.com/PipedreamHQ/pipedream/blob/master/components/aws/sources/common/utils.mjs#L34) here as a utility function for log text in `description` props

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -58,6 +74,161 @@ module.exports = {
...
+      options: [
+        "MarkdownV2",
+        "HTML",
+        "Markdown",
```
jcortes, line 95: Set these as constants too

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -58,6 +74,161 @@ module.exports = {
...
+    reply_to_message_id: {
+      type: "string",
+      label: "Original Message ID",
+      description: "Enter the ID of the original message.",
+      optional: true,
+    },
+    reply_markup: {
+      type: "string",
+      label: "Reply Markup",
+      description: "Enter additional interface options that are a JSON-serialized object including an inline keyboard, a custom reply keyboard, instructions to remove the reply keyboard or instructions to force a reply from the user, e.g. `{\"inline_keyboard\":[[{\"text\":\"Some button text 2\",\"url\":\"https://botpress.org\"}]]}` or `{\"keyboard\":[[\"Yes\",\"No\"],[\"Maybe\"]]}`. Note: keyboard cannot be used with channels.",
```
jcortes, line 120: Seems like there is a lot of options here, can you add like a link so the user can have more idea about these properties?

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -58,6 +74,161 @@ module.exports = {
...
+    height: {
+      type: "integer",
+      label: "Height",
+      description: "Enter the video height.",
+      optional: true,
+    },
+    length: {
+      type: "integer",
+      label: "Length",
+      description: "Enter the video width and height.",
```
jcortes, line 190: I didn't get this description, is it the duration of a video?

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -58,6 +74,161 @@ module.exports = {
...
+      optional: true,
+    },
+    type: {
+      type: "string",
+      label: "Media Type",
+      description: "Select the media type.",
+      options: [
+        "Photo",
+        "Video",
```
jcortes, line 199: The defaults should set from contants
```
TELEGRAM_BOT_API_UI_MEDIA_PHOTO
TELEGRAM_BOT_API_UI_MEDIA_VIDEO
```

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -89,5 +271,386 @@ module.exports = {
...
+     * username of the target channel (in the format `@channelusername`)
+     * @param {String} text - Text of the message to be sent, 1-4096 characters
+     * after entities parsing
+     * @param {Object} [opts] - An object containing additional configuration
+     * options for this method, as defined in [the API
+     * docs](https://core.telegram.org/bots/api#sendmessage)
+     * @returns The sent Message
+     */
+    async sendMessage(chatId, text, opts) {
+      return await this.sdk().sendMessage(chatId, text, opts);
```
jcortes, line 287: Seems like it's unecessary to have these `await` here. Only places I would call `await` is because either I would put it inside a `try/catch` or because there is a need to resolve a value before returning it
```suggestion
      return this.sdk().sendMessage(chatId, text, opts);
```

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -89,5 +271,386 @@ module.exports = {
...
+    async editMessageText(text, opts) {
+      const {
+        chatId,
+        messageId,
+        inlineMessageId,
+        ...extraOpts
+      } = opts;
+      if (!(chatId && messageId) && !inlineMessageId) {
+        throw new Error("chatId, messageId, or inlineMessageId is required");
+      }
```
jcortes, line 319: Maybe you should change the error message here since it seems like `chatId && messageId` and `inlineMessageId` are mutually exclusive
Also to make it more readable I would set a variable like `hasChatIdAndMessageId`
```suggestion
      const hasChatIdAndMessageId = chatId && messageId;
      if (!hasChatIdAndMessageId && !inlineMessageId) {
         throw new Error("chatId, messageId, or inlineMessageId is required");
      }
```

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -89,5 +271,386 @@ module.exports = {
...
+     * @param {Object} [opts] - An object containing additional configuration
+     * options for this method
+     * @param {Number|String} [opts.filename] - The name of the file to send
+     * @param {Number|String} [opts.contentType] - The MIME type of the file to
+     * send
+     * @param {...*} [opts.extraOpts] - Additional Telegram query options to be
+     * fed to the Telegram Bot API call, as defined in [the API
+     * docs](https://core.telegram.org/bots/api)
+     * @returns {Promise<TelegramBot.Message>} The sent message
+    */
```
jcortes, line 403: Not sure what is the reason you put two versions of the jsdoc?

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -89,5 +271,386 @@ module.exports = {
...
+        contentType,
+        ...extraOpts
+      } = opts;
+      return await sendFn(chatId, media, extraOpts, {
+        filename,
+        contentType,
+      });
+    },
+    /**
+     * @typedef {import("./constants.js").UIMediaType} UIMediaType
```
jcortes, line 435: Shouldn't this be inside the jsdoc below this one?

components/telegram_bot_api/telegram_bot_api.app.js
```diff
@@ -89,5 +271,386 @@ module.exports = {
...
+        [TELEGRAM_BOT_API_UI_MEDIA_VIDEO_NOTE]: sdk.sendVideoNote,
+        [TELEGRAM_BOT_API_UI_MEDIA_VOICE]: sdk.sendVoice,
+        [TELEGRAM_BOT_API_UI_MEDIA_STICKER]: sdk.sendSticker,
+      };
+      const sendFn = typeToSendFn[type].bind(sdk);
+      return this.sendMedia(sendFn, chatId, media, opts);
+    },
+    /**
+     * @type {SendMediaFn}
+     */
```
jcortes, line 473: Is this jsdoc really needed? same goes for the rest of functions like this one

jcortes (changes requested):
This is my first round with this PR since it's a big one. But it's looking good so far

components/telegram_bot_api/actions/delete-message/delete-message.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable camelcase */
+const telegram_bot_api = require("../../telegram_bot_api.app.js");
```
jcortes, line 2: Shouldn't all variable names be in lower capital case? like `telegramBotApi`?

components/telegram_bot_api/actions/delete-message/delete-message.js
```diff
@@ -0,0 +1,28 @@
...
+
+module.exports = {
+  key: "telegram_bot_api-delete-message",
+  name: "Delete a Message",
+  description: "Deletes a message",
```
jcortes, line 7: Can you add the api endpoint in the `description` prop. Same goes to the rest of action components

components/telegram_bot_api/actions/edit-media-message/edit-media-message.js
```diff
@@ -0,0 +1,87 @@
...
+    },
+    reply_markup: {
+      propDefinition: [
+        telegram_bot_api,
+        "reply_markup",
+      ],
+    },
+  },
+  async run() {
+    return await this.telegram_bot_api.editMessageMedia({
```
jcortes, line 77: `await` is unecessary here. Same goes to the rest of the action components

components/telegram_bot_api/actions/list-chats/list-chats.js
```diff
@@ -0,0 +1,72 @@
...
+    if (this.autoPaging && updates.length > 0) {
+      const lastUpdateId = updates[updates.length - 1].update_id;
+      // "Confirm" updates by calling API to get updates using an offset of the
+      // last `update_id + 1` See [documentation for getUpdates in the Telegram
+      // Bot API reference](https://core.telegram.org/bots/api#getupdates)
+      await this.telegram_bot_api.getUpdates({
+        offset: lastUpdateId + 1,
+        limit: 1,
+      });
+    }
```
jcortes, line 69: I didn't quite get how does this work. I mean you are calling only twice the `getUpdates` function in case `autoPaging` is true and there is still records in the list, but you are just making the request without reassgining the `updates` array? also what happens if there is more than 2 pages? and should you set a default limit for the whole pagination? like 30 records max? same questions go for `list-updates`

components/telegram_bot_api/actions/send-album/send-album.js
```diff
@@ -0,0 +1,44 @@
...
+    chatId: {
+      propDefinition: [
+        telegram_bot_api,
+        "chatId",
+      ],
+    },
+    media: {
+      type: "any",
+      label: "Media",
+      description: "A JSON-serialized array describing photos and videos to be sent, must include 2–10 items (e.g., `[{\"type\":\"photo\",\"media\":\"https://example.com/myImage.jpeg\"},{\"type\":\"video\",\"media\":\"/tmp/myVideo.mp4\"}]`)",
```
jcortes, line 21: In addition to the description I think is also necessary the link to the docs

jcortes (commented):
I've added a few more comments/questions thanks!

components/telegram_bot_api/telegram_bot_api.app.js
js07: I put two version of the jsdoc because [this version](https://github.com/PipedreamHQ/pipedream/blob/3f114c56dad3346420f7888ae5357e9b1f45ab85/components/telegram_bot_api/telegram_bot_api.app.js#L408) has an additional `sendFn` param. Ideally docs for the params wouldn't be repeated, but I'm not sure how to avoid duplicating docs for functions with mostly the same params. Maybe these methods could be refactored so they have the same arguments.

components/telegram_bot_api/telegram_bot_api.app.js
js07: I think this jsdoc could be separate or part of the one below, but it might be clearer/simpler as part of the one below. I'll combine the jsdocs.

components/telegram_bot_api/telegram_bot_api.app.js
js07: I added this jsdoc to document the types and descriptions of the parameters and return values. Do you think the code be simpler or more easily understood without them?

components/telegram_bot_api/actions/list-chats/list-chats.js
js07: According to the [Telegram Bot API docs](https://core.telegram.org/bots/api#getupdates):
> By default, updates starting with the earliest unconfirmed update are returned. An update is considered confirmed as soon as getUpdates is called with an offset higher than its update_id.

When `getUpdates` is called with an `offset`, following calls to `getUpdates` _without_ an `offset` will return updates with `update_id` >= (previously called) max `offset`.

- When `autoPaging` is `true`, this action "confirms" the updates by calling `getUpdates` with `offset = lastUpdateId + 1`.
- Each following execution of this action returns only updates that haven't yet been processed (`update_id >= lastOffset`)--the Telegram server handles pagination.

Do you think we should optionally paginate through all of the new updates and not just `limit` updates?

components/telegram_bot_api/telegram_bot_api.app.js
jcortes: I still don't get it maybe I'm a little slow here. But that means I can call `sendMedia` without `sendFn` function? Even though you are only calling it passing `sendFn` right?

components/telegram_bot_api/telegram_bot_api.app.js
js07: I converted the JS files in `telegram_bot_api` from CommonJS to ESModules.

components/telegram_bot_api/telegram_bot_api.app.js
js07: I created a file with this `const` called `update-types.mjs`, similar to [twist/event-types.js](https://github.com/PipedreamHQ/pipedream/blob/master/components/twist/event-types.js) or [todoist/resource-types.js](https://github.com/PipedreamHQ/pipedream/blob/master/components/todoist/resource-types.js). Is that okay, or would you suggest putting `updateTypes` in `constants.mjs`?

components/telegram_bot_api/telegram_bot_api.app.js
js07: I updated the `description` of `chatId`:
https://github.com/PipedreamHQ/pipedream/blob/c80b5afcf9851debaa7cab8efab46a3619aeaa1f/components/telegram_bot_api/telegram_bot_api.app.mjs#L32-L35

components/telegram_bot_api/telegram_bot_api.app.js
js07: Added a link to the docs in the `description` of `reply_markup`:

https://github.com/PipedreamHQ/pipedream/blob/c80b5afcf9851debaa7cab8efab46a3619aeaa1f/components/telegram_bot_api/telegram_bot_api.app.mjs#L80-L87

components/telegram_bot_api/telegram_bot_api.app.js
js07: Updated the `description` of `length`:

https://github.com/PipedreamHQ/pipedream/blob/c80b5afcf9851debaa7cab8efab46a3619aeaa1f/components/telegram_bot_api/telegram_bot_api.app.mjs#L168

components/telegram_bot_api/telegram_bot_api.app.js
jcortes: This is just an opinon. What I would do is to just call the needed function for example right from the action component in this case would be `send-audio-file`
```
this.sendMediaByType(TELEGRAM_BOT_API_UI_MEDIA_AUDIO, chatId, audio, opts);
```
in that way you can remove all those funcion definitions from with only this in the jsdoc
```
    /**
     * @type {SendMediaFn}
     */
```

components/telegram_bot_api/actions/send-album/send-album.js
js07: Added a link to the docs:

https://github.com/PipedreamHQ/pipedream/blob/c80b5afcf9851debaa7cab8efab46a3619aeaa1f/components/telegram_bot_api/actions/send-album/send-album.mjs#L22-L27

components/telegram_bot_api/actions/delete-message/delete-message.js
js07: Some apps like [Microsoft One Drive](https://github.com/PipedreamHQ/pipedream/blob/master/components/microsoft_onedrive/sources/common/base.js) and [Zoho CRM](https://github.com/PipedreamHQ/pipedream/blob/master/components/zoho_crm/sources/common/timer-based/base.js) currently use snake_case rather than camelCase for their app module names, but others like [Google Drive](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_drive/sources/common-webhook.js) and [Google Sheets](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_sheets/sources/common.mjs) use camelCase. It seems like camel case is preferred given the lint rules.

I changed `telegram_bot_api` to `telegramBotApi` and removed `/* eslint-disable camelcase */` from the action files.

components/telegram_bot_api/actions/list-chats/list-chats.js
jcortes: Well that was discussed some days ago on slack, but I will say I'll leave it to you. For me it looks good!

components/telegram_bot_api/telegram_bot_api.app.js
jcortes: I think a practical way to solve this is by doing what I suggested below with my opinon. But again that's just an opinion.

js07: I removed one of the versions of the jsdoc.

components/telegram_bot_api/telegram_bot_api.app.js
js07: I refactored the [`send<Media>` functions](https://github.com/PipedreamHQ/pipedream/blob/0d608e3513972888420b3f73fcfdcb6fdb8ab067/components/telegram_bot_api/telegram_bot_api.app.mjs#L413) to call `this.sendMedia` directly rather than through `this.sendMediaByType`, and removed the JSDocs. I was hoping to keep separate functions for `sendPhoto`, `sendAudio`, etc., like the [Telegram Bot Node.js module](https://github.com/yagop/node-telegram-bot-api/blob/5cc6629ca02c5160424be4615b281866fb56c027/doc/api.md#TelegramBot+sendPhoto) and [Telegram Bot API](https://core.telegram.org/bots/api#sendphoto) do. But we could remove the separate functions in favor of a `this.sendMediaByType` method if that's preferable.

jcortes: It looks great to me as they are. Thanks for making this change!

jcortes (approved):
It looks good to me! ready to go!

js07:
Thanks for reviewing @jcortes!

js07 (commented):
Tested all of the actions and affected sources and they work for me*.

*~~Not fully tested: props for **Promote a Chat Member** and **Restrict a Chat Member**, receiving `channel_message`/`edited_channel_message` events in **Channel Updates (Instant)** source~~
Edit: Tested props for **Promote a Chat Member** and **Restrict a Chat Member**.

PR #1570, opened 2021-08-10
New AWS S3 Event Sources

jverce:
* New S3 File
* Restored S3 File
* Deleted S3 File
* New S3 Event

Addresses #1194

Issue #1194, opened 2021-04-26
[FEATURE] Support S3 events


cfjedimaster:
**Is your feature request related to a problem? Please describe.**
Like Dropbox/OneDrive, let me kick off events when files are added/updated/deleted from a S3 bucket.

**Describe the solution you'd like**
Pretty much the above. :)

**Do you have a workaround?**
Nope.

Issue #1458, opened 2021-07-27
[ACTION] AWS new actions


michelle0927:
**Is there a specific app this action is for?**
AWS

**Please provide a link to the relevant API docs for the specific service / operation.**
https://docs.aws.amazon.com/sdk-for-javascript/

**Run Code** | Write javascript code and execute it in a Lambda function that we provide the source for
**Invoke Function** | Invokes a prebuilt function asynchronously providing only data
**Create Bucket** | Creates a new bucket
**Get File** | Gets a file from a bucket
**Upload File** | Uploads a file to a bucket
**List Files** | List all files in the specified path
**List Folders** | List all folders in the specified path

dylburger:
@andrewjschuang Can you tackle #2150 along with these actions?

andrewjschuang:
> @andrewjschuang Can you tackle #2150 along with these actions?
Yep, np

DilanAthukorala:
@andrewjschuang 
Observing error when trying to deploy app into Pipedream
![image](https://user-images.githubusercontent.com/102980146/163761794-9780c837-aae3-4df1-ae96-47b43a3ef009.png)

andrewjschuang:
> @andrewjschuang Observing error when trying to deploy app into Pipedream

Hey Dilan, this happens because all the imports summed up goes beyond Pipedream's size limit. Hadn't been able to think of a solution until now, but it will take some heavy refactoring of the directory files/structure

andrewjschuang:
@DilanAthukorala fixed the issues!

vunguyenhung:
Hi @andrewjschuang and everyone, all test cases has passed! Ready for release!

Please note that the test exclude the expression and curly brace issue on JSON string input described [here](https://pipedream-users.slack.com/archives/C01E5KCTR16/p1652115042492479). The issue will be fixed in another ticket!

Refer to the test report here for more information
[AWS_Test_Report_9_May_0028.pdf](https://github.com/PipedreamHQ/pipedream/files/8653887/AWS_Test_Report_9_May_0028.pdf)

PR #1797, opened 2021-10-15
AWS S3 event sources

compwright:
Replaces #1570 (I can't push commits to `@jverce`'s branch)

* New S3 File
* Restored S3 File
* Deleted S3 File
* New S3 Event

Addresses #1194


andrewjschuang:
Issue #1458 refactors AWS app actions and sources, so after it is approved this issue can also be refactored and submitted to review

PR #1875, opened 2021-11-03
[SOURCES/ACTIONS] Amara new sources/actions

jcortes:
# Actions
- [x] Add a video https://apidocs.amara.org/#add-a-video
- [x] List videos https://apidocs.amara.org/#list-videos (*I've opened a ticket [here](https://support.amara.org/support/tickets/1054082)*)
- [x] Get video details https://apidocs.amara.org/#view-video-details
- [x] Delete a video https://apidocs.amara.org/#delete-a-video (*In order to delete a video, it must be part of a team that you're an admin of.*)
- [x] List subtitle languages for a video  https://apidocs.amara.org/#list-subtitle-languages-for-a-video
- [x] Get details on a single subtitle language https://apidocs.amara.org/#get-details-on-a-single-subtitle-language
- [x] Create a subtitle language https://apidocs.amara.org/#create-a-subtitle-language
- [x] Update a subtitle language https://apidocs.amara.org/#update-a-subtitle-language
- [x] Fetch subtitles data https://apidocs.amara.org/#fetch-subtitles-data
- [x] Add new subtitles https://apidocs.amara.org/#add-new-subtitles (*I've opened a ticket [here](https://support.amara.org/support/tickets/1054099)*)
- [x] Delete subtitles https://apidocs.amara.org/#delete-subtitles (*It’s only allowed if the video is part of a team and the API user is an admin for that team.*)
- [x] Perform action https://apidocs.amara.org/#perform-actions
- [x] Get details on a team https://apidocs.amara.org/#get-details-on-a-team
- [x] List teams https://apidocs.amara.org/#list-teams

# Sources
https://apidocs.amara.org/#video-notifications

- [x] Subtitles published
 - (*I couldn't fully test this source because it seems like the only way to publish subtitles is through public videos, and it's kind of weird because the notifications endpoint requires that you need to pass the `team-slug` as the path parameter. So I still wonder what should be the workflow when there is a `team`  setup for private videos*)
- [x] Subtitles unpublished
 - (*I had the same problem as `Subtitles published` source*)
- [x] ~Video added~
- [x] ~Video removed~
- [x] Video made primary
 - *I couldn't test this because I didn't get the notification even when making the changes that should trigger the event*
- [x] Video moved project
 - *I couldn't test this because I didn't get the notification even when making the changes that should trigger the event*

Now the following sources have been added instead using this endpoint:
https://apidocs.amara.org/#team-activity

- [x] Video added
- [x] Video deleted


dylburger:

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,462 @@
...
+        return versions.map(({ version_number: versionNumber }) => ({
+          label: `Version ${versionNumber}`,
+          value: versionNumber,
+        }));
+      },
+    },
+    subtitles: {
+      type: "string",
+      label: "Subtitles",
+      description: "The subtitles to submit, as a string. The format depends on the `sub_format` param.",
```
js07, line 205:
```suggestion
      description: "The subtitles to submit, as a string, in **Subtitles format**",
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,462 @@
...
+      optional: true,
+    },
+    subtitlesUrl: {
+      type: "string",
+      label: "Subtitles URL",
+      description: "Alternatively, subtitles can be given as a text file URL. The format depends on the `sub_format` param.",
```
js07, line 211:
```suggestion
      description: "Alternatively, subtitles can be given as a text file URL. The format depends on **Subtitles format**.",
```

js07 (commented):
Looks good! I made some comments about minor issues and suggested refactoring parts of the code.

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,543 @@
...
+        }
+
+        const {
+          meta,
+          objects: videos,
+        } = await this.listVideos({
+          url,
+          params: {
+            limit: 20,
+            team: team ?? null,
```
js07, line 84: Could we change this to `team: team || null` so we don't have to use [`utils.emptyStrToUndefined(team)`](https://github.com/PipedreamHQ/pipedream/blob/0ea1c26a13cd0183a9bfd8d731075de5d5293678/components/amara/actions/update-subtitle-language/update-subtitle-language.mjs#L23) in each action's `videoId` prop definition?

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,543 @@
...
+          return [];
+        }
+
+        const {
+          meta,
+          objects: languages,
+        } =
+          await this.getVideoSubtitleLanguages({
+            url,
+            videoId,
```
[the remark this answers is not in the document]
jcortes: I would say just to return `[]` in case of error because for all cases the language need to exist in order to call the request. (`delete-subtitles, fetch-raw-subtitles, fetch-subtitles-data, get-subtitle-language-details, perform-action, update-subtitle-language`)

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,543 @@
...
+          language,
+        });
+
+        return actions.map(({
+          label, action,
+        }) => ({
+          label,
+          value: action,
+        }));
+      },
```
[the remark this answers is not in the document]
jcortes: Yeah I think returning empty array will do it

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,543 @@
...
+          await this.getSubtitleLanguage({
+            videoId,
+            language,
+          });
+
+        return versions.map(({ version_number: versionNumber }) => ({
+          label: `Version ${versionNumber}`,
+          value: versionNumber,
+        }));
+      },
```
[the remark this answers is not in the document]
jcortes: Yeah I think returning empty array will do it

components/amara/actions/fetch-subtitles-data/fetch-subtitles-data.mjs
```diff
@@ -0,0 +1,77 @@
...
+      versionNumber,
+      subFormat,
+    } = this;
+
+    const response = await this.amara.getSubtitles({
+      $,
+      videoId,
+      language,
+      params: {
+        version_number: versionNumber,
```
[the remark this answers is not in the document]
jcortes: Yeah it should give you the latest version. I've fixed it thanks!

components/amara/actions/list-videos/list-videos.mjs
```diff
@@ -0,0 +1,124 @@
...
+    },
+    team: {
+      description: "Filter by team. Passing in `null` will return only videos that are in the public area.",
+      propDefinition: [
+        amara,
+        "team",
+      ],
+    },
+    primaryAudioLanguageCode: {
+      description: "Filter team videos by video language (only works when the team parameter is also present).",
```
js07, line 29:
```suggestion
      description: "Filter team videos by video language (only works when **Team** is given a non-null value).",
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,558 @@
...
+    team:	{
+      type: "string",
+      label: "Team",
+      description: "Filter by team (e.g. `amplify` in `https://amara.org/en/teams/amplify/`). Leave blank for videos that are in the public area.",
+      optional: true,
+    },
+    videoUrl: {
+      type: "string",
+      label: "Video URL",
+      description: "The url for the video. Any url that Amara accepts will work here. You can send the URL for a file (e.g. `http:///www.example.com/my-video.ogv`), or a link to one of the accepted providers (YouTube, Vimeo).",
```
js07, line 17:
```suggestion
      description: "The URL for the video. Any URL that Amara accepts will work here. You can send the URL for a file (e.g. `http:///www.example.com/my-video.ogv`), or a link to one of the accepted providers (YouTube, Vimeo).",
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,558 @@
...
+          context: {
+            url: meta.next,
+          },
+        };
+      },
+    },
+    language: {
+      type: "string",
+      label: "Language",
+      description: "Language code for the language of the subtitles",
```
js07, line 105:
```suggestion
      description: "Language code for the language of the subtitles ([ISO 639-1 two-letter code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes))",
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,558 @@
...
+    },
+    videoUrl: {
+      type: "string",
+      label: "Video URL",
+      description: "The url for the video. Any url that Amara accepts will work here. You can send the URL for a file (e.g. `http:///www.example.com/my-video.ogv`), or a link to one of the accepted providers (YouTube, Vimeo).",
+    },
+    primaryAudioLanguageCode: {
+      type: "string",
+      label: "Primary audio language code",
+      description: "Language code for the main language spoken in the video.",
```
js07, line 22:
```suggestion
      description: "Language code for the main language spoken in the video ([ISO 639-1 two-letter code](https://en.wikipedia.org/wiki/List_of_ISO_639-1_codes))",
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,558 @@
...
+          .map((key) => ({
+            label: key,
+            value: constants.FORMAT_TYPES[key],
+          }));
+      },
+    },
+    versionNumber: {
+      type: "integer",
+      label: "Version number",
+      description: "Version number of the subtitles to fetch. Versions are listed in the VideoLanguageResouce request. If none is specified, the latest public version will be returned. If you want the latest private version (and have access to it) use “last”.",
```
js07, line 216:
```suggestion
      description: "Version number of the subtitles to fetch. Versions are listed in the [VideoLanguageResouce](https://apidocs.amara.org/#list-subtitle-languages-for-a-video) request. If none is specified, the latest public version will be returned. If you want the latest private version (and have access to it) use “last”.",
```

components/amara/sources/common.mjs
```diff
@@ -0,0 +1,102 @@
...
+      this.db.set(constants.LAST_URL, lastUrl);
+    },
+    getLastUrl() {
+      return this.db.get(constants.LAST_URL);
+    },
+    getResourceFn() {
+      throw new Error("getResourceFn is not implemented");
+    },
+    getResourceFnArgs() {
+      throw new Error("getResourceFn is not implemented");
```
js07, line 36:
```suggestion
      throw new Error("getResourceFnArgs is not implemented");
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,558 @@
...
+
+        nextResources = nextResponse.objects;
+        lastUrl = nextResponse.meta.next;
+        resources = resources.concat(nextResources);
+
+        if (callback) {
+          nextResources.forEach(callback);
+        }
+
+      } while (nextResources.length && resources.length < max);
```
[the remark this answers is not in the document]
jcortes: Yeah I think this will do it too, thanks!
```javascript
} while (lastUrl && resources.length < max);
```

components/amara/sources/common.mjs
```diff
@@ -0,0 +1,102 @@
...
+      resourceFn: this.getResourceFn(),
+      resourceFnArgs: this.getResourceFnArgs({
+        $,
+        url,
+      }),
+      limit: 40,
+      callback: this.processEvent,
+    });
+
+    this.setLastUrl(lastUrl);
```
js07, line 100: If more than the [default max](https://github.com/PipedreamHQ/pipedream/blob/7463b1b9957cd0edf5d75e9951b79a300a5b05d1/components/amara/constants.mjs#L7) of 60 event resources have been created in Amara, `lastUrl` will have an offset of 60 at `this.setLastUrl(lastUrl)`. If I'm not mistaken, on the next `run` (and following runs), the source may fetch and `$emit` duplicate events that have not been emitted in the last [100 cached events](https://pipedream.com/docs/components/api/#dedupe-strategies).

I think we could fix this issue by removing `url` from the `resourceFnArgs`, but events would be missed if there are more than `DEFAULT_MAX_ITEMS` new events between source runs.

Preferably, we could use an approach similar to in [Textlocal sources](https://github.com/PipedreamHQ/pipedream/blob/3f6545d06638d674c6b15b06a036abc48de67a8d/components/textlocal/sources/new-sent-api-message/new-sent-api-message.js#L40-L41) where we paginate items [until the last](https://github.com/PipedreamHQ/pipedream/blob/3f6545d06638d674c6b15b06a036abc48de67a8d/components/textlocal/textlocal.app.js#L93) emitted item (ID or timestamp) in the previous run.
What do you think?

js07 (commented):
Thanks for making those changes. I tested the actions and sources and they worked for me, aside from an issue I mentioned in a comment below.

components/amara/sources/common.mjs
```diff
@@ -0,0 +1,102 @@
...
+    },
+    generateMeta(resource) {
+      const [
+        eventType,
+        eventId,
+      ] = this.getEventTypeAndId(resource);
+      return {
+        id: JSON.stringify(resource),
+        summary: `${eventId} ${eventType}`,
+        ts: Date.parse(resource.data ?? resource.timestamp ?? Date.now()),
```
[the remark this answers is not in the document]
jcortes: Oh my bad, thanks!

components/amara/sources/common.mjs
jcortes: That's right but I think the problem is because I'm setting `null` into `lastUrl` at the end of the `run` so I can fix this by just doing
```javascript
        nextResources = nextResponse.objects;
        resources = resources.concat(nextResources);

        if (nextResponse.meta.next) {
          lastUrl = nextResponse.meta.next;
        }

        if (callback) {
          nextResources.forEach(callback);
        }

      } while (nextResponse?.meta.next && resources.length < max);
```

js07: It looks like [Notifications](https://apidocs.amara.org/#list-team-notifications) and [Activity](https://apidocs.amara.org/#team-activity) are in reverse creation order (descending `date`/`timestamp`), so I think we would be fetching older resources on each request if we remove setting `lastUrl` to `null`.

timestamp


The [Team activity](https://apidocs.amara.org/#team-activity) endpoint has an `after` query parameter, which we could use to include only activity after the most recent activity event emitted. But [Notifications](https://apidocs.amara.org/#list-team-notifications) doesn't seem to have an `after` or `sort` query parameter.

jcortes: Hi @js07  It seems like is not even sorting in descending order but ascending order contrary to what the docs say. So I did something similar to what you pointed out in the previous comment but without taking into account the `id` but the whole object as the identifier since there is no unique property I can use here. I'm going to wait for an actual account that has an Enterprise plan so I can test with real data the rest of notifications events. @dylburger is going to be helping me out on that.

joelmiles: I have an Enterprise account so if that's useful for helping move anything along quicker, I'm happy to assist where I'm able.  There's probably no substitute for seeing the real data in your own environment, but if there's utility in me helping by using our account, let me know what you'd like me to do.

components/amara/sources/common.mjs
```diff
@@ -0,0 +1,106 @@
...
+      type: "$.interface.timer",
+      label: "Polling schedule",
+      description: "How often to poll the Amara API",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+    team: {
+      description: "The team slug of the team to watch",
+      optional: true,
```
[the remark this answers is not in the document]
jcortes: Yes, the idea here is that if you set `team` as either `null` or empty it will return resources in the public area. I've changed the description to be clearer.

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,619 @@
...
+        }
+
+        if (callback) {
+          nextResources.forEach(callback);
+        }
+
+        // if resourceNotFound is true, then keeps looping until found
+        // if nextResponse?.meta.next is not null, then there are more resources to fetch
+        // if resources.length < max is the max number of resources to be fetched
+      } while (resourceNotFound && nextResponse?.meta.next && resources.length < max);
```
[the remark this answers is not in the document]
jcortes: Ok for your first paragraph, what will happen in the second run is, if there is more events created I'm actually taking into acount those since the sorting is in ascending order, so latest items will show up until `lastResourceStr` is found.

For your second paragraph, yeah I think is a good idea to add the `deploy` hook, so in that way I'm not calling all those resources again when updating the component.

For the third paragraph I'm not used to work with generators, I've always thought you can do everything with closures and functions and generators are just syntactic sugar in javascript. I'm aware there is some advantages in terms of memory and time performance in the browser context in this case. Either way is just a personal preference so I'm going to be switching to `Generators` in favor of the guidlines.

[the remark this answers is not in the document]
js07: I think using a callback function to process events, like the sources do currently, should work about the same as generators. But async generators allow us to express asynchronous code in a sequential/synchronous way that is easier to follow and understand... in the way promises or `async`/`await` improve readability and maintainability when replacing callbacks.

components/amara/sources/common.mjs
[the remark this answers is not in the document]
js07: When I leave `team` blank in a source's configuration, I get an error when the source runs. E.g. for **Video Deleted**, I get the error:
```

{
  "status": 404,
  "statusText": "Not Found",
  "data": {
    "error": "not-found",
    "path": "/api/teams//activity/"
  }
}
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,625 @@
...
+        for (const resource of nextResources) {
+          yield resource;
+        }
+
+        // if resourceNotFound is true, then keeps looping until found
+        // if nextResponse?.meta.next is not null, then there are more resources to fetch
+        const resourceNotFoundAndNexUrl = resourceNotFound && nextResponse?.meta.next;
+
+        // if resources.length < max is the max number of resources to be fetched
+        const keepFetching = resourceNotFoundAndNexUrl && resources.length < max;
```
js07, line 617: I think changing this method to a generator function allows us to simplify it by removing the call to `this.processResources()`.

```suggestion

      while (true) {
        const nextResponse = await resourceFn({
          ...resourceFnArgs,
          url: lastUrl,
          params: {
            ...resourceFnArgs.params,
            limit,
            offset,
          },
        });

        if (!nextResponse) {
          throw new Error("No response from the Amara API.");
        }

        let nextResources = nextResponse.objects;

        if (nextResponse.meta.next) {
          lastUrl = nextResponse.meta.next;
        }

        for (const resource of nextResources) {
          if (lastResourceStr && JSON.stringify(resource) === lastResourceStr) {
            return;
          }
          yield resource;
        }

        const keepFetching = nextResponse?.meta.next && resources.length < max;
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,625 @@
...
+
+      return {
+        resources: resources.concat(nextResources),
+        nextResources,
+        resourceNotFound,
+      };
+    },
+    /**
+     * getResourcesStream always gets the latest resources from the API.
+     * @param {Object} args - all arguments to pass to the `paginateResources` function
```
js07, line 554: **-Minor-**
```suggestion
     * @param {Object} args - all arguments to pass to the `getResourcesStream` function
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,625 @@
...
+            url: meta.next,
+          },
+        };
+      },
+    },
+    limit: {
+      type: "integer",
+      label: "Limit",
+      description: "Limit the number of results",
+      optional: true,
```
js07, line 159:
```suggestion
      optional: true,
      min: 1,
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,625 @@
...
+    },
+    offset: {
+      type: "integer",
+      label: "Offset",
+      description: "Start pagination after this number",
+      optional: true,
```
js07, line 165:
```suggestion
      optional: true,
      min: 0,
```

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,625 @@
...
+          return [];
+        }
+
+      },
+    },
+    max: {
+      type: "integer",
+      label: "Max records",
+      description: "Max number of records in the whole pagination",
+      optional: false,
```
js07, line 303:
```suggestion
      optional: false,
      min: 1,
```

js07 (commented):
I have a few suggestions, but overall it looks great!

components/amara/sources/common.mjs
```diff
@@ -0,0 +1,129 @@
...
+    timer: {
+      type: "$.interface.timer",
+      label: "Polling schedule",
+      description: "How often to poll the Amara API",
+      default: {
+        intervalSeconds: 60 * 15,
+      },
+    },
+    team: {
+      description: "The team slug of the team to watch. Passing in `null` or empty will return only resources that are in the public area.",
```
js07, line 17: Should the second sentence from this description be removed now that this prop has `optional: false`?

components/amara/amara.app.mjs
```diff
@@ -0,0 +1,589 @@
...
+     * to validate against. This parameter is only passed in from sources.
+     * @returns {Iterable} - Iterable that yields resources,
+     *  the first element is the last resource string in cache
+     */
+    async *getResourcesStream({
+      resourceFn,
+      resourceFnArgs,
+      offset,
+      limit = constants.DEFAULT_PAGE_LIMIT,
+      max = constants.DEFAULT_MAX_ITEMS,
```
js07, line 546: If there are more than `DEFAULT_MAX_ITEMS` (`300`) new events between runs of an event source, I think only the first 300 will be processed, and the rest will be missed. Should the default value for the `max` arg be removed?
```js
      max,
```
And change the `if` statement, https://github.com/PipedreamHQ/pipedream/blob/e9d0aac9692961d36a943f4d7e16fd770e404856/components/amara/amara.app.mjs#L583
to?:
```js
        if (!nextResponse?.meta.next || (max && resourcesCount >= max)) {
```

js07 (approved):

dannyroosevelt:
@jcortes @js07 can you confirm which actions/triggers have _not_ been successfully tested?

js07:
@dannyroosevelt All of the actions and triggers have worked for me. But in my last two reviews, I wasn't able to successfully test the **Video Added**, **Subtitles Published**, or **Video Made Primary** sources because Amara seemed to stop creating notifications for those events for my Amara team.

Amara API endpoints used to list events:
- Team Activity: `https://amara.org/api/teams/<team_slug>/activity/`
- Team Notifications: `https://amara.org/api/teams/<team_slug>/notifications/`

dannyroosevelt:
Great! @jcortes can you go ahead and merge to master? I'm going to enlist the help of the user who requested these to QA.

PR #1903, opened 2021-11-16
Docusign: update to use production url

michelle0927:
Updates Docusign sources and actions to work for production accounts instead of developer accounts.

### DIFF
--- a/components/docusign/actions/create-signature-request/create-signature-request.js
+++ b/components/docusign/actions/create-signature-request/create-signature-request.js
@@ -4,7 +4,7 @@ module.exports = {
   key: "docusign-create-signature-request",
   name: "Create Signature Request",
   description: "Creates a signature request from a template",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "action",
   props: {
     docusign,
--- a/components/docusign/docusign.app.js
+++ b/components/docusign/docusign.app.js
@@ -90,7 +90,7 @@ module.exports = {
     async getUserInfo() {
       return await this._makeRequest(
         "GET",
-        "https://account-d.docusign.com/oauth/userinfo",
+        "https://account.docusign.com/oauth/userinfo",
       );
     },
     async getBaseUri(accountId) {
--- a/components/docusign/sources/envelope-sent-or-complete/envelope-sent-or-complete.js
+++ b/components/docusign/sources/envelope-sent-or-complete/envelope-sent-or-complete.js
@@ -5,7 +5,7 @@ module.exports = {
   name: "Envelope Sent or Complete",
   description:
     "Emits an event when an envelope status is set to sent or complete",
-  version: "0.0.2",
+  version: "0.0.3",
   dedupe: "unique",
   props: {
     docusign,
--- a/components/docusign/sources/new-folder/new-folder.js
+++ b/components/docusign/sources/new-folder/new-folder.js
@@ -4,7 +4,7 @@ module.exports = {
   key: "docusign-new-folder",
   name: "New Folder",
   description: "Emits an event when a new folder is created",
-  version: "0.0.2",
+  version: "0.0.3",
   dedupe: "unique",
   props: {
     docusign,


dylburger:

dannyroosevelt:
@michelle0927 we're actually going to increase the scope of this a bit -- in order to support both developer and production accounts, we had to create an additional app for the developer use case. @dylburger had an idea on how we might be able to extend functionality for the existing Docusign components so they can work with either app, instead of having to clone all of them. I'll let him drop in his thoughts.

dylburger:
@michelle0927 the tl;dr is:

* I think we should have a common file that contains all shared props methods. We can keep this in the `docusign` app dir.
* Then the actual Docusign `*.app.js` files can import those from the common file, adding the actual app name slug in the file.
* Since the `key` will have to differ for components tied to each app, you'll have to create common files for each component and override the `key` / `version` / etc. for both the Docusign and Docusign Developer apps.

vellames-turing:
Hey @michelle0927. This stills in progress right?

michelle0927:
> Hey, @michelle0927. Looks like this is in progress right? I am asking this because I can't see any new commit after the previous review

@vellames-turing This is now ready for review.

components/docusign/sources/envelope-sent-or-complete/common.mjs
```diff
@@ -1,27 +1,14 @@
...
-  name: "Envelope Sent or Complete",
-  description:
-    "Emits an event when an envelope status is set to sent or complete",
-  version: "0.0.2",
+export default {
   dedupe: "unique",
   props: {
-    docusign,
     db: "$.service.db",
+    // eslint-disable-next-line pipedream/props-label,pipedream/props-description
```
vellames-turing, line 5: Could you please add a description and remove this lint-disable rule?

components/docusign/sources/new-folder/common.mjs
```diff
@@ -1,26 +1,14 @@
...
-  key: "docusign-new-folder",
-  name: "New Folder",
-  description: "Emits an event when a new folder is created",
-  version: "0.0.2",
+export default {
   dedupe: "unique",
   props: {
-    docusign,
     db: "$.service.db",
+    // eslint-disable-next-line pipedream/props-label,pipedream/props-description
```
vellames-turing, line 5: Its possible to add a description and remove this rule?

vellames-turing (changes requested):
Hello, I found some minor adjustments.
- There is an error message if I try to select a `template` without having a template. `TypeError: Cannot read property 'map' of undefined` (dev and prod)
- Could you please add a `summary` to the actions?
- Could you please add the documentation link to the actions description?

michelle0927:
> Hello, I found some minor adjustments and a bug.
> 
> * There is an error message if I try to select a `template` without having a template. `TypeError: Cannot read property 'map' of undefined` (dev and prod)
> * Could you please add a `summary` to the actions?
> * Could you please add the documentation link to the actions description?

For templates, I updated the prop to throw an error saying "No templates found" when no templates are returned.

vellames-turing (dismissed):
Looks great, thank you!

dannyroosevelt:
I saw some comments above about throwing errors when no templates are found, but I'm getting these errors for both prod and dev accts: 
image
image

michelle0927:
@dannyroosevelt I get the first error too for my prod account because it's a free account and doesn't have access to templates. I think the error being displayed has to do with how the new UI is looking for the error message? If I change `_makeRequest()` to this:
```
    async _makeRequest({ $, config }) {
      config.headers = this._getHeaders();
      try {
        return await axios($ ?? this, config);
      } catch(e) {
        throw new Error(e.response.data.message);
      }
    },
```
It displays the correct error:
`This User lacks sufficient permissions. User does not have the permision to the RequestTemplates method`

However, I'm not sure if that's an ideal solution. What do you think?

The error for the Developer version looks like there's a URL with a `/` missing between .com and templates. I'm not sure where that's coming from though. The only file that has "account-d.docusign.com" in it is `docusign_developer.app.mjs`, but it has "oauth/userinfo" after it:
`url: "https://account-d.docusign.com/oauth/userinfo",`

dannyroosevelt:
> 

@michelle0927 The action works fine for me in the developer app but I don't quite follow your recommendation / question for the production app. I'm still getting the below error.

What's the downside with your approach to display the proper error copy?

image

michelle0927:
@dannyroosevelt This is now updated to display the proper error.

vunguyenhung:

Test report
[DocuSign_Developer_1903.pdf](https://github.com/PipedreamHQ/pipedream/files/8708988/DocuSign_Developer_1903.pdf)
[DocuSign_1903.pdf](https://github.com/PipedreamHQ/pipedream/files/8708985/DocuSign_1903.pdf)

andrewjschuang (approved):

PR #1999, opened 2022-01-03
Fix Readme broken Llink to components API

DomHawken:
You might  also use /docs/components/api - unsure of your house style.


dylburger:
Thank you very much for the fix, @DomHawken !

PR #2022, opened 2022-01-06
Shopify App Partner integration

ctrlaltdylan:
Adds new Shopify Partner integration

* https://partner.shopify.com

Shopify Partners is the App Developer & Expert dashboard. You can track your app installs, referrals and app charges via this API.

### Sources

* New Transactions
* New Installs
* New Uninstalls


dylburger:

components/shopify_partner/README.md
```diff
@@ -0,0 +1,18 @@
...
+## Configure API Key
+
+In order to authorize Pipedream to access your Shopify Partner Account, you will need to generate a new API key for Pipedream to use on your behalf.
+
+Include these scopes:
+
+- `View financials`
+- `Manage apps`
+
+## Using Shopify Partner Sources
```
dylburger, line 12: Were you still planning to finish these sections or do you want to remove for now?

components/shopify_partner/README.md
```diff
@@ -0,0 +1,18 @@
+# Shopify Partner API Integration
+
+## Configure API Key
+
```
dylburger, line 4: In these instructions I normally like to link to the partner docs for creating API keys, in case users don't know how

components/shopify_partner/package.json
```diff
@@ -0,0 +1,17 @@
+{
```
dylburger, line 1: Thanks for adding this. When you merge to `master`, we should publish the package automatically.

Can you DM me your `npm` username? I'll add you to our `pipedream` npm organization.

components/shopify_partner/queries/getAppInstalls.js
```diff
@@ -0,0 +1,45 @@
+const { gql } = require("graphql-request");
```
dylburger, line 1: Now that you have the components working, can you try to convert to ESM? If you encounter the same errors you saw before, we should figure out what the blocker is in the API and address that.

components/shopify_partner/queries/getAppInstalls.js
```diff
@@ -0,0 +1,45 @@
...
+
+const query = gql`
```
dylburger, line 3: Does it make sense to link to the relevant Shopify GraphQL API docs so users can look up more about the query format?

components/shopify_partner/shopify_partner.app.js
```diff
@@ -0,0 +1,83 @@
+require("graphql/language");
+const get = require("lodash/get");
```
dylburger, line 2: Now that sources and workflows support Node 14, we prefer using the nullish coalescing operator in place of `lodash/get` to remove the dependency. I looked at the examples where you use `get` and I think nullish coalescing will work for each, but lmk if not.

components/shopify_partner/shopify_partner.app.js
```diff
@@ -0,0 +1,83 @@
...
+  methods: {
+    /**
+     * Handle a Shopify Partner GraphQL query
+     *
+     * Includes pagination, based off of a recusion or the last stored cursor in the $service.db
+     * @param {String} query - the query passed to the Shopify Partner API
+     * @param {Db} db - the Pipedream database for getting/setting cursors for pagination
+     * @param {String} mutation - the mutation passed to the Shopify Partner API
+     * @param {Object} variables - variables passed to a query or mutation
+     * @param {Funciton} handleEmit - handles event emission given the response data
```
dylburger, line 17: `Funciton` -> `Function`

components/shopify_partner/shopify_partner.app.js
```diff
@@ -0,0 +1,83 @@
...
+const { GraphQLClient } = require("graphql-request");
+
+module.exports = {
+  type: "app",
+  app: "shopify_partner",
+  methods: {
+    /**
+     * Handle a Shopify Partner GraphQL query
+     *
+     * Includes pagination, based off of a recusion or the last stored cursor in the $service.db
```
dylburger, line 12: "recusion" -> "recursion"

components/shopify_partner/shopify_partner.app.js
```diff
@@ -0,0 +1,83 @@
...
+        headers: {
+          "Content-Type": "application/json",
+          "X-Shopify-Access-Token": this.$auth.api_key,
+        },
+      });
+
+      // the key is unique to the source module, so we should always be getting the last message
+      const lastCursor = db.get(key);
+
+      console.log(`lastCursor: ${lastCursor}`);
```
dylburger, line 43: `db.get` and `db.set` should already show logs for the get / set (e.g. here it should automatically print the value retrieved for `key`). Does that show up in your source?

components/shopify_partner/shopify_partner.app.js
```diff
@@ -0,0 +1,83 @@
...
+          }
+          : {}),
+      });
+
+      console.log(JSON.stringify(data, null, 2));
+
+      if (data) {
+        handleEmit(data);
+        // the db becomes the source of truth for the last cursor
+        //   return null because undefined causes a bug
```
dylburger, line 59: Can you include this case in your bug report re: `undefined` keys, too? Sounds like it fails for both keys and values correct?

components/shopify_partner/shopify_partner.app.js
```diff
@@ -0,0 +1,83 @@
...
+        db.set(key, get(data, cursorPath, null));
+      }
+
+      // paginate the results recursively
+      //   the hasNextPagePath references how to traverse the response from Shopify
+      //   if hasNextPath is truthy, we continue
```
dylburger, line 65: This last line of the comment duplicates the logic in the code so I would recommend removing that specifically

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+
+module.exports = {
+  key: "shopify_partner-new-app-charges",
+  name: "New App Charges",
+  type: "source",
+  version: "0.0.1",
+  description:
+    "Emit new events when new app charges made to your partner account.",
+  props: {
+    shopify,
```
dylburger, line 12: The `shopify` app prop, `db` and `timer` can be abstracted into a common file for all three sources

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+    db: "$.service.db",
+    createdAtMin: {
+      type: "string",
```
dylburger, line 15: Are these props required or optional?

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+      label: "createdAtMin",
+    },
+    createdAtMax: {
+      type: "string",
+      description:
+        "Only include transactions up to this specific time (ISO timestamp)",
+      label: "createdAtMin",
+    },
+    timer: {
+      description: "How often this action should run",
```
dylburger, line 27: When I add descriptions to timers, I like to avoid Pipedream terms as much as possible (and this is a source), so I'd say something like "How often to poll for new app charges"

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+      type: "$.interface.timer",
+      label: "timer",
+      default: {
+        intervalSeconds: 60 * 60,
+      },
+    },
+  },
+  dedupe: "unique",
```
dylburger, line 35: Can be pulled into the common file

components/shopify_partner/sources/new_app_installs.js
```diff
@@ -0,0 +1,75 @@
...
+module.exports = {
+  key: "shopify_partner-new-app-installs",
+  name: "New App Installs",
+  type: "source",
+  version: "0.0.1",
+  description: "Emit new events when new shops install your app.",
+  props: {
+    shopify,
+    db: "$.service.db",
+    appId: {
```
dylburger, line 13: Does it make sense to include this as a part of the information we collect for a connected account? i.e. should a Shopify Partner account be specific to an (organization, app ID)?

components/shopify_partner/sources/new_app_installs.js
```diff
@@ -0,0 +1,75 @@
...
+      type: "string",
+      optional: false,
```
dylburger, line 15: This is the default, so I would remove

components/shopify_partner/sources/new_app_installs.js
dylburger: Separate question: is there a Shopify API to list app IDs? Can this prop be an async option so the user doesn't have to enter it manually?

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+    createdAtMax: {
+      type: "string",
+      description:
+        "Only include transactions up to this specific time (ISO timestamp)",
+      label: "createdAtMin",
+    },
+    timer: {
+      description: "How often this action should run",
+      type: "$.interface.timer",
+      label: "timer",
```
dylburger, line 29: I would also rename this something like "API Polling Frequency" to describe the timer's purpose

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+
+    await this.shopify.query({
+      db,
+      key: this.key,
+      query: getAppTransactions,
+      variables,
+      hasNextPagePath: "transactions.pageInfo.hasNextPage",
+      cursorPath: "transactions[0].cursor",
+      handleEmit: (data) => {
+        data.transactions.edges.map(({ node: { ...txn } }) => {
```
dylburger, line 58: Will `data.transactions.edges` always be defined / an array?

components/shopify_partner/sources/new_app_charges.js
```diff
@@ -0,0 +1,67 @@
...
+          this.$emit(txn, {
+            id: txn.id,
+            summary: `New successful app charge ${txn.id} that earned ${txn.netAmount.amount} ${txn.netAmount.currencyCode}`,
```
dylburger, line 61: Will all transactions have the amount and `currencyCode`?

dylburger: Text this long may get truncated in the sources UI. I like to print only the information unique to each event (i.e. the user knows these are all app charges, so you can remove "New successful app charge" and perhaps change to something like:

```
Transaction ${txn.id} earned ${txn.netAmount.amount} ${txn.netAmount.currencyCode}
```

components/shopify_partner/sources/new_app_installs.js
```diff
@@ -0,0 +1,75 @@
...
+      description: "gid://partners/App/<your App ID here>",
+      label: "Shopify App ID",
+    },
+    occurredAtMin: {
```
dylburger, line 19: Since `appId`, `occurredAtMin` and `occurredAtMax` are defined twice, you can move those to the app file and then reference the `propDefinition` within each component. See https://pipedream.com/docs/components/api/#prop-definitions-example

components/shopify_partner/sources/new_app_installs.js
```diff
@@ -0,0 +1,75 @@
...
+    },
+    occurredAtMax: {
+      type: "string",
+      description:
+        "Only include install events up to this specific time (ISO timestamp)",
+      label: "occurredAtMin",
+      optional: true,
+    },
+    timer: {
+      description: "How often this action should run",
```
dylburger, line 34: Same feedback as the prior timer

components/shopify_partner/sources/new_app_charges.js
dylburger: If you move into a common file, I guess you could keep the description generic to all sources, e.g. "How often to poll for new events" or something like that

components/shopify_partner/sources/new_app_uninstalls.js
```diff
@@ -0,0 +1,74 @@
...
+    await this.shopify.query({
+      db,
+      query: getAppUninstalls,
+      variables,
+      key: this.key,
+      handleEmit: (data) => {
+        data.app.events.edges.map(({ node: { ...event } }) => {
+          this.$emit(event, {
+            id: event.occurredAt,
+            summary: `Shopify shop ${event.shop.name} (${event.shop.myshopifyDomain}) uninstalled ${event.app.name}`,
```
dylburger, line 66: Same feedback re: `summary` as above

components/shopify_partner/sources/new_app_installs.js
```diff
@@ -0,0 +1,75 @@
...
+    await this.shopify.query({
+      db,
+      key: this.key,
+      query: getAppInstalls,
+      variables,
+      handleEmit: (data) => {
+        data.app.events.edges.map(({ node: { ...event } }) => {
+          this.$emit(event, {
+            id: event.occurredAt,
+            summary: `Shopify shop ${event.shop.name} (${event.shop.myshopifyDomain}) installed ${event.app.name}`,
```
dylburger, line 67: Same feedback re: `summary` as above

dylburger (changes requested):
Looking great! Let me know if you have any questions on the feedback

components/shopify_partner/shopify_partner.app.js
ctrlaltdylan: Yup agreed, just was dealing with that `undefined` bug and saw lodash.get was still being referencing in docs/other actions. Wasn't sure if that's the direct we were heading.

ctrlaltdylan: Actually, it kind of makes sense to use `get` in this case because lodash's get offers nested pathing, which I don't believe is available out of the box with ES6.

For example, the cursor for pagination for the uninstalls source is:

`const cursorPath = 'app.events.edges[0].cursor'`

It's not possible - afaik - to path in ES6 with this:

```javascript
const cursorPath = 'app.events.edges[0].cursor'
const cursor = response[cursorPath];
```

So in this case, get is nice because it supports nested array & object traversal with that dot notation.

components/shopify_partner/shopify_partner.app.js
ctrlaltdylan: 😅

components/shopify_partner/shopify_partner.app.js
ctrlaltdylan: Yup good call. This was cruft leftover when I was debugging the `undefined` value to `db.set` bug earlier.

components/shopify_partner/shopify_partner.app.js
ctrlaltdylan: Yea it is extra verb-age that's not really helpful. I'm going to remove the last 2 lines.

components/shopify_partner/shopify_partner.app.js
ctrlaltdylan: Ah right, good point. I think I only reported that the value was `undefined.` I'll update the issue.

components/shopify_partner/sources/new_app_charges.js
ctrlaltdylan: Donezo

components/shopify_partner/sources/new_app_charges.js
ctrlaltdylan: They are optional, but I thought the component docs mentioned that the `optional` param was implicitly true, but I can make it explict.

components/shopify_partner/sources/new_app_charges.js
ctrlaltdylan: Yup, even if there are no new transactions since the last execution the array will be empty.

components/shopify_partner/sources/new_app_charges.js
ctrlaltdylan: Sounds like it would work, but interestingly enough the dev command doesn't deploy:

```javascript

  key: "shopify_partner-new-app-installs",
  name: "New App Installs",
  type: "source",
  version: "0.0.1",
  description: "Emit new events when new shops install your app.",
  ...common,
  props: {
    ...common.props,
    db: "$.service.db",
    appId: {
      type: "string",
      optional: false,
      description: "gid://partners/App/<your App ID here>",
      label: "Shopify App ID",
    },
    occurredAtMin: {
      type: "string",
      description:
        "Only include install events after this specific time (ISO timestamp)",
      label: "occurredAtMin",
      optional: true,
    },
    occurredAtMax: {
      type: "string",
      description:
        "Only include install events up to this specific time (ISO timestamp)",
      label: "occurredAtMix",
      optional: true,
    },
  },
```

```bash
error         | component create api call responded with status: 400, body: {"error":"{\"name\":\"UserError\",\"message\":\"unsupported type for prop dedupe: unique\"}"}                                           │
Waiting for file change (or ^C to quit)...
```

components/shopify_partner/sources/new_app_installs.js
ctrlaltdylan: > Does it make sense to include this as a part of the information we collect for a connected account? i.e. should a Shopify Partner account be specific to an (organization, app ID)?

It's ad hoc, the `app.events` namespace in the GraphQL API requires a specific `app_id`, however the `transactions` namespace doesn't accept app IDs and its just transactions on the partner (or organization, same meaning).

> Separate question: is there a Shopify API to list app IDs? Can this prop be an async option so the user doesn't have to enter it manually?

I wish, I would like to add async options for a better UX, but I couldn't find an `Apps` connection in the query root. Kind of a bummer.

components/shopify_partner/sources/new_app_charges.js
ctrlaltdylan: Nice, I dig it

ctrlaltdylan: > Will all transactions have the amount and currencyCode?

Yes it's part of the `Transaction` object in the API.

components/shopify_partner/README.md
ctrlaltdylan: Added some docs, even though users won't see it.

components/shopify_partner/README.md
```diff
@@ -0,0 +1,26 @@
...
+This action will poll the Shopify Partner API for new app installs, requires an `app_id` in the Shopify GraphQL API GID format.
+
+### New App Uninstalls
+
+This action will poll the Shopify Partner API for new app installs, requires an `app_id` in the Shopify GraphQL GID format.
+
+### New Transactions
+
+This action will poll the Shopify Partner API for new app charges, including reoccuring, one time and usage charges.
```
dylburger, line 26: reoccuring -> reoccurring

components/shopify_partner/README.md
```diff
@@ -0,0 +1,26 @@
...
+Make sure to include these scopes:
+
+- `View financials`
+- `Manage apps`
+
+You will use this API key to authenticate your Pipedream workflows with Shopify Partners.
+
+### New App Installs
+
+This action will poll the Shopify Partner API for new app installs, requires an `app_id` in the Shopify GraphQL API GID format.
```
dylburger, line 18: Can you add "GID" to the `.wordlist.txt` file?

dylburger (approved):

components/shopify_partner/README.md
ctrlaltdylan: 📝 added

Issue #1483, opened 2021-07-28
[ACTION] PostgreSQL new actions


michelle0927:
**Is there a specific app this action is for?**
PostgreSQL

**New Row** \| Adds a new row
**Update Row** \| Updates an existing Row
**Find Row** \| Fins a row in a table via a lookup column
**Find Row via Custom Query** \| Finds a row in a table via a custom query you control
**Delete Row(s) from a Table** \| Deletes a row or rows from a table
**Execute a Query** \| Executes a query

michelle0927:
Related issue for new PostgreSQL event sources: https://github.com/PipedreamHQ/pipedream/issues/1034
These two issues should be completed together.

dannyroosevelt:
@michelle0927 this is ready for release! 

>Added via [Pipedream](https://pipedream.com/@/p_8rCkMkO?o=a&a=github)

Issue #1034, opened 2021-03-01
[SOURCE] PostgreSQL: New Event Sources


michelle0927:
**Describe the event source. What app, what event?**
PostgreSQL

**New Table** | Triggers when you add a new table.
**New Column** | Triggers when you add a new column.
**New Row** | Trigger when you add a new row.
**New or Updated Row** | Triggers when a row is added or modified.
**New Row (Custom Query)** | Triggers when new rows are returned from a custom query that you provide.

dannyroosevelt:
@michelle0927 this is ready for release! 

>Added via [Pipedream](https://pipedream.com/@/p_8rCkMkO?o=a&a=github)

PR #2024, opened 2022-01-06
PostgreSQL - new sources & actions

michelle0927:
New event sources and new actions for PostgreSQL. Addresses issues https://github.com/PipedreamHQ/pipedream/issues/1034 and https://github.com/PipedreamHQ/pipedream/issues/1483


dylburger:

components/postgresql/actions/execute-custom-query/execute-custom-query.mjs
```diff
@@ -0,0 +1,23 @@
...
+    postgresql,
+    query: {
+      propDefinition: [
+        postgresql,
+        "query",
+      ],
+    },
+  },
+  async run({ $ }) {
+    const res = await this.postgresql.executeQuery(this.query);
```
jcortes, line 19: Not sure if this will be secured against sql injection, since is not using any kind of [prepared statements](https://node-postgres.com/features/queries#prepared-statements) here. I would suggest to add support for prepared statemens so if you want you can take a look at this example with [MySql execute-query.mjs](https://github.com/PipedreamHQ/pipedream/blob/ac90bf536ce0e911ac0bc7b07d6ba2e23b09e18e/components/mysql/actions/execute-query/execute-query.mjs#L52)

components/postgresql/actions/new-row/new-row.mjs
```diff
@@ -0,0 +1,35 @@
...
+  async run({ $ }) {
+    const {
+      table,
+      rowValues,
+    } = this;
+    const columns = Object.keys(rowValues);
+    const values = Object.values(rowValues);
+    const res = await this.postgresql.insertRow(table, columns, values);
+    $.export("$summary", "New row inserted");
+    return res;
```
jcortes, line 33: You should consider the case when a user type an unknown column name
```suggestion
   try {
      const res = await this.postgresql.insertRow(table, columns, values);
      $.export("$summary", "New row inserted");
      return res;

    } catch (error) {
      $.export("$summary", "New row not inserted due to an error");
    }
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,193 @@
...
+    },
+    value: {
+      type: "string",
+      label: "Lookup Value",
+      description: "Value to search for",
+      async options({
+        table, column,
+      }) {
+        return this.getColumnValues(table, column);
+      },
```
jcortes, line 36: I'm not sure about getting all values from one table in this way because the table might have a huge number of records, so pagination would be a must here I guess. The other way I see it is by doing it with just a simple where string (e.g. [MySql delete-row](https://github.com/PipedreamHQ/pipedream/blob/ac90bf536ce0e911ac0bc7b07d6ba2e23b09e18e/components/mysql/actions/delete-row/delete-row.mjs#L27))

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,193 @@
...
+    },
+    rowValues: {
+      type: "object",
+      label: "Row Values",
+      description: "Enter the column names and respective values as key/value pairs",
```
jcortes, line 41: I guess here we can also give a clue to the user on how to add the key/value pairs when structure mode is off describing the value as `{{{columnName:"columnValue"}}}`?

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,193 @@
...
+      const client = new Client({
+        connectionString,
+      });
+      await client
+        .connect()
+        .catch((err) => console.error("Connection error", err.stack));
+      return client;
+    },
+    async endClient(client) {
+      client.end();
```
jcortes, line 64:
```suggestion
      return client.end();
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,193 @@
...
+    /**
+     * Executes SQL query and returns the resulting rows
+     * @param query {string} SQL query to execute
+     * @returns Array of rows returned from the given SQL query
+     */
+    async executeQuery(query) {
+      const client = await this.getClient();
+      const { rows } = await client.query(query);
+      await this.endClient(client);
+      return rows;
```
jcortes, line 75: I think you might consider getting errors running the query and if that happens you are not ending the connection so this is my suggestion:
```suggestion
      try {
        const { rows } = await client.query(query);
        return rows;
      } finally {
        await this.endClient(client);
      }
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,193 @@
...
+     * @param column {string} Column to filter by lastResult
+     * @param lastResult {string} A column value. Used to retrieve new rows since the
+     *  last time the table was queried.
+     * @returns Array of rows returned from the query
+     */
+    async getRows(table, column, lastResult = null) {
+      const query = `SELECT * FROM ${table} ${lastResult
+        ? "WHERE " + column + ">" + lastResult
+        : ""} ORDER BY ${column} DESC`;
+      const { rows } = await this.executeQuery(query);
```
jcortes, line 115:
```suggestion
      const rows = await this.executeQuery(query);
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,193 @@
...
+    },
+    /**
+     * Inserts a row in a table
+     * @param table {string} Name of database table to insert row into
+     * @param columns {array} Array of column names
+     * @param values {array} Array of values corresponding to the column names provided
+     * @returns The newly created row
+     */
+    async insertRow(table, columns, values) {
+      const { rows } = this.executeQuery(`INSERT INTO ${table} (${columns.join()}) VALUES (${values.join()}) RETURNING *`);
```
jcortes, line 155: You are missing the `await` here. Also you can remove the `.join()` function from the template literal since the array will be converted automatically into an string of names separated by comas.

Also there is a caveat on the sql query syntax here, because it seems like if I'm sending a column name that was declared as a VARCHAR it throws an error. So my suggestion is to map the values and force them to be enclosed by single quotes like this
```javascript
${values.map((v) => "'" + v + "'")
```
Otherwise you need to advice the user in the property description that he needs to set those values inside `''`

components/postgresql/sources/new-row/new-row.mjs
```diff
@@ -0,0 +1,60 @@
...
+        id: row[column],
+        summary: "New Row Added",
+        ts: Date.now(),
+      };
+    },
+  },
+  async run() {
+    const column = this._getColumn();
+    await this.newRows(this.table, column);
+  },
```
jcortes, line 59: There is an issue with this one because if I have a table with these columns `(id, name)` and I select the column `name` in the setup config, that means the rows will be order by `name`. So then let's supose I'm adding the first row as
```sql
INSERT INTO test_table (id, name) values (1, 'George');
```
and right after the insert I run the source, so in this case the `lastResult` values will be `George`. So then if I need to insert the next row like this
```sql
INSERT INTO test_table (id, name) values (2, 'Andrew');
```
and then the trigger gets called. It won't display this new record since the latest `name` inserted was `George`  and according to the sql statement here in `getRows` function
```sql
...
WHERE " + column + ">" + lastResult
...
```
it won't show the row with `Andrew` in the `name`.

So I think a practical solution for this is when we are deploying we should validate if the colum is `unique` if not, throws with error advicing this to the user. In that way the data would be more consistent when it comes to get new inserted rows. I was testing this on Zapier and it's the same aproach, they seem to not allow to select a column that is not unique before proceed with next step.

Another solution would be to make the user know that the `column` property should be datetime type the same aproach that you took for  `new-or-updated-row.mjs` in the property description.

Let me see what are your thoughts on this.

jcortes (changes requested):
I've just left a few comments/suggestions but overall looks good to me!

components/postgresql/postgresql.app.mjs
michelle0927: Updated to use pagination.

components/postgresql/sources/new-row/new-row.mjs
michelle0927: I updated `new-row.mjs` so that it doesn't use `lastResult`. It will just use the selected column to dedupe. I also added in a method to check that the column provided is unique (checks that it contains no duplicate values) which is used in sources `new-or-updated-row.mjs`, `new-row.mjs`, and `new-row-custom-query.mjs`.

components/postgresql/actions/delete-rows/delete-rows.mjs
```diff
@@ -0,0 +1,51 @@
...
+  async run({ $ }) {
+    const {
+      table,
+      column,
+      value,
+    } = this;
+    const rowCount = await this.postgresql.getRowCount(table);
+    await this.postgresql.deleteRows(table, column, value);
+    const newRowCount = await this.postgresql.getRowCount(table);
+    const rowsDeleted = rowCount - newRowCount;
```
jcortes, line 48: So here you can omit both `getRowCount` if you use `RETURNING *` in the delete sql statement

components/postgresql/actions/execute-custom-query/execute-custom-query.mjs
```diff
@@ -0,0 +1,43 @@
...
+    if (!Array.isArray(values)) {
+      throw new Error("No valid values provided. The values property must be an array.");
+    }
+
+    const numberOfValues = query?.match(/\$/g)?.length || 0;
+    if (values.length !== numberOfValues) {
+      throw new Error("The number of values provided does not match the number of values in the query.");
+    }
+
+    const res = await this.postgresql.executeQuery(query, values);
```
jcortes, line 39: If you change `executeQuery` as I suggested in `postgresql.app.mjs` I would refactor this one to
```suggestion
    const res = await this.postgresql.executeQuery({
      text: query,
      values,
    });
```

components/postgresql/actions/find-row-custom-query/find-row-custom-query.mjs
```diff
@@ -0,0 +1,46 @@
+import postgresql from "../../postgresql.app.mjs";
+
+export default {
```
jcortes, line 3: This particular component is kind of confusing becase the name suggests it's only for finding rows with a custom query but at the same time the query is too open to even run an insert/update/delete as well, so I think this one could easily be replaced by `execute-custom-query` don´t you think so?

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
+import pg from "pg";
```
jcortes, line 1: I would also suggest using [pg-format lib](https://www.npmjs.com/package/pg-format) to also prevent potential SQL Injection. It is mentioned [here as well](https://node-postgres.com/features/queries#parameterized-query)

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+          values,
+        }
+        : text;
+      try {
+        const { rows } = await client.query(query);
+        return rows;
+      } finally {
+        await this.endClient(client);
+      }
+    },
```
jcortes, line 98: As a suggestion I would leave the `query` object to be configurable outside of this function to make it available to other functions. It migth be helpful in case you want to add more properties to `query`.

```suggestion
    async executeQuery(query) {
      const client = await this.getClient();
      try {
        const { rows } = await client.query(query);
        return rows;
      } finally {
        await this.endClient(client);
      }
    },
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+      const rows = await this.executeQuery("SELECT table_name FROM information_schema.tables WHERE table_schema = 'public'");
+      return rows.map((row) => row.table_name);
+    },
+    /**
+     * Gets an array of column names in a table
+     * @param {string} table - Name of the table to get columns in
+     * @returns Array of column names
+     */
+    async getColumns(table) {
+      const rows = await this.executeQuery(`SELECT column_name FROM INFORMATION_SCHEMA.COLUMNS WHERE TABLE_NAME = '${table}'`);
```
jcortes, line 113: If you change `executeQuery` the way I suggested previously I would refactor this one to
```suggestion
      const rows = await this.executeQuery({
        text: `
          SELECT column_name FROM INFORMATION_SCHEMA.COLUMNS
            WHERE TABLE_NAME = $1
        `,
        values: [
          table,
        ],
      });
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+      return rows.map((row) => row.column_name);
+    },
+    /**
+     * Gets a table's primary key field
+     * @param {string} table - Name of the table to get the primary key for
+     * @returns Name of the primary key column
+     */
+    async getPrimaryKey(table) {
+      const rows = await this.executeQuery(`SELECT a.attname FROM pg_index i JOIN pg_attribute a ON a.attrelid = i.indrelid AND a.attnum = ANY(i.indkey) WHERE i.indrelid = '${table}'::regclass AND i.indisprimary`);
```
jcortes, line 122: If you change `executeQuery` the way I suggested previously I would refactor this one to
```suggestion
      const rows = await this.executeQuery({
        text: `
          SELECT a.attname FROM pg_index i
            JOIN pg_attribute a ON a.attrelid = i.indrelid
              AND a.attnum = ANY(i.indkey)
            WHERE i.indrelid = $1::regclass
              AND i.indisprimary
        `,
        values: [
          table,
        ],
      });
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+     * @param {string} column - Column to filter by lastResult
+     * @param {string} [lastResult] - A column value. Used to retrieve new rows since the
+     *  last time the table was queried.
+     * @returns Array of rows returned from the query
+     */
+    async getRows(table, column, lastResult = null) {
+      const query = `SELECT * FROM ${table} ${lastResult
+        ? "WHERE " + column + ">" + lastResult
+        : ""} ORDER BY ${column} DESC`;
+      return this.executeQuery(query);
```
jcortes, line 137:
```suggestion
      const select = "SELECT * FROM %I";
      const where = "WHERE %I > $1";
      const orderby = "ORDER BY %I DESC";

      const query = lastResult
        ? {
          text: format(`${select} ${where} ${orderby}`, table, column, column),
          values: [
            lastResult,
          ],
        }
        : format(`${select} ${orderby}`, table, column);

      return this.executeQuery(query);
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+    },
+    /**
+     * Gets a single row from a table based on a lookup column & value
+     * @param {string} table - Name of database table to query
+     * @param {string} column - Column to filter by value
+     * @param {string} value - A column value to search for
+     * @returns A single database row
+     */
+    async findRowByValue(table, column, value) {
+      const rows = await this.executeQuery(`SELECT * FROM ${table} WHERE ${column} = '${value}'`);
```
jcortes, line 147: If you change `executeQuery` the way I suggested previously I would refactor this one to
```suggestion
      const rows = await this.executeQuery({
        text: format("SELECT * FROM %I WHERE %I = $1", table, column),
        values: [
          value,
        ],
      });
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+    },
+    /**
+     * Gets the number of rows in a table
+     * @param {string} table - Name of database table to query
+     * @returns Count of how many rows exist in the table
+     */
+    async getRowCount(table) {
+      const response = await this.executeQuery(`SELECT COUNT(*) FROM ${table}`);
+      return response[0].count;
+    },
```
jcortes, line 158: I think this function won´t be necessary if you get the array of rows affected by `RETURNING *` in the delete sql statement.

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+    /**
+     * Deletes a row or rows in a table based on a lookup column & value
+     * @param {string} table - Name of database table to delete from
+     * @param {string} column - Column used to find the row(s) to delete
+     * @param {string} value - A column value. Used to find the row(s) to delete
+     */
+    async deleteRows(table, column, value) {
+      await this.executeQuery(`DELETE FROM ${table} WHERE ${column} = '${value}'`);
+    },
```
jcortes, line 167: If you change `executeQuery` the way I suggested previously I would refactor this one to
```suggestion
   async deleteRows(table, column, value) {
      return this.executeQuery({
        text: format("DELETE FROM %I WHERE %I = $1 RETURNING *", table, column),
        values: [
          value,
        ],
      });
    },
```
Also with this aproach you won´t need to call twice the function `getRowCount` in order to get the number of deleted rows

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+    /**
+     * Inserts a row in a table
+     * @param {string} table - Name of database table to insert row into
+     * @param {array} columns - Array of column names
+     * @param {array} values - Array of values corresponding to the column names provided
+     * @returns The newly created row
+     */
+    async insertRow(table, columns, values) {
+      return this.executeQuery(`INSERT INTO ${table} (${columns}) VALUES (${values.map((v) => "'" + v + "'")}) RETURNING *`);
+    },
```
jcortes, line 177: If you change `executeQuery` the way I suggested previously I would refactor this one to
```suggestion
    async insertRow(table, columns, values) {
      const placeholders = this.getPlaceholders({
        values,
      });
      return this.executeQuery({
        text: format(`
          INSERT INTO %I (${columns}) 
            VALUES (${placeholders})
            RETURNING *
        `, table),
        values,
      });
    },
```

with this additional function which will give you the necessary placeholder values for the query
```javascript
    getPlaceholders({
      values = [], fromIndex = 1,
    }) {
      return values.map((_, index) => `$${index + fromIndex}`);
    },
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+        val,
+      ]) => {
+        query += ` ${key} = '${val}',`;
+      });
+      // trim last comma
+      query = query.slice(0, -1);
+      query += ` WHERE ${primaryKey} =(SELECT ${primaryKey} FROM ${table} WHERE ${lookupColumn} = '${lookupValue}' ORDER BY ${primaryKey} LIMIT 1) RETURNING *`;
+      const response = await this.executeQuery(query);
+      return response[0];
+    },
```
jcortes, line 201: If you change `executeQuery` the way I suggested previously I would refactor this one to

```suggestion
    async updateRow(table, lookupColumn, lookupValue, rowValues) {
      const primaryKey = await this.getPrimaryKey(table);
      const columnsPlaceholders = this.getColumnsPlaceholders({
        rowValues,
        fromIndex: 2,
      });
      const response = await this.executeQuery({
        text: format(`
          UPDATE %I SET ${columnsPlaceholders}
            WHERE %I = (
              SELECT %I FROM %I
              WHERE %I = $1
              ORDER BY %I LIMIT 1
            ) RETURNING *
        `, table, primaryKey, primaryKey, table, lookupColumn, primaryKey),
        values: [
          lookupValue,
          ...Object.values(rowValues),
        ],
      });
      return response[0];
    },
```

with this additional function which will give you the necessary placeholder values for the query
```javascript
    getColumnsPlaceholders({
      rowValues = {}, fromIndex = 1,
    }) {
      return Object.keys(rowValues)
        .map((key, index) => `${key} = $${index + fromIndex}`);
    },
```

components/postgresql/postgresql.app.mjs
```diff
@@ -1,11 +1,215 @@
...
+    /**
+     * Gets all of the values for a single column in a table
+     * @param {string} table - Name of database table to query
+     * @param {string} column - Name of column to get values from
+     * @params {string} [limit] - maximum number of rows to return
+     * @params {integer} [offset] - number of rows to skip, used for pagination
+     * @returns Array of column values
+     */
+    async getColumnValues(table, column, limit = "ALL", offset = 0) {
+      const rows = await this.executeQuery(`SELECT ${column} FROM ${table} LIMIT ${limit} OFFSET ${offset}`);
```
jcortes, line 211: If you change `executeQuery` the way I suggested previously I would refactor this one to
```suggestion
      const rows = await this.executeQuery({
        text: format(`
          SELECT %I FROM %I
            LIMIT $1::numeric OFFSET $2::numeric
        `, column, table),
        values: [
          limit,
          offset,
        ],
      });
```

components/postgresql/sources/common.mjs
```diff
@@ -0,0 +1,68 @@
...
+      const rows = await this.postgresql.getRows(table, column, lastResult);
+      for (const row of rows) {
+        const meta = this.generateMeta(row, column);
+        this.$emit(row, meta);
+      }
+      this._setLastResult(rows, column);
+    },
+    async isColumnUnique(table, column) {
+      const query = `select count(*) <> count(distinct ${column}) as duplicate_flag from ${table}`;
+      const hasDuplicates = await this.postgresql.executeQuery(query);
```
jcortes, line 63: This sql can also be used with `pg-format`

components/postgresql/sources/new-row-custom-query/new-row-custom-query.mjs
```diff
@@ -0,0 +1,79 @@
...
+    if (values.length !== numberOfValues) {
+      throw new Error("The number of values provided does not match the number of values in the query.");
+    }
+
+    const isColumnUnique = await this.isColumnUnique(table, column);
+    if (!isColumnUnique) {
+      throw new Error("The column selected contains duplicate values. Column must be unique");
+    }
+
+    const rows = await this.postgresql.executeQuery(query, values);
```
jcortes, line 73: If you change `executeQuery` as I suggested in `postgresql.app.mjs` then you can change this to
```suggestion
    const rows = await this.postgresql.executeQuery({
      text: query,
      values,
    });
```

jcortes (changes requested):
Thanks for making all those changes. There si a few more comments request changes I would like you to address and that would be good enough. Thanks again!

components/postgresql/actions/find-row-custom-query/find-row-custom-query.mjs
michelle0927: I agree. `find-row-custom query` does the same thing as `execute-custom-query` but only returns the first result. I removed this component.

components/postgresql/sources/common.mjs
```diff
@@ -59,7 +60,9 @@ export default {
       this._setLastResult(rows, column);
     },
     async isColumnUnique(table, column) {
-      const query = `select count(*) <> count(distinct ${column}) as duplicate_flag from ${table}`;
+      const query = {
+        text: format("select count(*) <> count(distinct %I) as duplicate_flag from %I", column, table),
+      };
```
jcortes, line 65: Optionally you can also set query as a plain string by using `format`
```suggestion
      const query = format("select count(*) <> count(distinct %I) as duplicate_flag from %I", column, table);
```

jcortes (approved):
Thanks a lot for all those changes again, nice work!

dmathukorala:
Verification completed, issue observed in Update Row Action - Error: Relation Does not exist (Please find the screenshot attached)
image

michelle0927:
@dmathukorala The error was coming from the method `getPrimaryKey()` where the query was using a case insensitive version of the `table` name. I made an update that fixed the issue for me, so let me know if you're still seeing that error.

DilanAthukorala:
@michelle0927 Now observing different error for update_row action
image

michelle0927:
@DilanAthukorala I'm able to reproduce this issue when the table selected doesn't have a primary key constraint. I made an update to `updateRow()` to use the `lookupColumn` instead of the `primaryKey` so that this component should work in tables without a primary key.

DilanAthukorala:
Thanks @michelle0927 this works fine now. Could you also check **New or Updated Row** source I'm not getting any events after updating rows or adding new rows.

michelle0927:
> Thanks @michelle0927 this works fine now. Could you also check **New or Updated Row** source I'm not getting any events after updating rows or adding new rows.

@DilanAthukorala Is the `column` prop that you're entering "A datetime column, such as 'date_updated' or 'last_modified' that is set to the current datetime when a row is updated"?

michelle0927:
@dannyroosevelt This needs an approval before I can merge.

dannyroosevelt (approved):

dannyroosevelt:
> @dannyroosevelt This needs an approval before I can merge.

approved!

PR #2032, opened 2022-01-07
new Create a Text Post action for linkedin app

marcodali:
With this action it is posible now to create a text Post for the user of the connected account.
The only param needed is the text to be published

### DIFF
--- a/components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
+++ b/components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
@@ -1,0 +1,51 @@
+import axios from "axios";
+
+export default {
+  key: "linkedin_create-text-post-user",
+  name: "Create a Text Post (User)",
+  description: "Create shares on LinkedIn using text, URLs, and images. [See the docs](https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/share-on-linkedin#creating-a-share-on-linkedin) for more information",
+  version: "0.0.11",
+  type: "action",
+  props: {
+    name: {
+      type: "string",
+      label: "Text",
+      description: "Text to be posted on LinkedIn timeline",
+    },
+    linkedin: {
+      label: "LinkedIn",
+      description: "LinkedIn Integration",
+      type: "app",
+      app: "linkedin",
+    },
+  },
+  async run({ $ }) {
+    const data = {
+      "author": `urn:li:person:${this.linkedin.$auth.oauth_uid}`,
+      "lifecycleState": "PUBLISHED",
+      "specificContent": {
+        "com.linkedin.ugc.ShareContent": {
+          "shareCommentary": {
+            "text": this.name,
+          },
+          "shareMediaCategory": "NONE",
+        },
+      },
+      "visibility": {
+        "com.linkedin.ugc.MemberNetworkVisibility": "PUBLIC",
+      },
+    };
+    try {
+      await axios.post("https://api.linkedin.com/v2/ugcPosts", data, {
+        headers: {
+          "Authorization": `Bearer ${this.linkedin.$auth.oauth_access_token}`,
+          "X-Restli-Protocol-Version": "2.0.0",
+        },
+      });
+      $.export("$summary", "Successfully created a new Text Post as User");
+    } catch (err) {
+      console.error(err);
+      $.export("$summary", "It wasn't posible to create Post");
+    }
+  },
+};
--- a/components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
+++ b/components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
@@ -1,51 +1,30 @@
-import axios from "axios";
+import { axios } from "@pipedream/platform";
+import linkedin from "../../linkedin.app.mjs";
 
 export default {
-  key: "linkedin_create-text-post-user",
+  key: "linkedin-create-text-share-user",
   name: "Create a Text Post (User)",
-  description: "Create shares on LinkedIn using text, URLs, and images. [See the docs](https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/share-on-linkedin#creating-a-share-on-linkedin) for more information",
-  version: "0.0.11",
+  description: "Create shares on LinkedIn using text and URLs. [See the docs](https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/share-on-linkedin#creating-a-share-on-linkedin) for more information",
+  version: "0.0.1",
   type: "action",
   props: {
+    linkedin,
     name: {
-      type: "string",
-      label: "Text",
-      description: "Text to be posted on LinkedIn timeline",
-    },
-    linkedin: {
-      label: "LinkedIn",
-      description: "LinkedIn Integration",
-      type: "app",
-      app: "linkedin",
+      propDefinition: [
+        linkedin,
+        "name",
+      ],
     },
   },
   async run({ $ }) {
-    const data = {
-      "author": `urn:li:person:${this.linkedin.$auth.oauth_uid}`,
-      "lifecycleState": "PUBLISHED",
-      "specificContent": {
-        "com.linkedin.ugc.ShareContent": {
-          "shareCommentary": {
-            "text": this.name,
-          },
-          "shareMediaCategory": "NONE",
-        },
-      },
-      "visibility": {
-        "com.linkedin.ugc.MemberNetworkVisibility": "PUBLIC",
-      },
-    };
+    const payload = this.linkedin.getPayload("NONE", this.name);
+    let response = "It wasn't posible to create Post";
     try {
-      await axios.post("https://api.linkedin.com/v2/ugcPosts", data, {
-        headers: {
-          "Authorization": `Bearer ${this.linkedin.$auth.oauth_access_token}`,
-          "X-Restli-Protocol-Version": "2.0.0",
-        },
-      });
+      response = await axios($, this.linkedin.getRequestConfig(payload));
       $.export("$summary", "Successfully created a new Text Post as User");
     } catch (err) {
-      console.error(err);
-      $.export("$summary", "It wasn't posible to create Post");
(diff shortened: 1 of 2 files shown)


dylburger:

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
+import axios from "axios";
```
dylburger, line 1: You'll want to use `@pipedream/platform` axios here. See https://pipedream.com/docs/components/guidelines/#use-pipedream-platform-axios-for-all-http-requests

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+
+export default {
+  key: "linkedin_create-text-post-user",
```
dylburger, line 4: You can use `linkedin-` instead of an underscore. We use underscores to separate spaces in multi-word app names (like `google_calendar`), but dashes to separate app name + component name. See https://pipedream.com/docs/components/guidelines/#component-key-pattern

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+  name: "Create a Text Post (User)",
```
dylburger, line 5: What is the "user" in this context? I see from the MSFT docs that the action is called "Creating a Share on LinkedIn" there. Can you create a post as an admin vs. a user?

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+  description: "Create shares on LinkedIn using text, URLs, and images. [See the docs](https://docs.microsoft.com/en-us/linkedin/consumer/integrations/self-serve/share-on-linkedin#creating-a-share-on-linkedin) for more information",
```
dylburger, line 6: Appreciate the link to docs

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+  version: "0.0.11",
```
dylburger, line 7: It's hard when you're iterating on actions, but before you merge this, you'll want to change to `0.0.1`. See https://pipedream.com/docs/components/guidelines/#versioning

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+  type: "action",
+  props: {
```
dylburger, line 9: Prop definitions should typically live in the app file. See https://pipedream.com/docs/components/guidelines/#app-files

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+    name: {
+      type: "string",
+      label: "Text",
+      description: "Text to be posted on LinkedIn timeline",
+    },
+    linkedin: {
```
dylburger, line 15: Move to app file, as well

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
dylburger: I see that the title in the linked issue is tied to that, gotcha. @dannyroosevelt we said the action should be called "Create a Text Post (User)" — maybe that's what Zapier had? — but the API docs call it "Creating a Share". What do you think the action title should be?

dannyroosevelt: I'm not sure where that came from. Zapier is "Create Share Update," so let's go with "Create a Share (consumer)". I think the "user" is referencing NOT posting to a company page.

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+      label: "LinkedIn",
+      description: "LinkedIn Integration",
+      type: "app",
+      app: "linkedin",
+    },
+  },
+  async run({ $ }) {
+    const data = {
+      "author": `urn:li:person:${this.linkedin.$auth.oauth_uid}`,
```
dylburger, line 24: I would move this into the app file, as well, as a method that generates LinkedIn share payloads. We'll want to reuse this for other types of shares (e.g. images), where most of the payload will be the same.

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+          },
+          "shareMediaCategory": "NONE",
+        },
+      },
+      "visibility": {
+        "com.linkedin.ugc.MemberNetworkVisibility": "PUBLIC",
+      },
+    };
+    try {
+      await axios.post("https://api.linkedin.com/v2/ugcPosts", data, {
```
dylburger, line 39: Move this into the app file, and abstract all API requests using a `_makeRequest` method like so: https://github.com/PipedreamHQ/pipedream/blob/dbc0a14a790a00e8aca8ef5d2fd2cb0aa6c99570/components/github/github.app.js#L197

This keeps things that are common to all LinkedIn API requests — the base URL, the auth headers, the `X-Restli-Protocol-Version`, etc. — within a single method. Then you can rewrite this to pass specific data for this API request. For example, see how the `_makeRequest` method is used in GitHub here: https://github.com/PipedreamHQ/pipedream/blob/dbc0a14a790a00e8aca8ef5d2fd2cb0aa6c99570/components/github/github.app.js#L211

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
```diff
@@ -0,0 +1,51 @@
...
+        headers: {
+          "Authorization": `Bearer ${this.linkedin.$auth.oauth_access_token}`,
+          "X-Restli-Protocol-Version": "2.0.0",
+        },
+      });
+      $.export("$summary", "Successfully created a new Text Post as User");
```
dylburger, line 45: I would store the create post response and also `return` that after the `$summary` export, so the user has access to e.g. the post ID and other things that they can use in the rest of their workflow.

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
dylburger: Since this action specifically posts only text, you should remove references to images in the description. Can you also pass a URL as text here, or is that shared in a different way via the API?

dylburger (changes requested):
Thanks a lot for putting this together. Let me know if you have any questions.

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: There will be 2 actions very similar:

1. Create a Text Post (User)
2. Create a Text Post (Organization)

What should be the correct key name?

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: Yes, the **shareMediaCategory** can be one of:

1. NONE - The share does not contain any media, and will only consist of text.
2. ARTICLE - The share contains a URL.
3. IMAGE - The Share contains an image.

So I think I will remove "and images" from the description.

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
dannyroosevelt: Create a Text Share (User): `linkedin_create-text-share-user`
Create a Text Share (Organization): `linkedin_create-text-share-organization`

marcodali: ok! this action has the first name

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: done

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: implemented, now the $return_value is "id": "urn:li:share:XXX"

marcodali (commented):
All suggested changes have been implemented

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: ok!

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: changed!

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: ok!

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: there is now in the linkedin.app.mjs two methods:

1. getPayload()
2. getRequestConfig()

Maybe the name should be changed to "generate" instead of "get"?

components/linkedin/actions/create-text-post-user/create-text-post-user.mjs
marcodali: Can you review the solution I proposed to reuse the headers, url, method and inject payload? I think it also works very well

marcodali (commented):
Changes are done

dannyroosevelt:
Closing this out, will start from scratch for this set of components

PR #2036, opened 2022-01-07
Firebase_admin_sdk - Update Authentication

michelle0927:


dylburger:

components/firebase_admin_sdk/firebase_admin_sdk.app.mjs
```diff
@@ -76,52 +70,44 @@ module.exports = {
...
+        "https://www.googleapis.com/auth/cloud-platform",
+      ];
+      const jwtClient = await new googleAuth.JWT(
+        clientEmail,
+        null,
+        formattedPrivateKey,
+        SCOPES,
+        null,
       );
+      return jwtClient.authorize().then((tokens) => tokens.access_token);
```
jcortes, line 95: Since we are using `async/await` pattern aproach I recommend using it here to return the token
```suggestion
      const { access_token: accessToken } = await jwtClient.authorize();
      return accessToken;
```

jcortes (commented):
Overall looks good to me. However I've found an issue in the folowing scenario where the user wants to delete a record and the recreate it again with the same key name. The source will not show you that record because it got registered with the `dedupe: unitque` cache strategy. So I'm not sure what other strateggy we should use in this case. Also this is out of the scope so I think the PR is ok.

michelle0927:
> Overall looks good to me. However I've found an issue in the folowing scenario where the user wants to delete a record and then recreates it again with the same key name. The source will not show you that record because it got registered with the `dedupe: unique` cache strategy. So I'm not sure what other strategy we should use in this case. Also this is out of the scope so I think the PR is ok.

That's a good point. Since the database snapshot doesn't include any creation dates, we could just store all of the keys and compare them each run, but that's probably overkill?

jcortes:
> That's a good point. Since the database snapshot doesn't include any creation dates, we could just store all of the keys and compare them each run, but that's probably overkill?

I do agree. I would say let's just keep it as it is right now!

jcortes (approved):
Looks good to me!

dannyroosevelt:
go ahead and merge when you're ready @michelle0927

PR #2042, opened 2022-01-10
Replace node-sass with sass instead (M1 issues)

ctrlaltdylan:
# Replace node-sass with sass package

Having issues with building `node-sass` even with the arch flag, using the drop in replacement `sass` library instead.


dylburger (approved):
lgtm! Thanks for cleaning this up and adding the docs

PR #2056, opened 2022-01-12
[DOCS] New workflow builder

dylburger:


docs/docs/.vuepress/config.js
```diff
@@ -23,8 +23,27 @@ module.exports = {
...
+    nav: [
+      {
+        text: "v2",
+        className: "docs-version",
+        ariaLabel: "Docs Version Menu",
+        items: [
+          {
+            text: "v2",
+            link: "https://pipedream.com/docs-v2",
+            internal: true,
```
dylburger, line 35: Was this property set `true` to avoid the external link icon, or does it serve another purpose?

docs/docs/.vuepress/config.js
```diff
@@ -52,35 +70,52 @@ module.exports = {
         ],
       },
       {
-        title: "Workflows",
+        title: "Code",
```
dylburger, line 73: In the hierarchy of concepts, do you think Workflows should go above Code instead? That way the order of the concepts in the sidebar go: Overview -> Workflows -> things you can do in workflows (like code, connect accounts, etc.)

docs/docs/limits/README.md
```diff
@@ -89,10 +89,6 @@ Currently, most of the [limits that apply to HTTP triggers](#http-triggers) also
 
 The only limit that differs between email and HTTP triggers is the payload size: the body of HTTP requests is limited to `{{$site.themeConfig.PAYLOAD_SIZE_LIMIT}}`, where the total size of an email sent to a workflow - its body, headers, and attachments - is limited to `{{$site.themeConfig.EMAIL_PAYLOAD_SIZE_LIMIT}}` on the default interface.
```
dylburger, line 90: You can probably remove "on the default interface" at the end of the last sentence, since there's only a single interface now.

docs/docs/limits/README.md
```diff
@@ -89,10 +89,6 @@ Currently, most of the [limits that apply to HTTP triggers](#http-triggers) also
...
 
-**However, you can send emails up to `30MB` in size using the address `[YOUR EMAIL ENDPOINT]@upload.pipedream.net`**. [Read more here](/workflows/steps/triggers/#sending-large-emails).
-
-This `30MB` limit cannot be raised.
```
dylburger, line 94: Keep this since we've received questions about it. Some limits (like QPS) can be raised and others cannot, so we like to call that out. We need a better matrix of this [like AWS has](https://docs.aws.amazon.com/lambda/latest/dg/gettingstarted-limits.html).

docs/docs/.vuepress/theme/components/NavLink.vue
```diff
@@ -17,7 +18,8 @@
     @focusout="focusoutAction"
   >
     {{ item.text }}
-    <OutboundLink v-if="isBlankTarget" />
+    <span v-if="item.badge" class="badge" :class="item.badgeVariation">{{ item.badge }}</span>
```
dylburger, line 21: Thanks for all of these changes

docs/docs/.vuepress/theme/components/NavLinks.vue
```diff
@@ -141,6 +145,14 @@ export default {
   .repo-link {
     margin-left: 0.75rem;
   }
+
+  .docs-version {
+    .dropdown-title, .nav-link {
+      font-family: 'JetBrains Mono', monospace;
```
dylburger, line 151: This visually distinguished it well for me.

Separately, should we add the "New" text you have in the drop-down menu to the "v2" label in the nav, as well? If not not a big deal, but that might also bring attention to it for users who want to explore.

Screen Shot 2022-01-21 at 2 03 41 PM

docs/docs/code/README.md
```diff
@@ -0,0 +1,20 @@
+# Writing Code in Workflows
+
+_This document details how to use code in workflow steps, without leaving your browser. If you're building a [component](/components/), please [reference the component API docs](/components/api/)._
```
dylburger, line 3: "workflow steps" should be a link

docs/docs/code/README.md
```diff
@@ -0,0 +1,20 @@
...
+
+Pipedream comes with thousands of prebuilt [triggers](/workflows/steps/triggers/) and [actions](/components/actions/). Often, these will be sufficient for building simple workflows.
```
dylburger, line 5: Maybe add:

```
Pipedream comes with thousands of prebuilt [triggers](/workflows/steps/triggers/) and [actions](/components/actions/) for [hundreds of apps](https://pipedream.com/apps)
```

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
+# Writing Bash in Code Steps
+
+Prefer to write quick scripts in bash? We've got you covered.
```
dylburger, line 3: Capitalize "bash" here

dylburger: I saw some other places in this doc where Bash should be capitalized, as well. I'll make sure it's cleaned up in the product, too.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+You can run any bash in a Pipedream step within your workflows.
+
+::: warning
+Bash steps are available in a limited alpha release.
+
+You can still run arbitrary bash scripts, including [sharing data between steps](/code/bash/#sharing-data-between-steps) as well as [accessing environment variables](/code/bash/#using-environment-variables).
```
dylburger, line 10: I would shorten to:

```
You can still run arbitrary bash scripts, [share data between steps](/code/bash/#sharing-data-between-steps), and  [access environment variables](/code/bash/#using-environment-variables).
```

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+However, features available in [Node.js steps](/code/nodejs) like `$.respond`, `$.end`, and `$.auth` are not yet available in bash. If you have any questions please [contact support](https://pipedream.com/support).
```
dylburger, line 12: I would recommend reframing this section to talk about the features in the abstract, instead of the specific methods on the `$`. Maybe something like

> However, you can't connect accounts, return HTTP responses, or take advantage of other features available in the Node.js environment at this time.

Maybe also add "If you have any questions **, find any bugs, or have any feedback** ..." in the last sentence.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+:::
+
+## Adding a Bash code step
+
+1. Click the + icon to add a new step
+2. Click "Custom Code"
+3. In the new step, select the bash runtime in language dropdown
+
+## Logging & debugging
```
dylburger, line 21: Maybe at the end of this section, note that "any standard output or standard error should also appear in the Logs section of the results"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Available binaries
+
+Bash steps come with many common & useful binaries preinstalled and available in `$PATH` for you to use out of the box. These binaries include but aren't limited to:
+
+* `curl` for making HTTP requests
+* `jq` for manipulating and viewing JSON data
+* `git` for interacting with remote repositories
+
+Unfortunately it is not possible to install from a package manager like `apt` or `yum`.
```
dylburger, line 42: "Unfortunately it is not possible to install **packages**"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+If you need a package pre-installed in your bash steps, [just ask us](https://pipedream.com/support).
+
+Otherwise, you can use `/tmp` to download and install from source.
```
dylburger, line 46: "download and install **the software**"

dylburger: "use **the** `/tmp` **directory**"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## Making an HTTP request
+
+`curl` is already preinstalled in bash steps, we recommend using it for making HTTP requests in your code for sending or requesting data from APIs or webpages.
```
dylburger, line 50: This sentence is a little long. Consider splitting up into two like:

> `curl` comes preinstalled in Bash steps. You can use it to make HTTP requests that send or request data from APIs or webpages.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Making a GET request
+
+You can use `curl` to perform GET requests from websites or APIs directly.
```
dylburger, line 54: Small style point: I like to wrap `GET` and other API methods in backticks.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+::: tip
+Use the `--silent` flag with `curl` to suppress extra extra diagnostic information that `curl` produces when making requests.
+
+This enables you to only worry about the body of the response so you can visualize it with tools like `echo` or `jq`. 
+:::
+
+### Making a POST request
+
+`curl` handles making POSTs requests as well. The `-X` flag allow you to specify the HTTP method you'd like to use for an HTTP request.
```
dylburger, line 73: word choice on "handles making". Maybe "`curl` can also make `POST` requests."

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+The `-d` flag is for passing data in the POST request.
+
+```bash
+curl --silent -X POST https://postman-echo.com/post -d 'name=Bulbasaur&id=1'
+
+# to store the API response in a variable, interpolate the response into a string and store it in variable
```
dylburger, line 80: Very very small nit but I like to capitalize the comments when they begin a complete sentence

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+RESPONSE=`curl --silent -X POST https://postman-echo.com/post -d 'name=Bulbasaur&id=1'`
+
+# now the response is stored as a variable
+echo $RESPONSE
+```
+
+### Using API key authentication
+
+Some APIs require you to use a secret API key to prove your access.
```
dylburger, line 89: "Some APIs require you to **authenticate with** a secret API key"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+`curl` has an `-h` flag where you can pass your API key as a token.
+
+Below is an example of searching Twitter for mentions of a specific Twitter handle. Twitter requires a valid API key passed in the header of the request.
```
dylburger, line 93: You could shorten to "For example, here's how to retrieve mentions from the Twitter API:"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+```bash
+curl --silent -X POST -h "authorization:Bearer $(<your api key here>)" https://api.twitter.com/2/users/@pipedream/mentions
```
dylburger, line 96: As another small style point, I normally capitalize header names and place a value between the `:` and the value:

```
"Authorization: Bearer"
```

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+## Sharing data between steps
+
+A step can accept data from other steps in the same workflow, or pass data downstream to others.
```
dylburger, line 101: "others" -> "other steps". As a new user I could read "others" as referring to workflows, since that ends the last phrase.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+This makes your steps even more powerful, you can compose new workflows and reuse steps.
```
dylburger, line 103: Consider removing this sentence, since this section isn't about composition. Then you can move straight from: "A step can receive data from other steps and pass it to others" -> here's how to use data from other steps.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+### Using data from another step
+
+In bash scripts, data from the initial workflow trigger and other steps are available in the `$PIPEDREAM_STEPS` environment variable.
+
+In this example, we'll pretend this data is coming into our HTTP trigger via a POST request.
+
+```json
+{
+  "id": 1,
+  "name": 'Bulbasaur",
```
dylburger, line 114: Change first single-quote to double

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+  "type": "plant"
+}
+```
+
+In our bash script, we can access this data via the `$PIPEDREAM_STEPS` variable. Specifically, this data from the POST request into our workflow is available in the `trigger` object.
```
dylburger, line 119: `$PIPEDREAM_STEPS` is now an environment variable that points to a file (Giao just made this change), so you'll need to modify this section and change the examples below to use something like:

```bash
jq .trigger.event < $PIPEDREAM_STEPS
```

dylburger: You may also want to link to [these docs](https://pipedream.com/docs/workflows/events/#event-format) re: the format of the `trigger` object, and briefly explain that that contains the data that triggered your workflow.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+```bash
+echo $PIPEDREAM_STEPS | jq .trigger.event
+
+# results in { id: 1, name: "Bulbasaur", type: "plant" }
+```
+
+::: tip
+The period (`.`) in front the `trigger.event` in the example is not a typo. This is to define the starting point for `jq` to start traversing down the JSON in the HTTP response.
```
dylburger, line 128: "start traversing down" -> "traverse"

(I hope these suggestions are helpful. I suffer from lengthy writing so I'm always on the lookout for places to cut words when I review docs PRs)

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+:::
+
+### Sending data downstream to other steps
+
+To share data created, retrieved, transformed or manipulated by a step to others downstream, append it to the `$PIPEDREAM_EXPORTS` variable.
```
dylburger, line 133: "To share data created, retrieved, transformed or manipulated by a step to others downstream" -> "To return data for future steps to use"

dylburger: Like `$PIPEDREAM_STEPS` above, `$PIPEDREAM_EXPORTS` now points to a file, so you'll also need to update this section

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+An example speaks a thousand words, so here's one passing data from an API to the bash step.
```
dylburger, line 135: You can probably remove this sentence entirely and just let the example follow from the first sentence of the section.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+# Retrieve the data from an API and store it in a variable
+DATA=`curl --silent https://pokeapi.co/api/v2/pokemon/charizard`
+
+# Write data to $PIPEDREAM_EXPORTS to share with steps downstream
+EXPORT="key:json=${DATA}"
+echo $EXPORT >> $PIPEDREAM_EXPORTS
+```
+
+::: tip
+Don't worry, the special `key` string in the `EXPORT` will automatically reference the current step's name.
```
dylburger, line 147: This is actually not true. `key` refers to the actual name of the property exported from the step. If I set it to `key`:

Screen Shot 2022-01-21 at 2 42 08 PM

If I set to `foo:json=...`:

Screen Shot 2022-01-21 at 2 40 23 PM

In the context of the step, we always know `$PIPEDREAM_EXPORTS` refers to exports for the current step (e.g. `steps.code`). The key you pass at the beginning just defines the property under that (e.g. `steps.code.key`).

It's also worth noting that the `:json` after the key actually tells the execution environment that the data is JSON and that we should format it as a structured object in exports.

If I just do `key=value`, this is what the export looks like:

Screen Shot 2022-01-21 at 2 38 22 PM

When I do `key:json=value`, I get:

Screen Shot 2022-01-21 at 2 38 05 PM

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+To access them, just append the `$` in front of the environment variable name.
+
+```bash
+echo $POKEDEX_API_KEY
+```
+
+Or an even more useful example, using the stored environment variable to make an authenticated API request.
+
+```bash
+curl --silent -X POST -h "authorization:Bearer $TWITTER_API_KEY" https://api.twitter.com/2/users/@pipedream/mentions
```
dylburger, line 165: Same small feedback re: `Authorization` header format as above

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+```
+
+## Raising exceptions
+
+You may need to stop your step immediately. You can use the normal `exit` function available in bash to quit the step prematurely.
+
+```bash
+  echo "Exiting now!" 1>&2
```
dylburger, line 173: Indent these two lines over

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+  exit 1
+```
+
+This will exit the step and output the error message to `stderr` which will appear in the results of the step in the workflow.
```
dylburger, line 177: Maybe note that it will exit the step _and_ end the execution of the workflow.

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+## File storage
+
+If you need to download and store files you can place them in the `/tmp` directory.
```
dylburger, line 181: Add comma after "files"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+
+### Writing a file to /tmp
+
+For example to download a file to `/tmp` using `curl`
```
dylburger, line 185: Comma after "For example"

docs/docs/code/bash/README.md
```diff
@@ -0,0 +1,199 @@
...
+# Download the current weather in Cleveland in PNG format
+curl --silent https://wttr.in/Cleveland.png --output /tmp/weather.png
+
+# output the contents of /tmp to confirm the file is there
+ls /tmp
+```
+
+### `/tmp` limitations
+
+The `/tmp` directory can store up to 512 megabytes of storage. Also the storage may be wiped or may not exist between workflow executions.
```
dylburger, line 197: Can you create a new var in `config.js` to store the max size of the `/tmp` dir? We reference this in a few places.

dylburger: Actually, we should do a couple more things:

- Move all limits to [this section](https://pipedream.com/docs/limits/#disk) of the limits docs, then refer users there from this section.
- Reference the existing language [here](https://pipedream.com/docs/workflows/steps/code/nodejs/working-with-files/#managing-tmp-across-workflow-runs) re: how `/tmp` files remain across workflow runs.

docs/docs/code/nodejs/README.md
```diff
@@ -1,26 +1,18 @@
-# Code
+# Writing Node.js in Code Steps
```
dylburger, line 1: I like this new intro section

docs/docs/code/nodejs/README.md
```diff
@@ -1,26 +1,18 @@
-# Code
```
dylburger, line 1: Many of the code in this doc won't work in Node.js steps in bob, since they run in the component code environment. A few of the specific changes you should make:

* [Adding a code step section](https://pipedream-docs-redirect-do-not-edit-git-doc-b8b90e-pipedreamers.vercel.app/docs-v2/code/nodejs/#adding-a-code-step) — screenshot / instructions will need to be modified, the code example with `this.test` doesn't work, etc.
* [Passing parameters to code steps](https://pipedream-docs-redirect-do-not-edit-git-doc-b8b90e-pipedreamers.vercel.app/docs-v2/code/nodejs/#passing-parameters-to-code-steps) section needs to reference props instead of params (we should probably have a simple example of a Node.js component with a prop here)
* [async function declaration](https://pipedream-docs-redirect-do-not-edit-git-doc-b8b90e-pipedreamers.vercel.app/docs-v2/code/nodejs/#async-function-declaration) section should be removed. we may want to replace it with a section that discusses how `steps` is passed to the `run` method of components to show users how to access event data in here.
* We should also add a new section on how to link apps to steps via app props in the component, how that surfaces the connected account section of the step, etc. That section should probably live [in this doc](https://pipedream.com/docs/workflows/steps/code/auth/#the-auths-object), and we can link to that doc from here.
* We should remove the screenshot in the **Logs** section and change the language to note that logs are under the Results > Logs section of the output.

dylburger: The link to the updated `$.respond()` docs still references `$respond` (no `.`)

dylburger: The link to the docs on `$.send.http` still reference `$send` (also no `.`)

dylburger: The "managing state" section links to a 404 (since you removed the original `$checkpoint` doc). I would actually recommend keeping that doc and putting it under the Node.js sub-section, and just re-working the examples to use `$.service.db`. More comments on that below.

dylburger: Rework the `$end` section to use `$.flow.exit`

dylburger: [Using secrets in code](https://pipedream-docs-redirect-do-not-edit-git-doc-b8b90e-pipedreamers.vercel.app/docs-v2/code/nodejs/#using-secrets-in-code) section refers to making workflows public, so we'll want to clean up that section.

docs/docs/.vuepress/config.js
```diff
@@ -52,35 +70,52 @@ module.exports = {
...
         collapsable: false,
+        children: [
+          "/code/",
+          {
+            title: "Node.js",
+            children: ["/code/nodejs/"],
```
dylburger, line 79: I believe you'll need to add the docs you `git mv`d under this `children` section (e.g. workflow state, making HTTP requests, etc.)

docs/docs/code/nodejs/README.md
dylburger: I see what happened with some of the links. It looks like you `git mv`d the files but didn't update some of the references here (e.g. to the workflow state doc).

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+**Anything you can do in Python can be done in a Pipedream Workflow**. This includes using any of the [350,000+ PyPi packages available](https://pypi.org/) in your Python powered workflows. 
+
+Pipedream supports [Python v{{$site.themeConfig.PYTHON_VERSION}}](https://www.python.org) in workflows.
+
+::: warning
+Python steps are available in a limited alpha release.
+
+You can still run arbitrary python code, including [sharing data between steps](/code/python/#sharing-data-between-steps) as well as [accessing environment variables](/code/python/#using-environment-variables).
+
+However, features available in [Node.js steps](/code/nodejs) like `$.respond`, `$.end`, and `$.auth` are not yet available in bash. If you have any questions please [contact support](https://pipedream.com/support).
```
dylburger, line 12: Same comments here as in the Bash section

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+:::
+
+
+## Adding a Python code step
+
+1. Click the + icon to add a new step
+2. Click "Custom Code"
+3. In the new step, select the Python language runtime in language dropdown
```
dylburger, line 20: Bold **Python**

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+You can still run arbitrary python code, including [sharing data between steps](/code/python/#sharing-data-between-steps) as well as [accessing environment variables](/code/python/#using-environment-variables).
+
+However, features available in [Node.js steps](/code/nodejs) like `$.respond`, `$.end`, and `$.auth` are not yet available in bash. If you have any questions please [contact support](https://pipedream.com/support).
+:::
+
+
+## Adding a Python code step
+
+1. Click the + icon to add a new step
+2. Click "Custom Code"
```
dylburger, line 19: As a small point of style, I prefer bolding text vs. quotation marks when I'm referring to buttons / menu options. See https://developers.google.com/style/text-formatting

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+3. In the new step, select the Python language runtime in language dropdown
+
+## Logging & debugging
+
+You can use `print` at any time in a Python code step to view information as the script is running.
```
dylburger, line 24: "view" -> "log"

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+The output for the `print` logs will appear in the `Results` section just beneath the code editor.
+
+<div>
+<img alt="Python print log output in the results" src="./images/print-logs.png">
+</div>
+
+## Using 3rd party packages
```
dylburger, line 32: "3rd party" -> "third-party"

I get this guidance from [Dreyer's English](https://www.amazon.com/Dreyers-English-Utterly-Correct-Clarity/dp/0812995708), who goes by the rule of thumb that most numbers should be written as words in this context.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+You can use any packages from [PyPi](https://pypi.org) in your Pipedream workflows. This includes popular choices such as:
+
+* `requests` for making HTTP requests
```
dylburger, line 36: Worth linking to packages?

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+* `sqlalchemy` for retrieving or inserting data in a SQL database
+* `pandas` for working with complex datasets
+
+To use a PyPi package, just include it in your step's code:
+
+`import requests`
```
dylburger, line 42: Use multiline code here with `python` language declaration

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+## Making an HTTP request
+
+We recommend using the popular `requests` HTTP client package available in Python to send HTTP requests.
+
+No need to run `pip install`, just `import requests` at the top of your step's code and it's available for your code to use.
+
+### Making a GET request
+
+GET requests typically are for retrieving data from an API. Below is an example d
```
dylburger, line 56: "Below is an example d"

dylburger: Add backticks around `GET`

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+```python
+import requests
+
+url = 'https://swapi.dev/api/people/1'
+
+r = requests.get(url)
+
+# to see the response in your Pipedream step logs:
```
dylburger, line 65: Capitalize beginning of comments

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+url = 'https://api.twitter.com/2/users/@pipedream/mentions'
+
+headers { 'Authorization': f"Bearer {token}"}
+r = requests.get(url, headers=headers)
+
+print(r.text)
+```
+
+```note
```
dylburger, line 110: Did you mean for this to be a Vuepress tip?

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+# This a POST request to this URL will echo back whatever data we send to it
+url = 'https://postman-echo.com/post'
+
+data = {"name": "Bulbasaur"}
+
+r = requests.post(url, data)
+
+# to see the response in your Pipedream step logs:
+print(r.text)
```
dylburger, line 85: Most people will probably want to return the data for use in future steps. Maybe put the sharing data between steps section above this to introduce that concept, and then use the pipedream Python package to export the response data in this example.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+# to see the response status code in your Pipedream step logs:
+print(r.status)
+```
+
+### Using API key authentication
+
+If an particular service requires you to use an API key, you can pass it via the headers of the request.
```
dylburger, line 93: I would just use the same verbiage from the Bash section to introduce all of these concepts. I'll also mimic that in the Golang doc.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+This proves your identity to the service so you can interact with it:
+
+```python
+import requests
+
+token = 'replace this with your API key'
```
dylburger, line 100: Reference the use of environment variables here so we don't advocate users storing keys in code.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+```note
+Not all APIs use a `Authorization: Bearer <api key>` header format.
+
+Refer to the documentation of the API you're working with for their specific authentication format.
+```
+
+### Sending files
+
+You can send also files within a step.
```
dylburger, line 118: "send also" -> "also send"?

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+An example of sending a previously stored file in the workflow's `/tmp` directory: 
+
+```python
+# Retrieving a previously saved file from workflow storage
+files = {'image': ('python-logo.png', open('/tmp/python-logo.png', 'rb')}
```
dylburger, line 124: I believe you have an extra `(` in front of `python-logo.png`:

```
'image': ('python-logo.png'
```

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+r = requests.post(url='https://api.imgur.com/3/image', files=files)
+```
+
+## Sharing data between steps
+
+You can share data between steps in your workflows. Data can be passed forward to the next steps, and it can be used by other steps downstream.
```
dylburger, line 131: Like above, I would mimic the language in the Bash section to keep each of these introductions consistent.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+```json
+{
+  "id": 1,
+  "name": "Bulbasaur",
+  "type": "plant"
+}
+```
+
+Every Python step can import the `pipedream.script_helpers` module. This module contains data from other steps in the workflow via `steps` dictionary.
```
dylburger, line 149: "This module **exposes** data from **prior** steps..."

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+pokemon_type = steps["trigger"]["event"]["type"]
+
+print(f"{pokemon_name} is a {pokemon_type} type Pokemon")
+```
+
+### Sending data downstream to other steps
+
+The trigger steps automatically export their data for other steps to use downstream.
+
+Just pass the data to the `export` module from `pipedream.script_helpers`.
```
dylburger, line 167: `export` is a function so I would say:

> Call the `export` function to return data from a step

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+from pipedream.script_helpers import (steps, export)
+
+pokemon = { name: "Pikachu" }
+
+export('pokemon', pokemon)
+```
+
+Now this `pokemon` is accessible to downstream steps within `steps.code.pokemon`
+
+::: warning
```
dylburger, line 180: This is a good warning and I didn't see it in the Bash step. I would actually recommend re-writing the step exports docs [here](https://pipedream-docs-redirect-do-not-edit-git-doc-b8b90e-pipedreamers.vercel.app/docs-v2/workflows/steps/#exporting-data-in-code-steps) to include:

* Links to each language-specific doc on how to export from different language runtimes
* A section on Limitations like this

Then you could link to that **Limitations** section so you only keep the JSON-serializable note in one place.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+Not all data types can be stored in the `steps` data shared between workflow steps.  Only JSON serializable types can be passed, which includes:
+
+* lists 
+* dictionaries
+  :::
+
+## Using environment variables
+
+You can leverage any [environment variables defined in your Pipedream account](/environment-variables/#environment-variables) in a python step. This is useful for keeping your secrets out of code as well as keeping them flexible to swap API keys without having to update each step individually.
```
dylburger, line 189: python -> "Python"

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+`os.environ['ENV_NAME_HERE']` will raise an error that stops your workflow if that key doesn't exist in your Pipedream account.
+
+Whereas `os.environ.get('ENV_NAME_HERE')` will _not_ throw an error and instead returns an empty string. 
+
+If your code relies on the presence of a environment variable, consider using `os.environ['ENV_NAME_HERE']` instead.
+:::
+
+## Handling errors
+
+You may find a need to exit a workflow early. In a Python step, just a `raise` an error to halt a step's execution.
```
dylburger, line 219: "find a need" -> "need"

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+
+```python
+raise NameError('Something happened that should not. Exiting early.')
+```
+
+All exceptions from your Python code will appear in the logs area of the results.
```
dylburger, line 226: We should settle on a single way to discuss the **Results** and **Logs** section. Since Google's style guide recommends bolding UI elements, I would recommend bolding and capitalizing in all cases.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+## File storage
+
+Not only can you run Python code in workflows, but you can also store files as well. This means you can upload or download photos, retrieve datasets or accept files from an HTTP request and much more.
```
dylburger, line 230: "Not only can you run Python code in workflows, but you can also store files as well" -> "You can also store files in Python code steps"

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+The `/tmp` directory is accessible from your workflow steps for saving and retrieving files.
+
+You have full access to read & write both files & directories in `/tmp`.
```
dylburger, line 234: `&` -> "and"

This guidance also comes from Dreyer's English. Ampersands are generally reserved for formal titles and proper nouns (like law firms), but I always use "and" otherwise.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+with open('/tmp/python-logo.png', 'wb') as f:
+  # save the content of the HTTP response into the file
+  f.write(r.content)
+```
+
+Now `/tmp/python-logo.png` holds the official Python logo.
+
+### Reading a file from /tmp
+
+You can also open files you have previously stored in the `/tmp` directory. Let's open the `python-logo.png` we downloaded in the previous example.
```
dylburger, line 254: "Let's open the `python-logo.png` **file**"

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+import os
+
+with open('/tmp/python-logo.png') as f:
+  # put the contents of the file into a variable
+  file_data = f.read()
+```
+
+### Listing files in /tmp
+
+If you need to check what files are currently in `/tmp` you can list them and print the results to the workflow console:
```
dylburger, line 266: "workflow console" -> "**Logs** section of **Results"** to keep the reference consistent.

docs/docs/code/python/README.md
```diff
@@ -0,0 +1,279 @@
...
+
+```python
+import os
+
+# Prints the files in the tmp directory
+print(os.listdir('/tmp'))
+```
+
+### `/tmp` limitations
```
dylburger, line 275: Like the feedback in the **Bash** section, move this to its own section and link from here.

docs/docs/components/actions/README.md
```diff
@@ -22,6 +22,12 @@ For example, here's how to add the HTTP `GET` Request action:
 <img alt="Adding HTTP GET request action" src="./images/adding-http-get-request-action.gif">
 </div>
 
+::: warning
+Legacy actions are no longer available in the latest version of workflows. 
+
+This is a [new version](https://pipedream.com/community/t/actions-improved-open-source-now-in-beta/606) of Pipedream Actions available.  Legacy Actions haven been deprecated. Get started with our [quickstart](/components/quickstart/nodejs/actions/) for new actions or learn how to [migrate](/components/migrating/) legacy actions.
```
dylburger, line 28: "This is a" -> "There is a"

"Legacy Actions haven been deprecated" -> "Legacy Actions have been deprecated"

docs/docs/components/api/README.md
```diff
@@ -1,8 +1,8 @@
 # Component API Reference
 
-This document was created to help developers author and use [Pipedream components](/components/). You can develop [sources](/components/quickstart/nodejs/sources/) (workflow triggers) and [actions](/components/quickstart/nodejs/actions/) using the component API. You can publish components to your account for private use, or [contribute them to the Pipedream registry](/components/guidelines/) for anyone to run.
+This document was created to help developers author and use [Pipedream components](/components/). Not only can develop [sources](/components/quickstart/nodejs/sources/) (workflow triggers) and [actions](/components/quickstart/nodejs/actions/) using the component API, but you can also develop [Node.js steps](/code/nodejs) right in your workflows - without leaving your browser! You can publish components to your account for private use, or [contribute them to the Pipedream registry](/components/guidelines/) for anyone to run.
```
dylburger, line 3: "Not only can **you**"

docs/docs/destinations/README.md
```diff
@@ -23,16 +23,16 @@ The docs below discuss features common to all Destinations. See the [docs for a
 
 ### Using destinations in workflows
 
-You can send data to Destinations in [Node.js code steps](/workflows/steps/code/), too, using `$send` functions.
+You can send data to Destinations in [Node.js code steps](/workflows/steps/code/), too, using `$.send` functions.
```
dylburger, line 26: `/workflows/steps/code/` 404s since you renamed the file.

docs/docs/destinations/README.md
```diff
@@ -23,16 +23,16 @@ The docs below discuss features common to all Destinations. See the [docs for a
...
+`$.send` is an object provided by Pipedream that exposes destination-specific functions like `$.send.http()`, `$.send.s3()`, and more. **This allows you to send data to destinations programmatically, if you need more control than the default actions provide**.
 
-Let's use `$send.http()` to send an HTTP POST request like we did in the Action example above. [Add a new Action](/components/actions/#adding-a-new-action), then search for "**Run Node.js code**":
+Let's use `$.send.http()` to send an HTTP POST request like we did in the Action example above. [Add a new Action](/components/actions/#adding-a-new-action), then search for "**Run Node.js code**":
 
 [Create an endpoint URL](https://requestbin.com), adding the code below to your code step, with the URL you created:
 
 ```javascript
-$send.http({
+$.send.http({
```
dylburger, line 35: These examples won't work as written, since you only have access to the `$` within the `run` method of Node.js components. So you'll just need to wrap with the standard `defineComponent` code.

docs/docs/destinations/README.md
dylburger: I didn't comment in the other docs, but we'll need to modify the links there, too.

docs/docs/destinations/README.md
dylburger: I didn't comment in the other docs, but we'll need to modify the code there, too.

docs/docs/workflows/steps/triggers/README.md
```diff
@@ -288,12 +288,12 @@ When you're processing HTTP requests, you often don't need to issue any special
...
 
-If you need to issue a custom HTTP response from a workflow, **you can use the `$respond()` function in a Code or Action step**.
+If you need to issue a custom HTTP response from a workflow, **you can use the `$.respond()` function in a Code or Action step**.
 
-`$respond()` takes a single argument: an object with properties that specify the body, headers, and HTTP status code you'd like to respond with:
+`$.respond()` takes a single argument: an object with properties that specify the body, headers, and HTTP status code you'd like to respond with:
 
 ```javascript
-$respond({
+$.respond({
```
dylburger, line 296: You'll also need to modify this example to use `defineComponent` + the `run` method.

I would also call out here that returning HTTP responses are only available in the Node.js runtime.

docs/docs/workflows/steps/triggers/README.md
```diff
@@ -321,20 +321,20 @@ If you need to issue an HTTP response in the middle of a workflow, see the secti
 You can issue an HTTP response within a workflow, and continue the rest of the workflow execution, by setting the `immediate` property to `true`:
 
 ```javascript
-await $respond({
+await $.respond({
```
dylburger, line 324: Update with `defineComponent` + `run` method.

docs/docs/code/python/README.md
ctrlaltdylan: I was debating that, I wanted a monospace font for the package name, didn't want the link styling to override that. Let me try it out thought.

ctrlaltdylan: Format looks good. I'll link all package 👍

docs/docs/code/python/README.md
ctrlaltdylan: Good call, thought it was silly to make a multi-line doc for one doc, but for formatting sake we should do it.

docs/docs/code/python/README.md
ctrlaltdylan: I keep forgetting that markdown supports multiple formats in headlines & links. Updated.

docs/docs/code/python/README.md
ctrlaltdylan: I did, thanks for the catch.

docs/docs/code/python/README.md
ctrlaltdylan: True, perhaps a fast follow could be we link all mentions of Logs & Results in one section.

Side note, Google's style guide? Is that part of the build or some other tool?

docs/docs/code/python/README.md
ctrlaltdylan: Thanks, maybe I should look into a grammar GH action plugin 😅 

Just one minor change I made to the suggestion: "You can also store files in Python code steps" -> "You can also store and read files with Python steps."

My thinking is, we're not literally storing the file in a code step we're leveraging Python to manage files.

docs/docs/destinations/README.md
ctrlaltdylan: Ah I see, I thought the `defineComponent` line was implict and wasn't editable like the old `run` function in the past workflow examples. Sure I'll update all to match.

ctrlaltdylan: Probably should make a note in the Components reference to make sure developers understand they are defining a component and the `defineComponent` function isn't necessary.

docs/docs/code/python/README.md
ctrlaltdylan: Yea good call, I was trying to introduce the full suite of HTTP verbs, but we should never condone bad security practices even if it's for a partial example.

Should never assume a person will read the docs in full, most people including me like to take snippets and move on to solve their problem.

docs/docs/code/python/README.md
ctrlaltdylan: Agreed, "automatically" is not accurate.

docs/docs/code/python/README.md
ctrlaltdylan: Good call, added that to the newest branch update.

docs/docs/destinations/README.md
dylburger: Yes if that works feel free. You may also need to add equivalent redirects in `vercel.json` so we point Google to the new link.

docs/docs/.vuepress/config.js
ctrlaltdylan: This is part of that hack I had to do in order for the links to escape the parent path, it's a long story.

docs/docs/.vuepress/config.js
ctrlaltdylan: Yup 100% agreed. The user journey should start with Workflows before Code.

docs/docs/.vuepress/config.js
```diff
@@ -33,156 +52,94 @@ module.exports = {
...
+        children: 
+        [
           "/workflows/",
           "/workflows/steps/",
           "/workflows/steps/triggers/",
-          "/components/actions/",
-          "/workflows/steps/code/",
-          "/workflows/steps/params/",
-          "/workflows/steps/code/state/",
-          "/workflows/steps/code/nodejs/http-requests/",
```
dylburger, line 65: Did you get a chance to edit these docs (HTTP requests, working with files) and add them to the new Node.js code section?

docs/docs/code/nodejs/README.md
```diff
@@ -34,9 +26,13 @@ JavaScript is one of the [most used](https://insights.stackoverflow.com/survey/2
 You can add any Node.js code in the editor that appears. For example, try:
 
 ```javascript
-console.log("This is Node.js code");
-this.test = "Some test data";
-return "Test data";
+defineComponent({
+  async run({ steps, $ }) {
+    console.log("This is Node.js code");
+    this.test = "Some test data";
```
dylburger, line 32: `this.test` won't work. `$.export("test", "Some test data")` is the equivalent.

docs/docs/code/nodejs/README.md
```diff
@@ -49,12 +45,14 @@ You can make code steps reusable by allowing them to accept parameters. Instead
...
-You'll notice an [`async` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) declaration that appears when you add a new code step:
+You'll notice an [`async` function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/async_function) declaration that appears when you add a new Node.js code step:
 
 ```javascript
-async (event, steps) => {
-  // this node.js code will execute when your workflow is triggered
-};
+defineComponent({
+  async run({ steps, $ }) {
+      // this node.js code will execute when your workflow is triggered
```
dylburger, line 53: Capitalize Node.js in the comment

docs/docs/code/nodejs/README.md
```diff
@@ -65,9 +63,11 @@ This communicates a couple of key concepts:
...
-  // this node.js code will execute when your workflow is triggered
-};
+defineComponent({
+  async run({ steps, $ }) {
+    // this node.js code will execute when your workflow is triggered
+  })
+});
 ```
 
 When you use [step parameters](/workflows/steps/#passing-data-to-steps-step-parameters), Pipedream passes the `params` object (named pairs of param key and its associated value) to the function.
```
dylburger, line 73: Similar to the above, this also needs to be changed to reflect props instead of params, no function signature

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+This functionality (`$.service.db`) is limited to only Node.js code steps at this time.
+
+Other step languages like [Python](/code/python), [bash](/code/bash) and [Go](/code/go) do not have this feature available yet.
+
+For more information on what functionality is available for those languages, please refer to their documentation.
+:::
+
+### Injecting the database
+
+By default Node.js steps do not contain the database service. It needs to be injected by defining it as a `prop`.
```
dylburger, line 223: Comma after "By default"

I would say, "By default, Node.js steps **don't have access to** the database service"

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+```javascript
+defineComponent({
+  props: {
+    // Define that the "db" variable in our component is a database
+    db: "$.service.db",
+  },
+  async run({ steps, $ }) {
+    // Now we can access the database at "this.db"
+    this.db
```
dylburger, line 233: Change to `this.db.set("name", "Luke")`

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+:::tip
+`props` injects variables under `this` scope in components.
+
+In the above example we essentially instructed that this step needs the database injected into the `this.db` prop. 
+:::
+
+### Using the database
+
+Once the database is injected into the component, we can use it to both store & retrieve data generated from workflow steps. The 2 functions available are to `set` and `get` the data.
```
dylburger, line 246: "store **and** retrieve"

dylburger: You can combine these two sentences to shorten, maybe something like this:

> Once you inject the database into the component, you can use it to both store (`set`) and retrieve (`get`) data.

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+```javascript
+export default defineComponent({
+  props: {
+    "db": "$.service.db",
+  },
+  async run({ steps, $ }) {
+    // By default, all database entries are undefined.
+    // It's wise to set a default value so our code as an initial value to work with
+    const counter = this.db.get('counter') || 0;
```
dylburger, line 292: Since we're on Node 14, I would prefer `??` to `||`

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+    
+    // On the first run "counter" will be and we'll increment it to 1
```
dylburger, line 294: "On the first run "counter" will be" -> did you mean to say 0?

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+    // The next run will increment the counter to 2, and so forth
+    this.db.set('counter', counter + 1);
+  },
+})
+```
+
+### Dedupe data example
+
+This database is also useful for storing data from APIs from prior runs to prevent acting on duplicate data, or data that's been seen before.
```
dylburger, line 303: You can remove "from APIs" to say just "storing data from prior runs"

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+For example, this workflows trigger contains an email address from a potential new customer. But we want to track all emails collected so we don't send a welcome email twice:
```
dylburger, line 305: Add apostrophe to "workflows"

I would also say "this workflow's trigger **data**"

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+```javascript
+export default defineComponent({
+  props: {
+    "db": "$.service.db",
+  },
+  async run({ steps, $ }) {
+    const email = steps.trigger.context.new_customer_email;
+    // Retrieve the past recorded emails from other runs
+    const emails = this.db.get('emails') || [];
```
dylburger, line 315: `||` -> `??`

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+For example, this workflows trigger contains an email address from a potential new customer. But we want to track all emails collected so we don't send a welcome email twice:
+
+```javascript
+export default defineComponent({
+  props: {
+    "db": "$.service.db",
+  },
+  async run({ steps, $ }) {
+    const email = steps.trigger.context.new_customer_email;
```
dylburger, line 313: `steps.trigger.event.body` (`context` can't be modified by users)

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+    // Retrieve the past recorded emails from other runs
+    const emails = this.db.get('emails') || [];
+
+    // If the current email being passed from our webhook is already in our list, exit early
+    if(emails.includes(email)) {
+      $.exit('Already welcomed this user');
```
dylburger, line 319: `$.flow.exit`

docs/docs/code/nodejs/README.md
```diff
@@ -200,7 +206,131 @@ You can return HTTP responses from [HTTP-triggered workflows](/workflows/steps/t
...
+
+    // Add the current email to the list of past emails so we can detect it in the future runs
+    this.db.set('emails', [...emails, email]);
+  },
+})
+```
+
+### `$.service.db` limitations
+
+The `$.service.db` is only currently available in Node.js code steps. It is not yet available in other languages like Go, bash or Python.
```
dylburger, line 330: Capitalize bash

docs/docs/code/nodejs/README.md
```diff
@@ -211,33 +341,45 @@ Sometimes you want to end your workflow early, or otherwise stop or cancel the e
 - You only want to run your workflow for users in the United States. If you receive a request from outside the U.S., you don't want the rest of the code in your workflow to run.
 - You may use the `user_id` contained in the event to look up information in an external API. If you can't find data in the API tied to that user, you don't want to proceed.
 
-**In any code step, calling the `$end()` function will end the execution of the workflow immediately.** No remaining code in that step, and no code or destination steps below, will run for the current event.
+**In any code step, calling the `$.end()` function will end the execution of the workflow immediately.** No remaining code in that step, and no code or destination steps below, will run for the current event.
```
dylburger, line 344: Need to modify this section to use `$.flow.exit` instead of `$.end`

docs/docs/code/nodejs/README.md
```diff
@@ -211,33 +341,45 @@ Sometimes you want to end your workflow early, or otherwise stop or cancel the e
...
-$end("Event doesn't have the correct schema");
+defineComponent({
+  async run({ steps, $ }) {
+    $.end("Event doesn't have the correct schema");
+  })
+});
 ```
 
-This message will appear in the Inspector in the **Messages** column for the event where `$end()` was called:
+This message will appear in the Inspector in the **Messages** column for the event where `$.end()` was called:
```
dylburger, line 365: You can remove this section / screenshot since we don't actually write the end message anywhere yet in Bob

docs/docs/code/nodejs/auth/README.md
```diff
@@ -6,58 +6,66 @@ This document applies only to [Node.js workflow code steps](/workflows/steps/cod
 
 When you use [prebuilt actions](/components/actions/) tied to apps, you don't need to write the code to authorize API requests. Just [connect your account](/connected-accounts/#connecting-accounts) for that app and run your workflow.
 
-But sometimes you'll need to [write your own code](/workflows/steps/code/). You can also connect apps to custom code steps, using the auth information to authorize requests to that app.
+But sometimes you'll need to [write your own code](/code/nodejs). You can also connect apps to custom code steps, using the auth information to authorize requests to that app.
 
-For example, you may want to send a Slack message from a step. We use Slack's OAuth integration to authorize requests to Slack, so you can connect your account to this step and send a message using the access token Pipedream generates for you at `auths.slack.oauth_access_token`:
+For example, you may want to send a Slack message from a step. We use Slack's OAuth integration to authorize requests to Slack, so you can connect your account to this step and send a message using the access token Pipedream generates for you at `$.auths.slack.oauth_access_token`:
```
dylburger, line 11: Like `$.service.db`, auth data is exposed to components by adding the associated app prop (just like in normal components). So you'll want to rewrite these sections accordingly.

Then you can reference auth data via `this.slack.oauth_access_token` (for example), not `$.auths`.

docs/docs/code/nodejs/working-with-files/README.md
```diff
@@ -20,9 +20,13 @@ Use the [`fs` module](https://nodejs.org/api/fs.html) to write data to `/tmp`:
 import fs from "fs"
 import { file } from 'tmp-promise'
 
-const { path, cleanup } = await file();
-await fs.promises.appendFile(path, Buffer.from("hello, world"))
-await cleanup();
```
dylburger, line 25: These modified examples look great. I didn't see where you linked to this file in `config.js` so I think it's still not accessible.

docs/docs/connected-accounts/README.md
```diff
@@ -7,11 +7,15 @@ For example, you can connect to Slack from Pipedream (via their OAuth integratio
 ```javascript
 import { WebClient } from '@slack/web-api';
 
-const web = new WebClient(auths.slack.oauth_access_token)
-return await web.chat.postMessage({
-  text: "Hello, world!",
-  channel: "#general",
-})
+defineComponent({
+  async run({ steps, $ }) {
```
dylburger, line 11: Need to rewrite this to use the `slack` app prop (see comments on the auth doc above)

docs/docs/.vuepress/config.js
ctrlaltdylan: Yes, all of these sections are added to Node JS docs:

* HTTP requests - `/docs-v2/code/nodejs/#making-http-requests-from-your-workflow`
* State -` /docs-v2/code/nodejs/#managing-state`
* Code - `/docs-v2/code/nodejs
* Params - `/code/nodejs/#passing-parameters-to-code-steps` (this particular section links to a more enriched section in `/workflows/steps/#passing-data-to-steps-step-parameters`

docs/docs/code/nodejs/README.md
ctrlaltdylan: Ugh of course. On it.

docs/docs/code/nodejs/README.md
ctrlaltdylan: Fixed all 3 occurrences

docs/docs/code/nodejs/README.md
```diff
@@ -65,9 +63,11 @@ This communicates a couple of key concepts:
 If you're using [step parameters](/workflows/steps/#passing-data-to-steps-step-parameters) or [connect an account to a step](/connected-accounts/#from-a-code-step), you may notice two new parameters passed to the function signature, `params` and `auths`:
```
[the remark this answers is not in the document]
ctrlaltdylan: Link clean up and params -> props migrated

docs/docs/code/nodejs/README.md
ctrlaltdylan: Params migrated to props 👍

docs/docs/code/nodejs/README.md
ctrlaltdylan: 👍

docs/docs/code/nodejs/README.md
ctrlaltdylan: 👍

docs/docs/code/nodejs/README.md
ctrlaltdylan: Nice, I like it.

docs/docs/code/nodejs/README.md
ctrlaltdylan: Whoops good catch.

docs/docs/code/nodejs/README.md
ctrlaltdylan: 👍

docs/docs/code/nodejs/README.md
ctrlaltdylan: It's always muddy to me when possession happens, I was told before it doesn't apply to nouns that aren't people. But I agree it makes it more clear we're talking about a single workflow.

docs/docs/code/nodejs/README.md
ctrlaltdylan: 👍

docs/docs/code/nodejs/README.md
ctrlaltdylan: Whoops, I didn't know that. Is there a reference doc I'm missing that outlines this about the trigger params from timers/http?

docs/docs/code/nodejs/README.md
ctrlaltdylan: Capitalized all bash mentions

docs/docs/code/nodejs/README.md
ctrlaltdylan: Are there other `$.flow` options and do they need documented?

docs/docs/code/nodejs/auth/README.md
ctrlaltdylan: Can you double check that?

I've tried in my workflow and API keys are only available under `this.slack.$auths`.

```js
  props: {
    slack: {
      type: 'app',
      app: 'slack'
    }
  },
  async run({ steps, $ }) {
    console.log(this.slack.oauth_access_token);
  }
});

```

docs/docs/code/nodejs/README.md
ctrlaltdylan: C'mon Lambda upgrade already.

docs/docs/code/bash/README.md
dylburger: @ctrlaltdylan did you see the rest of these comments in this file?

docs/docs/code/README.md
dylburger: @ctrlaltdylan Did you see the rest of these comments in this file? All small

docs/docs/limits/README.md
dylburger: @ctrlaltdylan Did you see the comments in this file?

docs/docs/code/nodejs/auth/README.md
ctrlaltdylan: `this.appName.$auths` is the way currently, closing out 👍

docs/docs/code/nodejs/README.md
ctrlaltdylan: https://pipedream.com/docs-v2/workflows/events/#event-format

docs/docs/code/bash/README.md
ctrlaltdylan: I was trying to use more plain language, but ended up making it wordy :(

docs/docs/code/bash/README.md
ctrlaltdylan: Good call.

docs/docs/code/bash/README.md
ctrlaltdylan: Very helpful, thanks

docs/docs/code/python/README.md
ctrlaltdylan: Actually, to make it closer to the real formatting of the option in the dashboard, I think it makes the most sense to say `python`

docs/docs/.vuepress/theme/components/NavLinks.vue
ctrlaltdylan: I don't think the ROI is worth it. We're going to need to upgrade the whole layout and instead of incrementally hacking on my hacks.

The theme shows it's age and would really like to use a Tailwind powered theme instead of hacking styles within components.

dylburger: Totally fine

PR #2194, opened 2022-02-07
Update README.md

brianmak3:
The function has an extra second last bracket that  makes it throw a syntax error.


dannyroosevelt:
@ctrlaltdylan tagging you on this

ctrlaltdylan:
@dannyroosevelt thanks, and thanks for reporting the issue @brianmak3 .

It seems like these examples are v2, which are not supposed to be in v1. Perhaps there was a mismerge or @brianmak3 did you mean to merge into our v2 docs?

brianmak3:
Yeah it is an error in v2
Tried to follow during testing and it threw a syntax error

PR #2207, opened 2022-02-08
Adding /docs/v1 pathing

ctrlaltdylan:
Adding `/docs/v1/` as a compatible path.

https://docs.google.com/document/d/1ZiBDL0qug-5TBMqNS328d7TPgDf3J6uLONAMxfr8BR0/edit#


docs/docs/.vuepress/config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   title: "",
   head: [["link", { rel: "icon", href: "/favicon.ico" }]],
   description: "Pipedream Documentation - Connect APIs, remarkably fast",
-  base: "/docs/",
+  base: "/docs/v1/",
```
dylburger, line 8: Remind me - how will this be deployed? If we merge this into `master`, it will clobber v2, so are we going to follow a strategy like:

1. Create a branch called `docs-v${n}-pathing` when we cut a new version
2. Point `/docs/v${n}` to the Vercel deployment tied to that branch from the `pipedream-proxy`?
3. That branch remains in the repo in its state (that way we can always backport fixes to that branch).

Or are you envisioning something different?

ctrlaltdylan: Correct, I left a couple comments about that in the internal repo where I thought the routing should be explict:

https://github.com/PipedreamHQ/pipedreamin/pull/2993/files

That was the strategy I was following, the downside is whatever version is latest, we have to update the `docs-v{$n}-pathing` in tandem too. 

So 2 mains in the repo, the first points to `/docs` and the second points to `/v${latest-version}`, but if you know of a better way I'm all ears.

dylburger: Awesome. I should have reviewed all the PRs before I provided feedback! That works and makes sense.

dylburger (approved):
Code looks good. I think you can close the PR and the Vercel host will still point to this branch (I could be forgetting... do whatever makes the most sense here)

ctrlaltdylan:
I guess we never truly deploy this version since it hosts `/docs/v1` via this specific branch. Live on branch.

PR #2208, opened 2022-02-08
Docs canonical pathing plugin

ctrlaltdylan:
Adds a plugin that will automatically ignore the versioning  of `/v2/` and instruct google the preferred URL is the `/docs/` path.


docs/vercel.json
```diff
@@ -1,11 +1,11 @@
 {
   "rewrites": [
     {
-      "source": "/docs-v2",
```
dylburger, line 4: Just since we're linking to it in a few places already in the wild, can you keep a rule for `/docs-v2` and do a permanent redirect to `/docs/v2`? https://vercel.com/support/articles/does-vercel-support-permanent-redirects

Then we'll want to keep the `pipedream-proxy` entry for `/docs-v2` (in addition to adding `/docs/v2`).

Since it's a permanent redirect, I have a feeling we can remove that code after some time and Google will know how to handle `/docs-v2` (I am not sure how that works).

dylburger: This comment is technically a duplicate of the one I put in the `pipedreamin` repo. As long as we handle `/docs-v2` in one place, that is sufficient.

dylburger (approved):
I have no strong opinion on whether we handle the old `/docs-v2` redirect here or in the `pipedream-proxy` code. That is the only major comment I had.

tjk (dismissed):
lgtm

dylburger:
@ctrlaltdylan This one was merged into master separately, right? Can it be closed?

ctrlaltdylan:
@dylburger that's right, just a bit of cruft to clean up. Thanks!

PR #2307, opened 2022-02-24
Fix $.flow.end is not a function

nauxliu:


ctrlaltdylan:
Hi @nauxliu thanks for the contribution and catch of my typo!

Issue #1464, opened 2021-07-27
[ACTION] Firebase/Firestore new actions


michelle0927:
**Is there a specific app this action is for?**
Firebase Admin SDK

**Please provide a link to the relevant API docs for the specific service / operation.**
https://firebase.google.com/docs/reference/admin

**Create a Document** | Creates a new document
**List Documents** | Lists documents in a collection
**Update a Document** | Updates or inserts a document
**Create or Replace Firebase Realtime Database Record** | Creates or replaces a child object within your Firebase Realtime Database

dannyroosevelt:
@michelle0927 this is ready for release! 

>Added via [Pipedream](https://pipedream.com/@/p_8rCkMkO?o=a&a=github)

PR #2311, opened 2022-02-24
Firebase/Firestore new actions

michelle0927:
**Create a Document** | Creates a new document
**List Documents** | Lists documents in a collection
**Update a Document** | Updates a document
**Create or Replace Firebase Realtime Database Record** | Creates or replaces a child object within your Firebase Realtime Database

Addresses issue https://github.com/PipedreamHQ/pipedream/issues/1464


dylburger:

components/firebase_admin_sdk/actions/create-document/create-document.mjs
```diff
@@ -0,0 +1,38 @@
...
+      const firebase = this.firebase.getApp();
+      const collection = await firebase.firestore().collection(this.collection);
+      const doc = await collection.add(this.data);
+      $.export("$summary", `Successfully added document ${doc._path.segments[1]}`);
+      return doc;
+    } catch (err) {
+      throw new Error(err);
+    } finally {
+      this.firebase.deleteApp();
+    }
```
jcortes, line 36: With the suggestion I made here. You can now do something like
```suggestion
    const response =
      await this.firebase.createDocument({
        collection: this.collection,
        data: this.data,
      });
    $.export("$summary", `Successfully added document ${response._path.segments[1]}`);
    return response;
```

components/firebase_admin_sdk/actions/create-realtime-db-record/create-realtime-db-record.mjs
```diff
@@ -0,0 +1,39 @@
...
+      const record = this.firebase.getApp().database()
+        .ref(this.path);
+      await record.update(this.data);
+      $.export("$summary", "Successfully created record");
+    } catch (err) {
+      throw new Error(err);
+    } finally {
+      this.firebase.deleteApp();
+      await new Promise(resolve => setTimeout(resolve, 5000));
+    }
```
jcortes, line 37: With the suggestion I made here. You can now do something like
```suggestion
    const response =
      await this.firebase.createRealtimeDBRecord({
        path: this.path,
        data: this.data,
      });
    $.export("$summary", "Successfully created record");
    return response;
```

jcortes (changes requested):
Hi @michelle0927 I made a few suggestions as request changes I didn't test them out but I hope you can and let me know if that works, other than than I think it looks pretty good to me.

michelle0927:
@jcortes I implemented your suggestions for the most part. I was having issues trying to pass the promises without executing them, so instead of `appHandler`, I created a common `async run` function that initializes and deletes the app.

andrewjschuang (dismissed):
Thanks for these actions, @michelle0927!
I've added some comments below.

I was looking at the docs and it seems that there is support for [listening for real-time changes](https://firebase.google.com/docs/firestore/query-data/listen) for `collections`, `docs`, and `queries`. Maybe we can add issues for updating the current sources?

components/firebase_admin_sdk/firebase_admin_sdk.app.mjs
```diff
@@ -1,6 +1,7 @@
 import admin from "firebase-admin";
 import { axios } from "@pipedream/platform";
 import googleAuth from "google-auth-library";
+import { Firestore } from "@google-cloud/firestore";
```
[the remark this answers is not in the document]
michelle0927: This is needed due to the way we install packages with `optionalDependencies`. See Slack https://pipedream-users.slack.com/archives/CPTJYRY5A/p1646765861830569

components/firebase_admin_sdk/actions/create-realtime-db-record/create-realtime-db-record.mjs
```diff
@@ -0,0 +1,38 @@
...
+  methods: {
+    async getResponse() {
+      return this.firebase.createRealtimeDBRecord(this.path, this.data);
+    },
+    emitSummary($) {
+      $.export("$summary", "Successfully created record");
+    },
+    async deleteApp() {
+      this.firebase.deleteApp();
+      await new Promise((resolve) => setTimeout(resolve, 5000));
```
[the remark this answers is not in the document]
michelle0927: This is the only action that accesses the Firebase Realtime Database instead of the Firestore Database and uses `admin.app().database()`. The others finish without errors, but without the `sleep` here, this will display a message saying the code was still running when it stopped.

components/firebase_admin_sdk/actions/common/base.mjs
```diff
@@ -0,0 +1,22 @@
+import firebase from "../../firebase_admin_sdk.app.mjs";
+
+export default {
+  props: {
+    firebase,
+  },
+  methods: {
+    async deleteApp() {
```
vellames-turing, line 8: I am wondering if we can remove this function here and call `this.firebase.deleteApp();` directly. Looks like `deleteApp` just call another function

vellames-turing (commented):
Just one question :)

components/firebase_admin_sdk/actions/create-realtime-db-record/create-realtime-db-record.mjs
```diff
@@ -35,4 +35,14 @@ export default {
...
     },
   },
+  async run({ $ }) {
+    try {
+      await this.firebase.initializeApp();
+      const response = await this.getResponse();
+      this.emitSummary($);
+      return response;
+    } finally {
+      await this.deleteApp();
```
vellames-turing, line 45: Maybe this is supposed to be `this.firebase.deleteApp` since we deleted the `deleteApp` from `app`?

vellames-turing: Oh, ok, my bad

vellames-turing (approved):

DilanAthukorala:
**create_realtime_db_record**

@michelle0927  This keeps loading without giving output

image

michelle0927:
> **create_realtime_db_record**
> 
> @michelle0927 This keeps loading without giving output
> 

@DilanAthukorala I'm having trouble reproducing this. Do you know if the `data` is actually added to the database before it freezes?

DilanAthukorala:
> > **create_realtime_db_record**
> > @michelle0927 This keeps loading without giving output
> 
> @DilanAthukorala I'm having trouble reproducing this. Do you know if the `data` is actually added to the database before it freezes?

@michelle0927 I re-checked and this was only observed when there was no Realtime DB. We can ignore it since it's an edge case. Thanks.

PR #2330, opened 2022-02-26
Git hooks

dylburger:


.github/workflows/pull-request-checks.yaml
```diff
@@ -11,14 +11,84 @@ on:
...
+      name: Check git diff for version changes
+      run: |-
+        commit_contained_components_that_did_not_modify_version=0
+        for f in ${{ steps.changed_files.outputs.all }}
+        do
+          ext="${f##*.}"
+          # Only run this check on modified sources or actions, excluding common files
+          if ([[ "$ext" == "js" ]] || [[ "$ext" == "mjs" ]]) && \
+             ([[ "${f}" == *components/**/sources/* ]] || [[ "${f}" == *components/**/actions/* ]]) && \
+             [[ "${f}" != *common*.*js ]] && [[ "${f}" != **/actions/common/* ]] && [[ "${f}" != **/sources/common/* ]]
```
js07, line 60: I think this condition to include only component files should work for almost every file.
**Minor issue**:
There are a small number of **non-component** files in the repo that pass this condition, which would cause the check to fail. For example, these files pass the `if` condition, but don't contain `version:` (in the diff):
- [components/datadog/sources/payload-format.js](https://github.com/PipedreamHQ/pipedream/blob/f38ef1994a174fe51c3df4c6c3971a4357de9395/components/sendgrid/sources/events/engagement-event-types.js)
- [components/sendgrid/sources/events/delivery-event-types.js](https://github.com/PipedreamHQ/pipedream/blob/f38ef1994a174fe51c3df4c6c3971a4357de9395/components/sendgrid/sources/events/delivery-event-types.js)

**Potential fix**:
Add a condition to the `if` statement which checks if the file contains "version:", so that files without "version:" are excluded from the check. E.g.,
```sh
             && [[ "$(<${f})" == *"version:"* ]]
```
Drawback: Component files missing a "version:" will be skipped. (I think this should rarely be an issue because components require a `version` property to be published via the CLI.)

**Other potential fixes**:
- Require, as a guideline, that new/updated non-component JS files, in the `source` or `action` folders, be nested in the `common` folder
- Improve component detection (the `if` condition) by checking features/properties of the file (e.g., an object as `default export`, inclusion of `key`, `type`, `version` properties, etc.)
- For now, ignore failed checks for non-component files if they occur, since they should be uncommon

.github/workflows/pull-request-checks.yaml
```diff
@@ -11,14 +11,84 @@ on:
...
+          then
+            export BASE_COMMIT=${{steps.base_commit.outputs.result}}
+            export HEAD_COMMIT=${{steps.head_commit.outputs.result}}
```
js07, line 63: Is there a reason these variables are `export`ed?

.github/workflows/pull-request-checks.yaml
```diff
@@ -11,14 +11,84 @@ on:
...
+      uses: actions/github-script@v6
+      with:
+        script: return context.payload.pull_request?.base?.sha
+        result-encoding: string
+    - name: Return HEAD commit
+      id: head_commit
+      uses: actions/github-script@v6
+      with:
+        script: return context.payload.pull_request?.head?.sha
+        result-encoding: string
```
js07, line 46: **-Minor/Comment-**
I think we could get the base_commit and head_commit from the `github` [context](https://docs.github.com/en/actions/learn-github-actions/contexts) using an [expression](https://docs.github.com/en/actions/learn-github-actions/expressions) instead.

E.g., using the [`env` context](https://docs.github.com/en/actions/learn-github-actions/contexts#env-context):
```sh
    - id: git_diff_on_components
      name: Check git diff for version changes
      env:
        head_commit: ${{ github.event.pull_request.head.sha }}
        base_commit: ${{ github.event.pull_request.base.sha }}
      run: |-
        commit_contained_components_that_did_not_modify_version=0
        for f in ${{ steps.changed_files.outputs.all }}
        do
          ext="${f##*.}"
          # Only run this check on modified sources or actions, excluding common files
          if ([[ "$ext" == "js" ]] || [[ "$ext" == "mjs" ]]) && \
             ([[ "${f}" == *components/**/sources/* ]] || [[ "${f}" == *components/**/actions/* ]]) && \
             [[ "${f}" != *common*.*js ]] && [[ "${f}" != **/actions/common/* ]] && [[ "${f}" != **/sources/common/* ]] && \
             [[ "$(<${f})" == *"version:"* ]]
          then
            BASE_COMMIT=${{env.base_commit}}
            HEAD_COMMIT=${{env.head_commit}}
```

dylburger:
Great feedback @js07 ! Fixed.

PR #2356, opened 2022-03-02
Sent via PD prototype

dannyroosevelt:
The goal is to productize this functionality and extend to the below actions:

Slack app:
- send-block-kit-message
- send-custom-message
- send-direct-message
- send-group-message
- send-message-private-channel
- send-message-public-channel

Discord Webhook app:
- send-message-advanced
- send-message-with-file
- send-message


dylburger:

components/discord_webhook/actions/send-message-advanced/send-message-advanced.js
```diff
@@ -39,9 +39,28 @@ module.exports = {
...
+      default: true,
+      label: "Include link to workflow",
+      description: "Defaults to `true`, includes a link to the workflow at the end of your Discord message.",
+    },
   },
   async run() {
-    const content = this.message;
+    let content = this.message;
+
+    let link = `https://pipedream.com/@/${process.env.PIPEDREAM_WORKFLOW_ID}`;
```
dannyroosevelt, line 53: these are new env vars that Giao added yesteday, to populate the workflow URL and the line below will append the specific event ID, so you can go directly to the specific event that triggered the message

components/slack/actions/send-message-public-channel/send-message-public-channel.js
```diff
@@ -47,11 +47,48 @@ module.exports = {
...
+        ],
+      };
+
+    if (this.include_sent_via_pipedream_flag == true) {
+      this.blocks = [
+        {
+          "type": "section",
+          "text": {
+            "type": "mrkdwn",
+            "text": this.text,
```
dannyroosevelt, line 81: this chokes when I send an object via the text prop (only when `include_sent_via_pipedream_flag` == `true`)

components/discord_webhook/actions/send-message-advanced/send-message-advanced.js
dannyroosevelt: oh but actually I realized right now these env vars are only exposed to the alpha environment, which you don't have access to (yet). so don't be alarmed when those come back as `undefined`.

components/discord_webhook/actions/send-message-advanced/send-message-advanced.js
```diff
@@ -1,58 +1,49 @@
...
+    let { embeds } = this;
 
     if (!content && !embeds) {
       throw new Error("This action requires at least 1 message OR embeds object. Please enter one or the other above.");
     }
 
+    if (includeSendMessageViaPipedreamFlag) {
+      embeds = [
+        ...(embeds ?? []),
+        this.makeSentViaPipedreamEmbed(),
```
js07, line 43: This adds an [embedded object ](https://discord.com/developers/docs/resources/channel#embed-object), containing the "sent via Pipedream" text in its `description`, to the message's array of [`embeds`](https://birdie0.github.io/discord-webhooks-guide/discord_webhook.html).
With the embed, the Discord message looks like:
discord-sent-via

For comparison, this message is from a previous commit, and has "sent via Pipedream" appended to the message content:
discord-sent-via-text

dannyroosevelt: Nice, thanks for putting that together! I think I like the original version better, but I'll run it by Tod tomorrow to get his take and let you know.

dannyroosevelt:
@js07 other than the pending design question above, are you all set on this?

js07:
I converted the Slack and Discord Webhook actions from CJS to ES Modules and increased the number of channels loaded in the **Channel** prop's drop down menu.

components/discord_webhook/actions/send-message-advanced/send-message-advanced.js
dannyroosevelt: Ok let's go w/ the original option of 2 line breaks (but no italics). I think you have this now as `getSentViaPipedreamText`, right? Let's go with that but just omit the surrounding `*...*`

components/slack/slack.app.mjs
```diff
@@ -404,7 +404,7 @@ export default {
       const params = {
         types,
         cursor,
-        limit: 10,
+        limit: 200,
```
dannyroosevelt, line 407: @js07 can we paginate through all channels and return the whole list, so the user never has to click "load more"? and to confirm, we'll probably have to rev the version for any action that uses the `channel` prop, right? (and the trigger too?) cc @dylburger

dylburger: Does Slack have a channel search API? Then we could have the prop use that instead of returning all channels and doing search client-side

components/discord_webhook/actions/send-message-advanced/send-message-advanced.js
js07: I changed the "Sent via pipedream" text back to the original option without italics.

A minor comment/issue with adding the "Sent via pipedream" text to the message content - the text appears above custom embeds.
For example:
send-via-pipedream-w-embed


With this **Embeds** prop in **Send Message Advanced**
embeds

dannyroosevelt: Ah good catch --  I hadn't tested sending text **and** an embed. That isn't ideal but I think it's acceptable.

components/slack/slack.app.mjs
js07: It looks like Slack has a [channel search API](https://api.slack.com/methods/admin.conversations.search#markdown), but it's limited Enterprise Grid organizations (and requires the `admin.conversations:read` scope).

dannyroosevelt: Ok so that's probably not a viable route right now then

dannyroosevelt:
@js07 FYI `process.env.PIPEDREAM_WORKFLOW_ID` and `process.env.PIPEDREAM_TRACE_ID` are out in prod now, so you are not blocked on those anymore.

PR #2394, opened 2022-03-09
Exports messages that were not delivered

### LINKED ISSUE #201: Discord action: if we exceed the API rate limit, action should return messages we failed to send
It might make sense to separate the streams of output, for example in { stdout, stderr }.

andrewjschuang:
**Enhancement**
Messages that failed to be delivered are exported through the variable name `unsent`.

Resolves #201

### DIFF
--- a/components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
+++ b/components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
@@ -37,6 +37,15 @@ export default {
       throw new Error("This action requires at least 1 message OR embeds object. Please enter one or the other above.");
     }
 
+    const params = {
+      avatarURL,
+      embeds,
+      message,
+      threadID,
+      username,
+      includeSentViaPipedream,
+    };
+
     let content = message ?? "";
 
     if (includeSentViaPipedream) {
@@ -46,21 +55,16 @@ export default {
       content += `\n\n${this.getSentViaPipedreamText()}`;
     }
 
-    const params = {
-      avatarURL,
-      embeds,
-      content,
-      threadID,
-      username,
-    };
-
     try {
       // No interesting data is returned from Discord
-      await this.discordWebhook.sendMessage(params);
+      await this.discordWebhook.sendMessage({
+        ...params,
+        content,
+      });
       $.export("$summary", "Message sent successfully");
     } catch (err) {
-      $.export("$summary", `${err}`);
       $.export("unsent", params);
+      throw err;
     }
   },
 };
--- a/components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
+++ b/components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
@@ -5,7 +5,7 @@ export default {
   key: "discord_webhook-send-message-advanced",
   name: "Send Message (Advanced)",
   description: "Send a simple or structured message (using embeds) to a Discord channel",
-  version: "0.2.1",
+  version: "0.3.0",
   type: "action",
   props: {
     ...common.props,
@@ -23,7 +23,7 @@ export default {
       ],
     },
   },
-  async run() {
+  async run({ $ }) {
     const {
       message,
       avatarURL,
@@ -46,13 +46,21 @@ export default {
       content += `\n\n${this.getSentViaPipedreamText()}`;
     }
 
-    // No interesting data is returned from Discord
-    await this.discordWebhook.sendMessage({
+    const params = {
       avatarURL,
       embeds,
       content,
       threadID,
       username,
-    });
+    };
+
+    try {
+      // No interesting data is returned from Discord
+      await this.discordWebhook.sendMessage(params);
+      $.export("$summary", "Message sent successfully");
+    } catch (err) {
+      $.export("$summary", `${err}`);
+      $.export("unsent", params);
+    }
   },
 };
--- a/components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
+++ b/components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
@@ -37,26 +37,21 @@ export default {
       throw new Error("This action requires at least 1 message OR embeds object. Please enter one or the other above.");
     }
 
-    const params = {
-      avatarURL,
-      embeds,
-      message,
-      threadID,
-      username,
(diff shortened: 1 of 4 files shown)


dylburger:

components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
```diff
@@ -46,13 +46,21 @@ export default {
...
       username,
-    });
+    };
+
+    try {
+      // No interesting data is returned from Discord
+      await this.discordWebhook.sendMessage(params);
+      $.export("$summary", "Message sent successfully");
+    } catch (err) {
+      $.export("$summary", `${err}`);
```
js07, line 62: I think the [`$summary`](https://pipedream.com/docs/components/guidelines/#guidelines-patterns) should use plain language to summarize what happened. E.g., "Failed to send the message. See the `unsent` [export](https://pipedream.com/docs/components/api/#returning-data-from-steps) for the unsent message."?  
Typically a `$summary` isn't emitted when an action fails.

components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
```diff
@@ -46,13 +46,21 @@ export default {
...
+      $.export("unsent", params);
```
js07, line 63: Should the data for the `unsent` export be the user-configured prop values rather than the `params` object? If the `includeSentViaPipedream` prop is `true`, `content` is different from `this.message`. Emitting an object with the prop values instead would allow users to resend the message using those prop values.

components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
```diff
@@ -46,13 +46,21 @@ export default {
...
       threadID,
       username,
-    });
+    };
+
+    try {
+      // No interesting data is returned from Discord
+      await this.discordWebhook.sendMessage(params);
+      $.export("$summary", "Message sent successfully");
+    } catch (err) {
```
js07, line 61: Re-throw the caught `err` (or a [formatted error](https://github.com/PipedreamHQ/pipedream/blob/55236b3aa993cbcb545e245803d8654c6358b0a2/components/airtable/airtable.app.js#L71)) so the workflow job fails and the error gets sent to system notifications or the global error workflow?

js07 (commented):
@andrewjschuang I made a few comments that apply to each of the actions.

components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
andrewjschuang: Addressed

components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
andrewjschuang: Addressed

components/discord_webhook/actions/send-message-advanced/send-message-advanced.mjs
andrewjschuang: Addressed

components/discord_webhook/actions/send-message-with-file/send-message-with-file.mjs
```diff
@@ -58,24 +66,19 @@ export default {
...
 
-    if (includeSentViaPipedream) {
-      if (typeof content !== "string") {
-        content = JSON.stringify(content);
-      }
-      content += `\n\n${this.getSentViaPipedreamText()}`;
+    try {
+      // No interesting data is returned from Discord
+      await this.discordWebhook.sendMessageWithFile({
+        ...params,
```
js07, line 72: I think we ideally shouldn't pass the prop values (in [`params`](https://github.com/PipedreamHQ/pipedream/blob/0bf71aa27a8649d22b464be6331560fc477b9a75/components/discord_webhook/actions/send-message-with-file/send-message-with-file.mjs#L51-L59)) that aren't used by `sendMessageWithFile` to `sendMessageWithFile`, to make the code more easily understood. (Also applies to the other actions)

Idea/suggestion:
- Include only params used by `sendMessageWithFile` in `params`
- To export the [user input props](https://pipedream.com/docs/components/api/#user-input-props) on failure (as the message), create a method in `send-message-common.mjs` that gets the user input props

For example:
```js
  // `send-message-common.mjs`
  methods: {
    getUserInputProps(omit = [
      "discordWebhook",
    ]) {
      return Object.keys(this)
        .filter((k) => typeof this[k] !== "function" && !omit.includes(k))
        .reduce((props, key) => {
          props[key] = this[key];
          return props;
        }, {});
    },
```
```js
    } catch (err) {
      const unsentMessage = this.getUserInputProps();
      $.export("unsent", unsentMessage);
      throw err;
    }
```

components/discord_webhook/actions/send-message-common.mjs
```diff
@@ -39,6 +39,14 @@ export default {
     },
   },
   methods: {
+    appendPipedreamText(message) {
```
js07, line 42: I like moving the logic of this function to the common file so it can be reused 👍.

js07 (commented):
@andrewjschuang I made a comment about removing unused params passed to the `this.discordWebhook.sendMessage` methods, but otherwise it looks good to me.

components/discord_webhook/actions/send-message-with-file/send-message-with-file.mjs
andrewjschuang: Addressed

dannyroosevelt:
@js07 @andrewjschuang any thoughts on how to test this? I'm inclined to just publish since it seems pretty low risk.

andrewjschuang:
> @js07 @andrewjschuang any thoughts on how to test this? I'm inclined to just publish since it seems pretty low risk.

Yeah, I tested it by looping messages until discord threw an rate limiting error

js07:
> @js07 @andrewjschuang any thoughts on how to test this? I'm inclined to just publish since it seems pretty low risk.

I tested the sources by adding code to throw an error in the [`try` block](https://github.com/andrewjschuang/pipedream/blob/d761308ae3e09f5a33ca885d46c64fece7b9862d/components/discord_webhook/actions/send-message/send-message.mjs#L22). They all worked for me.

send-message-error

Issue #2396, opened 2022-03-10
[BUG] FraudLabsPro send SMS verification broken


ctrlaltdylan:
**Describe the bug**

When attempting to send a 2FA SMS message with FraudlabsPro's built in action, I receive an error about attempting to access an `apiKey` value on an undefined property:

![CleanShot 2022-03-10 at 09 55 25@2x](https://user-images.githubusercontent.com/2694734/157690145-64124f34-248d-46d4-9f10-e85b261c77de.png)

**To Reproduce**
1. Create a Fraudlabs pro action with a valid SMS API key from FraudLabsPro
2. Attempt to send a 2FA token
3. Receive the error

**Expected behavior**
I expect the FraudLabsPro action to send a 2FA key to the SMS phone number.

**Additional Context**

My Fraudlabs Pro account is piercepipedream.

@fraudlabspro pinging you directly for more visibility.

ctrlaltdylan:
I found the cause of the issue.

`utils.promisify` is being used to convert the Fraudlabs Pro individual API calls from a callback strategy to a Promise strategy.

However, the `promisify` has a different signature: `(err, data) => {}` than what's being passed in this implementation `params, (err, data) => {}`.

I just wrapped in Promise the old fashioned way and it's working for me:

```javascript
    async sendSmsVerification(params) {
      return new Promise((resolve, reject) => {
        this.smsVerification().sendSMS(params, (err, data) => {
          if (err) {
            reject(err);
          }
          resolve(data);
        });
      });
    },
```

fraudlabspro:
Hi [ctrlaltdylan](https://github.com/ctrlaltdylan),

Thanks for bring us the attention! Since you had made the PR, we won't touch the code. Thanks again for the code changes suggestion.

dannyroosevelt:
@ctrlaltdylan can you help verify the fix? If it looks good you or @js07 can merge.

ctrlaltdylan:
@dannyroosevelt , I actually used this PR for this week's Speed Run, so I know it works.

https://pipedream.com/blog/verify-woocommerce-orders-with-sms-codes-with-fraudlabs-pro/

@js07 if you want to switch it to the `.bind` interface and retest it's up to you. But I know the `Promise` way is working.

js07:
@dannyroosevelt @ctrlaltdylan I switched it to the `.bind` interface and retested the **Send SMS Verification** and **Get Verification Result** actions, and they both work for me.

[...]

PR #2397, opened 2022-03-10
Removing broken promisify implementation

ctrlaltdylan:
Fix for broken Fraudlabs Pro actions:

https://github.com/PipedreamHQ/pipedream/issues/2396

### DIFF
--- a/components/fraudlabs_pro/fraudlabs_pro.app.mjs
+++ b/components/fraudlabs_pro/fraudlabs_pro.app.mjs
@@ -1,4 +1,5 @@
 import fraudlabspro from "fraudlabspro-nodejs";
+import { promisify } from "util";
 
 export default {
   type: "app",
@@ -16,24 +17,14 @@ export default {
       return new fraudlabspro.SMSVerification(this.$auth.api_key);
     },
     async sendSmsVerification(params) {
-      return new Promise((resolve, reject) => {
-        this.smsVerification().sendSMS(params, (err, data) => {
-          if (err) {
-            reject(err);
-          }
-          resolve(data);
-        });
-      });
+      const smsVerification = this.smsVerification();
+      const sendSMS = promisify(smsVerification.sendSMS).bind(smsVerification);
+      return sendSMS(params);
     },
     async verifyOtp(params) {
-      return new Promise((resolve, reject) => {
-        this.smsVerification().verifyOTP(params, (err, data) => {
-          if (err) {
-            reject(data);
-          }
-          resolve(data);
-        });
-      });
+      const smsVerification = this.smsVerification();
+      const verifyOTP = promisify(smsVerification.verifyOTP).bind(smsVerification);
+      return verifyOTP(params);
     },
   },
 };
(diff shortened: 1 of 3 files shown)


dylburger:

components/fraudlabs_pro/fraudlabs_pro.app.mjs
```diff
@@ -17,12 +16,24 @@ export default {
       return new fraudlabspro.SMSVerification(this.$auth.api_key);
     },
     async sendSmsVerification(params) {
-      const promise = promisify(this.smsVerification().sendSMS);
-      return promise(params);
```
js07, line 21: I think when `promise(params)` is called, the function's (`sendSMS`'s) `this` is no longer bound to `this.smsVerification()`. So [`this.apiKey`](https://github.com/fraudlabspro/fraudlabspro-nodejs/blob/fd54077bb54d4dd2f150b3a8e1f1d517504113b4/src/fraudlabspro.js#L240) throws an error because `this` is `undefined`.

To fix the error, I think we can bind the return value from `this.smsVerification()` to the promisified `sendSMS` function before calling it. Ex:
```js
  methods: {
    smsVerification() {
      return new fraudlabspro.SMSVerification(this.$auth.api_key);
    },
    async sendSmsVerification(params) {
      const smsVerification = this.smsVerification();
      const sendSMS = promisify(smsVerification.sendSMS).bind(smsVerification);
      return sendSMS(params);
    },
    async verifyOtp(params) {
      const smsVerification = this.smsVerification();
      const verifyOTP = promisify(smsVerification.verifyOTP).bind(smsVerification);
      return verifyOTP(params);
    },
```

js07 (commented):
@ctrlaltdylan The fix works for me! I just made a comment about using `promisify` and [`.bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/bind) to slightly simplify the code.

components/fraudlabs_pro/fraudlabs_pro.app.mjs
ctrlaltdylan: @js07 Oh right, I forgot about scoping `this`. I didn't see a mention in of the `params` prop to the `promisify` function but this way works too.

js07: I changed these methods to use [`promisify()`](https://nodejs.org/dist/latest-v8.x/docs/api/util.html#util_util_promisify_original) and [`bind()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind) instead of [`new Promise`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#constructor) in https://github.com/PipedreamHQ/pipedream/pull/2397/commits/5d32c725b32ecac1a5c408a2ee9dcad04cb29e68.

js07 (approved):
From https://github.com/PipedreamHQ/pipedream/issues/2396#issuecomment-1065320018

> I switched it to the `.bind` interface and retested the **Send SMS Verification** and **Get Verification Result** actions, and they both worked for me.

Issue #1510, opened 2021-07-29
[ACTION] Strava new actions


michelle0927:
**Is there a specific app this action is for?**
Strava

**Please provide a link to the relevant API docs for the specific service / operation.**
https://developers.strava.com/docs/reference/

**Create an Activity** \| Creates an activity
**Get an Activity** \| Retrieves details of an activity by its ID
**List Activities** \| Lists your activities
**Update an Activity** \| Updates an activity by its ID
**Find Activity** \| Finds one of your activities based on the ID or Name. Useful for getting activity updates as well as kudos and segment information
**Create Totals and Stats Report** \| Creates report that includes recent (last 4 weeks), year to date, and all time stats of your activities
**Get a Stats Report** \| Retrieves your recent, year to date, and all activity stats

djs-djs:
I can assist with this issue--thanks

feyzullah:
Strava has an npm package named 'strava_api_v3'. I could not import it. Can anyone please make it available?

dannyroosevelt:
@feyzullah this is ready for release! 

>Added via [Pipedream](https://pipedream.com/@/p_8rCkMkO?o=a&a=github)

PR #2452, opened 2022-03-18
Added new actions, corrected linter errors in sources #1510

feyzullah:
#1510 
Added actions:
*create-activity
*get-activity-by-id
*get-stats
*list-activities
*update-activity
Linter errors has been fixed in sources

### DIFF
--- a/components/strava/actions/create-activity/create-activity.mjs
+++ b/components/strava/actions/create-activity/create-activity.mjs
@@ -3,7 +3,7 @@ import strava from "../../strava.app.js";
 export default {
   name: "Create Activity",
   description: "Creates a manual activity for an athlete, requires activity:write scope. See `https://developers.strava.com/docs/reference/`",
-  key: "strava_create-activity",
+  key: "strava-create-activity",
   version: "0.0.1",
   type: "action",
   props: {
--- a/components/strava/actions/create-activity/create-activity.mjs
+++ b/components/strava/actions/create-activity/create-activity.mjs
@@ -1,0 +1,88 @@
+import strava from "../../strava.app.js";
+
+export default {
+  name: "Create Activity",
+  description: "Creates a manual activity for an athlete, requires activity:write scope. See `https://developers.strava.com/docs/reference/`",
+  key: "strava_create-activity",
+  version: "0.0.12",
+  type: "action",
+  props: {
+    strava,
+    name: {
+      type: "string",
+      label: "The name of the activity",
+      description: "The name of the activity",
+    },
+    type: {
+      type: "string",
+      label: "Type of activity",
+      description: "Type of activity. For example - Run, Ride etc.",
+    },
+    start_date_local: {
+      type: "string",
+      label: "Start date of activity",
+      description: "ISO 8601 formatted date time without miliseconds e.g. `2013-10-20T19:20:30+01:00`",
+    },
+    elapsed_time: {
+      type: "integer",
+      label: "Elapsed time",
+      description: "Elapsed time in seconds",
+    },
+    description: {
+      type: "string",
+      label: "Description of the activity",
+      description: "Description of the activity",
+      optional: true,
+    },
+    distance: {
+      type: "string",
+      label: "Distance",
+      description: "Must be a float",
+      optional: true,
+    },
+    trainer: {
+      type: "integer",
+      label: "Trainer",
+      description: "Set to 1 to mark as a trainer activity",
+      optional: true,
+    },
+    commute: {
+      type: "integer",
+      label: "Commute",
+      description: "Set to 1 to mark as commute",
+      optional: true,
+    },
+    hide_from_home: {
+      type: "boolean",
+      label: "Hide from home",
+      description: "Set to true to mute activity",
+      optional: true,
+    },
+  },
+  async run({ $ }) {
+    const isoDateRegexp = /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(([+-](\d{2}):(\d{2})|Z))$/;
+    const data = {
+      name: this.name,
+      type: this.type,
+      start_date_local: this.start_date_local,
+      elapsed_time: this.elapsed_time,
+    };
+    if (this.description)
+      data.description = this.description;
+    if (this.distance)
+      data.distance = this.distance;
+    if (this.trainer)
+      data.trainer = this.trainer;
+    if (this.commute)
+      data.commute = this.commute;
+    if (this.hide_from_home)
+      data.hide_from_home = this.hide_from_home;
+    if (this.distance && isNaN(parseFloat(this.distance)))
+      throw new Error("Please provide a float for `Distance`");
+    if (!isoDateRegexp.test(this.start_date_local))
+      throw new Error("Please provide `Start date of activity` in required format");
+    const resp = await this.strava.createNewActivity($, data);
+    $.export("$summary", "Successfully added activity");
+    return resp;
+  },
+};
--- a/components/strava/actions/create-activity/create-activity.mjs
+++ b/components/strava/actions/create-activity/create-activity.mjs
@@ -2,7 +2,7 @@ import strava from "../../strava.app.js";
 
 export default {
   name: "Create Activity",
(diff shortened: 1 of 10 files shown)


dylburger:

feyzullah:
More changes will be done

components/strava/actions/create-activity/create-activity.mjs
```diff
@@ -0,0 +1,88 @@
+import strava from "../../strava.app.js";
+
+export default {
+  name: "Create Activity",
+  description: "Creates a manual activity for an athlete, requires activity:write scope. See `https://developers.strava.com/docs/reference/`",
```
dylburger, line 5: You can remove the "requires activity:write scope" section, since we've included that in our default authorization request

components/strava/actions/create-activity/create-activity.mjs
```diff
@@ -0,0 +1,88 @@
...
+    },
+    type: {
+      type: "string",
+      label: "Type of activity",
+      description: "Type of activity. For example - Run, Ride etc.",
+    },
+    start_date_local: {
+      type: "string",
+      label: "Start date of activity",
+      description: "ISO 8601 formatted date time without miliseconds e.g. `2013-10-20T19:20:30+01:00`",
```
dylburger, line 24: miliseconds -> milliseconds

components/strava/actions/create-activity/create-activity.mjs
```diff
@@ -0,0 +1,88 @@
...
+    description: {
+      type: "string",
+      label: "Description of the activity",
+      description: "Description of the activity",
+      optional: true,
+    },
+    distance: {
+      type: "string",
+      label: "Distance",
+      description: "Must be a float",
```
dylburger, line 40: Can you add an example at the end, like "e.g. `4.3`"

components/strava/actions/create-activity/create-activity.mjs
```diff
@@ -0,0 +1,88 @@
...
+  },
+  async run({ $ }) {
+    const isoDateRegexp = /^(\d{4})-(\d{2})-(\d{2})T(\d{2}):(\d{2}):(\d{2})(([+-](\d{2}):(\d{2})|Z))$/;
+    const data = {
+      name: this.name,
+      type: this.type,
+      start_date_local: this.start_date_local,
+      elapsed_time: this.elapsed_time,
+    };
+    if (this.description)
```
dylburger, line 70: You can actually combine this with the object above, since props should default to `undefined`, so including them in the object as properties won't hurt if they're not defined:

```javascript
const data = {
  name: this.name,
  type: this.type,
  start_date_local: this.start_date_local,
  elapsed_time: this.elapsed_time,
};
```

dylburger: Or even:

```
const { name, type, description } = this
const data = {
  name,
  type,
  description,
}

(and so on, for the remaining properties)

File: components/strava/actions/list-activities/list-activities.mjs [hunk 5]
```diff
@@ -0,0 +1,47 @@
...
+      description: "An epoch timestamp to use for filtering activities that have taken place before a certain time",
+      optional: true,
+    },
+    after: {
+      type: "integer",
+      label: "After",
+      description: "An epoch timestamp to use for filtering activities that have taken place after a certain time",
+      optional: true,
+    },
+    page: {
```
dylburger, line 23: We actually recommend handling pagination within the action / method, and simply fetching all records given the other criteria (e.g. `before` / `after`). This way a user doesn't have to ever think about pagination themselves - it's just handled by Pipedream.

Then, you can set an optional prop called `maxItems` that specifies the maximum number of items the action fetches. This should not have a default value (i.e. it should be `undefined`), so that the action fetches _all_ items from the API endpoint by default. Then if the user wants to list only the first N items, they can optionally set `maxItems` to N.

components/strava/actions/update-activity/update-activity.mjs
```diff
@@ -0,0 +1,32 @@
+import strava from "../../strava.app.js";
+
+export default {
+  name: "Update Activity",
+  description: "Updates the given activity that is owned by the authenticated athlete. Requires activity:write. Also requires activity:read_all in order to update Only Me activities. See `https://developers.strava.com/docs/reference/`",
```
dylburger, line 5: You can remove "Requires activity:write. Also requires activity:read_all in order to update Only Me activities"

components/strava/sources/activity-created/activity-created.js
```diff
@@ -4,13 +4,16 @@ module.exports = {
   key: "strava-activity-created",
   name: "Activity Created",
   description: "Emits an event when a new activity is created",
```
dylburger, line 6: Did you setup ESLint locally in your editor? I like to do that so I catch these errors there.

components/strava/strava.app.js
```diff
@@ -1,21 +1,36 @@
...
       opts.url = `https://www.strava.com/api/v3${
-        path[0] === "/" ? "" : "/"
+        path[0] === "/" ?
+          "" :
+          "/"
       }${path}`;
-      return await axios(opts);
+      return opts;
+    },
+    makeQueryString(params) {
```
dylburger, line 22: `axios` should accept a `params` object, from which they create a query string. So you shouldn't need this method, and you should be able to pass any query params in `params`. Let me know if that works

components/strava/strava.app.js
```diff
@@ -31,5 +46,56 @@ module.exports = {
...
+        await this._makeAPIRequestWithPlatform($, {
+          method: "GET",
+          path: `/activities${this.makeQueryString(data)}`,
+        })
+      );
+    },
+    async updateActivity($, data) {
+      const {
+        id,
+        ...dataRest
```
dylburger, line 81: Can the user pass an `id` property in the update request? i.e. could that clobber the `id` property you pass here, that you use as the activity ID?

components/strava/strava.app.js
```diff
@@ -31,5 +46,56 @@ module.exports = {
         })
       ).data;
     },
+    async createNewActivity($, data) {
+      return (
```
dylburger, line 50: I don't believe the extra parens between `return` and `await` are necessary — is that right?

components/strava/actions/list-activities/list-activities.mjs
feyzullah: In Strava's docs(https://developers.strava.com/docs/reference/#api-Activities-getLoggedInAthleteActivities) it is written that the default page size is 30. So as I understand, if we don't provide `per_page` field, the API will return first page with size 30.
In this situation, to handle this kind of paging, should we call the API with something like per_page=INFINITY?

components/strava/strava.app.js
feyzullah: I deleted the method and used params object. It works well

components/strava/strava.app.js
feyzullah: updated to activityId

components/strava/strava.app.js
feyzullah: Of course. I removed extra parentheses

components/strava/actions/update-activity/update-activity.mjs
feyzullah: Done

components/strava/sources/activity-created/activity-created.js
feyzullah: It has some warnings, I am correcting them

dylburger (approved):

components/strava/actions/list-activities/list-activities.mjs
dylburger: Sorry, I missed this. In general I do this:

1. Find out what the max page size is for the given endpoint
2. Write the app file method to fetch pages of that max size, until there are no more records. Return _all_ items across all pages to the user.
3. If the user sets the `maxItems` prop, your method should paginate until you find `maxItems` items, then return those items.

Users don't care about pagination and they explicitly don't want to handle it, so if we handle it for them and just return all results back, normally that's what they want. If they want to limit the result set, they can always set `maxItems`. And if they really want to handle pagination themselves, they can always write that code manually. But for actions, we like to abstract it.

Does that help clarify?

PR #2474, opened 2022-03-23
Add opensea events source

wong2:
the API: https://docs.opensea.io/reference/retrieving-asset-events


js07:
@wong2 Thanks for the PR! I see you're using the OpenSea API in your event source. We've created a basic integration for OpenSea here:
https://pipedream.com/apps/opensea

You can connect your OpenSea account to Pipedream [here](https://pipedream.com/accounts), and use the connected account in sources and actions instead of your `apiKey` prop. In `events.mjs`:
1. Remove the `apiKey` prop
```diff
-    apiKey: {
-      type: "string",
-      label: "API Key",
-      description: "OpenSea API key",
-    },
```
3. Import the `opensea` [app prop](https://pipedream.com/docs/components/api/#app-props)
```diff
  import { axios } from "@pipedream/platform";
+ import opensea from "../../opensea.app.mjs";
```
4. Add the opensea [app prop](https://pipedream.com/docs/components/api/#app-props) to the source's props
```diff
   props: {
+    opensea,
     timer: {
```
5. Get the API key from `this.opensea.$auth.api_key`
```diff
-    const apiKey = this.apiKey;
+    const apiKey = this.$auth.opensea.api_key;
```

dylburger:

wong2:
Thanks for the review

wong2:
Fixed!

components/opensea/sources/events/events.mjs
```diff
@@ -0,0 +1,55 @@
...
+        "listings",
+      ],
+      label: "Event Type",
+      description: "OpenSea event type",
+    },
+  },
+  dedupe: "greatest",
+  type: "source",
+  methods: {},
+  async run({ $ }) {
```
andrewjschuang, line 35: The `$` reference is exposed to actions only, while sources receive the triggered `event`.
Since `event` is not used, it may be omitted here.
See in docs: [Source Invocation](https://pipedream.com/docs/components/api/#sources)

```suggestion
  async run() {
```

components/opensea/sources/events/events.mjs
```diff
@@ -0,0 +1,55 @@
+import { axios } from "@pipedream/platform";
```
andrewjschuang, line 1: > General app logic and API requests may be located in the app.mjs file.

As seen in the comment below

```suggestion
```

components/opensea/sources/events/events.mjs
```diff
@@ -0,0 +1,55 @@
...
+import opensea from "../../opensea.app.mjs";
+
+export default {
+  name: "New OpenSea Collection Events",
+  version: "0.0.1",
+  key: "opensea-events",
```
andrewjschuang, line 7: In general, the same value as the component's `name` can be used

```suggestion
  key: "opensea-new-collection-events",
```

components/opensea/sources/events/events.mjs
```diff
@@ -0,0 +1,55 @@
...
+      options: [
+        "sales",
+        "listings",
+      ],
+      label: "Event Type",
+      description: "OpenSea event type",
+    },
+  },
+  dedupe: "greatest",
+  type: "source",
```
andrewjschuang, line 33: `type` and `dedupe` attributes may be moved up above `props` for easier visualization

components/opensea/sources/events/events.mjs
```diff
@@ -0,0 +1,55 @@
...
+      type: "string",
+      options: [
+        "sales",
+        "listings",
+      ],
+      label: "Event Type",
+      description: "OpenSea event type",
+    },
+  },
+  dedupe: "greatest",
```
andrewjschuang, line 32: The `dedupe` strategy `greatest` works well with apps that use incrementing `ids` such as [GitHub](https://pipedream.com/docs/components/quickstart/nodejs/sources/#dedupe-the-events).

I'm not sure if OpenSea `event` `ids` are incremental; I couldn't find it in the docs.
If not, since the [retrieving events endpoint](https://docs.opensea.io/reference/retrieving-asset-events) accepts the parameter `occurred_after`, in order to prevent duplicate events a `db` prop may be used to cache the `timestamp` of the last emitted `event`.

For example,

```javascript
props: {
  db: "$.service.db",
},
methods: {
  getLastTimestamp() {
    return this.db.get("lastTimestamp");
  },
  setLastTimestamp(ts) {
    this.db.set("lastTimestamp", ts);
  },
},
async run() {
  ...

  let lastTimestamp = this.getLastTimestamp();
  const resp = this.opensea.retrieveEvents({
    contract,
    eventType,
    lastTimestamp,
  });

  resp.asset_events
    .filter((event) => event.created_date > lastTimestamp)
    .forEach((event) => {
      this.$emit(event, {
        id: event.id,
        ts: +new Date(event.created_date),
      });
      lastTimestamp = event.created_date;
    });
  this.setLastTimestamp(lastTimestamp);

  ...
},
```

components/opensea/sources/events/events.mjs
```diff
@@ -0,0 +1,55 @@
...
+  type: "source",
+  methods: {},
+  async run({ $ }) {
+    const contract = this.contractAddress;
+    const eventType = this.eventType === "sales"
+      ? "successful"
+      : "created";
+    const apiKey = this.opensea.$auth.api_key;
+    const url = `https://api.opensea.io/api/v1/events?only_opensea=false&asset_contract_address=${contract}&event_type=${eventType}`;
+    const resp = await axios($, {
```
andrewjschuang, line 42: I think this API request could support `pagination` using the `cursor` parameter to be able to retrieve all new events, something like:

```javascript
let cursor = null;
do {
  const resp = this.opensea.retrieveEvents({
    ...,
    cursor,
  });
  cursor = resp.cursor;
} while (cursor);
```

andrewjschuang (changes requested):
Thanks for making this pull request, @wong2!

I've added some comments since it's the first time contributing, but overall it looks very good to me! 

You can move this card to the `Doing` column and once finished, move it back to `Ready for PR Review` and request a new review.

Let me know if there are any questions. Look forward to seeing more!

components/opensea/sources/events/events.mjs
wong2: The event ids are incremental.

components/opensea/sources/events/events.mjs
wong2: The api does support pagination, however, a simple loop will retrieve all history events, we need to set some stop condition.

wong2: OK, figured this out.

wong2:
@andrewjschuang Hi, I've fixed all suggested issues.

andrewjschuang:
Thank you for the changes, @wong2!

I have a couple of questions below:

> The api does support pagination, however, a simple loop will retrieve all history events, we need to set some stop condition.

Does the API response omit the `cursor` parameter when the last page is reached? So that when iterating the last page, `cursor === null` would break the `while` loop?

> The event ids are incremental.

Cool, **and if the above is** `true`, then we could save the `cursor` (when `not null`) to the db instead of the timestamp. Since event `ids` and `created_time` are incremental, then the event pages returned by the API should always be in the same order and with the same cursor.

For example:
```javascript
let cursor = this.getCursor(); // null on the first run and not null after that if pages > 1
do {
  const resp = await this.opensea.retrieveEvents({
    contract: this.contractAddress,
    eventType,
    cursor,
  });

  // emitEvents

  if (cursor) {
    this.setCursor(cursor);
  }
  cursor = resp.next; // will be null in the last page
} while (cursor);
```

wong2:
I think the current implementation already does the work. And i think we should not store cursor in the db, because it's a pagination detail.

andrewjschuang (approved):

dannyroosevelt:
@andrewjschuang feel free to merge

Issue #2158, opened 2022-02-01
[FEATURE] Allow email send action to reply to trigger email


dzg:
I have workflows triggered by emails to `...@pipedream.net`

It would be great if there were a way to **Reply** to that same email at the end of workflow with my results.

Currently when using `email_me` action – the sender of email is `notifications@pipedream.com` which causes the email to be 'de-threaded' (at least within Gmail) and so does not appear as a reply to the original triggering email, but rather as a new 'conversation' (even if Subject is the same.)

Thanks!

Referenced at https://pipedream.com/community/t/reply-to-email-trigger/1865/2

eric-brechemier:
I have a similar issue, using Thunderbird to send and receive SMS from Twilio as emails. I would like to reply to the email received from `notifications@pipedream.com` to my inbound pipedream email address, but the `Reply-To` header is currently set to `noreply@pipedream.com` which requires two extra steps to do so:

1. Confirm that I want to reply to the email, clicking `Reply anyway` in the dialog prompt from Thunderbird which warns me that the sender is using a `noreply` address which will ignore my answer.
2. Replace the `To` address in the editing window for the reply, removing `noreply@pipedream.com` to replace it with the inbound email address. Although I have recorded this address as a contact, this extra step is cumbersome.

I would appreciate to be able to customize the `Reply-to` address in the configuration of the `email_me` step.

eric-brechemier:
@dzg I have just checked that emails remain threaded in Gmail when replying to an email using its `Reply-To` address:
I sent an email with a `Reply-To` address different from the `From` address to a Gmail account. I replied to the email from the Gmail account. Gmail automatically used the `Reply-To` address and the reply was attached to the same thread.

In short: having the ability to customize the `Reply-To` address would also solve the issue that you described in Gmail.

PR #2488, opened 2022-03-25
Provide reply_to property to $.send.email() (#2158)

eric-brechemier:


dylburger:

dylburger:
@eric-brechemier This is awesome! But a custom `Reply-To` header isn't yet supported in our `$.send.email` service. Were you implementing a mock PR to show us the API you'd like to see for this feature?

eric-brechemier:
@dylburger Let's say it's a first step in the right direction ;) I did look for the source of `$.send.email()` but have not found it yet… Is this part of your private implementation of the platform?

I went as far as the [platform project](https://github.com/PipedreamHQ/platform/pull/29) but only found type definitions there.

dylburger:
Great thanks. Yes, `$.send.email` is part of our private, production platform. We're in the middle of some work to improve it, which covers sending email to any address (within some basic limits), setting a custom `Reply-To`, etc. We'll update that issue when we make this available.

Since the implementation of that may not use `$.send.email`, are you OK if I close this PR for now? We use this list as a part of our backlog, and since there isn't any action for us to take on this PR, I'd like to move it into a different state. If we end up making more of the code for `$.send.email` public, etc., we can always re-open in the future.

eric-brechemier:
@dylburger Thank you for the clarification. I am looking forward to seeing these improvements to the email_me task. With regards to allowing more destination email addresses, here is my feedback: I used a specific email address when I created my pipedream account and I did not expect to be restricted to that particular email address for notifications to myself. I would rather use a different email address (I have my own domain) for separate workflows to allow filtering in my email client. I would currently need to create a separate account for each of these destinations, which I was planning to do, part of a common organization. In short, the current behavior is not blocking for my use case, but would need to be clarified at sign-up: the email address that you are using is the one where you will receive email notifications.

> Since the implementation of that may not use `$.send.email`, are you OK if I close this PR for now?

Sure, as long as you keep the original issue open, feel free to close my PRs.

dylburger:
Great thanks! All of that makes sense and ideally would be clarified / covered under the new model.

PR #2564, opened 2022-04-06
Fixing vercel preview links

ctrlaltdylan:
- Empty commit to try different preset
- Adding sourcemaps
- Empty commit to try different preset
- Vercel rewrites experiment


ctrlaltdylan:
@dylburger this will fix the issue with the preview refresh links on Vercel.

Long story short, the `base: /docs/` is breaking any subpath URLs.

In production, our proxy will rewrite `/docs/*` to the root path on our Vercel deployment. However in preview mode, we don't have that.

So I added a `vercel.json` rule to fix the issue. However, we can also solve this issue by removing the `base: /docs/` config altogether too.

PR #2579, opened 2022-04-08
Update video url trackers property

mjabali:


dylburger:

components/symbl_ai/actions/post-video-url/post-video-url.mjs
```diff
@@ -122,8 +122,8 @@ export default {
           enableAllTrackers: this.enableAllTrackers,
           enableSpeakerDiarization: this.enableSpeakerDiarization,
           diarizationSpeakerCount: this.diarizationSpeakerCount,
-          trackers: JSON.parse(this.trackers),
-          channelMetadata: JSON.parse(this.channelMetadata),
+          trackers: JSON.parse(this.trackers ?? "[]"),
+          channelMetadata: JSON.parse(this.channelMetadata ?? "[]"),
```
jcortes, line 126: I think you might consider first sanitize the property like in [this example](https://github.com/PipedreamHQ/pipedream/blob/master/components/amara/actions/list-teams/list-teams.mjs#L32) and then using `||` as a fallback instead.
```suggestion
          trackers: JSON.parse(this.trackers || "[]"),
          channelMetadata: JSON.parse(this.channelMetadata || "[]"),
```
because if the user leaves an empty space character in the input box you will get an error like
```
Uncaught SyntaxError: Unexpected end of JSON input
```

jcortes (changes requested):
Hi @mjabali I've just added a minor request here. Thanks for your contributions!

components/symbl_ai/actions/post-video-url/post-video-url.mjs
mjabali: Thanks a lot @jcortes.

mjabali:
@ctrlaltdylan is there any chance you can help me getting a final review here and merging this PR?

PR #2637, opened 2022-04-18
Fix broken SES send email action

ctrlaltdylan:
Attempt to fix the bug on the SES send email action:

CleanShot 2022-04-18 at 11 20 27@2x


seidtgeist:
FYI, until this is fixed:

I’ve worked around this by setting the value as an array of strings in a code step and using its return value.

hervenivon:
@seidtgeist I'm new to Pipedream, can you share your array of Strings code step please?

andrewjschuang (changes requested):
@ctrlaltdylan the action works fine with these changes!
This just needs a rebase on top of master to fix the PR checks

PR #2647, opened 2022-04-19
Filter Update Actions

### LINKED ISSUE #2643: [Refactor] Filter actions
We should make a couple changes to the current Filter action:

1. Add an optional "Reason" prop, and set the `default` to `Condition met for exiting [or continuing] workflow`
2. Break out the current Filter action into 3 different actions and get rid of the `continue` prop:
- "Continue workflow based on a condition" (`filter-continue-workflow-condition`)
- "End workflow based on a condition" (`filter-end-workflow-condition`)
- "End workflow" (`filter-end-workflow`)

andrewjschuang:
Splits Filter into 3 actions:

1. Continue Workflow on Condition
2. End Workflow on Condition
3. End Workflow on Custom Condition

Resolves #2643


dylburger:

vellames-turing (approved):

dannyroosevelt:
@andrewjschuang I think for the binary conditions, the logic disappeared that did not render the `Value to compare against` field, right? 
image

andrewjschuang:
> @andrewjschuang I think for the binary conditions, the logic disappeared that did not render the `Value to compare against` field, right?

In your case it should be:
Value to evaluate: `url`,
Value to compare against: `{{steps.search_issues.$return_value[0]}}`

Or it can be:
Screen Shot 2022-04-25 at 12 53 08 PM

dannyroosevelt:
> > > > @andrewjschuang I think for the binary conditions, the logic disappeared that did not render the `Value to compare against` field, right?
> > > 
> > > 
> > > @dannyroosevelt, in your case it should be: Value to evaluate: `url`, Value to compare against: `{{steps.search_issues.$return_value[0]}}`
> > 
> > 
> > In the case of the `url` key not being found, I get an error in both configurations, instead of calling `$.flow.exit()`
> 
> @dannyroosevelt what is the value of `steps.search_issues.$return_value`?

In this case it's `[]`

andrewjschuang:
> In this case it's `[]`

@dannyroosevelt oh so the error occurs because `steps.search_issues.$return_value[0]` is `undefined`. It is necessary to check if `steps.search_issues.$return_value[0] !== null && steps.search_issues.$return_value[0].url !== null` as two consecutive steps

dannyroosevelt:
Shouldn't this get evaluated as `null`? 
image

andrewjschuang:
> Shouldn't this get evaluated as `null`?

It should. Just pushed a fix

dannyroosevelt (approved):

PR #2681, opened 2022-04-22
Added config action to setup variables for a workflow

jcortes:
The idea is to give support to any workflow with variables that can be set with a simple action component step that in a following step we can reference with something like
```
  async run({ steps, $ }) {
    return steps.helper_functions_export_variables.config;
  },
})
```


vellames-turing (approved):

dannyroosevelt:
@jcortes is there a reason to `$.export` config in addition to just `return`ing it?

Also, this should go in the `helper_functions` app, not `helpers`.

components/helper_functions/actions/export-variables/export-variables.mjs
```diff
@@ -0,0 +1,17 @@
+export default {
+  key: "helper_functions-export-variables",
+  name: "Export Variables",
+  description: "Export variables for use in your workflow",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    config: {
```
dannyroosevelt, line 8: can you also import `helper_functions` so the action inherits the app icon?

components/helper_functions/actions/export-variables/export-variables.mjs
```diff
@@ -0,0 +1,17 @@
+export default {
```
dannyroosevelt, line 1: `import helper_functions from "../../helper_functions.app.mjs"`

components/helper_functions/actions/export-variables/export-variables.mjs
dannyroosevelt: so that the action inherits the app icon

dannyroosevelt (changes requested):
Can you just make sure to import the app in the action props, so the action inherits the app icon? Then let's ship it!

dannyroosevelt (approved):

PR #2767, opened 2022-05-04
Adding configuration error example

ctrlaltdylan:


dannyroosevelt:
@ctrlaltdylan I noticed a few issues w/ this documentation that should be fixed:

- `throw ConfigurationError()` I think needs to be `throw new ConfigurationError()`
- Looks like you have to `import { ConfigurationError } from "@pipedream/platform";`
- Your code is checking to see if `@` is **present**, then throwing the error, but I think you want the opposite, hah
- Missing the closing `})`
- The screenshot of the ConfigurationError in the UI is pretty grainy.

ctrlaltdylan:
@dannyroosevelt revision #2 here: https://github.com/PipedreamHQ/pipedream/pull/2769

PR #2807, opened 2022-05-06
ZeroTier Action: Update Network Member

zalmanlew:


dylburger:

dylburger:

components/zerotier/zerotier.app.mjs
```diff
@@ -32,6 +32,48 @@ export default {
         }));
       },
     },
+    memberHidden: {
+      type: "boolean",
+      label: "Member Hidden?",
+      description: "Whether or not the member is hidden in the UI",
+      optional: true,
+      default: null,
```
michelle0927, line 40: Since `memberHidden` is boolean, the default should be either `true` or `false`. Another option would be to not include any default value. Then if the user doesn't enter anything, the value would be undefined.
```suggestion
      default: false,
```

components/zerotier/zerotier.app.mjs
```diff
@@ -32,6 +32,48 @@ export default {
...
+    },
+    memberName: {
+      type: "string",
+      label: "Member Name",
+      description: "User defined name of the member",
+      optional: true,
+      default: null,
```
michelle0927, line 47: Since `memberName` is a string, the default value should be a string. I think in this case, it would be best not to include a default value.
```suggestion
```

components/zerotier/zerotier.app.mjs
```diff
@@ -32,6 +32,48 @@ export default {
...
+    },
+    memberDescription: {
+      type: "string",
+      label: "Member Description",
+      description: "User defined description of the member",
+      optional: true,
+      default: null,
```
michelle0927, line 54: Since `memberDescription` is a string, the default value, if included, would need to be a string. I think it's probably best not to include a default value here.
```suggestion
```

michelle0927 (changes requested):
Thanks for the contribution! I left a few suggestions.

components/zerotier/zerotier.app.mjs
```diff
@@ -32,6 +32,48 @@ export default {
...
+      description: "Allow the member to be a bridge on the network",
+      optional: true,
+      default: null,
+    },
+    memberNoAutoAssignIps: {
+      type: "boolean",
+      label: "Do Not Auto-Assign IPs",
+      description: "Pass `TRUE` to exempt this member from the IP auto assignment pool on a Network",
+      optional: true,
+      default: null,
```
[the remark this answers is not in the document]
zalmanlew: @michelle0927 the reason default is specified as `null` is because of how the ZT API works. If you pass any key as `null` it will skip it over and not update the value for that key (this allows the user to ignore those fields they don't want to update and be assured it won't update them).

- If it's passed as a boolean of true or false it will actually update the respective value to the boolean defined.
- If it's passed as `undefined` it returns a 500 error.

As far as I can tell, the only option to send values that you don't want to update is to pass them as `null` - that's why I set it like that.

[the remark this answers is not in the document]
zalmanlew: https://docs.zerotier.com/central/v1/#operation/updateNetworkMember

zalmanlew:
@michelle0927 regarding the review - the reason default is specified as null is because of how the ZT API works. If you pass any key as null it will skip it over and not update the value for that key (this allows the user to ignore those fields they don't want to update and be assured it won't update them).

If it's passed as a boolean of true or false it will actually update the respective value to the boolean defined.
If it's passed as undefined it returns a 500 error.
As far as I can tell, the only option to send values that you don't want to update is to pass them as null - that's why I set it like that.

https://docs.zerotier.com/central/v1/#operation/updateNetworkMember

michelle0927 (approved):

DilanAthukorala:
Verification Completed

dannyroosevelt:
This is ready for release!

Issue #1527, opened 2021-08-02
[ACTION / SOURCE] Postmark new actions


michelle0927:
**Is there a specific app this action is for?**
Postmark

**Please provide a link to the relevant API docs for the specific service / operation.**
https://postmarkapp.com/developer/api/overview

### Triggers
- New message received
- New email opened

### Actions
- Send an email
- Send an email with template

GTFalcao:
@andrewjschuang @lcaresia Thank you both for your very thorough reviews. I've checked and replied to all the comments, and significantly improved the code reusability and adherence to the good practices that you've suggested.

Hassan-Abbas-SQA:
In action: Send an email > HTML body must be required to receive "New Email Opened" trigger events.
please update the hints for this scenario

GTFalcao:
To make sure the user is not caught off-guard by this [limitation documented by Postmark](https://postmarkapp.com/developer/user-guide/tracking-opens#limitations), I've updated the descriptions for the following props:

[Source: new email opened] Track opens by default
[Action: send single email] HTML Body
[Action: (both)] Track opens

@Hassan-Abbas-SQA

dannyroosevelt:
This is ready for release!

PR #2823, opened 2022-05-10
[ACTION / SOURCE] #1527 - Postmark new actions

GTFalcao:
# Sources
- **New message received** (created new source)
- **New email opened** (created new source)

# Actions
- **Send an email** (updated existing action)
- **Send an email with template** (created new action)


dylburger:

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+    async setServerInfo(params) {
+      return await axios
+        .put(
+          "https://api.postmarkapp.com/server",
+          params,
+          {
+            headers: this.getHeaders(),
+          },
+        )
+        .then(({ data }) => data);
```
lcaresia, line 205:
```suggestion
        return axios(this, {
            method: "put",
            path: "https://api.postmarkapp.com/server",
            params,
            headers: this.getHeaders(),
        })
```

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+        "custom_headers",
+        "track_opens",
+        "track_links",
+        "attachments",
+        "metadata",
+        "message_stream",
+      ];
+    },
+    async sharedActionRequest($, action, endpoint, uniqueProps) {
+      return await axiosPipedream($, {
```
lcaresia, line 166:
```suggestion
      return await axios($, {
```

components/postmark/sources/new-email-opened/new-email-opened.mjs
```diff
@@ -0,0 +1,51 @@
...
+        PostFirstOpenOnly: this.postFirstOpenOnly,
+        TrackOpens: this.trackOpensByDefault,
+      });
+    },
+    async deactivate() {
+      return this.postmark.setServerInfo({
+        OpenHookUrl: "",
+      });
+    },
+  },
```
lcaresia, line 43: You can move all common things of sources to a common file.

Example: https://github.com/PipedreamHQ/pipedream/blob/master/components/asana/sources/common/common.mjs

components/postmark/sources/new-email-opened/new-email-opened.mjs
```diff
@@ -0,0 +1,51 @@
...
+  async run(event) {
+    this.http.respond({
+      status: 200,
+    });
+
+    this.$emit(event);
```
lcaresia, line 49: You need emit a event with a `id`, `summary` and `ts`.

Example: https://github.com/PipedreamHQ/pipedream/blob/aedf057dde1d963fd39fa7206c14314cc937602f/components/asana/sources/new-story/new-story.mjs#L49-L53

components/postmark/sources/new-inbound-email-received/new-inbound-email-received.mjs
```diff
@@ -0,0 +1,36 @@
...
+      return this.postmark.setServerInfo({
+        InboundHookUrl: this.http.endpoint,
+      });
+    },
+    async deactivate() {
+      return this.postmark.setServerInfo({
+        InboundHookUrl: "",
+      });
+    },
+  },
```
lcaresia, line 28: You can move all common things of sources to a common file.

Example: https://github.com/PipedreamHQ/pipedream/blob/master/components/asana/sources/common/common.mjs

components/postmark/sources/new-inbound-email-received/new-inbound-email-received.mjs
```diff
@@ -0,0 +1,36 @@
...
+  async run(event) {
+    this.http.respond({
+      status: 200,
+    });
+
+    this.$emit(event);
```
lcaresia, line 34: You need emit a event with a `id`, `summary` and `ts`.

Example: https://github.com/PipedreamHQ/pipedream/blob/aedf057dde1d963fd39fa7206c14314cc937602f/components/asana/sources/new-story/new-story.mjs#L49-L53

lcaresia (dismissed):

components/postmark/actions/send-single-email/send-single-email.mjs
```diff
@@ -1,92 +1,50 @@
...
+  };
+});
 
 export default {
   key: "postmark-send-single-email",
-  name: "Send an Email with Postmark to a Single Recipient",
-  description: "Send an Email with Postmark to a Single Recipient",
-  version: "0.1.1",
+  name: "Send an email",
+  description: "Send an email with Postmark",
```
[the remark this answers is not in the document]
andrewjschuang: I think the `name` can be `Send Email` and the description is fine according to the [docs](https://pipedream.com/docs/components/guidelines/#action-name)

```suggestion
  name: "Send Email",
  description: "Send an email to a Single Recipient",
```

components/postmark/actions/send-single-email/send-single-email.mjs
```diff
@@ -1,92 +1,50 @@
...
+
+let objSharedProps = {};
+postmark.methods.listSharedProps().forEach((propName) => {
+  objSharedProps[propName] = {
+    propDefinition: [
+      postmark,
+      propName,
+    ],
+  };
+});
```
andrewjschuang, line 12: Good thinking in reusing common props!
There is a pattern [described in the docs](https://pipedream.com/docs/components/guidelines/#common-files-optional) that we use to achieve this though. You can take a look at [Google Drive](https://github.com/PipedreamHQ/pipedream/tree/master/components/google_drive) and [AWS](https://github.com/PipedreamHQ/pipedream/tree/master/components/aws) components as examples

components/postmark/actions/send-single-email/send-single-email.mjs
```diff
@@ -1,92 +1,50 @@
 // legacy_hash_id: a_8KiGrJ
```
andrewjschuang, line 1: This line can be removed

```suggestion
```

components/postmark/actions/send-single-email/send-single-email.mjs
```diff
@@ -1,92 +1,50 @@
...
+});
 
 export default {
   key: "postmark-send-single-email",
-  name: "Send an Email with Postmark to a Single Recipient",
-  description: "Send an Email with Postmark to a Single Recipient",
-  version: "0.1.1",
+  name: "Send an email",
+  description: "Send an email with Postmark",
+  version: "0.1.4",
```
andrewjschuang, line 18: Typically the version would be bumped by one `patch` version, to `0.1.2`

But in this case since it has some kind of structural changes, I think we could bump the `minor` version to `0.2.0`. See [docs](https://pipedream.com/docs/components/guidelines/#versioning)

```suggestion
  version: "0.2.0",
```

components/postmark/actions/send-single-email/send-single-email.mjs
```diff
@@ -1,92 +1,50 @@
...
-        "Bcc": this.bcc_email,
-        "Subject": this.subject,
-        "Tag": this.tag,
-        "HtmlBody": this.html_body,
-        "TextBody": this.text_body,
-        "ReplyTo": this.reply_to,
-        "TrackOpens": this.track_opens,
-        "TrackLinks": this.track_links,
-      },
+    return this.postmark.sharedActionRequest($, this, "email", {
```
andrewjschuang, line 44: If the `Common Files` pattern is used, then instead of passing `this`, the common props would be accessible in this action, and you can pass them inside `{}`

components/postmark/actions/send-single-email/send-single-email.mjs
andrewjschuang: As seen in this [comment](https://github.com/PipedreamHQ/pipedream/pull/2823/files#r873974530)

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+        "Set message stream ID that's used for sending. If not provided, message will default to the outbound transactional stream.",
+      optional: true,
+    },
+  },
   methods: {
     // this.$auth contains connected account data
     authKeys() {
       console.log(Object.keys(this.$auth));
     },
+    async listTemplates() {
```
andrewjschuang, line 111: I would suggest trying to simplify this method, something like this:

```js
    async listTemplates($ = this) {
      const { data } = await axios($, {
        url: "https://api.postmarkapp.com/templates?count=500&offset=0",
        headers: this.getHeaders(),
        method: "GET",
      });

      if (data.TotalCount > 0) {
        return data.Templates
          .filter((obj) => obj.TemplateType === "Standard")
          .map((obj) => ({
            label: obj.Name,
            value: obj.Alias,
          }));
      }

      return [];
    },
```

Pipedream can handle if there are no options to be selected.

Also, if possible, [we prefer to use `await` instead of `promise chaining`](https://pipedream.com/docs/code/nodejs/async/#await-all-promises)

andrewjschuang: And instead of requesting for all options in one API call, it would be better to paginate them like [this example](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_drive/google_drive.app.mjs#L64) or refer to [here in the docs](https://pipedream.com/docs/components/guidelines/#async-options)

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+        "reply_to",
+        "custom_headers",
+        "track_opens",
+        "track_links",
+        "attachments",
+        "metadata",
+        "message_stream",
+      ];
+    },
+    async sharedActionRequest($, action, endpoint, uniqueProps) {
```
andrewjschuang, line 165: I think having a shared action request method is great, but also have specific methods for each action, for example:

```js
postmark.mjs

async sendEmailWithTemplate($, data) {
  return this.sharedActionRequest($, endpoint = "email/withTemplate", data);
}
```

And inside the action:

```js
send-email-with-template.mjs

async run({ $ }) {
  ...
  const response = await this.postmark.sendEmailWithTemplate($, data);
  $.export("$summary", "Some summary here");
  return response;
}
```

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+          Attachments: action.attachments?.map((str) => {
+            let params = str.split("|");
+            return params.length === 3
+              ? {
+                Name: params[0],
+                Content: params[1],
+                ContentType: params[2],
+              }
+              : JSON.parse(str);
+          }),
```
andrewjschuang, line 190: Could this be in a separate method?

Actually, could `Name`, `Content` and `ContentType` be separate props and `Attachments` object be constructed inside `async run()`?

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
+import axios from "axios";
+import { axios as axiosPipedream } from "@pipedream/platform";
+
 export default {
   type: "app",
   app: "postmark",
-  propDefinitions: {},
+  propDefinitions: {
+    from_email: {
```
andrewjschuang, line 8: Typically we use `camelCase` for prop names even if the API accepts only `snake_case`. It would be _converted_ when constructing the `params` or `data` object. For example:

```js
propDefinitions: {
  fromEmail: {
    ...
  },
  ...
},
```

```js
component.mjs

props: {
  fromEmail: {
    propDefinition: [
      postmark,
      "fromEmail",
    ],
  },
}.
...
async run({ $ }) {
  const params = {
    from_email: this.fromEmail,
    ...
  };
  ...
}

```

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
```diff
@@ -0,0 +1,51 @@
...
+  version: "0.1.23",
+  type: "action",
+  props: {
+    postmark,
+    template_alias: {
+      type: "string",
+      label: "Template",
+      description: "The template to use for this email.",
+      options() {
+        return this.postmark.listTemplates();
```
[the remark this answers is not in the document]
andrewjschuang: Also is missing `async` keyword

components/postmark/postmark.app.mjs
andrewjschuang: Also remove `await` here

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
```diff
@@ -0,0 +1,51 @@
...
+      propName,
+    ],
+  };
+});
+
+export default {
+  key: "postmark-send-email-with-template",
+  name: "Send an email with template",
+  description: "Send an email with Postmark using a template",
+  version: "0.1.23",
```
andrewjschuang, line 17: Should this version be `0.0.1`?

```suggestion
  version: "0.0.1",
```

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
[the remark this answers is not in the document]
GTFalcao: > We prefer stay the props that use `options()` in the app file.
I kept it here because only this action uses it, but if that's the standard I will move it to the app.

> Also is missing `async` keyword
I did not keep the async as `postmark.listTemplates` already returns a `Promise`, but will add it if that is a standard across options.

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
```diff
@@ -0,0 +1,51 @@
...
+    propDefinition: [
+      postmark,
+      propName,
+    ],
+  };
+});
+
+export default {
+  key: "postmark-send-email-with-template",
+  name: "Send an email with template",
```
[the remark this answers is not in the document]
GTFalcao: Updated the names for all four components.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+      type: "string",
+      label: "Message stream",
+      description:
+        "Set message stream ID that's used for sending. If not provided, message will default to the outbound transactional stream.",
+      optional: true,
+    },
+  },
   methods: {
     // this.$auth contains connected account data
     authKeys() {
```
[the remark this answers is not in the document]
GTFalcao: Removed, thanks.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+          {
+            label:
+              "An error ocurred while fetching the list of templates for this Postmark account. Please try again.",
+            value: -2,
+          },
+        ]);
+    },
+    getHeaders() {
+      return {
+        "X-Postmark-Server-Token": this.$auth.api_key,
```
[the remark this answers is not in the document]
GTFalcao: Created the method.

components/postmark/sources/new-email-opened/new-email-opened.mjs
GTFalcao: Good reminder. I completely missed this part in the Docs and searched for it after your comment.

Added the event metadata with id (generated by combining email id with timestamp from event data), summary (email subject) and timestamp (timestamp from event data)

components/postmark/sources/new-inbound-email-received/new-inbound-email-received.mjs
GTFalcao: Added event metadata as with the other source.

components/postmark/actions/send-single-email/send-single-email.mjs
GTFalcao: Removed.

components/postmark/actions/send-single-email/send-single-email.mjs
GTFalcao: Good reminder. I increased the version numbers as needed to update the component in early development, but will avoid committing such changes and make sure the version is updated/created to/at the correct number on a pull request.

components/postmark/postmark.app.mjs
GTFalcao: Moving it to a separate method, certainly agree it keeps things more organized.

The problem with making the props separate is it does not feel intuitive or user-friendly at all, in my opinion. Namely because there can be any number of attachments, each of which need these three props. Given an example of three attachments, in the current implementation:

```
Attachments: [
  "{ Name: 'readme.txt', Content: att1content, ContentType: 'text/plain' }",
  "{ Name: 'photo.png', Content: att2content, ContentType: 'image/png' }",
  "{ Name: 'index.html', Content: att3content, ContentType: 'text/html' }" ]
```

And by making them separate props, the individual data is in a more straightforward format, but the attachments are "split into pieces":

Attachment Names: [
  "readme.txt",
  "photo.png",
  "index.html" ]
Attachment Contents: [
  att1content,
  att2content,
  att3content ]
Attachment ContentTypes: [
  "text/plain",
  "image/png",
  "text/html" ]

I could not think of a better way to implement this, but any ideas are welcome, as I do not think the current implementation is optimal either.

components/postmark/postmark.app.mjs
GTFalcao: Adjusting the prop names. These were props from the existing Postmark action that I moved to the app's propDefinitions, but did not pay attention to the naming - in fact, this API expects `PascalCase` object properties, so they were already being converted. Nonetheless, you provided perfect examples 👍

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
GTFalcao: Adjusted version numbers to 0.0.1 for the new action and two new sources.

components/postmark/sources/new-inbound-email-received/new-inbound-email-received.mjs
GTFalcao: Good examples! I created a `common` module for the sources and managed to share a good deal of their inner workings via this file.

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
```diff
@@ -0,0 +1,51 @@
+import postmark from "../../postmark.app.mjs";
+
+let objSharedProps = {};
+postmark.methods.listSharedProps().forEach((propName) => {
```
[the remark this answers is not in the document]
GTFalcao: Thanks - after reading and experimenting, I agree using common modules here is indeed very handy! I was initially using the app file for common code, but that neither covered all the cases I needed, nor felt optimal for some of the ones I used it for - such as having to duplicate this initializing with a dummy object; it just didn't feel right.

components/postmark/postmark.app.mjs
GTFalcao: Created individual methods for each of the two actions on the `app`.

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
```diff
@@ -0,0 +1,51 @@
...
+        "The model to be applied to the specified template to generate the email body and subject.",
+    },
+    inline_css: {
+      type: "boolean",
+      label: "Inline CSS",
+      description:
+        "By default, if the specified template contains an HTMLBody, Postmark will apply the style blocks as inline attributes to the rendered HTML content. You may opt-out of this behavior by passing false for this request field.",
+      optional: true,
+    },
+    ...objSharedProps,
```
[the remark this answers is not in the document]
GTFalcao: Updated to reflect the common module standard.

components/postmark/actions/send-email-with-template/send-email-with-template.mjs
```diff
@@ -0,0 +1,51 @@
...
+  },
+  async run({ $ }) {
+    return this.postmark.sharedActionRequest($, this, "email/withTemplate", {
```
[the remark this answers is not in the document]
GTFalcao: Adjusted so that the common properties are obtained via a common method, thus no longer needing the action to be passed.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+        "Content-Type": "application/json",
+        "Accept": "application/json",
+      };
+    },
+    listSharedProps() {
```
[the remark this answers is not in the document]
GTFalcao: Moved the props to common module and removed this (no longer needed) method.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+      optional: true,
+    },
+  },
   methods: {
     // this.$auth contains connected account data
     authKeys() {
       console.log(Object.keys(this.$auth));
     },
+    async listTemplates() {
+      return await axios({
```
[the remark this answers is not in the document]
GTFalcao: Change applied

components/postmark/postmark.app.mjs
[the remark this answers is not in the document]
GTFalcao: Change applied

components/postmark/postmark.app.mjs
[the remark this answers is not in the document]
GTFalcao: Change applied

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+              ? {
+                Name: params[0],
+                Content: params[1],
+                ContentType: params[2],
+              }
+              : JSON.parse(str);
+          }),
+          Metadata: action.metadata,
+          MessageStream: action.message_stream,
+        },
```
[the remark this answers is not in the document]
GTFalcao: Change applied

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
+import axios from "axios";
```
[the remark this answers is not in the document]
GTFalcao: I initially had problems to pass the $ parameter since it seemed to be available only in the `run` method, but I followed the provided examples and eliminated the "duplicate" axios module as recommended.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,208 @@
...
+import { axios as axiosPipedream } from "@pipedream/platform";
```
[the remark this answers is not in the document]
GTFalcao: Change applied

components/postmark/postmark.app.mjs
GTFalcao: Change applied

components/postmark/postmark.app.mjs
GTFalcao: Changes applied.

components/postmark/actions/send-single-email/send-single-email.mjs
GTFalcao: Adjusted so that the common properties are obtained via a common method, thus no longer needing the action to be passed.

components/postmark/postmark.app.mjs
GTFalcao: Simplified the method considerably, and removed the promise chaining.

Great call on the pagination. Implemented that as well.

components/postmark/postmark.app.mjs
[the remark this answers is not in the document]
andrewjschuang: Sure, we generally use `this` as fallback if `$` is not available

components/postmark/postmark.app.mjs
andrewjschuang: Yeah, I see your point. We don't have a great way to do this yet, so generally we try to keep the number of options low, but in this case I think it's fine for now

andrewjschuang (changes requested):
@GTFalcao thanks for the changes, Looks very good!

Added just a few more comments.

You can also make `sharedActionRequest()` accept a custom http `method` and call it on `GET - listTemplates()` and `PUT - setServerInfo()` requests

components/postmark/sources/common.mjs
```diff
@@ -0,0 +1,49 @@
...
+      customResponse: true,
+    },
+  },
+  methods: {
+    getWebhookType() {
+      throw new Error("Component is missing Webhook type definition");
+    },
+    getWebhookProps() {
+      return {};
+    },
```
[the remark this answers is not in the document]
GTFalcao: The reason for this is to keep the `activate` hook shared, even though not all Postmark webhooks have custom properties to pass to the server. In this case, `new-email-opened` has custom properties but `new-inbound-email-received` does not, so it "inherits" this method from the common module.

This way, if other Postmark sources are implemented in the future, they only need to implement this method if they have custom server properties to pass when activating.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,71 @@
...
+  },
   methods: {
-    // this.$auth contains connected account data
-    authKeys() {
-      console.log(Object.keys(this.$auth));
+    _apikey() {
+      return this.$auth.api_key;
+    },
+    async listTemplates(page) {
+      const amountPerPage = 3;
```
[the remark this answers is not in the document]
GTFalcao: I was not quite sure what value to go for here, but that's a good point on the number of API calls. Changing it to 10.

components/postmark/postmark.app.mjs
```diff
@@ -1,11 +1,71 @@
...
+
+      return data.TotalCount
+        ? data.Templates.map((obj) => {
+          return {
+            label: obj.Name,
+            value: obj.Alias,
+          };
+        })
+        : [];
+
```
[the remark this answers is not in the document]
GTFalcao: Yes, that makes sense, the `options`-specific object should be created in the `options` method. Adjusting to reflect that.

I was working with `prevContext` initially instead of `page`, and thought it was not needed to return that when using `page`. Adjusting the return value as you suggested.

components/postmark/sources/common.mjs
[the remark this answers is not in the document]
andrewjschuang: Yeah, that's fine! I could see that happening in future requests

andrewjschuang (dismissed):
@GTFalcao thanks for the changes! LGTM!

vellames-turing (approved):

PR #2824, opened 2022-05-10
75 watch an url

alysonturing:
**New event when the content of the URL changes.**

**Emit body only:**
If set as true the emitted item will contain only the response body, otherwise, it will include the request status code.

**Emit as array:**
If the request responds with an array, this source will emit changes individually for each item of the array.

FYI:
This source uses a prop of type `http_request` that is not supported by the CLI;
For testing, use the web interface, see: https://pipedream-users.slack.com/archives/G01J50YGJGP/p1652141349624249?thread_ts=1652140519.202669&cid=G01J50YGJGP

### DIFF
--- a/components/http/package-lock.json
+++ b/components/http/package-lock.json
@@ -1,0 +1,31 @@
+{
+    "name": "@pipedream/http",
+    "version": "0.3.3",
+    "lockfileVersion": 2,
+    "requires": true,
+    "packages": {
+        "": {
+            "name": "@pipedream/http",
+            "version": "0.3.3",
+            "license": "MIT",
+            "dependencies": {
+                "object-hash": "^3.0.0"
+            }
+        },
+        "node_modules/object-hash": {
+            "version": "3.0.0",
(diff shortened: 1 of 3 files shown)


alysonturing:
This source uses a prop of type `http_request` that is not supported by the CLI;
so you should test using the web interface, see: https://pipedream-users.slack.com/archives/G01J50YGJGP/p1652141349624249?thread_ts=1652140519.202669&cid=G01J50YGJGP

michelle0927 (approved):
Looks good!

DilanAthukorala:
@alysonturing API Query params are not being parsed correctly, please look at the example below. 
URL: http://universities.hipolabs.com/search?**country=United+States**
image

alysonturing:
@DilanAthukorala please try again to using space instead of the plus sign
like this:
image

DilanAthukorala:
@alysonturing Thanks!

Verification completed for this change.

Issue #69, opened 2020-05-18
[TRIGGER / ACTION] Chatbot


dylburger:
App: `chatbot`

### Triggers
- New message https://www.chatbot.com/docs/webhooks/

### Actions
- Create user https://www.chatbot.com/docs/users/#create-user
- List users https://www.chatbot.com/docs/users/#list-users

lcaresia:
@dylburger or @dannyroosevelt Can you give me more details about this task? Thansk!

dannyroosevelt:
> @dylburger or @dannyroosevelt Can you give me more details about this task? Thansk!

@lcaresia I just updated the ticket with some details and I'm adding `chatbot` as an app right now.

dannyroosevelt:
This is ready for release!

PR #2873, opened 2022-05-17
[TRIGGER / ACTION] Chatbot #69

lcaresia:


andrewjschuang:
Resolves #69

components/chatbot/chatbot.app.mjs
```diff
@@ -1,11 +1,50 @@
...
+    },
+    _baseApiUrl() {
+      return "https://api.chatbot.com";
+    },
+    async _makeRequest(path, options = {}, $ = this) {
+      const config = {
+        ...options,
+        url: `${this._baseApiUrl()}/${path}`,
+        headers: {
+          Authorization: `Bearer  ${this._developerAccessToken()}`,
```
andrewjschuang, line 19: Extra space

```suggestion
          Authorization: `Bearer ${this._developerAccessToken()}`,
```

components/chatbot/chatbot.app.mjs
```diff
@@ -1,11 +1,50 @@
...
+      return this._makeRequest("users", {
+        method: "post",
+        data,
+      }, $);
+    },
+    async getUsers({ $ } = {}) {
+      const response = await this._makeRequest("users", {}, $);
+
+      return response.data;
+    },
```
andrewjschuang, line 48: **Question**

Will this return all users or is pagination needed?

components/chatbot/actions/create-user/create-user.mjs
```diff
@@ -0,0 +1,59 @@
...
+      label: "User ID",
+      description: "The user ID",
+      type: "string",
+    },
+    email: {
+      label: "Email",
+      description: "The email of the user",
+      type: "string",
+      optional: true,
+    },
```
andrewjschuang, line 21: > Note: userId is not required when attributes.default_email is provided
https://www.chatbot.com/docs/users/#create-user

It seems that if `email` is supplied, `userId` is not necessary. I think having the `email` prop as `required` instead of `userId` would be better

andrewjschuang (changes requested):
@lcaresia thanks for the submission!

Left just a few comments!

components/chatbot/chatbot.app.mjs
lcaresia: Now will return all users!

components/chatbot/chatbot.app.mjs
```diff
@@ -1,11 +1,71 @@
...
+          },
+        }, $);
+        const users = response.data;
+
+        after = users?.length >= 40
+          ? users[users.length - 1].id
+          : null;
+
+        allUsers = allUsers.concat(users);
+      } while (after);
```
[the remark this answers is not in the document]
andrewjschuang: Also it would be nice to define a constant for the default per page

andrewjschuang (changes requested):
@lcaresia thanks for the changes!

I've added a comment about pagination

andrewjschuang (dismissed):

andrewjschuang:
@lcaresia thanks, LGTM! (I have not tested my suggestion)

DilanAthukorala:
@lcaresia Getting following error when publishing action

image

lcaresia:
@DilanAthukorala I commited a suggestion without validate, now will work!

andrewjschuang (approved):

PR #2974, opened 2022-05-26
Gorgias Update Sources to Webhooks

### LINKED ISSUE #2947: [TRIGGER] Update Gorgias Sources to use webhooks instead of polling endpoints
We should update Gorgias sources to use HTTP webhooks instead of polling endpoints.
https://developers.gorgias.com/reference/the-integration-object

However, this may be a good opportunity to preserve the ability to also fetch historical events upon source creation.

andrewjschuang:
- [X] Substitute the timer from triggers to webhooks.
- [X] Fetches historical events on deploy.
- [X] Sync `gorgias` and `gorgias_oauth` component versions

Webhooks in Gorgias are called Integrations

Resolves #2947.

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/2974"></a>


dylburger:

vellames-turing (approved):
lgtm

andrewjschuang:
After some extensive testing, I noticed a few bugs coming from the [Integrations API](https://developers.gorgias.com/reference/the-integration-object) (webhooks).


**[1] Updated Tickets don't trigger an Integration call. (Trigger: Ticket Updated)**


**[2] Integrations HTTP headers has unexpected formats. (Trigger: Ticket Created)**

There are two options when creating a new Integration (as seen in the images below):
  - request_content_type: `application/json`

json-webhook-code
*application/json request_content_type*

  - request_content_type: `application/x-www-form-urlencoded`

form-webhook-code
*application/x-www-form-urlencoded request_content_type*

The requested data comes through the `event.query` attribute.

As seen in the image below, using `application/json`, `event.query` has a weird format. The `key` is the whole `id: ${id}` object, and the `value` is an empty string.

json-event
*application/json event*

json-event-query
*application/json query*

json-logs-error


Unexpectedly, I had to use `application/x-www-form-urlencoded` since it was the one that sent a valid JSON so the `id` could be extracted correctly.

form-event
*application/x-www-form-urlencoded event*

form-event-query
*application/x-www-form-urlencoded query*

form-logs-success


**[3] Integrations HTTP call sends data with Python code instead of JSON. (Trigger: Ticket Message Created)**

The content that comes inside `event.query` is not actually a valid JSON object, it is a Python Dictionary encapsulated in a string.

This is the content of `event.query.messages`:
```
"[{'id': 6618275, 'uri': '/api/tickets/3653515/messages/6618275/', 'message_id': '', 'ticket_id': 3653515, 'external_id': '', 'public': True, 'channel': 'email', 'via': 'help-center', 'source': {'type': 'email', 'to': [{'name': '', 'address': 'andrew.c@turing.com'}], 'from': {'name': '', 'address': 'andrew.c@turing.com'}}, 'sender': {'id': 15003231, 'email': 'andrew.c@turing.com', 'name': '', 'firstname': '', 'lastname': '', 'meta': ''}, 'integration_id': '', 'intents': [], 'rule_id': '', 'from_agent': False, 'receiver': {'id': 15003231, 'email': 'andrew.c@turing.com', 'name': '', 'firstname': '', 'lastname': '', 'meta': ''}, 'subject': 'Test Subject', 'body_text': '', 'body_html': 'Test Message', 'stripped_text': '', 'stripped_html': '', 'stripped_signature': '', 'headers': '', 'attachments': '', 'actions': '', 'meta': '', 'created_datetime': '2022-05-30T15:26:36.914430+00:00', 'sent_datetime': '2022-05-30T15:26:36.914430+00:00', 'failed_datetime': '', 'opened_datetime': '', 'last_sending_error': '', 'is_retriable': True, 'sentiments': [], 'text': ''}]"
```

Evidence for being a Python Dictionary: `'public': True` and single quotes.

dannyroosevelt:
> 

Thanks for flagging and for the thorough explanation, Andrew. I just reached out to someone from Gorgias to see if we can chat with a dev from their team directly.

dannyroosevelt:
@andrewjschuang heard back from Gorgias eng -- let me know if it's easy to respond here and I'll forward your responses along. Otherwise I can try and get the two of you connected directly.

[2] Integrations HTTP headers has unexpected formats. (Trigger: Ticket Created)

> I'm not sure how you're passing the query string, but when using "application/json" with the HTTP GET method, the expected called URL will be something like:
[https://test.com/endpoint?{"key":"value"}](https://test.com/endpoint?%7B%22key%22:%22value%22%7D)

> While when using the "application/x-www-form-urlencoded" format, we call:
https://test.com/endpoint?key=value

> Since it seems you're reading data by parsing the query string in a pretty usual way, I'd recommend keeping the form format.

> But all in all, even passing data in GET is not really 100% supported on our end (you can see in the front-end that the "body" is not shown when using GET, for example), is there any reason you cannot use POST instead?

[3] Integrations HTTP call sends data with Python code instead of JSON. (Trigger: Ticket Message Created)

>Indeed, this is something that is currently not possible today, you unfortunately need to pass fields one by one. Do you need maybe fields? I added it to our internal tracker as it indeed makes sense to serialize as JSON in this context rather than the default Python string formatter.

dannyroosevelt:
> > @andrewjschuang heard back from Gorgias eng -- let me know if it's easy to respond here and I'll forward your responses along. Otherwise I can try and get the two of you connected directly.
> 
> Thanks @dannyroosevelt,
> 
> Analyzing the response from the Gorgias eng:
> 
> [1] Resolved - receiving webhook requests
> 
> [2] Resolved - changed integration to POST and application/json
> 
> [3] Pending - we need to parse `ticket-messages`, which is an array of JSON objects with many attributes. Since we want to emit the whole message, we would want all of the attributes.
> 
> We can ship it like this and when they start to serialize as JSON, then we can change the `ticket-message-created` source. But there may be some events that will cause errors with this trigger. The other triggers won't be affected.

Oh and actually they just followed up again:
>About the last point, since you want to pass the whole data to your endpoint, I just remembered you can also use just "{{context}}" (and only this, not { "context": {{context}} } or similar) as a JSON body. This will dump the whole ticket/user/message in the payload, which sounds like what you'd like to receive.

andrewjschuang:
> Oh and actually they just followed up again:
> 
> > About the last point, since you want to pass the whole data to your endpoint, I just remembered you can also use just "{{context}}" (and only this, not { "context": {{context}} } or similar) as a JSON body. This will dump the whole ticket/user/message in the payload, which sounds like what you'd like to receive.

Works well, even better. I'll make the adjustments.

dannyroosevelt:
@andrewjschuang while you're in the code, is it possible for the Team ID prop to be a `string[]`, so I can select multiple?
Screenshot by Dropbox Capture

andrewjschuang:
> @andrewjschuang while you're in the code, is it possible for the Team ID prop to be a `string[]`, so I can select multiple?

Are you referring to #3088?

andrewjschuang:
I've made the adjustments and am marking it as `Ready for PR Review` since there were some changes

components/gorgias_oauth/actions/create-customer/create-customer.mjs
```diff
@@ -3,11 +3,12 @@ import overrideApp from "../../common/override-app.mjs";
...
 overrideApp(base);
 
+// same version as base
+// eslint-disable-next-line pipedream/required-properties-version
 export default {
   ...base,
   key: "gorgias_oauth-create-customer",
   name: "Create Customer",
   description: "Create a new customer. [See the docs](https://developers.gorgias.com/reference/post_api-customers)",
-  version: "0.0.1",
```
vellames-turing, line 11: Sorry. I don't get it exactly even with the comment. Why are you removing the version here and in other files?

vellames-turing (commented):
Hi there, I just need a clarification, please

components/gorgias_oauth/actions/create-customer/create-customer.mjs
andrewjschuang: This action has the same behaviour defined in `gorgias-create-customer`. If the code from `gorgias-create-customer` is changed and published, we have to rev the version of `gorgias_oauth-create-customer` and publish it too.

To avoid forgetting to do this, removing `version` will make this action always have the same version as the `base` and be published together.

vellames-turing (dismissed):

DilanAthukorala:
@andrewjschuang

1. Observing following issue when adding the source - New Ticket Message
image



2. After creating 'New Ticket Updated Source', we can see the events received previously for the **'New Ticket Created'** trigger. We are expecting to receive only the updated events for this source. 
image



3. For adding the auth connection details please add documentation for how to adding 'domain' with example. 

image

andrewjschuang:
@DilanAthukorala thanks for the feedback.

1. Changed to emit the event since to fetch the ticket message the API requires the `ticket_id`, but the received event does not contain the `ticket_id`.
2. Gorgias considers new tickets as updated too; they are sending the events through the webhook.
3. @dannyroosevelt I'm not sure where to add the description as requested, can you help us here?

dannyroosevelt:
Re: the domain description, yes that is accessible internally, and I made that change. Thanks for the suggestion @DilanAthukorala

DilanAthukorala:
Verification completed and this is ready for release.

feyzullah (approved):

PR #2979, opened 2022-05-26
Data store add create record if not found

### LINKED ISSUE #2963: [Updates] Add create record if not found in Has Key to Data Store Action
Extend [this functionality](https://github.com/PipedreamHQ/pipedream/blob/master/components/data_stores/actions/get-record/get-record.mjs#L27) to [has-key](https://github.com/PipedreamHQ/pipedream/blob/master/components/data_stores/actions/has-key/has-key.mjs) action.

We should also probably update the name/key of that action to be more descriptive: "Check if key exists"

andrewjschuang:
Resolves #2963.

### DIFF
--- a/components/data_stores/actions/add-update-record/add-update-record.mjs
+++ b/components/data_stores/actions/add-update-record/add-update-record.mjs
@@ -4,7 +4,7 @@ export default {
   key: "data_stores-add-update-record",
   name: "Add or update a single record",
   description: "Add or update a single record in your [Pipedream Data Store](https://pipedream.com/data-stores/).",
-  version: "0.0.5",
+  version: "0.0.6",
   type: "action",
   props: {
     app,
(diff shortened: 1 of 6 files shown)


vellames-turing (changes requested):
Could you rename the actions, please?
Something like `get-record-or-create` and `check-key-or-create`
I am not very good with names haha
cc @dannyroosevelt

vellames-turing (dismissed):
tks

DilanAthukorala:
@andrewjschuang If the key does not exist the newly created value is created as String type even the if I passed an Object or an Array value. 
image

andrewjschuang:
> @andrewjschuang If the key does not exist the newly created value is created as String type even the if I passed an Object or an Array value.

@DilanAthukorala should be fixed now!

DilanAthukorala:
Verification completed.

feyzullah (approved):
approved

Issue #2849, opened 2022-05-13
Request to add [Action] Woocommerce: Update product(s) information


andr3selite:
It would be great if there could be an action to **update product(s) information** on **Woocommerce**.

https://woocommerce.github.io/woocommerce-rest-api-docs/#update-a-product

dylburger:
Thanks for the request! Did you know that anyone can [author and publish their own Pipedream actions](https://pipedream.com/docs/components/quickstart/nodejs/actions/)? Would you be interested in helping out with that? We're happy to answer any questions or support you in any way.

If you don't have the time, would you be willing to reach out to the app provider to see if they're interested in building these? When we work with an app partner, it improves the quality of the integration, and it helps when this kind of request comes from a customer like you.

dylburger:
We've added this issue to our backlog. You can follow this issue for updates!

PR #2998, opened 2022-05-30
Request to add [Action] Woocommerce: Update product(s) information #2849

lcaresia:


feyzullah (dismissed):
LGTM!

vunguyenhung:

[WooCommerce_2998.pdf](https://github.com/PipedreamHQ/pipedream/files/8824638/WooCommerce_2998.pdf)

lcaresia:
@vunguyenhung I made some updates, you can test now!

vunguyenhung:
Hello @dylburger, I want to request us to change the description on Woocomerce Account Connect window
Screen Shot 2022-06-03 at 19 22 28
I would suggest change it to 
"
Find your API Keys at WooCommerce > Settings > Advanced > REST API
For url, enter the domain of your store. For example, if your store is at https://mystore.com/wp-json/wc/v2/orders then enter https://mystore.com below.
"

The part I want to change is "then enter **https://mystore.com below**"

The reason is: when testing Woocomerce, I need to setup a new site that doesn’t have HTTPs, but in the [code here](https://github.com/PipedreamHQ/pipedream/blob/master/components/woocommerce/woocommerce.app.mjs#L151) we always use HTTPs. This require me to change the code to allow HTTP, or register my site with HTTPs which is more effort than needed to test this action

I think other users having a need to evaluate Pipedream would face the same issue as me, so I suggest to allow the user to include the `http` or `https` protocol along with the url

I have discussed this request with @lcaresia.

dylburger:
@vunguyenhung That makes sense. With these changes, we'll just need to push the code changes first, and we'll need to handle existing users who haven't entered a protocol. So you'll need to e.g. check if the URL doesn't start with a protocol and add `https://` in that case. Something like:

```javascript
let url = this.$auth.url
if (!/^(?:(http(s?):\/\/)/.test(url)) {
  url = `https://${u}`;
}
```

This should also handle users who don't enter protocols, which will definitely happen even with the new instructions (we see this a lot).

Let me know if that makes sense. Once you make the changes and merge to master, let me know on Slack and I can make the change. Tracking the change at #3049.

vunguyenhung:
Thanks you for confirming @dylburger! Your suggestion makes sense to me!

Hello @lcaresia boss, could you consider that changes that @dylburger suggested
```
let url = this.$auth.url
if (!/^(?:(http(s?):\/\/)/.test(url)) {
  url = `https://${url}`;
}
```
This would make all Woocomerce actions works either with `yoursite.com` or `http://yoursite.com`. In case that user input `yoursite.com`, we will automatically replace to `https://yoursite.com` internally.

vunguyenhung:

Test report
[WooCommerce_2024.pdf](https://github.com/PipedreamHQ/pipedream/files/8853549/WooCommerce_2024.pdf)

feyzullah (approved):
Approved

PR #3003, opened 2022-05-30
new source

luancazarine:
Source
   - New list contact


components/emailoctopus/emailoctopus.app.mjs
```diff
@@ -1,11 +1,70 @@
...
+          page,
+        },
+      });
+    },
+    async getContacts({ listId }) {
+      return this._makeRequest({
+        method: "GET",
+        path: `/lists/${listId}/contacts`,
+      });
+    },
```
feyzullah, line 60: According to [this doc](https://emailoctopus.com/api-documentation/lists/get-all-contacts), all results are not returned, where `page` defaults to `1` and `limit` defaults to `100`. Is it possible to have more than `100` contacts? If so can we just make sure that we are returning all contacts here? [This](https://github.com/PipedreamHQ/pipedream/blob/5a4392ae161a9f60ecda6cd501240b5af87f09a8/components/gorgias/gorgias.app.mjs#L127) can be a good example.

feyzullah (changes requested):
Hi @luancazarine I added just 2 comments. Can you please check?

feyzullah (approved):
Thanks for the updates, approved

vunguyenhung:

Test report
[Email_Octopus_3303.pdf](https://github.com/PipedreamHQ/pipedream/files/8843823/Email_Octopus_3303.pdf)

PR #3068, opened 2022-06-06
docs python magic comments

ctrlaltdylan:
- Adding step exports section and link
- Adding magic comment docs


ctrlaltdylan:
@dylburger what do you think about these changes?

https://pipedream-docs-bygvsz3ld-pipedreamers.vercel.app/docs/code/python/import-mappings/#using-magic-comments

* Removed the ability to request for Python imports
* Placed magic comments at the top, with an example

Issue #791, opened 2021-01-13
[ACTION] Survey Monkey


todsac:
**Is there a specific app this action is for?**
SurveyMonkey

**Please provide a link to the relevant API docs for the specific service / operation.**
Get a Collector - Get the specific collector information.
Get a Response - Get the specific response information. Paid SurveyMonkey account is required (Advantage or higher).
Get a Survey - Get the specific survey information.
Get My Info - Returns the current user’s account details including their plan.
Search for a collector - Returns a list of collectors.
Search for a response - Returns a list of responses.
Search for a survey - Returns a list of surveys.

dannyroosevelt:
This is ready for release!

PR #3147, opened 2022-06-14
#791 - Survey Monkey new actions

GTFalcao:
# New Actions
- **List Surveys**
- **List Responses**
- **List Collectors**
- **Get Response Details**
- **Get Collector Details**

# Modified Action
- **Find Survey**
  - Changed name and description to **Get Survey Details** to maintain a clear pattern between the actions, but did not rename the actual component key `find-survey` - is this the correct approach?

# Modified Source
- **Custom Webhook Events**
  - Adjusted the code to add support for pagination with `_paginatedRequest` returning an array instead of a single request response. Previously, only the first 50 values (default SurveyMonkey response page size) would be received by `getSurveys` and `getCollectors`


dylburger:

components/survey_monkey/sources/custom-webhook-events/custom-webhook-events.mjs
```diff
@@ -4,7 +4,7 @@ import common from "../common/base.mjs";
 export default {
   ...common,
   name: "New Custom webhook events",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "source",
   key: "survey_monkey-custom-webhook-events",
   description: "Emmit new custom webhook event",
```
jcortes, line 10:
```suggestion
  description: "Emit new custom webhook event",
```

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -9,16 +9,52 @@ export default {
...
           value: survey.id,
         }));
       },
     },
+    responseId: {
+      type: "string",
+      label: "Response",
+      description: `Select one of the above survey's **Responses** to get details of.
+      \\
+      Alternatively, you can provide a custom *Response ID*.`,
```
jcortes, line 29:
```suggestion
      description: `Select one of the above survey's **Responses** to get details of. Alternatively, you can provide a custom *Response ID*.`,
```

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -9,16 +9,52 @@ export default {
...
+          value: response.id,
+        }));
+      },
+    },
+    collectorId: {
+      type: "string",
+      label: "Collector",
+      description: `Select one of the above survey's **Collectors** to get details of.
+      \\
+      Alternatively, you can provide a custom *Collector ID*.`,
```
jcortes, line 46:
```suggestion
      description: `Select one of the above survey's **Collectors** to get details of. Alternatively, you can provide a custom *Collector ID*.`,
```

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -50,13 +86,7 @@ export default {
...
     async _makeRequest(customConfig) {
       const {
-        $,
-        url,
-        path,
-        method,
-        params,
-        data,
-        ...otherConfig
+        $, url, path, method, params, data, ...otherConfig
```
jcortes, line 89:
```suggestion
        $ = this, url, path, method, params, data, ...otherConfig
```
and then you can call `axios` like
```javascript
return axios($, config);
```

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -9,16 +9,52 @@ export default {
     survey: {
       type: "string",
       label: "Survey",
-      description: "Survey where the action will be performed.",
+      description: `Select a **Survey** from the list.
+      \\
+      Alternatively, you can provide a custom *Survey ID*.`,
```
jcortes, line 14:
```suggestion
      description: `Select a **Survey** from the list. Alternatively, you can provide a custom *Survey ID*.`,
```

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -73,25 +103,86 @@ export default {
...
+          ...params,
+          path: `${path}?page=${page}&per_page=${amountPerPage}`,
+        });
+
+        if (response.data) values.push(...response.data);
+        else throw new Error(response);
+      } while (page++ < Math.ceil(response.total / amountPerPage));
+
+      return values;
+    },
```
jcortes, line 128: To simplify a bit more taking advantage of `axios` properties:
```suggestion
    async _paginatedRequest(args = {}) {
      // https://api.surveymonkey.net/v3/docs?javascript#pagination
      // using default SurveyMonkey pagination - 50 per page
      const amountPerPage = 50;
      const values = [];

      let page = 1;
      let response;

      do {
        response = await this._makeRequest({
          params: {
            page,
            per_page: amountPerPage,
          },
          ...args,
        });

        if (response.data) {
          values.push(...response.data);
        } else {
          throw new Error(response);
        }
      } while (page++ < Math.ceil(response.total / amountPerPage));

      return values;
    },
```

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -73,25 +103,86 @@ export default {
...
+    async getUserInfo({ $ }) {
+      return this._makeRequest({
+        method: "GET",
+        path: "/users/me",
+        $,
+      });
+    },
```
jcortes, line 135: You can simplify here in that way you don't have to set `$` symbol no each function. This same advice applies to all other functions with this approach.
```suggestion
    async getUserInfo(args = {}) {
      return this._makeRequest({
        method: "GET",
        path: "/users/me",
        ...args,
      });
    },
```

jcortes (changes requested):
HI @GTFalcao I've just added a few request changes. Overall it's looking good so far. thanks

components/survey_monkey/sources/custom-webhook-events/custom-webhook-events.mjs
GTFalcao: Great catch! Fixed that typo.

components/survey_monkey/survey_monkey.app.mjs
GTFalcao: Agreed! Applied this change.

components/survey_monkey/survey_monkey.app.mjs
GTFalcao: That makes sense, but most of the methods use some of the `args` to assemble the URL, for example:
`async getResponseDetails({
      surveyId, responseId, $,
    }) {
      return this._makeRequest({
        method: "GET",
        path: `/surveys/${surveyId}/responses/${responseId}`,
        $,
      });
    },`
In this case, since surveyId and responseId will need to be destructured from the `args`, I prefer to destructure all of them right away; this also immediately shows what the method uses.

Do you think this change should be applied only for `getUserInfo` and `getSurveys` which require no parameters other than $? Or do you think there is a better approach to keep all the methods with the same pattern?

components/survey_monkey/survey_monkey.app.mjs
GTFalcao: Great suggestion! I overlooked the use of axios `params` completely. Changed to use them as per your suggestion.

components/survey_monkey/survey_monkey.app.mjs
GTFalcao: This was done to provide a line break in the prop description - do you think it'd be best to remove that line break then?

Here's a visual comparison of the suggested change:
image

components/survey_monkey/survey_monkey.app.mjs
jcortes: Yeah in theses case what you can do is
```javascript
    async getResponseDetails({
      surveyId, responseId, ...args
    } = {}) {
      return this._makeRequest({
        method: "GET",
        path: `/surveys/${surveyId}/responses/${responseId}`,
        ...args,
      });
    },
```
in that way you don't deal with `$` here either

components/survey_monkey/survey_monkey.app.mjs
```diff
@@ -101,27 +101,32 @@ export default {
...
 
-      return axios($ || this, config);
+      return axios($, config);
     },
-    async _paginatedRequest(params) {
+    async _paginatedRequest(args = {}) {
       // https://api.surveymonkey.net/v3/docs?javascript#pagination
       // using default SurveyMonkey pagination - 50 per page
-      const amountPerPage = 50;
+      const amountPerPage = 1;
```
jcortes, line 109: Why did you change `50` to `1`?

jcortes (changes requested):
Hi @GTFalcao I have one minor observation about pagination. Thanks

components/survey_monkey/survey_monkey.app.mjs
GTFalcao: Sure! That works well and is even more clear about which args are explicitly needed for this method.

Agreed and applied!

components/survey_monkey/survey_monkey.app.mjs
GTFalcao: That was a mistake - I had changed it for testing and should not have commited that change. Fixed this, thanks!

jcortes (commented):
HI @GTFalcao LGTM! Ready for QA!

DilanAthukorala:
Verification completed for newly added survey monkey actions and no issues observed.

michelle0927 (approved):

PR #3156, opened 2022-06-15
Add @pipedream/platform to monorepo

### LINKED ISSUE #3005: [FEATURE] Move @pipedream/platform axios to this monorepo, add debug flag
This ticket captures a few items:

- We should move [the `@pipedream/platform` library ](https://github.com/PipedreamHQ/platform) to this monorepo, publish with pnpm, etc.
- We should add support for a `debug` flag. See below thread from Slack for context:

---

1. The output of the core API method used to fetch the data, so we see what the API returned. For webhook-based sources this may not be relevant, but if we make additional API calls to fetch additional data, we should log the output of all of that.
2. If you’re transforming data significantly, we should log the output of those variables
3. Log the data you’re attempting to emit, so we see what the source should have emitted (if the call to this.$emit fails, we may not get good logs to show that, so it’s helpful to know if the data you’re emitting is empty, or if the data is present but we failed to emit for another reason.

Ideally #1 would be solved in a single place (e.g. if we used `@pipedream/platform` axios, you could just pass a debug param to that, which would automatically log the serialized axios response.

We could also add better logging automatically for #3 in the platform.

andrewjschuang:
- [x]  Move @pipedream/platform to monorepo
- [x]  Add debug flag to log data output in @pipedream/platform/axios
- [x]  Throw ConfigurationError if `body` is passed to @pipedream/platform/axios
- [x]  Fix null step export
- [ ]  ~Change pre-commit hooks to build and test all and automatically add modified `pnpm-lock.yaml`~

Resolves #3005.

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/3156"></a>


dylburger:

michelle0927 (approved):
LGTM

michelle0927:
@andrewjschuang Approved, but there are a couple conflicts that will need to be resolved before mreging.

js07 (commented):
@andrewjschuang Looks great overall and the changes work for me. I made some comments about updating package minor versions and moving some of the pre-commit hook script to pre-push to speed up pre-commit.

helpers/package.json
```diff
@@ -17,6 +17,11 @@
   "publishConfig": {
     "access": "public"
   },
+  "scripts": {
+    "prebuild": "rm -rf dist",
+    "build": "tsc",
+    "test": "npm run build"
```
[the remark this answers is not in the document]
andrewjschuang: Just to see if the build fails but I agree it's better to remove it

andrewjschuang:
@js07 thanks for the detailed suggestions!

.husky/pre-push
```diff
@@ -0,0 +1,7 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npm run test-all
+npm run platform-dist
+npx pnpm install -r
+git add pnpm-lock.yaml
```
js07, line 7: If `pnpm install` makes changes to `pnpm-lock.yaml` when this hook runs, `pnpm-lock.yaml` is added to staged files for commit, but not included in the pushed changes. Is that intended?

js07 (commented):
@andrewjschuang The changes look good. I made a couple comments.

.husky/pre-push
andrewjschuang: Right, it was supposed to be in `pre-commit`

andrewjschuang:
@js07 I've made the changes. Can you please check them? Thanks

andrewjschuang:
Since #3225 improved the git hooks flow, I've reverted the changes I had made to the husky hooks.

js07 (commented):
This PR adds a `docs/package-lock.json` file. I think the `docs` package uses yarn to manage dependencies (see [docs/README.md](https://github.com/PipedreamHQ/pipedream/blob/master/docs/README.md)) rather than npm and has a `yarn.lock` file. Could we remove `docs/package-lock.json`?

andrewjschuang:
I've made both changes. Maybe I shouldn't have merged with `master`, I'm not sure. If so, I'll happily `push --force` removing the merge.

platform/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@pipedream/platform",
+  "version": "0.11.0",
```
js07, line 3: @andrewjschuang Thanks for making those changes!

I suggested incrementing the major or minor version in a prior comment. Could we actually increment the major version (to `1.0.0`) to decrease the potential impact of these changes?

js07: After merging this PR, could you do some smoke testing with some components that use `@pipedream/platform` to check that everything works as expected?

andrewjschuang: I'll test some before merging too

js07 (dismissed):

vellames-turing (approved):
lgtm

js07 (approved):

PR #3205, opened 2022-06-21
Run GitHub Action on PR Merge

### LINKED ISSUE #3170: [Enhancement] Improve visibility for component publishing successes and failures
We need to fix the workflow that alerts the team when we publish new components to the registry. Specifically, we need to solve these problems:
- Alert the user who merged the PR and on-call PR reviewer when a component failed to publish (and why)
- Alert the Pipedream team whenever any new component is successfully published to the registry
- Let the community know when we publish new components to the registry (these should include links and should be more human readable)

Task:
Update the [publish-components script](https://github.com/PipedreamHQ/pipedream/blob/master/.github/workflows/publish-components.yaml) or build a new workflow in the `pd-component-dev` org that more cleanly parses the output of the publish event, so that I can build a couple Slack messages using these variables:
- Success or error (if error, provide the error)
- App name
- Component name
- Component key
- Component type (action or source)
- File path in pipedream repo
- GitHub handle of whoever merged the PR
- Link to the PR

andrewjschuang:
Run Publish Components Action on PR Merge and send more info to the workflow.
The workflow in pd-component-dev org handles the inputs and sends the Slack messages.

Example output message from the workflow:

Screen Shot 2022-06-21 at 4 32 53 PM

Resolves #3170.

### DIFF
--- a/.github/workflows/publish-components.yaml
+++ b/.github/workflows/publish-components.yaml
@@ -1,13 +1,10 @@
 on:
-  pull_request:
+  push:
     branches:
       - master
-    types:
-      - closed
 
 jobs:
   publish-components:
-    if: github.event.pull_request.merged == true
     name: Publish Components to Pipedream Registry
     runs-on: ubuntu-latest
     steps:
@@ -104,7 +101,6 @@ jobs:
           rm $HOME/.config/pipedream/config
 
   publish-typescript-components:
-    if: github.event.pull_request.merged == true
     name: Publish TypeScript components
     runs-on: ubuntu-latest
 
--- a/.github/workflows/publish-components.yaml
+++ b/.github/workflows/publish-components.yaml
@@ -1,10 +1,13 @@
-on: 
-  push:
+on:
+  pull_request:
     branches:
       - master
+    types:
+      - closed
 
 jobs:
   publish-components:
+    if: github.event.pull_request.merged == true
     name: Publish Components to Pipedream Registry
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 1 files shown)


andrewjschuang:
Example output message from the workflow:

.github/workflows/publish-components.yaml
```diff
@@ -1,10 +1,13 @@
-on: 
-  push:
+on:
+  pull_request:
```
js07, line 2: I think the workflow was previously configured to run on `push` because the workflow would run from the context of the forked repo and lose access to secrets used in the workflow when triggered by the `pull_request` event. And it allows pushes of commits directly to `master` to trigger this workflow.

Could we trigger this workflow on `push` and maybe find the associated pull request using a GitHub API request in the workflow in the `pd-component-dev` org?

js07 (commented):
@andrewjschuang The workflow in `pd-component-dev` looks great! The output is so much readable and actionable.

.github/workflows/publish-components.yaml
andrewjschuang: I tried finding a way, but I'll take a deeper look at this.

andrewjschuang:
@js07 I changed back to push events, and got the PR info from the workflow

andrewjschuang:
@js07 I've updated the workflow to support both PR and non-PR commits.

To be able to fully test the output without breaking I think we need to merge and create a http post step in the current workflow to forward the data to the new one

js07 (approved):
@andrewjschuang Looks good to merge. After merging, we can test the workflow like you mentioned, or update the `ENDPOINT` secret to the workflow's endpoint.

PR #3300, opened 2022-07-01
Gitpodify!

axonasif:
I'm not sure if I could push to https://github.com/PipedreamHQ/pipedream/pull/3296, so making a new PR 😅 

In #3296, I see that you're trying to use an user ENVIRONMENT VARIABLE from inside a dockerfile. Which is actually very dangerous since this is a public repo even if we make that work. We can do it in a different way.

I think you tried doing that [here](https://github.com/PipedreamHQ/pipedream/blob/23c7cafb5bd9cd96c2659d7abdd24d6f4e4b6f76/.gitpod.yml#L23) already but it didn't work because of [Gitpod's prebuild behavior](https://www.gitpod.io/docs/prebuilds#workspace-directory-only) 😅 

This change will also make the flow more interactive.
(Note: Please squash merge it BTW)

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/3300"></a>

### DIFF
--- a/.gitpod.Dockerfile
+++ b/.gitpod.Dockerfile
@@ -1,0 +1,6 @@
+# You can find the new timestamped tags here: https://hub.docker.com/r/gitpod/workspace-full/tags
+FROM gitpod/workspace-full:2022-05-08-14-31-53
+
+USER root
+RUN cd /tmp && curl https://cli.pipedream.com/linux/amd64/latest/pd.zip --output pd.zip \
+    	&& unzip pd.zip -d /usr/local/bin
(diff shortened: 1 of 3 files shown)


dylburger:

andrewjschuang:
@ctrlaltdylan requesting a review from you since you're in the loop on this

ctrlaltdylan:
Thanks again @axonasif ! 

Works on my account, and I didn't lose my home dir.

Appreciate you going the extra mile on this! Clearly my devops skillz are a little rusty 😅

ctrlaltdylan:
@axonasif could you merge your fork with the latest `master`? I can't merge until it's up to date unfortunately. Would like to give you contribution credit, I would do it myself otherwise!

axonasif:
Done 👍

PR #3359, opened 2022-07-07
@pipedream/platform axios console.log if unable to export

andrewjschuang:


platform/dist/axios.js
```diff
@@ -112,12 +112,11 @@ function stepExport(step, message, key) {
     if (step) {
         if (step.export) {
             step.export(key, message);
+            return;
         }
         else {
```
vellames-turing, line 117: I think we don't need the else statement since you are `return` on `if`

platform/dist/axios.js
```diff
@@ -112,12 +112,11 @@ function stepExport(step, message, key) {
     if (step) {
         if (step.export) {
             step.export(key, message);
+            return;
```
vellames-turing, line 115: Using this `return`, the `console.log` will not be called in this case, just confirming: this is the idea?

vellames-turing (changes requested):
Hi there, I left some comments, please consider the same comments for the `ts`file

vellames-turing (approved):

PR #3461, opened 2022-07-18
publish app/component readme to marketplace

TheBestMoshe:
If a README is added to an app, action, or trigger directory, it will get published to the Pipedream marketplace.

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/3461"></a>


TheBestMoshe:
I based the workflow on the [Publish Components](https://github.com/PipedreamHQ/pipedream/blob/master/.github/workflows/publish-components.yaml) workflow. 

I did not do much in terms of error handling and catching errors in the API. I see that in that workflow [we are logging the success and errors to an endpoint](https://github.com/PipedreamHQ/pipedream/blob/f22a7c88fe2a39b3a62a8f3fe9fb30c424604ab0/.github/workflows/publish-components.yaml#L101) (a PD workflow?). Should we do that here as well?

scripts/updateMarketplaceReadme.mts
```diff
@@ -0,0 +1,79 @@
...
+
+    if (pathSegments.length === 3) {
+      key = pathSegments[1];
+    } else if (pathSegments.length === 5) {
+      key = pathSegments[1] + "-" + pathSegments[3];
+    } else {
+      // TODO handle invalid file path
+      console.warn(
+        `"${readmePath}" is an invalid path. Cannot determine if this is an app or a component. Skipping...`
+      );
```
TheBestMoshe, line 44: Currently, I'm not doing anything for invalid readme paths. Should we be pushing an alert anywhere for this, or just log it and move on?

tjk: These warns will likely not be seen -- it might be easy to create a sentry app for this public repo and then provide the key via secrets and just `Sentry.captureMessage`?

tjk: (if you do this i recommend making the message static and passing the variable as an `extra: {...}` so that the grouping is easier maybe)

scripts/updateMarketplaceReadme.mts
```diff
@@ -0,0 +1,79 @@
...
+      ) {
+        marketplaceContentSet(
+          key: $key
+          markdownB64: $markdownB64
+          path: $path
+        ) {
+          marketplaceContent {
+            id
+          }
+          errors
```
tjk, line 62: Would be nice to check errors

scripts/updateMarketplaceReadme.mts
```diff
@@ -0,0 +1,79 @@
...
+      }
+    `;
+
+    const variables = {
+      key,
+      markdownB64: b64,
+      path: readmePath
+    }
+
+    const response = await pdClient.request(query, variables)
```
tjk, line 73: a graphql "network error" (eg. top-level errors that cause non-200) will break at a random point -- would it be better to try/catch and collect the errors to sentry (like the first warning and checking the internal `errors`) but still proceed for the others?

PR #3892, opened 2022-07-27
Adding PoC auth and data store docs

ctrlaltdylan:
## Changes

* Adds references to new Data Stores & App Connections in the main Python docs under `/code/python/`
* Adds specific Data Store usage docs `/code/python/using-data-stores/`
* Adds specific Connecting App usage docs `/code/python/auth`

### Scratch list

✅  Add screenshots
✅  Update ToC links in the `/code/python/` to the auth and data store references

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/3892"></a>


docs/docs/code/python/README.md
```diff
@@ -19,6 +19,26 @@ However, you can't connect accounts, return HTTP responses, or take advantage of
 2. Click **Custom Code**
 3. In the new step, select the `python` language runtime in language dropdown
 
+## Python Code Step Structure
+
+A new Python Code step will have the following structure, with a `hander` method and a `pd` argument passed into it:
```
dylburger, line 24: hander -> handler

docs/docs/code/python/README.md
```diff
@@ -19,6 +19,26 @@ However, you can't connect accounts, return HTTP responses, or take advantage of
...
+
+```python
+
+def handler(pd: "pipedream"):
+  # Prints "Hello, World!" to the step's logs
+  print("Hello, World!")
```
dylburger, line 30: Since the logs tab isn't highlighted in the results by default, you might consider modifying these examples to export the data using `pipedream.script_helpers`

docs/docs/code/python/auth/README.md
```diff
@@ -0,0 +1,103 @@
...
+```python
+from slack_sdk import WebClient
+
+def handler(pd: "pipedream"):
+  token = pd.inputs["slack"]["$auth"]["oauth_access_token"]
+  # Authentication details for all of your apps are accessible under the special pd.inputs["slack"] variable:
+  console.log(pd.inputs["slack"]["$auth"])
+```
+
+`this.appName.$auth` contains named properties for each account you connect to the associated step. Here, we connected Slack, so `this.slack.$auth` contains the Slack auth info (the `oauth_access_token`).
```
dylburger, line 69: References `this.app.$auth` which isn't exposed in Python right?

docs/docs/code/python/using-data-stores/README.md
```diff
@@ -0,0 +1,230 @@
+---
+short_description: Store and read data with data stores.
+thumbnail: https://res.cloudinary.com/pipedreamin/image/upload/v1646763735/docs/icons/icons8-database-96_iv1oup.png
+---
+
+# Using Data Stores 
+
```
dylburger, line 7: We should probably link to the general data stores docs near the top in case users don't know what they are in general

dylburger (changes requested):
lgtm, added just a couple of comments

docs/docs/code/python/README.md
ctrlaltdylan: Good point, but it seems that in this new `handler` paradigm we've included a nice `pd.export()` method instead.

docs/docs/code/python/auth/README.md
ctrlaltdylan: Nice catch!

ctrlaltdylan: Found some other invalid node references, fixing those.

Issue #3877, opened 2022-07-26
Please, add "Retrieve a page property item" to the Notion API actions.[ACTION]


hubertse:
This is a request for Notion API. See Documentation:
https://developers.notion.com/reference/retrieve-a-page-property

ljdatasci:
I would like to be assigned to this issue. Where do I start?

andrewjschuang:
Hi @ljdatasci, thank you for prompting to help!
You can start by reading the [Pipedream Docs](https://pipedream.com/docs/components/guidelines/) and checking out the [Notion Actions](https://github.com/PipedreamHQ/pipedream/tree/master/components/notion/actions).
Feel free to ask any questions if needed!

ljdatasci:
Some developers are using GitPod to submit a PR.  Should I also use GitPod to submit my PR? It's not specifically stated in the Pipedream Docs.

andrewjschuang:
Hi @ljdatasci, are you currently working on this?

ljdatasci:
Yes.  I'm adding a method to extract the database property ids and paginate them in the action, so that the user can select from a list of property ids by name.

andrewjschuang:
Cool, feel free to move the card [in the project](https://github.com/PipedreamHQ/pipedream/projects/1) to **Ready for PR Review** when you submit the PR!

ljdatasci:
I have some questions about async options.  Do you have time to discuss my changes on the my fork of the repo https://github.com/ljdatasci/pipedream?

ctrlaltdylan:
PR here: https://github.com/PipedreamHQ/pipedream/pull/3945

PR #3945, opened 2022-08-06
3877 notion page property item

ljdatasci:
I removed the async options from the propertyId.  In order to get a paginated list of propertyIds, you need to request them from the Retrieve a Database endpoint, for the pages that are part of a database and have a database_id [https://developers.notion.com/reference/retrieve-a-page-property](url). If the page is not part of a database, then the only propertyId that can be used is 'title'.  The user can create a step prior to Retrieve a Database and pass a propertyId as a step variable.  Below is a screenshot of the results of the Retrieve P
Screen Shot 2022-08-06 at 9 01 51 AM
age Property Item action:  

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/3945"></a>


dylburger:

dylburger:

ctrlaltdylan:
This is a PR for https://github.com/PipedreamHQ/pipedream/issues/3877

ctrlaltdylan:
I have moved the corresponding issue for this PR into the QA column, it should be picked up by a QA engineer and they'll leave feedback. Thanks again for submitting a PR! :)

andrewjschuang (changes requested):
Hi @ljdatasci, thanks for the contribution!

Looks good overall, I've added just a few comments.

We have some linting rules that can automatically be applied following the instructions [here](https://pipedream.com/docs/components/guidelines/#local-checks). Can you please revert the changes in `notion.app.mjs` that are not related to the new action?

andrewjschuang:
Just wondering, would your use case be resolved using the `Retrieve Page` action?

components/notion/actions/retrieve-page-property-item/retrieve-page-property-item.mjs
```diff
@@ -0,0 +1,27 @@
+import notion from '../../notion.app.mjs';
+
+export default {
+  key: 'notion-retrieve-page-property-item',
+  name: 'Retrieve Page Property Item',
+  description:
+    'Retrieves a property_item object for a given page_id and property_id.',
+  version: '0.0.21',
```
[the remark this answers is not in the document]
ljdatasci: According to the linting instructions, I run `pnpm install` at the root of the project, run `pnpm install -r` to install the required dependencies, then I can run `npx eslint action/retrieve-page-property-item`? I'm new to using pnpm.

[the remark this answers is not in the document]
andrewjschuang: Yeah, exactly. If you run `npx eslint action/retrieve-page-property-item --fix`, then it will automatically fix most of the linting rules, if not all of them

ljdatasci:
Yes the use case would be resolved using the Retrieve Page actions (see
screenshot).  As I'm reviewing the Notion API documentation again, the
property keys can be retrieved from the Retrieve a Database endpoint.  The
documentation says that prior to version 2022-06-28, the Page Object
displayed the page properties.  As of version 2022-06-28, the Page Object
displays the page properties only when there's a Create or Update page
request.

From my test, the user can get the property IDs from both the Retrieve
Database action and Retrieve a Page action.  Going forward with version
2022-06-28, the user can only get the property IDs from the Page Object if
there's a Create or Update request.  So it seems that it would be best if
the user runs one of the two actions and passes the property ID as a step
variable.  I tried a paginated list of property ID, but with the newest
version, the Retrieve a Page action would not work unless run the Create or
Update a Page action first.  What do you think?

components/notion/actions/retrieve-page-property-item/retrieve-page-property-item.mjs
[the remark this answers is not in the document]
ljdatasci: I ran the eslint command and I don't see any changes made the to the component.  The version number occurred from testing the action in Pipedream several times, several different ways.  Is this an issue?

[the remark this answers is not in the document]
andrewjschuang: Probably most of the linting issues are located in the app file, `notion.app.mjs`. For example, strings should use double quotes instead of single quotes.

Regarding the component version, this is known limitation. Sources support using the `pd dev` command which doesn't require updating the component version for every change and publish for testing, but actions don't support this yet. When publishing to the global registry (after merging to master), we apply the [versioning rules](https://pipedream.com/docs/components/guidelines/#versioning).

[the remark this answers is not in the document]
ljdatasci: fixed and added to branch

components/notion/actions/retrieve-page-property-item/retrieve-page-property-item.mjs
```diff
@@ -0,0 +1,27 @@
+import notion from '../../notion.app.mjs';
+
+export default {
+  key: 'notion-retrieve-page-property-item',
+  name: 'Retrieve Page Property Item',
+  description:
+    'Retrieves a property_item object for a given page_id and property_id.',
```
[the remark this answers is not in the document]
ljdatasci: fixed and added to branch

ljdatasci:
I updated the above to say 'property object' not 'property keys'.

andrewjschuang:
Thanks for the explanation and the links. What if we add a `databaseId` prop before the `pageId` prop and use `async options()` to fetch the database properties and display them in the `propertyId` prop?

ljdatasci:
We would need to check the parent.type on the Page object.  If parent.type
is 'workspace' or 'page_id', then the only property ID allowed is 'title'.
If parent.type is 'database_id', then we use the database_id from the Page
Object, run a Retrieve a Database action, and return the property IDs.

andrewjschuang:
> We would need to check the parent.type on the Page object. If parent.type is 'workspace' or 'page_id', then the only property ID allowed is 'title'. If parent.type is 'database_id', then we use the database_id from the Page Object, run a Retrieve a Database action, and return the property IDs.

Sounds good to me!

ljdatasci:
I'll start working on it and add the code to the async options for the
pageId.

andrewjschuang:
Thank you!

ljdatasci:
> Thank you!

I'm getting error 
{"name":"UserError","message":"bad options response for prop: propertyId"} and I'm sure that my code is correct?  I'm trying to pull the propertyIds from response object.  Are other users having the same issue?

andrewjschuang:
> I'm getting error {"name":"UserError","message":"bad options response for prop: propertyId"} and I'm sure that my code is correct? I'm trying to pull the propertyIds from response object. Are other users having the same issue?

The error is that `const { propIds }` extracts `undefined` in this line below:
`const { propIds } = propKeys.length === 1 && propKeys.includes("title")`
It should be `const propIds = `

ljdatasci:
> > I'm getting error {"name":"UserError","message":"bad options response for prop: propertyId"} and I'm sure that my code is correct? I'm trying to pull the propertyIds from response object. Are other users having the same issue?
> 
> The error is that `const { propIds }` extracts `undefined` in this line below: `const { propIds } = propKeys.length === 1 && propKeys.includes("title")` It should be `const propIds = `

Thank you.  Here are the results of my testing.  I can pull the property Ids for the pages whose  parent type is 'database_id as seen in the screenshot below:
Screen Shot 2022-08-09 at 12 03 40 PM

But when I try to pull a property Id for a page whose parent type is 'page_id' or 'workspace', nothing returns.  What should return is a single property ID of 'title':
Screen Shot 2022-08-09 at 12 04 07 PM

I committed my changes to the branch in the fork of the repo on my account.

andrewjschuang:
Sorry for the delay, for some reason I missed the notification.

In this line, `response.properties` should be just `response`:

```js
const { properties } = parentType === "database_id"
  ? await this.notion.retrieveDatabase(response.parent.database_id)
  : response;
```

ljdatasci:
> Sorry for the delay, for some reason I missed the notification.
> 
> In this line, `response.properties` should be just `response`:
> 
> ```js
> const { properties } = parentType === "database_id"
>   ? await this.notion.retrieveDatabase(response.parent.database_id)
>   : response;
> ```

It's working now.  Here are the results: 
Screen Shot 2022-08-09 at 6 35 20 PM

Screen Shot 2022-08-09 at 6 42 12 PM

components/notion/notion.app.mjs
```diff
@@ -29,6 +29,42 @@ export default {
...
+
+          const propEntries = Object.entries(properties);
+          const propIds  = propEntries.length === 1 && Object.values(propEntries)[0][1].id === "title"
+            ?
+            propEntries.map((prop) => {
+              return {
+                label: prop[1].type,
+                value: prop[1].id,
+              };
+            })
```
andrewjschuang, line 55: **Minor/Optional**
Use parenthesis instead of `return` statement

```suggestion
            propEntries.map((prop) => ({
              label: prop[1].type,
              value: prop[1].id,
            }))
```

components/notion/notion.app.mjs
```diff
@@ -29,6 +29,42 @@ export default {
...
+            : propEntries.map((prop) => {
+              return {
+                label: prop[1].name,
+                value: prop[1].id,
+              };});
```
andrewjschuang, line 60: **Minor/Optional**
Use parenthesis instead of `return` statement

```suggestion
            : propEntries.map((prop) => ({
                label: prop[1].name,
                value: prop[1].id,
              }));
```

components/notion/notion.app.mjs
```diff
@@ -47,9 +83,10 @@ export default {
         parentId, parentType,
       }) {
         try {
-          const { properties } = parentType === "database"
-            ? await this.retrieveDatabase(parentId)
-            : await this.retrievePage(parentId);
+          const { properties } =
+            parentType === "database"
+              ? await this.retrieveDatabase(parentId)
+              : await this.retrievePage(parentId);
```
andrewjschuang, line 89: Can you revert this back to its original state?

```suggestion
          const { properties } = parentType === "database"
            ? await this.retrieveDatabase(parentId)
            : await this.retrievePage(parentId);
```

components/notion/notion.app.mjs
```diff
@@ -60,7 +97,8 @@ export default {
     archived: {
       type: "boolean",
       label: "Archive page",
-      description: "Set to true to archive (delete) a page. Set to false to un-archive (restore) a page.",
+      description:
+        "Set to true to archive (delete) a page. Set to false to un-archive (restore) a page.",
```
andrewjschuang, line 101: Can you revert this back to its original state?

```suggestion
      description: "Set to true to archive (delete) a page. Set to false to un-archive (restore) a page.",
```

components/notion/notion.app.mjs
```diff
@@ -104,8 +142,9 @@ export default {
     },
     _extractPageTitleOptions(pages) {
       return pages.map((page) => {
-        const propertyFound = Object.values(page.properties)
-          .find((property) => property.type === "title" && property.title.length > 0);
+        const propertyFound = Object.values(page.properties).find(
+          (property) => property.type === "title" && property.title.length > 0,
+        );
```
andrewjschuang, line 147: Can you revert this back to its original state?

```suggestion
        const propertyFound = Object.values(page.properties)
          .find((property) => property.type === "title" && property.title.length > 0);
```

components/notion/notion.app.mjs
```diff
@@ -196,16 +241,14 @@ export default {
         };
         const response = await this.queryDatabase(databaseId, params);
         const {
-          results: pages,
-          next_cursor: nextCursor,
+          results: pages, next_cursor: nextCursor,
```
andrewjschuang, line 244: Can you revert this back to its original state?

```suggestion
          results: pages,
          next_cursor: nextCursor,
```

components/notion/notion.app.mjs
```diff
@@ -225,18 +268,21 @@ export default {
 
       do {
         const {
-          results,
-          next_cursor: nextCursor,
-        } = await this.listBlockChildren(block.id, params);
+          results, next_cursor: nextCursor,
+        } =
+          await this.listBlockChildren(block.id, params);
```
andrewjschuang, line 273: Can you revert this back to its original state?

```suggestion
          next_cursor: nextCursor,
        } = await this.listBlockChildren(block.id, params);
```

components/notion/notion.app.mjs
```diff
@@ -225,18 +268,21 @@ export default {
...
-        .forEach((c, i) => {
-          children[i].children = c;
-        });
+      (
+        await Promise.all(
+          children.map((child) => this.retrieveBlockChildren(child)),
+        )
+      ).forEach((c, i) => {
+        children[i].children = c;
+      });
```
andrewjschuang, line 285: Can you revert this back to its original state?

```suggestion
      (await Promise.all(children.map((child) => this.retrieveBlockChildren(child))))
        .forEach((c, i) => {
          children[i].children = c;
        });
```

andrewjschuang (changes requested):
Looking good, I've left a few comments below.

andrewjschuang (approved):

vunguyenhung:

Test report
[Notion_3945.pdf](https://github.com/PipedreamHQ/pipedream/files/9351734/Notion_3945.pdf)

andrewjschuang:
Merging on behalf of @ljdatasci.
Thanks again for the contribution!

PR #4171, opened 2022-09-06
Big Cartel new sources

feyzullah:
* New Order
* New Product

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/4171"></a>

### DIFF
--- a/components/big_cartel/big_cartel.app.mjs
+++ b/components/big_cartel/big_cartel.app.mjs
@@ -1,11 +1,49 @@
+import { axios } from "@pipedream/platform";
+
 export default {
   type: "app",
   app: "big_cartel",
   propDefinitions: {},
   methods: {
-    // this.$auth contains connected account data
-    authKeys() {
-      console.log(Object.keys(this.$auth));
+    _getUrl(path) {
+      return `https://api.bigcartel.com/v1${path}`;
+    },
+    _getAccessToken() {
+      return this.$auth.oauth_access_token;
+    },
+    _getAccountId() {
+      return this.$auth.account_id;
+    },
+    _getHeaders(headers = {}) {
+      return {
+        "Authorization": `Bearer ${this._getAccessToken()}`,
+        "Accept": "application/vnd.api+json",
+        "Content-type": "application/vnd.api+json",
+        "User-Agent": "@PipedreamHQ/pipedream v0.1",
+        ...headers,
+      };
+    },
+    async _makeRequest({
+      $ = this, path, headers, ...otherConfig
+    } = {}) {
+      const config = {
+        url: this._getUrl(path),
+        headers: this._getHeaders(headers),
+        ...otherConfig,
+      };
+      return axios($, config);
+    },
+    async getOrders(args = {}) {
+      return this._makeRequest({
+        path: `/accounts/${this._getAccountId()}/orders`,
+        ...args,
+      });
+    },
+    async getProducts(args = {}) {
+      return this._makeRequest({
+        path: `/accounts/${this._getAccountId()}/products`,
+        ...args,
+      });
     },
   },
 };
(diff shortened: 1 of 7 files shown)


components/big_cartel/common/utils.mjs
```diff
@@ -0,0 +1,37 @@
...
+        url: next,
+      });
+      for (const resource of nextResources[resourceKey]) {
+        yield resource;
+      }
+      if (!nextResources[resourceKey]?.length) {
+        return;
+      }
+      next = nextResources?.links?.next;
+    }
```
jcortes, line 35: Maybe you can consider not to repeat code with something similiar to
```suggestion
    let next;
    while (true) {
      const nextResources = await resourceFn({
        ...resourceFnArgs,
        params: {
          ...resourceFnArgs?.params,
        },
        url: next,
      });
      for (const resource of nextResources[resourceKey]) {
        yield resource;
      }
      if (!nextResources[resourceKey]?.length) {
        return;
      }
      next = nextResources?.links?.next;
    }
```

components/big_cartel/sources/new-order/new-order.mjs
```diff
@@ -0,0 +1,33 @@
...
+  key: "big_cartel-new-order",
+  name: "New Order Event",
+  description: "Emit new events when a new order is created. [See the docs here](https://developers.bigcartel.com/api/v1#get-all-orders)",
+  version: "0.0.1",
+  type: "source",
+  ...common,
+  props: {
+    app,
+    ...common.props,
+  },
```
jcortes, line 14: I think you can remove this and add the `app` prop in `sources/common.mjs` right?

components/big_cartel/sources/new-order/new-order.mjs
```diff
@@ -0,0 +1,33 @@
+import app from "../../big_cartel.app.mjs";
+import common from "../common.mjs";
+
+export default {
+  key: "big_cartel-new-order",
+  name: "New Order Event",
+  description: "Emit new events when a new order is created. [See the docs here](https://developers.bigcartel.com/api/v1#get-all-orders)",
+  version: "0.0.1",
+  type: "source",
```
jcortes, line 9: `dedupe` prop is missing, I guess in these cases it should be set to `unique`

jcortes (changes requested):
Hi @feyzullah I've just adde a few request changes and it will be good to go! Thanks!

components/big_cartel/big_cartel.app.mjs
```diff
@@ -1,11 +1,49 @@
...
+      return {
+        "Authorization": `Bearer ${this._getAccessToken()}`,
+        "Accept": "application/vnd.api+json",
+        "Content-type": "application/vnd.api+json",
+        "User-Agent": "@PipedreamHQ/pipedream v0.1",
+        ...headers,
+      };
+    },
+    async _makeRequest({
+      $ = this, path, headers, ...otherConfig
```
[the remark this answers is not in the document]
feyzullah: > I think you should include de prop `url` since you are using it from `utils.getResourcesStream` right?
Actually it is coming as `otherConfig` and overrides `url` in `config`. If I add `url` here I need to check it once more.

components/big_cartel/sources/new-order/new-order.mjs
feyzullah: Oh I guess I forgot to add that. Good catch, thank you!

jcortes (commented):
Hi @feyzullah LGTM! Ready for QA!

vunguyenhung:

Test report
[Big_Cartel_4171.pdf](https://github.com/PipedreamHQ/pipedream/files/9527627/Big_Cartel_4171.pdf)

jcortes (approved):
Approved!

PR #4189, opened 2022-09-08
[SOURCE] Eversign

### LINKED ISSUE #3862: [Sources] Eversign
### App
eversign

### Triggers
* Document Completed
* Document Sent

vellames:
Closes #3862 

Since the free account is really limited. I was able to create a source where user can handle any Eversign event

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/4189"></a>


jcortes (commented):
Hi @vellames I think there are some actions that you still can implement. What about these ones?
https://eversign.com/api/documentation/methods#use-template
https://eversign.com/api/documentation/methods#download-original-pdf

vellames:
> Hi @vellames I think there are some actions that you still can implement. What about these ones? https://eversign.com/api/documentation/methods#use-template https://eversign.com/api/documentation/methods#download-original-pdf

Humm, I can because I need a paid account to set up a template for example, and the API calls are very limited for a free account. I was able to do the sources because the webhook notification does not count as an API call for the App. Also, I have a very limited number of documents to create in the free account. What do you think about approving this one so we can open another one asking for a paid version so that I can implement the other actions?

michelle0927 (changes requested):
It's not ideal, but I see why you set it up this way. Just a couple things, and then good to go!

components/eversign/package.json
```diff
@@ -0,0 +1,17 @@
+{
+  "name": "@pipedream/eversign",
+  "version": "0.0.1",
```
michelle0927, line 3: I think the package.json should list crypto as a dependency.

components/eversign/sources/new-event/new-event.mjs
```diff
@@ -0,0 +1,85 @@
+import crypto from "crypto";
+import app from "../../eversign.app.mjs";
+import options from "../../options.mjs";
+
+export default {
+  key: "eversign-new-event",
+  type: "source",
+  name: "New Event",
+  description: "Emit new event when a new event occurs in EverSign",
```
michelle0927, line 9: I think we should include some instructions here for the user. Something like:
```suggestion
  description: "Emit new event when a new event occurs in EverSign. To begin receiving events, navigate to your Developer dashboard, and enter the URL of this source/workflow as the Webhook URL. https://eversign.com/api/documentation/webhooks",
```

michelle0927 (changes requested):
It's not ideal, but I see why you set it up this way. Just a couple things, and then good to go!

michelle0927 (approved):
Looks good!

Issue #4152, opened 2022-09-02
[BUG] property.map.value is not a function, when pushing data to notion database.


GiZmoFalcon:
I'm facing an issue here where the UI shows the properties selected, I can fill the value for each property, 

<img width="794" alt="image" src="https://user-images.githubusercontent.com/16624267/183203956-d47f8d73-43f8-4c8d-a855-e8f487fb6546.png">

but it errors out after running saying ```property.map.value``` is not a function.

And in the "inputs" section of the request, I see that the property field is empty. 

<img width="539" alt="image" src="https://user-images.githubusercontent.com/16624267/183202054-94022b6c-7e86-463e-adc3-7afb5af6c951.png">

```js 
TypeError
property.value.map is not a function

DETAILS
    at Object.convertToNotion (file:///tmp/__pdg__/dist/code/4e0eaa7dbadfe9d398cc7f7ffbc8570f4bcf3770bd7362e8184bd6accc62cd9c/code/common/notion-page-properties.mjs:52:36)
    at Object._convertPropertiesToNotion (file:///tmp/__pdg__/dist/code/4e0eaa7dbadfe9d398cc7f7ffbc8570f4bcf3770bd7362e8184bd6accc62cd9c/code/actions/common/base-page-builder.mjs:98:60)
    at Object._buildNotionPageProperties (file:///tmp/__pdg__/dist/code/4e0eaa7dbadfe9d398cc7f7ffbc8570f4bcf3770bd7362e8184bd6accc62cd9c/code/actions/common/base-page-builder.mjs:128:19)
    at Object.buildPageProperties (file:///tmp/__pdg__/dist/code/4e0eaa7dbadfe9d398cc7f7ffbc8570f4bcf3770bd7362e8184bd6accc62cd9c/code/actions/common/base-page-builder.mjs:155:19)
    at Object.buildPage (file:///tmp/__pdg__/dist/code/4e0eaa7dbadfe9d398cc7f7ffbc8570f4bcf3770bd7362e8184bd6accc62cd9c/code/actions/create-page-from-database/create-page-from-database.mjs:62:31)
    at Object.run (file:///tmp/__pdg__/dist/code/4e0eaa7dbadfe9d398cc7f7ffbc8570f4bcf3770bd7362e8184bd6accc62cd9c/code/actions/create-page-from-database/create-page-from-database.mjs:73:23)
    at process.processTicksAndRejections (internal/process/task_queues.js:95:5)
[...]

PR #4404, opened 2022-10-04
[BUG] property.map.value is not a function, when pushing data to notion database. #4152

lcaresia:


dylburger:

components/notion/actions/create-page-from-database/create-page-from-database.mjs
```diff
@@ -69,6 +70,10 @@ export default {
     },
   },
   async run({ $ }) {
+    if (this?.Tags && typeof this?.Tags === "string") {
```
andrewjschuang, line 73: @lcaresia the reason for `buildPageProperties()` is so that this kind of conversion doesn't have to be copy and pasted in every action. Please look at [this common file here](https://github.com/PipedreamHQ/pipedream/blob/master/components/notion/common/notion-page-properties.mjs#L52).

andrewjschuang (changes requested):
@lcaresia the limitation in adding a check for `Tags` specifically is that a Notion page can have infinite other names for `multi_select` properties. There is no way to parse all of them.

Did you check on the [common file](https://github.com/PipedreamHQ/pipedream/blob/master/components/notion/common/notion-page-properties.mjs#L52) I mentioned earlier?

**Every** `multi_select` will pass through this `convertToNotion()` method. That way the **only place the fix is needed** is here:

image

So please change the `convertToNotion()` method to check if `property.value` is a `string`, because it's calling directly `.map` from it.

**Also, this has to be done in every other `string[]` type in this file**, because they all have the same issue. **Ideally**, what do you think of creating an **auxiliary function** that does the check if it's a `string` and converts it to an `array`? So that the code is not copied and pasted many times and there is only **one place to change** when needed.

For example:

```js
function parseArray(value) {
  if (typeof value === "string") {
    return JSON.parse(value);
  }
  return value;
}
```
And then:
```js
convertToNotion: (property) => ({
  multi_select: parseArray(property.value).map((name) => ({
    name,
  })),
}),
```

lcaresia:
Sorry man, I miss this, I make confusion with the common files.

Is resolved now!

andrewjschuang (approved):
Thank you, LGTM!

I fixed the versions, could you verify them before/after submitting next time please?

Issue #3645, opened 2022-07-25
Gumroad


jverce:
### App
Gumroad

### Triggers
* Cancellation
* New Product
* Refund
* Sale

PR #4474, opened 2022-10-11
Gumroad #3645

lcaresia:


components/gumroad/sources/common/common.mjs
```diff
@@ -0,0 +1,67 @@
...
+    const lastTimestamp = this._getLastTimestamp();
+    this._setLastTimestamp();
+
+    let page = 0;
+
+    while (true) {
+      const response = await this.getResources({
+        params: {
+          page,
+          after: lastTimestamp,
```
jcortes, line 53: I'm just wondering if this actually works because let's say a user made 1 sale and the source got trigger once so it gets one item in the list that is emitted, so the latest timestamp stored is for today, then another user makes another sale in the same day so that means it won't get any sale items for today because it will only get items after today date?
Or is it that after filter is inclusive?

jcortes (commented):
Hi @lcaresia I've just added one observation other than that I think this implementation is duplicated in [this other PR](https://github.com/PipedreamHQ/pipedream/pull/4484/files#diff-fec31cf5ed21edef4646d1b62ed5da026a8b3f8034ac81820a4ad1fc2b77212d) so I guess it's ok but my advice would be not to mix PRs in the future, it will be harder for reviewers to check

components/gumroad/sources/common/common.mjs
lcaresia: Yes, it's inclusive!

jcortes (commented):
Hi @lcaresia LGTM! Ready for QA!

jcortes (approved):
Approved!

PR #4623, opened 2022-10-31
4473 sendbird actions

alysonturing:


dylburger:

components/sendbird/sendbird.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+    reverse: {
+      type: "boolean",
+      label: "Reverse",
+      description: "Determines whether to sort the results in reverse chronological order. If set to **true**, messages appear in reverse chronological order where the newest comes first and the oldest last. (Default: **false**)",
+      optional: true,
+    },
+    senderIds: {
+      type: "string[]",
+      label: "Sender Ids",
+      description: "Restricts the search scope to only retrieve messages sent by one or more users with the specified IDs listed in a comma-separated string.",
```
GTFalcao, line 77: The description suggests this is a comma-separated string, but we're presenting the user with (one or more) options to select.

Also, I noticed it's passed as-is, rather than converted from an array to this comma-separated string (with `Array.join()` or a similar method), but perhaps either the SDK or their API handles this?

Regardless, I think the description should match the user's experience, for example:
```suggestion
      description: "Restricts the search scope to only retrieve messages sent by one or more users selected from the list.",
```

components/sendbird/sendbird.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+      label: "Next Limit",
+      max: 200,
+      min: 0,
+      description: "Specifies the number of sent messages to retrieve after `message_ts`. For example, if `message_ts=1484202848298`, then `next_limit=50` returns 50 messages sent from 1484202848299 (message_ts + 1).",
+      optional: true,
+    },
+    include: {
+      type: "boolean",
+      label: "Include",
+      description: "Determines whether to include messages sent exactly on the specified `message_ts` in the results. (Default: **true**)",
```
GTFalcao, line 65: Since the description states a default value, perhaps that should be set as the `default` property of the prop, for example:
```suggestion
      description: "Determines whether to include messages sent exactly on the specified `message_ts` in the results. (Default: **true**)",
      default: true,
```
so as to make use of the Pipedream UI for default values.

components/sendbird/actions/send-message/send-message.mjs
```diff
@@ -0,0 +1,138 @@
...
+      return {
+        url: {
+          type: "string",
+          label: "URL",
+          description: "Specifies the URL of the file hosted on the server of your own or other third-party companies.",
+        },
+        fileName: {
+          type: "string",
+          label: "File Name",
+          description: "the file name can be set to any string you want using the file name request property. If no file name is supplied, the name response property defaults to an empty string.",
```
GTFalcao, line 99: A tiny suggestion: capitalize the description as all others are
```suggestion
          description: "The file name can be set to any string you want using the file name request property. If no file name is supplied, the name response property defaults to an empty string.",
```

components/sendbird/sendbird.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+    channelType: {
+      type: "string",
+      label: "Channel Type",
+      description: "Specifies the type of the channel. Acceptable values are **open_channels** and **group_channels**.",
+      options: constants.CHANNEL_TYPE_OPTS,
+    },
+    applicationId: {
+      type: "string",
+      label: "Application Id",
+      description: "Specifies the unique ID of the Application, To find the application ID, sign in to [Sendbird Dashboard](https://dashboard.sendbird.com/auth/signin), go to **Settings > Application > General**, and then check Application ID.",
```
GTFalcao, line 17: Tiny grammar suggestion
```suggestion
      description: "Specifies the unique ID of the Application. To find the application ID, sign in to [Sendbird Dashboard](https://dashboard.sendbird.com/auth/signin), go to **Settings > Application > General**, and then check Application ID.",
```

components/sendbird/sendbird.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+    include: {
+      type: "boolean",
+      label: "Include",
+      description: "Determines whether to include messages sent exactly on the specified `message_ts` in the results. (Default: **true**)",
+      optional: true,
+    },
+    reverse: {
+      type: "boolean",
+      label: "Reverse",
+      description: "Determines whether to sort the results in reverse chronological order. If set to **true**, messages appear in reverse chronological order where the newest comes first and the oldest last. (Default: **false**)",
```
GTFalcao, line 71:
```suggestion
      description: "Determines whether to sort the results in reverse chronological order. If set to **true**, messages appear in reverse chronological order where the newest comes first and the oldest last. (Default: **false**)",
      default: false,
```

components/sendbird/sendbird.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+    },
+    message: {
+      type: "string",
+      label: "Message",
+      description: "Specifies the content of the message.",
+    },
+    sendPush: {
+      type: "boolean",
+      label: "Send Push",
+      description: "Determines whether to send a push notification for the message to the members of the channel (applicable to group channels only). Unlike text and file messages, a push notification for an admin message is not sent by default. (Default: true)",
```
GTFalcao, line 125:
```suggestion
      description: "Determines whether to send a push notification for the message to the members of the channel (applicable to group channels only). Unlike text and file messages, a push notification for an admin message is not sent by default. (Default: true)",
      default: true,
```

GTFalcao (changes requested):
Hi @alysonturing , thanks for your contribution!

I've left a few comments, mostly small adjustments regarding grammar and UI/UX.

components/sendbird/sendbird.app.mjs
alysonturing: we cannot do that, because it's going to send the value by default, even tho the property is optional, the API of sendbird can set it as default if it needs, for example, there are properties that cannot be set along another properties, in this case, not set the property is not the same as set an default value even tho it's the same value that they say as its default.

components/sendbird/sendbird.app.mjs
alysonturing: we cannot do that, because it's going to send the value by default, even tho the property is optional, the API of sendbird can set it as default if it needs, for example, there are properties that cannot be set along another properties, in this case, not set the property is not the same as set an default value even tho it's the same value that they say as its default.

components/sendbird/sendbird.app.mjs
alysonturing: we cannot do that, because it's going to send the value by default, even tho the property is optional, the API of sendbird can set it as default if it needs, for example, there are properties that cannot be set along another properties, in this case, not set the property is not the same as set an default value even tho it's the same value that they say as its default.

alysonturing:
Hello Guilherme, I fixed the changes that you asked, but regarding the properties with default value, we cannot do that, there are some properties that cannot be sent along with other properties, and if you put as default, it will be always sent even if its in hide state.
its unpredictable, so I believe that it’s best to rely on the app choice for default value usage, unless we want to specify a default nonetheless… which is not the case.

I’ll share with you an example of another app that presented that behavior
https://github.com/PipedreamHQ/pipedream/pull/4516#issuecomment-1282904766

GTFalcao (approved):
Hi @alysonturing , that makes sense, if there's any possible incompatibility between parameters then I completely agree we should only pass the props explicitly set by the user.

Looks good to me! 👏

DilanAthukorala:
@alysonturing Following issues are observed;

1. Description should be corrected
image


2. send_message action does not list group channels in **channel_url** prop
image


3. send_message action Mentioned User Ids should be an array type prop
image

GTFalcao (approved):
New changes approved!

DilanAthukorala:
Verification completed, ready for release.

PR #4723, opened 2022-11-11
[GH Actions Check] Enable multiple apps for a component

andrewjschuang:
Should be ok using multiple apps in a component, but should check for corresponding nameslug app.

Example: `deployhq-deploy-failed` should contain at least `deployhq` app; can contain other apps.

Noticed this because I'm currently working on a **Google Calendar** component that also needs **Pipedream** app auth.

### DIFF
--- a/scripts/checkComponentAppProp.js
+++ b/scripts/checkComponentAppProp.js
@@ -14,8 +14,12 @@ function checkComponentKey(component, nameslug) {
     console.error(`[!] ${key} - missing app prop`);
     err = true;
   } else if (props.length > 1) {
-    console.error(`[!] ${key} - more than one app prop not expected`);
-    err = true;
+    if (props.map((prop) => prop.app)
+      .filter((prop) => prop === nameslug)
+      .length === 0) {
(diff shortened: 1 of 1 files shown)


andrewjschuang:
@tjk would you mind reviewing this one?

scripts/checkComponentAppProp.js
```diff
@@ -14,8 +14,12 @@ function checkComponentKey(component, nameslug) {
     console.error(`[!] ${key} - missing app prop`);
```
tjk, line 14: This function should be renamed right? It's not `checkingComponentKey` and it could be simplified?

```js
function checkComponentHasAppProp(component, appNameSlug) {
  const matching = Object.values(component.props).filter((prop) => prop.type === "app" && prop.app === appNameSlug);
  if (!matching.length) {
    console.error(`[!] ${component.key} - missing app prop for ${appNameslug}`)
    err = true
  }
}
```

tjk (changes requested):

tjk (approved):

PR #4757, opened 2022-11-16
#3667 - Kanbanflow Actions & Sources

### LINKED ISSUE #3667: KanbanFlow
### App
KanbanFlow

### Triggers
* New Task
* Task Moved

### Actions
* Create Task

GTFalcao:
# New Actions
- **Create Task**

# New Sources
- **Task Created**
- **Task Moved**

Closes #3667

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/4757"></a>


dylburger:

components/kanbanflow/app/kanbanflow.app.ts
```diff
@@ -0,0 +1,123 @@
+import { defineApp } from "@pipedream/types";
+import { axios } from "@pipedream/platform";
+import {
+  Board, CreateHookParams, CreateTaskParams, HttpRequestParams, OptionsObject, Task, User, Webhook,
+} from "../common/types";
+import { getOptionsDescription } from "../common/constants";
```
jcortes, line 6: Optional:
Let me suggest to move this util function to another path like `../common/utils`
```suggestion
import { getOptionsDescription } from "../common/utils";
```

components/kanbanflow/sources/task-moved/task-moved.ts
```diff
@@ -0,0 +1,76 @@
...
+        newData.oldSwimlane = swimlanes.find(
+          ({ uniqueId }) => uniqueId === swimlaneChange.oldValue,
+        )?.name;
+        newData.newSwimlane = swimlanes.find(
+          ({ uniqueId }) => uniqueId === swimlaneChange.newValue,
+        )?.name;
+      }
+
+      delete newData.changedProperties;
+      return newData;
```
jcortes, line 73: Don't you think is better to just return the data as it is from kanbanflow api? instead of modifying the payload object?

jcortes (commented):
Hi @GTFalcao I've just left one suggestion and an observation in one of the source components. Thanks!

components/kanbanflow/sources/task-moved/task-moved.ts
GTFalcao: I could leave the changedProperties object instead of deleting it, as the user could use their values for future steps. I'll undo the 'delete' line here.

But I think it's important to add the names for the columns and swimlanes, instead of just the IDs (as the API returns them). Otherwise, the user would have to write their own code to match these IDs with columns and swimlanes returned from the 'get board' endpoint.

jcortes (commented):
Hi @GTFalcao LGTM! Ready for QA!

DilanAthukorala:
@GTFalcao In Create Task action, unable to pass integer values in Additional Options
image
image
image

GTFalcao:
Hey @DilanAthukorala there are a few options regarding this:
- we can process the additionalOptions prop so that each param is properly typed;
- we can add these remaining options as props in the action - I believe this makes more sense than the above;
- we can change additionalOptions to a string, where the user can input a JSON-stringified object that will be parsed.

In my opinion, having additionalOptions only makes sense if the user can use any new properties that are added to the app's API in the future, not requiring an update to the Pipedream component to do so - so the last option (expecting a JSON-stringified object) would fulfill this.

Note that the `dates` param, if added as a prop, would still need to be a JSON string, since it is an array of objects, which is not currently supported by Pipedream's interface. So I think the most flexible way is to keep additionalOptions, but as a string to be parsed, instead of an object.

What do you think would be best here?

DilanAthukorala:
@GTFalcao Thanks for the explanation so you are suggesting we keep the current implementation where the additionalOptions are passed as string, and if user requires to pass specific properties which are not string then we add that to the interface as a prop?

GTFalcao:
> @GTFalcao Thanks for the explanation so you are suggesting we keep the current implementation where the additionalOptions are passed as string, and if user requires to pass specific properties which are not string then we add that to the interface as a prop?

The current implementation is an object. It'd be more flexible as a string (JSON-stringified object). This way values that are supposed to be numbers, objects, arrays etc. can be parsed from the string.

DilanAthukorala:

jcortes (approved):
Approved!

PR #4782, opened 2022-11-18
3806 testmonitor

luancazarine:
Triggers:
  - New Issue
  - New Test Result

Actions
  - Find an Issue
  - Find a Project
  - Find a Test Result

<a href="https://gitpod.io/#https://github.com/PipedreamHQ/pipedream/pull/4782"></a>


dylburger:

components/testmonitor/actions/find-issue/find-issue.mjs
```diff
@@ -0,0 +1,38 @@
+import common from "../common/base.mjs";
+
+export default {
+  ...common,
+  key: "testmonitor-find-issue",
+  name: "Find an Issue",
+  description: "Retrieve an issue using its identifier. [See the docs here](https://docs.testmonitor.com/#tag/Issues/operation/GetIssue)",
```
jcortes, line 7: Observation:
I was wondering if all these actions actually are refering to kind of a search based on somo text query and filters like you can see in their docs [here](https://docs.testmonitor.com/#tag/Issues/operation/GetIssueCollection)

components/testmonitor/sources/common/base.mjs
```diff
@@ -0,0 +1,87 @@
...
+  },
+  async run() {
+    const dateTime = this._getLastDateTime();
+    const func = this.getFunc();
+
+    const params = {
+      "filter[created_at][start]": dateTime,
+      "filter[created_at][end]": new Date(),
+      "project_id": this.projectId,
+      "order[code]": "desc",
```
jcortes, line 77: I think is better to order by an specified field depending on the component type (e.g. `new-resource`, `updated-resource` etc), something like:
```suggestion
      [`order[this.getOrderField()]`]: "desc",
```
in that way you can set that field accordingly:
```javascript
getOrderField() {
  return "created_at";
},
```
or
```javascript
getOrderField() {
  return "updated_at";
},

File: components/testmonitor/sources/common/base.mjs [hunk 3]
```diff
@@ -0,0 +1,87 @@
...
+      }
+    },
+  },
+  async run() {
+    const dateTime = this._getLastDateTime();
+    const func = this.getFunc();
+
+    const params = {
+      "filter[created_at][start]": dateTime,
+      "filter[created_at][end]": new Date(),
```
jcortes, line 75: I think is not necessary this line right?
```suggestion
```

components/testmonitor/sources/common/base.mjs
```diff
@@ -0,0 +1,87 @@
...
+      default: {
+        intervalSeconds: DEFAULT_POLLING_SOURCE_TIMER_INTERVAL,
+      },
+    },
+  },
+  methods: {
+    _getLastDateTime() {
+      return this.db.get("dateTime");
+    },
+    _setDateTime(dateTime) {
```
jcortes, line 28:
```suggestion
    _setLastDateTime(dateTime) {
```

components/testmonitor/sources/common/base.mjs
```diff
@@ -0,0 +1,87 @@
...
+      const { data } = await func(params);
+
+      for (const item of data) {
+        if (!dateTime || moment(item.created_at).isAfter(dateTime)) {
+          this._setDateTime(item.created_at);
+          dateTime = item.created_at;
+        }
+        this.$emit(item, this.getDataToEmit(item));
+      }
+    },
```
jcortes, line 67: I think this version would be less confussing what do you think?
```suggestion
    async activate() {
      const func = this.getFunc();
      const params = {
        "project_id": this.projectId,
        [`order[${this.getOrderField()}]`]: "desc",
        "limit": 20,
      };

      const { data = [] } = await func(params);

      const [
        lastItem,
      ] = data;

      if (lastItem) {
        this._setLastDateTime(lastItem[this.getOrderField()]);
      }

      for (const item of data) {
        this.$emit(item, this.getDataToEmit(item));
      }
    },
```

components/testmonitor/sources/common/base.mjs
```diff
@@ -0,0 +1,87 @@
...
+    _setDateTime(dateTime) {
+      this.db.set("dateTime", dateTime);
+    },
+    async processEvent({
+      params, dateTime, func,
+    }) {
+      const items = this.testmonitor.paginate({
+        fn: func,
+        params,
+      });
```
jcortes, line 37: I would advise not to paginate in this function which role should only be as the name suggests. So there is a number of examples you can take a look at that you can use as a reference ([activecampaign](https://github.com/PipedreamHQ/pipedream/blob/master/components/activecampaign/sources/common/base.mjs#L21-L24), [amara](https://github.com/PipedreamHQ/pipedream/blob/master/components/amara/sources/common.mjs#L112-L115), [aweber](https://github.com/PipedreamHQ/pipedream/blob/master/components/aweber/sources/common.mjs#L72-L75) etc)

jcortes (commented):
Hi @luancazarine I've added a few request changes so we can standarize better the implementation according to components already tested. Thanks!

components/testmonitor/actions/common/base.mjs
```diff
@@ -0,0 +1,102 @@
...
+      ],
+      optional: true,
+    },
+    page: {
+      propDefinition: [
+        testmonitor,
+        "page",
+      ],
+      optional: true,
+    },
```
michelle0927, line 19: Rather than `limit` and `page` being props, the actions should paginate through the results. I'd suggest adding a `max` prop where the user can set the maximum number of results to return, and then the action should paginate through and return a maximum of `max` results.

components/testmonitor/sources/common/base.mjs
```diff
@@ -0,0 +1,83 @@
+import { DEFAULT_POLLING_SOURCE_TIMER_INTERVAL } from "@pipedream/platform";
+import moment from "moment";
```
michelle0927, line 2: The package.json file should include `moment` as a dependency. You can add it by running `npm install moment` in the /components/testmonitor directory.

michelle0927 (changes requested):
Looking good. Left a couple comments.

components/testmonitor/testmonitor.app.mjs
```diff
@@ -1,11 +1,171 @@
...
+        } = await fn(params);
+        for (const d of data) {
+          yield d;
+        }
+
+        if (maxResults && ++count === maxResults) {
+          return count;
+        }
+        lastPage = !last_page;
+      } while (lastPage);
```
michelle0927, line 168: I don't see that this is actually paginating. We need to use `page` and increment it to get each subsequent page. https://docs.testmonitor.com/#section/Requests/Pagination. Also, the `lastPage` prop isn't boolean. It's actually the page number of the last page.

```suggestion
    async *paginate({
      fn, params = {}, maxResults = null,
    }) {
      let count = 0;
      let lastPage = false;

      params = {
        ...params,
        page: 1,
      }

      do {
        const {
          data,
          meta: { last_page },
        } = await fn(params);
        for (const d of data) {
          yield d;
        }

        if (maxResults && ++count === maxResults) {
          return count;
        }

        lastPage = last_page == params.page;
        params.page++;
      } while (!lastPage);
    },
```

components/testmonitor/yarn.lock
```diff
@@ -0,0 +1,53 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
+# yarn lockfile v1
+
+
+"@pipedream/platform@^0.10.0":
+  version "0.10.0"
```
michelle0927, line 6: I don't think we need this `yarn.lock` file. I think we can remove it.

michelle0927 (changes requested):
Thanks for making those updates. I'm not sure about the pagination though. See my suggestion below.

michelle0927 (approved):
Approved!

Issue #3532, opened 2022-07-25
Ably


jverce:
### App
Ably

### Actions
* Publish Message
* Publish Push Notification

PR #4810, opened 2022-11-23
Ably #3532

lcaresia:


components/ably/actions/publish-message/publish-message.mjs
```diff
@@ -0,0 +1,43 @@
...
+    eventName: {
+      label: "Event Name",
+      description: "The event name that will be published",
+      type: "string",
+    },
+    messageData: {
+      label: "Message Data",
+      description: "The data of the message that will be published",
+      type: "string",
+    },
```
michelle0927, line 25: Looks like `channelName`, `eventName`, and `messageData` props are being used in both `publish-message` and `publish-push-notification`. Could we put these in the app file as propDefinitions?

michelle0927 (dismissed):
Looks good! Just one minor nitpick.

andrewjschuang (dismissed):
LGTM!

Just one question, is there a case in which `response.messageId` is not returned in the response?
```
    if (response.messageId) {
      $.export("$summary", `Successfully pushed notification with ID ${response.messageId}`);
    }
```

vunguyenhung:
Hi @lcaresia, I've added a commit to fix the Publish Push Notification and verified that all test cases are passed! Ready for release! 

Test report
[Ably_3532.pdf](https://github.com/PipedreamHQ/pipedream/files/10109569/Ably_3532.pdf)

andrewjschuang (approved):

PR #4940, opened 2022-12-13
[ACTION] Add inReplyTo & references to gmail-send-email

jc-tzn:
The "References" and "In-Reply-To" headers are necessary to keep an email answer in the same thread as the original email. 

Related documentation:
- Gmail API: https://developers.google.com/gmail/api/guides/sending#sending_messages
- RFC 2822 standard: https://www.rfc-editor.org/rfc/rfc2822#appendix-A.2
- Nodemailer Mailcomposer: https://nodemailer.com/extras/mailcomposer/#e-mail-message-fields

### DIFF
--- a/components/gmail/actions/send-email/send-email.mjs
+++ b/components/gmail/actions/send-email/send-email.mjs
@@ -6,7 +6,7 @@ export default {
   key: "gmail-send-email",
   name: "Send Email",
   description: "Send an email from your Google Workspace email account",
-  version: "0.0.7",
+  version: "0.0.12",
   type: "action",
   props: {
     gmail,
@@ -100,7 +100,7 @@ export default {
         opts.references = inReplyTo;
         opts.threadId = repliedMessage.threadId;
       } catch (err) {
-        throw new Error(`\`${this.inReplyTo}\` is not a valid message ID!`);
+        opts.threadId = this.inReplyTo;
       }
     }
 
--- a/components/gmail/actions/send-email/send-email.mjs
+++ b/components/gmail/actions/send-email/send-email.mjs
@@ -6,7 +6,7 @@ export default {
   key: "gmail-send-email",
   name: "Send Email",
   description: "Send an email from your Google Workspace email account",
-  version: "0.0.6",
+  version: "0.0.7",
   type: "action",
   props: {
     gmail,
--- a/components/gmail/actions/send-email/send-email.mjs
+++ b/components/gmail/actions/send-email/send-email.mjs
@@ -66,13 +66,7 @@ export default {
     inReplyTo: {
       type: "string",
       label: "In Reply To",
-      description: "Specify the message-id this email is replying to.",
-      optional: true,
-    },
-    references: {
-      type: "string",
-      label: "References",
-      description: "Specify the list of message-id in the thread the email is replying to (space-separated string).",
+      description: "Specify the message-id this email is replying to. To use this prop with async options please use `Gmail(Developer App)` `Send Email` component.",
       optional: true,
     },
   },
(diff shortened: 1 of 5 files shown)


dylburger:

dylburger:

components/gmail/actions/send-email/send-email.mjs
```diff
@@ -63,6 +63,12 @@ export default {
       description: "Add any attachments you'd like to include as objects. The `key` should be the **filename** and the `value` should be the **url** for the attachment. The **filename** must contain the file extension (i.e. `.jpeg`, `.txt`) and the **url** is the download link for the file.",
       optional: true,
     },
+    inReplyTo: {
+      type: "string",
+      label: "In Reply To",
+      description: "Specify the message-id this email is replying to. To use this prop with async options please use `Gmail(Developer App)` `Send Email` component.",
```
andrewjschuang, line 69:
```suggestion
      description: "Specify the `message-id` this email is replying to. To use this prop with `async options` please use `Gmail (Developer App)` `Send Email` component.",
```

Also we should remove the `replyTo` prop.

andrewjschuang (approved):
Looks great! I just added some nitpicks

andrewjschuang (changes requested):
Actually, I copied the `message-id` from `async options` and pasted to the **Gmail** app and tried sending the email, but it said that the value is not a valid message ID.

The **Gmail Developer App** worked fine though.

feyzullah:
> Actually, I copied the `message-id` from `async options` and pasted to the **Gmail** app and tried sending the email, but it said that the value is not a valid message ID.
> 
> The **Gmail Developer App** worked fine though.

Oops! This is again because of the insufficient scopes. Let me remove it from there and add directly to threadId.

feyzullah:
Now this should be able to reply non-replied emails. If a replied message-id would be given Gmail API will return an error because `threadId` is not existing. We can remove this prop from Gmail app and keep it in Gmail (Developer App). What do you think @andrewjschuang?

andrewjschuang:
Oh, so in the **Gmail** app `threadId` won't work? If so, yeah we can leave it only on the **Gmail Developer App**

feyzullah:
> Oh, so in the **Gmail** app `threadId` won't work? If so, yeah we can leave it only on the **Gmail Developer App**

It works for the first message. If replying to a reply, it will fail. Then I am removing the prop from Gmail app

andrewjschuang:
> It works for the first message. If replying to a reply, it will fail. Then I am removing the prop from Gmail app

Ok, I see. After testing a little bit more I have a few suggestions:

- `opts.inReplyTo` and `opts.references` are actually not needed (delete lines 96-97, 99-100).
- `inReplyTo` prop description in **Gmail** app: add that it must be the `message-id` of the first message sent in the thread.
- In the **Gmail** app, if the error is `Requested entity was not found` (`message-id` was not for the first message in the thread), change the error message to: **Unable to find email thread. `In Reply To` must be the `message-id` of the first message of the thread.**

Other than that looks great to me!

feyzullah:
> 

@andrewjschuang I have applied second and third suggestions. However when I don't send `inReplyTo` and `refenreces`, sender side is able to cluster all under one thread ID but receiver side is not. The following happens for sender side;
sender

And for the receiver side;
receiver

Am I missing something here? Or is there some other config to be done?

andrewjschuang:
> @andrewjschuang I have applied second and third suggestions. However when I don't send `inReplyTo` and `refenreces`, sender side is able to cluster all under one thread ID but receiver side is not.

Oh ok, sorry please keep it that way

andrewjschuang (approved):

PR #4999, opened 2022-12-19
[DigitalOcean Spaces] New components

### LINKED ISSUE #4829: [Triggers and Actions] DigitalOcean Spaces
**Name of app / service**
DigitalOcean Spaces (it has it's own keys, so it's a bit different than the regular DigitalOcean API

**Link to developer documentation**
https://docs.digitalocean.com/reference/api/spaces-api/

We can likely reuse most of the code from the S3 components since it was designed to be interchangeable

> The Spaces API aims to be interoperable with Amazon’s AWS S3 API. In most cases, when using a client library, setting the “endpoint” or “base” URL to ${REGION}.digitaloceanspaces.com and generating a Spaces key pair to replace your AWS IAM key pair will allow you to use Spaces in place of S3.
> 
> Spaces provides support for create, read, update, and delete (CRUD) operations for both buckets (Spaces) and objects as well as the ability to define access control lists (ACLs). Some S3 features are not supported... https://docs.digitalocean.com/reference/api/spaces-api/#aws-s3-compatibility

**Are there specific actions, or triggers, you'd like to see for this app? Please let us know here or use the Action and Trigger issue templates to open requests for each!**
More or less the same triggers and actions we have for AWS S3

**Triggers**
- New file uploaded
- Deleted file

**Actions**
- Upload file (from /tmp, from base64, from URL)
- Delete file

andrewjschuang:
**Triggers:**

- File Uploaded
- File Deleted

**Actions:**

- List (Bucket) Files
- Upload File (from /tmp, from base64, from URL)
- Delete Files

Sources are using time interval since base AWS triggers are using SNS Topics and not webhooks.

Resolves #4829.


lcaresia (approved):
LGTM!

PR #5005, opened 2022-12-20
AWS DynamoDB New Source - New DynamoDB Stream Event

### LINKED ISSUE #79: [TRIGGER] DynamoDB Stream

michelle0927:
Resolves #79


lcaresia (dismissed):
LGTM!

vunguyenhung:

[DynamoDB_5005_1827.pdf](https://github.com/PipedreamHQ/pipedream/files/10277228/DynamoDB_5005_1827.pdf)

vunguyenhung:
Hello @michelle0927, I've tested this again and still faced the same issue that only one event got emitted


Could you try to reproduce on your side? I do the exact steps as in the previous report

michelle0927:
@vunguyenhung I think the issue might be related to whether the current stream is enabled or not. When a stream is disabled, it is no longer able to receive events. When "Enable" is clicked, it creates a brand new stream. I updated the `streams` prop to only return enabled streams. Additionally, if the stream becomes disabled, the logs will display an error.

vunguyenhung:
Hello @michelle0927, could you confirm that you reproduced the issue on your side? I've tested it today and couldn't receive any events (your improvement on the stream seems good though). Here're the steps I've taken

1. I created a new AWS Dynamo DB table
image

2. Then I enabled the stream feature
image

3. Then I created a new Pipedream source using the new stream
image


4. Then on the Dynamo DB, I inserted multiple items, switch back to Pipedream to fetch for new events, still I only see one event



5. I created some more items on Dynamo DB, switch back to Pipedream to fetch for new events, still I see only one new event for one item. **I expect it to emit one event for each new item**
image
image


Could you try to reproduce this on your side and get back to me?

lcaresia (approved):

Issue #3790, opened 2022-07-25
Spondyr


jverce:
### App
Spondyr

### Triggers
* Spondyr Processed

### Actions
* Create Spondyr

lcaresia:
@jverce @dannyroosevelt 

I tried to `create a new Spondyr` in Zapier platform, but I can't find what exatly is this.

Can you explain me better?

Thanks!

Hassan-Abbas-SQA:
TestPassed
![image](https://user-images.githubusercontent.com/104946677/209856233-d3383655-f906-48a6-a3de-36a23c5defe2.png)

TestFailed
![image](https://user-images.githubusercontent.com/104946677/209856252-37534225-7810-46d2-9539-862b64dd101a.png)

Hassan-Abbas-SQA:
![image](https://user-images.githubusercontent.com/104946677/210002066-9e671d11-dbc0-4c57-bc37-4f46c44b7dca.png)

PR #5044, opened 2022-12-27
Spondyr #3790

luancazarine:
Triggers
  - Spondyr Processed

Actions
  - Create Spondyr

### DIFF
--- a/components/spondyr/actions/create-spondyr/create-spondyr.mjs
+++ b/components/spondyr/actions/create-spondyr/create-spondyr.mjs
@@ -1,0 +1,55 @@
+import spondyr from "../../spondyr.app.mjs";
+
+export default {
+  key: "spondyr-create-spondyr",
+  name: "Create Spondyr",
+  version: "0.0.3",
+  description: "Generate and optionally deliver correspondence. [See the docs here](https://client.spondyr.io/#/Public/Public/Documentation?Section=send-api)",
+  type: "action",
+  props: {
+    spondyr,
+    transactionType: {
+      propDefinition: [
+        spondyr,
+        "transactionType",
+      ],
+    },
+    eventType: {
+      propDefinition: [
+        spondyr,
+        "eventType",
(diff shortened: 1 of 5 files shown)


components/spondyr/actions/create-spondyr/create-spondyr.mjs
```diff
@@ -0,0 +1,55 @@
+import spondyr from "../../spondyr.app.mjs";
+
+export default {
+  key: "spondyr-create-spondyr",
+  name: "Create Spondyr",
+  version: "0.0.3",
```
alysonturing, line 6: please use version 0.0.1

components/spondyr/actions/create-spondyr/create-spondyr.mjs
```diff
@@ -0,0 +1,55 @@
...
+    isGenerateOnly: {
+      type: "boolean",
+      label: "Is Generate Only",
+      description: "Set this flag to true to pause/defer delivery until a followup call is made via the Deliver API. This is useful for scenarios where you wish to view/preview the generated correspondence and validate things before allowing the correspondence to be delivered.",
+      optional: true,
+    },
+    data: {
+      type: "object",
+      label: "Data",
+      description: "It should contain your application's specific data, which should be in the same format as you defined for the `Transaction Type`",
```
alysonturing, line 35: is it possible to include a doc link to the expected format of this object? or an example of the expected object?

alysonturing (changes requested):
Thanks for your contribution, I just noticed few points, please verify if the comments added are applicable.

components/spondyr/actions/create-spondyr/create-spondyr.mjs
luancazarine: actually this can be any data that makes sense to the user

alysonturing (approved):
LGTM!

PR #5084, opened 2023-01-02
Action to search multiple Todoist tasks

Alejandro-FA:
Currently, the only option to search Todoist tasks with a Pipedream action is to use [find-task](https://github.com/PipedreamHQ/pipedream/blob/master/components/todoist/actions/find-task/find-task.mjs). That action is designed to return just 1 single task given an exact name, and create it if it does not exist.

While this is useful for certain scenarios, it does not cover other possible needs (like retrieving all tasks that have a given label). This can be accomplished with the proposed action, which receives some filtering options and returns all the matching tasks from a Todoist account.


dylburger:

pipedream-component-development:

pipedream-component-development:

components/todoist/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,64 @@
...
+    } = this;
+    const tasks = await this.todoist.getActiveTasks({
+      $,
+      params: {
+        label: label,
+        project_id: project,
+        section_id: section,
+      },
+    });
+    let result = tasks.filter((task) => (task.content).includes(name));
```
GTFalcao, line 55: I believe you meant this instead:
```suggestion
    let result = tasks.filter((task) => (task.content.includes(name)));
```

Also, here you could check if `name` actually received a value from the user - else, no need to filter at all:
```suggestion
    let result = name ? tasks.filter((task) => (task.content.includes(name))) : tasks;
```

components/todoist/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,64 @@
...
+    let summary = `${result.length} task${result.length == 1
+      ? ""
+      : "s"} found`;
+    $.export("$summary", summary);
+    return result.length > 0
+      ? result
+      : undefined;
```
GTFalcao, line 62: For consistency, I would return the value itself even if it has zero length - so as to avoid breaking future steps that may rely on an array result from this component.
```suggestion
    return result;
```

GTFalcao (commented):
Hi @Alejandro-FA , thank you for your contribution!

I've made a few comments regarding small adjustments. Let me know if you have any questions.

Also, the `todoist` app's `package.json` file should have its version incremented (i.e. from `0.0.5` to `0.1.0`) to reflect the package being updated with this new component.

Alejandro-FA:
I've updated the `package.json` file to version `0.1.0` and applied your proposed changes. If there's anything more than I can do I'm open for it.

GTFalcao (commented):
Thanks @Alejandro-FA , this looks good!

Using a RegExp is a cool idea but I agree with keeping it simple as a string. If this is ever needed in the future, it can be included as a separate property.

Code is approved and I'm moving this issue to the QA stage.

DilanAthukorala:
@Alejandro-FA Please check the following issue

Unable to search tasks by Label
image

Alejandro-FA:
I've checked the issue of not being able to search tasks by label. The error was derived from the `label` prop, which returns the label ID instead of the label name (sorry I didn't catch this before). In this particular API call, we need to pass the label name instead of the ID.

So as I see it, there are 2 possible solutions:

1. To do an extra API call to get the label information given the label ID:

```node
const label_name = (await this.todoist.getLabels({id: label})).name;
const tasks = await this.todoist.getActiveTasks({
  $,
  params: {
    label: label_name,
    project_id: project,
    section_id: section,
  },
});
```

2. To create a new `labelName` prop either in the action file or the `todoist.app.mjs` file. Here is an example of implementation in the action file:

```node
labelName: {
  type: "string",
  label: "Label",
  description: "Select a label to filter results by",
  optional: true,
  async options() {
    return (await todoist.getLabels({})).map((label) => label.name);
  },
},
```

I prefer the second option simply because it doesn't require the extra API call. However, it is a bit redundant and it breaks the preferred style of only using prop definitions from the `.app` file. At the moment I've committed a solution using the first approach, but I'm obviously open to change if you desire so.

DilanAthukorala:
@Alejandro-FA Thanks for the changes, This looks good!

 Read for release!

GTFalcao (commented):
Nicely done @Alejandro-FA ! Thanks for the deep dive into this!

The two options you presented make sense, though in the first one we are querying for information at runtime that we already had before (in the async options call). So using options with the value as the name instead of the id would make more sense (your second option).

But it turns out you don't have to duplicate that prop - there is already a `labelString` propDefinition implemented that uses the labels' names instead of their ids. It's a multi-option selection (type `string[]`) but you could use propDefinition and override the type to use `string`, thus fitting your needs without needing to make an extra API call nor define a new prop.

I've commented below on the code. Let me know if it makes sense.

Alejandro-FA:
Yes of course, it makes perfect sense, it didn't occur to me that we could override the prop type as well. Thank you!

GTFalcao:
@DilanAthukorala let me know if/when this can be merged - the code is ready and approved

DilanAthukorala:
@GTFalcao Sorry I missed your comment. I verified the changes again. No issues Observed, Ready for merge.

GTFalcao (approved):
@DilanAthukorala my mistake as well - I am approving and merging this now.

Thank you once more for the contribution @Alejandro-FA 🚀

PR #5088, opened 2023-01-03
Supabase update to Select Row action

### LINKED ISSUE #5059: [ACTION] Pipedream >> Supabase queries
Pipedream >> Supabase queries

I would like to be able to query Supabase database/table by
- show latest N records
- show records where column contains "querystring"
Thanks

michelle0927:
Resolves #5059

### DIFF
--- a/components/supabase/supabase.app.mjs
+++ b/components/supabase/supabase.app.mjs
@@ -26,7 +26,7 @@ export default {
       description: "Enter the column names and values as key/value pairs",
     },
     filter: {
-      type: "integer",
+      type: "string",
       label: "Filter",
       description: "Filter type for the query",
       options: constants.FILTER_OPTIONS,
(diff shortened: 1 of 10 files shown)


components/supabase/supabase.app.mjs
```diff
@@ -24,18 +25,143 @@ export default {
...
       const client = await this._client();
-      const { data } = await client
+      const args = {
+        table: opts.table,
+        column: opts.column,
+        value: opts.value,
+        orderBy: opts.orderBy,
+        ascending: opts.sortOrder === "ascending"
+          ? true
+          : false,
```
GTFalcao, line 48:
```suggestion
        ascending: opts.sortOrder === "ascending",
```

components/supabase/supabase.app.mjs
```diff
@@ -24,18 +25,143 @@ export default {
...
+      table, column, value, orderBy, ascending, max,
+    }) {
+      return client
+        .from(table)
+        .select()
+        .eq(column, value)
+        .order(orderBy, {
+          ascending,
+        })
+        .limit(max);
```
GTFalcao, line 80: I think we could centralize the common parts here, since the only thing that changes between each of these is the filtering method. How about something like:

```
async baseFilterStart(client, { table }) {
  return client
    .from(table)
    .select();
},
async baseFilterEnd(obj, {
  ascending, max, orderBy
}) {
  return obj
    .order(orderBy, {
      ascending,
    })
    .limit(max);
},
```

So the equalTo method, for example, would be changed to only the unique filtering method:
```suggestion
    async equalTo(obj, {
      column, value,
    }) {
      return obj
        .eq(column, value);
```

So the response would be obtained by calling baseFilterStart, equalTo and baseFilterEnd, each with the result of the previous method. If all of these have to be promises (and not just the initial query), this succession of three methods would require chaining `.then` or `await`ing each of them.

But I think it's best to keep the shared logic in shared methods - what do you think? @michelle0927

GTFalcao (commented):
Hi @michelle0927 , thanks for your contribution!

I made a comment about reusing the common logic on the filtering methods - let me know if that sounds right or not.

GTFalcao (approved):
Looks good to me! 👏

DilanAthukorala:

PR #5121, opened 2023-01-05
Firebase: allow user to select database region

### LINKED ISSUE #5118: [IMPROVEMENT] Firebase: allow to change Database URL
Users might create the Firebase database in a different region that requires different Database URL. For example, the database created in EU might have the database URL:`https://stripe-example-199c9-default-rtdb.europe-west1.firebasedatabase.app`. 

But in [the Pipedream application code here](https://github.com/PipedreamHQ/pipedream/blob/146802e44c9aafd8bbba59a34b5f0c79ef2edef0/components/firebase_admin_sdk/firebase_admin_sdk.app.mjs#L79), it is always use default URL: `https://${projectId}-default-rtdb.firebaseio.com/`

Add an optional Database URL prop that allow the user to input customized database url

**Do you have a workaround?**
No

**Additional context**
First reported by a user here: https://pipedream.com/community/t/firebase-database-lives-in-a-different-region/4119

michelle0927:
Resolves #5118

### DIFF
--- a/components/firebase_admin_sdk/actions/update-document/update-document.mjs
+++ b/components/firebase_admin_sdk/actions/update-document/update-document.mjs
@@ -5,7 +5,7 @@ export default {
   key: "firebase_admin_sdk-update-documents",
   name: "Update Documents",
   description: "Updates a Document. [See the docs here](https://googleapis.dev/nodejs/firestore/latest/DocumentReference.html#update)",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "action",
   props: {
     ...common.props,
@@ -13,6 +13,9 @@ export default {
       propDefinition: [
         common.props.firebase,
         "collection",
+        (c) => ({
+          region: c.databaseRegion,
+        }),
       ],
     },
     document: {
@@ -21,6 +24,7 @@ export default {
         "document",
         (c) => ({
           collection: c.collection,
+          region: c.databaseRegion,
         }),
       ],
     },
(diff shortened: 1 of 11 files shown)


components/firebase_admin_sdk/actions/update-document/update-document.mjs
```diff
@@ -2,17 +2,20 @@ import common from "../common/base.mjs";
 
 export default {
   ...common,
-  key: "firebase_admin_sdk-update-documents",
+  key: "firebase_admin_sdk-update-document",
```
GTFalcao, line 5: Are users able to update their workflows despite the component key being changed?

GTFalcao (commented):
Hi @michelle0927 , thanks for your contribution, looks good to me!

I just have one question regarding changing the component key - are users able to update the component in their workflows despite this change? Not sure if this could be a problem.

michelle0927:
> Hi @michelle0927 , thanks for your contribution, looks good to me!
> 
> I just have one question regarding changing the component key - are users able to update the component in their workflows despite this change? Not sure if this could be a problem.

@GTFalcao I think they'd have to recreate the step in a workflow. The other option is to leave the key as `firebase_admin_sdk-update-documents` and change the file and folder names to `update-documents`. Which do you think would be better?

GTFalcao (approved):
@michelle0927 in that case, seems it'd be less troublesome to keep it this way 😅 approved

PR #5145, opened 2023-01-10
[ACTIONS/SOURCES] Plivo new actions sources

jcortes:
Dev and QA notes:

I decided to split `Make a call and play audio` into two different actions because I didn't find a way to make it work in just one go. That's mainly because once you make a call, a callback to `answer_url` property should be trigger in order to get the real `call_uuid` for then to play audio on it, so because it is put in an action there is no way to wait or listen for that response. And I did the same for `Make a call and speak text`.


components/plivo/actions/play-audio/play-audio.mjs
```diff
@@ -0,0 +1,46 @@
...
+      ],
+    },
+    url: {
+      type: "string",
+      label: "Media URL",
+      description: "A URL linking to an `.mp3` or `.wav` file.",
+    },
+  },
+  methods: {
+    playAudioOnCall(args = []) {
```
luancazarine, line 24: Couldn't you put this method in the app's base file?

components/plivo/actions/speak-text/speak-text.mjs
```diff
@@ -0,0 +1,48 @@
...
+    text: {
+      propDefinition: [
+        app,
+        "text",
+      ],
+      description: "The text that needs to be spoken.",
+    },
+  },
+  methods: {
+    speakText(args = []) {
```
luancazarine, line 26: Same question here, Couldn't you put this method in the app's base file?

luancazarine (changes requested):
Hello @jcortes. Thank you for your contribution! I've just add a minor suggestion.

luancazarine (approved):
LGTM!

PR #5167, opened 2023-01-12
[RegFox] New Triggers

### LINKED ISSUE #3758: RegFox
### App
RegFox

### Triggers
* List Forms
* New Deposit Payment
* New Registrant

andrewjschuang:
**Triggers:**

- Form Published
- Deposit Completed
- Registrant Applied

Found a bug in the RegFox API regarding creation of Webhooks, so prefixed the webhook-based sources with `common-` so they don't get published. RegFox doesn't have an ETA for the fix.

Resolves #3758.


components/regfox/sources/common/base.mjs
```diff
@@ -0,0 +1,81 @@
...
+      const response = await this.regfox.createWebhook({
+        url: this.http.endpoint,
+        events: this.eventTypes(),
+        forms: this.forms.map((form) => ({
+          formId: form,
+        })),
+      });
+
+      console.log(`Webhook ID: ${response.id} created succesfully`);
+      this.setWebhookId(response.id);
```
michelle0927, line 52:
```suggestion
      this._setWebhookId(response.id);
```

components/regfox/sources/common/base.mjs
```diff
@@ -0,0 +1,81 @@
...
+      this.setSigningSecret(response.signingSecret);
```
michelle0927, line 53: I don't see the signing secret being used anywhere.
```suggestion
```

components/regfox/sources/common/base.mjs
```diff
@@ -0,0 +1,81 @@
...
+    },
+    async deactivate() {
+      const id = this.getWebhookId();
```
michelle0927, line 56:
```suggestion
      const id = this._getWebhookId();
```

components/regfox/sources/common/base.mjs
```diff
@@ -0,0 +1,81 @@
...
+    },
+    _setWebhookId(id) {
+      this.db.set("webhookId", id);
+    },
+    _getSigningSecret() {
+      return this.db.get("signingSecret");
+    },
+    _setSigningSecret(signingSecret) {
+      this.db.set("signingSecret", signingSecret);
+    },
```
michelle0927, line 73: I don't see the signing secret being used anywhere.
```suggestion
```

michelle0927 (changes requested):
Just a couple comments

components/regfox/sources/common/base.mjs
andrewjschuang: It's still not being used because there's a bug on the RegFox API so webhooks can't be created through the API. Don't worry, the webhooks components won't be published.

andrewjschuang: It's still not being used because there's a bug on the RegFox API so webhooks can't be created through the API. Don't worry, the webhooks components won't be published.

components/regfox/sources/common/base.mjs
andrewjschuang: It's still not being used because there's a bug on the RegFox API so webhooks can't be created through the API. Don't worry, the webhooks components won't be published.

michelle0927 (dismissed):
Approved!

DilanAthukorala:
Verification completed and ready for release.

- Instant Webhooks are not supported by RegFox
- Deposit & Registrant sources require payment, therefore unable to trigger event but all other sources and behaviors are verified.

michelle0927 (approved):

PR #5250, opened 2023-01-20
[Pipedream Platform Axios] Add more info to the axios error

### LINKED ISSUE #5218: [FEATURE] Send the `response` property to the global error handler for AxiosErrors
I have a workflow, called Error Handler, that is triggered anytime any other workflows has an uncaught exception. 

Among other things, the Error Handler logs the properties of the error for debugging. This works well for most errors except those triggered by Axios.

For most errors, the message and the stack trace are enough to debug but for some AxiosError there is another property that is really useful: `error.response`. For example:

```
{
  message: 'Request failed with status code 400',
  name: 'AxiosError',
  code: 'ERR_BAD_REQUEST',
  response: {
    status: 400,
    statusText: 'Bad Request',
    headers: {...},
    data: { 
      error: {
        message: 'Unsupported post request...',
        type: 'GraphMethodException',
        code: 100,
      }
    }
  }
}
```

The response property (when it exists) should be sent to the error handler along with the message and the stack trace.

**Do you have a workaround?**
Wrapping every Axios requests in a try-catch then throwing a new error with the response property concatenated to the message property.

andrewjschuang:
Workaround in `@pipedream/platform axios` that changes from:

image
to:
image

Resolves #5218.

### DIFF
--- a/platform/dist/axios.js
+++ b/platform/dist/axios.js
@@ -104,6 +104,7 @@ async function default_1(step, config, signConfig) {
     }
     catch (err) {
         if (err.response) {
+            convertAxiosError(err);
             stepExport(step, err.response, "debug");
         }
         throw err;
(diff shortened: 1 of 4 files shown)


andrewjschuang:
Successfully tested by the customer.

Issue #5299, opened 2023-01-24
[SOURCE] Add Emoji prop to New Reaction Added Slack Source


dannyroosevelt:
Let's add an optional prop to the New Reaction Added Slack Source so users can filter events on the specific emoji(s) they care about.

Hassan-Abbas-SQA:
![image](https://user-images.githubusercontent.com/104946677/214391390-670108cb-0983-4c89-83c9-d33a5a95eba1.png)

PR #5305, opened 2023-01-24
[SOURCE] Add Emoji prop to New Reaction Added Slack Source #5299

lcaresia:


components/slack/sources/new-reaction-added/new-reaction-added.mjs
```diff
@@ -44,14 +44,25 @@ export default {
...
       ],
     },
+    iconEmoji: {
+      propDefinition: [
+        common.props.slack,
+        "icon_emoji",
+      ],
+      description: "Provide an emoji to use as filter the events. E.g. `fire`",
+      optional: false,
+    },
```
michelle0927, line 54: What do you think about letting users select multiple emojis to emit events for? You could also make this prop optional, so that if the user doesn't select any emojis, it'll emit events for all emojis.
```suggestion
    iconEmoji: {
      propDefinition: [
        common.props.slack,
        "icon_emoji",
      ],
      description: "Provide an emoji to use as filter the events. E.g. `fire`",
      type: "string[]",
    },
```

components/slack/sources/new-reaction-added/new-reaction-added.mjs
```diff
@@ -44,14 +44,25 @@ export default {
...
   methods: {
     ...common.methods,
     getSummary() {
       return "New reaction added";
     },
     async processEvent(event) {
-      if ((this.ignoreBot) && (event.subtype == "bot_message" || event.bot_id)) {
+      if (
+        ((this.ignoreBot) && (event.subtype == "bot_message" || event.bot_id)) ||
+        (this.iconEmoji && event.reaction !== this.iconEmoji)
```
michelle0927, line 64: With `this.iconEmoji` as an array, you could check if it includes the event reaction.
```suggestion
        (this.iconEmoji?.length > 0 && !this.iconEmoji.includes(event.reaction))
```

michelle0927 (changes requested):
Looks great! Just one suggestion.

michelle0927 (approved):
LGTM!

michelle0927 (approved):

PR #5066, opened 2022-12-30
Twitter Upload Media Action

michelle0927:
Resolves #5036 

Note: Uploading large files may require increasing the Workflow's timeout setting. 
Settings -> Execution Controls -> Timeout

PR #5308, opened 2023-01-24
Twitter Upload Media Action bug fix

michelle0927:
Error "SegmentIndex must be between 0 and 999." observed occurring for 4MB files. This PR increases the segment size to reduce the number of segments to less than 999 and prevent the error from occurring.

See #5066

### DIFF
--- a/components/twitter/actions/upload-media/upload-media.mjs
+++ b/components/twitter/actions/upload-media/upload-media.mjs
@@ -6,7 +6,7 @@ export default {
   key: "twitter-upload-media",
   name: "Upload Media",
   description: "Upload images or other media types to Twitter. Returns a `media_id_string` to be used with the Create Tweet action. [See the docs here](https://developer.twitter.com/en/docs/tutorials/uploading-media)",
-  version: "0.0.3",
+  version: "0.0.2",
   type: "action",
   props: {
     twitter,
--- a/components/twitter/actions/upload-media/upload-media.mjs
+++ b/components/twitter/actions/upload-media/upload-media.mjs
@@ -23,6 +23,17 @@ export default {
         return Object.values(mime._types);
       },
     },
+    mediaCategory: {
+      type: "string",
+      label: "Media Category",
+      description: "The category that represents how the media will be used",
+      options: [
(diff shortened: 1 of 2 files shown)


components/twitter/actions/upload-media/upload-media.mjs
```diff
@@ -41,7 +41,7 @@ export default {
     });
 
     // APPEND
-    const size = 5012;
+    const size = 8000;
```
andrewjschuang, line 44: As per [the docs](https://developer.twitter.com/en/docs/tutorials/uploading-media) the max chunk size is 1MB, should we use this value?

```suggestion
    const size = 1000000;
```

michelle0927: @andrewjschuang I found that setting the value higher resulted in an error saying that the request headers were too large.

andrewjschuang: Higher than 8000?

andrewjschuang (dismissed):

andrewjschuang (approved):
LGTM!

Hassan-Abbas-SQA:
TEST PASSED
200 Seconds 4.5 mb mp4 video
image

gabriel-kaam:
Hello @michelle0927 , me again 🥲

I have the same issue with this file

Capture d’écran 1657

But this time , the file can be properly uploaded via the Twitter UI , so the format is correct .

Video is 14.7mo , fresh out of a CloudConvert conversion .

Here's the MP4 file is you want to try: https://transfer.sh/tvnYmC/7E4267863858A5842A69CA17A1D7ED92_video_dashinit.mp4 

I feel bad for disturbing you with my problems ; is there a way for me to enable a Debug mode on my Workflow to get more insights on what's going wrong ?

michelle0927:
@gabriel-kaam I'm so sorry you keep seeing these errors. I've created one more PR (https://github.com/PipedreamHQ/pipedream/pull/5445) that should fix the issue. After updating, I was able to successfully upload and create a tweet with the .mp4 video you linked to above.

PR #5309, opened 2023-01-24
5259 featureaction blogger create post action new prop

luancazarine:
-  Added prop labels to create/update post


components/blogger/actions/update-post/update-post.mjs
```diff
@@ -37,13 +37,19 @@ export default {
...
   async run({ $ }) {
     const {
       blog,
       post,
       title,
       content,
+      labels,
     } = this;
 
     let postData;
```
michelle0927, line 55: I think we need to get the postData if labels doesn't exist.

`if (!content || !title) {`
should be:
`if (!content || !title || !(labels?.length > 0)) {`

michelle0927 (changes requested):
Just one comment!

components/blogger/actions/update-post/update-post.mjs
michelle0927: Just realized it has to exist because it's not optional. Could we make it optional in case someone wants to update just the title or content of a post?

michelle0927 (approved):

PR #5314, opened 2023-01-25
Feature/datastore 5223

### LINKED ISSUE #5223: Search Data Store key for particular value [ACTION]
This Data Store lacks a feature to check for the existence of a record within a key. I can do this inside Python but I wonder if a dedicated action would be faster/lighter.

Say for example we have a few values stored in a data store and I want to find a match (say 12345):
```
[
  67890,
  12345,
  7063578509386
]
```
I've tried every Data Store action relating to reading records but they dump back the whole JSON, Pipedream Filter is unable to search the records so cannot provide a match to stop execution with. so I guess an action such as  
`Search key for matching value`  
with a TRUE/FALSE return should do the trick. 

Any help on this would be great, I have no issue with the Python method but this seems a fairly easy win for those just wanting to avoid coding.

vellames:
Closes #5223


michelle0927 (dismissed):
LGTM!
FYI, there are 2 datastore PRs open right now, so whichever one merges last will need to update the `package.json` version again!

michelle0927 (approved):

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,114 @@
...
+    },
+    query: {
+      type: "string",
+      label: "Query",
+      description: "The query to filter the records.",
+    },
+    contains: {
+      type: "boolean",
+      label: "Contains",
+      description: "If true, the query will be used to filter the records that contains the query. If false, the query will be used to filter the records that are equal to the query.",
```
andrewjschuang, line 25:
```suggestion
      description: "If `true`, the query will be used to filter the records that contains the query. If `false`, the query will be used to filter the records that are equal to the query.",
```

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,114 @@
...
+      optional: true,
+    },
+    caseInsensitive: {
+      type: "boolean",
+      label: "Case Insensitive",
+      description: "if true, the query will be used to filter the records based on case-Insensitivity, if false, the query will be used to filter the records that are equal to the query.",
```
andrewjschuang, line 31:
```suggestion
      description: "If `true`, the query will be used to filter the records based on case-insensitivity. If `false`, the query will be used to filter the records that are equal to the query.",
```

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,114 @@
...
+      objData,
+      this.query,
+      this.contains,
+      this.caseInsensitive,
+    );
+
+    $.export("$summary", `Successfully returned ${filteredKeys.length} key(s).`);
+    return {
+      hasRegisteredKeys: filteredKeys.length > 0,
+      keys: filteredKeys,
```
andrewjschuang, line 111: Would it be possible for us to consider returning the values as well?
Since using `contains` may result in values that are different from the query, it might be beneficial to include them.

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,114 @@
...
+          filteredKeys.push(path.slice(0, -1));
+          return;
+        }
+
+        if (!contains && caseInsensitive
+          && String(obj).toLocaleLowerCase() == query.toLowerCase()
+        ) {
+          filteredKeys.push(path.slice(0, -1));
+          return;
+        }
```
andrewjschuang, line 81: I find this section of the code somewhat challenging to understand.

There are numerous base cases for the recursive portion, which can be a bit overwhelming. Additionally, the second part contains an excessive amount of conditions.

I am also not entirely sure of the logic behind the combination of the path and the index.

Furthermore, it appears that some of the code is repeated, which could potentially be improved.

Lastly, I'm not sure if `query` is being included in the scope of the helper function.

andrewjschuang (changes requested):
I took the initiative to review this code as I noticed it could benefit from some modifications.

components/data_stores/actions/get-record-keys/get-record-keys.mjs
andrewjschuang: There are other ways to make the code more readable. For example, create methods:
```js
containsQuery(obj, query, caseSensitive = false) {
  obj = String(obj);
  return caseSensitive
    ? obj.includes(query)
    : obj.toLowerCase().includes(query.toLowerCase());
}

matchesQuery(obj, query, caseSensitive = false) {
  obj = String(obj);
  return caseSensitive
    ? obj === query
    : obj.toLowerCase() === query.toLowerCase();
}
```

And then use the returned `boolean` value to
```js
if (matched) {
  filteredKeys.push(path.slice(0, -1));
  values.push(obj);
}
```

andrewjschuang: This is just my preference but for me it is clearer to use `caseSensitive` instead of `caseInsensitive`. Especially when having to negate it twice, e.g. `!caseInsensitive`.

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,133 @@
...
+      label: "Query",
+      description: "The query to filter the records.",
+    },
+    contains: {
+      type: "boolean",
+      label: "Contains",
+      description: "If `true`, the query will be used to filter the records that contains the query. If `false`, the query will be used to filter the records that are equal to the query.",
+      optional: true,
+    },
+    caseInsensitive: {
```
andrewjschuang, line 28:
```suggestion
    caseSensitive: {
```

Optional, but highly recommended. [See this comment.](https://github.com/PipedreamHQ/pipedream/pull/5314#discussion_r1090873831)

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,133 @@
...
+      obj,
+      query,
+      contains = false,
+      caseInsensitive = false,
+      path = "",
+    ) {
+      const filteredKeys = [];
+      const values = [];
+
+      const helper = (obj, path) => {
```
andrewjschuang, line 46: This helper method can be outside `getAllKeysWithRecord()`.

components/data_stores/actions/get-record-keys/get-record-keys.mjs
```diff
@@ -0,0 +1,133 @@
...
+        // `contains` is true and `caseInsensitive` is false
+        if (contains && !caseInsensitive && String(obj).includes(query)) {
+          filteredKeys.push(path.slice(0, -1));
+          values.push(obj);
+          return;
+        }
+
+        // `contains` is false and `caseInsensitive` is true
+        if (!contains && caseInsensitive
+          && String(obj).toLocaleLowerCase() == query.toLowerCase()
```
andrewjschuang, line 88: Is there a reason for it to be `.toLocaleLowerCase()`?

vellames: No reason, didn't notice that

andrewjschuang (approved):
LGTM!

Issue #5290, opened 2023-01-23
[BUG] PostgreSQL "New Row" trigger fails to connect when SSL is required on database


george-bitdotio:
**Describe the bug**
When attempting to use the PostgreSQL integration, certain triggers, namely the "New Row" trigger, fail to connect to the database, while others are working ("New Table", for example). The observed error is:
```
Error:
Connection terminated unexpectedly
    at Connection.<anonymous> (/var/task/node_modules/pg/lib/client.js:132:73)
    at Object.onceWrapper (events.js:519:28)
    at Connection.emit (events.js:400:28)
    at Socket.<anonymous> (/var/task/node_modules/pg/lib/connection.js:107:12)
    at Socket.emit (events.js:412:35)
    at endReadableNT (internal/streams/readable.js:1333:12)
    at processTicksAndRejections (internal/process/task_queues.js:82:21)
```

I believe this is due to certain queries in the trigger not respecting the user's selection of the `rejectUnauthorized` flag and therefore not connecting using SSL.

**To Reproduce**
1. Create a database on [bit.io](https://bit.io)
2. Attempt to create a new source using the "New Row" trigger on one of the tables in your database
3. See error

Hassan-Abbas-SQA:
@george-bitdotio I am able to reproduce the bug
![image](https://user-images.githubusercontent.com/104946677/214375092-f35e8c16-c4e4-4959-a2a4-409ca7326b4f.png)

Hassan-Abbas-SQA:
@lcaresia iam still facing the issue
![image](https://user-images.githubusercontent.com/104946677/214906363-f40bb171-9a61-4d14-84ba-ac52c1584a6b.png)

lcaresia:
@Hassan-Abbas-SQA Hey! To work you need set `Reject Unauthorized` as `false` to work. This is some rule of bit.io

Hassan-Abbas-SQA:
> @Hassan-Abbas-SQA Hey! To work you need set `Reject Unauthorized` as `false` to work. This is some rule of bit.io

![image](https://user-images.githubusercontent.com/104946677/214911435-6fa00898-b59f-4976-a068-cc4266ea46ad.png)

Hassan-Abbas-SQA:
@lcaresia i was able to create source but getting this error 
![image](https://user-images.githubusercontent.com/104946677/214945472-16e227a8-372e-4f94-ba7d-92b188d928b8.png)

[...]

PR #5325, opened 2023-01-26
[BUG] PostgreSQL "New Row" trigger fails to connect when SSL is required on database #5290

lcaresia:


michelle0927 (approved):
Approved!

andrewjschuang (approved):

PR #5357, opened 2023-01-30
[BUG] Hubspot sources not setting "after" correctly

### LINKED ISSUE #5354: [BUG] Hubspot sources not setting "after" correctly
Hubspot sources are not always setting the `db` prop, `after` to the most recent timestamp.
See [Slack conversation](https://pipedream-users.slack.com/archives/CPTJYRY5A/p1674945743930749) for additional context.

michelle0927:
Resolves #5354

### DIFF
--- a/components/hubspot/package.json
+++ b/components/hubspot/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/hubspot",
-  "version": "0.5.8",
+  "version": "0.5.9",
   "description": "Pipedream Hubspot Components",
   "main": "hubspot.app.mjs",
   "keywords": [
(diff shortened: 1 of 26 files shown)


andrewjschuang (approved):
LGTM, but if `this._setAfter(ts)` is called after every relevant item, would it spam `db.set` in the source logs?

michelle0927:
> LGTM, but if `this._setAfter(ts)` is called after every relevant item, would it spam `db.set` in the source logs?

It's called if the timestamp found is greater than any previously found. Since the results are sorted from newest to oldest, it should usually only have to be called once.

PR #5358, opened 2023-01-30
5169 clickup new tasks filter events

alysonturing:
Expected behavior (as said by the customer):
It should only emit events within the Space and List selected but currently, it emit events for all tasks created in the entire Workspace.

Resolution: new trigger `New Task Advanced (Instant)` it will only emit events matching it's filter


andrewjschuang (approved):
LGTM!
Just one small improvement would be making `new-task-advanced` import `new-task`, so that `new-task.methods` aren't repeated

DilanAthukorala:

Issue #5331, opened 2023-01-27
[Components] statuspage


dannyroosevelt:
### App
statuspage
https://developer.statuspage.io/

**Triggers**
1. New Incident 
2. Updated Incident 

**Actions**
1. Create Incident
2. Update Incident
3. Create Maintenance Window

andrewjschuang:
@lcaresia can you link the PR here?

andrewjschuang:
Nvm, got it

Hassan-Abbas-SQA:
All Functional Tests Passed but here is some findings please fix them
in this trigger : at Component IDS prop description please add some more information regarding how customer should add custom list of ids provide with an example in description
![image](https://user-images.githubusercontent.com/104946677/216138438-e1f93fd1-40f4-4010-8952-c1ee02a46282.png)


Triggers
![image](https://user-images.githubusercontent.com/104946677/216136673-4863a9c7-51e8-4b31-9541-62b79f080a57.png)
![image](https://user-images.githubusercontent.com/104946677/216136723-2b24290d-e359-4c91-87f0-4a5eec1ffeeb.png)

Actions
![image](https://user-images.githubusercontent.com/104946677/216137584-0fc4f1a4-f492-4102-951b-af847821dc8e.png)

Hassan-Abbas-SQA:
@lcaresia thats nice thanks mate TestPassed
<img width="680" alt="image" src="https://user-images.githubusercontent.com/104946677/216378547-bcaac8e2-0a46-44ca-8ecb-534e6ca0eb47.png">

dannyroosevelt:
@Hassan-Abbas-SQA @lcaresia I don't have context on this, but do these IDs have to be passed manually? Do we not support async options for that prop? And are they passed as an array?

Hassan-Abbas-SQA:
> @Hassan-Abbas-SQA @lcaresia I don't have context on this, but do these IDs have to be passed manually? Do we not support async options for that prop? And are they passed as an array?

@dannyroosevelt we support async options for that prop too actually i did this manual test along with async because what if someone tries to add IDs manually then it should be easy for them
<img width="1044" alt="image" src="https://user-images.githubusercontent.com/104946677/216397436-28999459-ee2d-4726-bb5b-51911538807a.png">

PR #5380, opened 2023-02-01
[Components] statuspage #5331

lcaresia:


andrewjschuang:
Resolves #5380.

components/statuspage/actions/create-incident/create-incident.mjs
```diff
@@ -0,0 +1,68 @@
...
+      label: "Name",
+      description: "The name of the incident",
+      type: "string",
+    },
+    status: {
+      label: "Status",
+      description: "The status of the incident. (e.g. `resolved`, `investigating`)",
+      type: "string",
+      options: constants.INCIDENT_STATUSES,
+    },
```
andrewjschuang, line 28: Can this prop be in the app file?

components/statuspage/actions/update-incident/update-incident.mjs
```diff
@@ -0,0 +1,58 @@
...
+          pageId: c.pageId,
+        }),
+      ],
+    },
+    status: {
+      label: "Status",
+      description: "The new status of the incident (e.g. `resolved`, `investigating`)",
+      type: "string",
+      options: constants.INCIDENT_STATUSES,
+    },
```
andrewjschuang, line 32: Can this prop be in the app file?

components/statuspage/sources/new-incident-updated/new-incident-updated.mjs
```diff
@@ -0,0 +1,37 @@
...
+          limit: 20,
+        },
+      });
+
+      incidents.reverse().forEach(this.emitEvent);
+    },
+    emitEvent(event) {
+      const data = event?.incident ?? event;
+
+      if (data.incident_updates.length <= 1) {
```
andrewjschuang: Is there are reason we skip when there is one incident only?

components/statuspage/sources/incident-updated/incident-updated.mjs
```diff
@@ -0,0 +1,37 @@
...
+          limit: 20,
+        },
+      });
+
+      incidents.reverse().forEach(this.emitEvent);
+    },
+    emitEvent(event) {
+      const data = event?.incident ?? event;
+
+      if (data.incident_updates.length <= 1) {
```
andrewjschuang, line 26: Is there a reason we skip when there is only one?

andrewjschuang (approved):
LGTM!
I left just a few questions below

components/statuspage/sources/incident-updated/incident-updated.mjs
lcaresia: Yes, because if just have one it's a new incident, not an updated.

andrewjschuang (approved):

PR #5427, opened 2023-02-07
[ACTIONS/SOURCES]: Braze new sources actions

### LINKED ISSUE #5345: [Components] braze
### App
braze
https://www.braze.com/docs/developer_guide/home

**Triggers**
1. New User Event
2. User Attribute Change

**Actions**
1. Send a Message
2. Create/Update User
3. Track an Event

jcortes:
Hi @dannyroosevelt 
I couldn't test these components because In order to create an account in Braze's platform I needed to write an email to support just to let them know I was interested in implementing some components using Pipedream, but I didn't get any reply at all from Braze team.
On the other hand I didn't find any endpoint or documentation to support either polling or webhooks source components. So this PR is going to be blocked for these reasons.


GTFalcao:
Closing as outdated and blocked

PR #5446, opened 2023-02-09
Update README.md

dannyroosevelt:


dannyroosevelt:
@dylburger @ctrlaltdylan teeny tiny patch to address a mistake in the credits limits

dannyroosevelt:
Also, not sure why that first check is failing

ctrlaltdylan:
@dannyroosevelt nice catch. Yea that is odd. There's another exact link on the same page that passes the check. Will take a look in another PR if needed.

PR #5447, opened 2023-02-09
Hubspot - Bug fix to New Deal In Stage

michelle0927:
Resolves issue [discussed in Slack](https://pipedream-users.slack.com/archives/CPTJYRY5A/p1675934856084529) with the source New Deal In Stage emitting events that have been updated, but not recently moved to the stage.


vellames (approved):

DilanAthukorala:
I was not able to reproduce the issue with previous version of the source, since the problem is observed only for older accounts with existing deals. 

The new changes are verified and no issues observed. Ready for release!

Issue #5443, opened 2023-02-09
[BUG] Github action script checking version of app file


andrewjschuang:
**Describe the bug**
The [Github action script that checks for component versions](https://github.com/PipedreamHQ/pipedream/blob/master/.github/actions/git-diff-on-components/src/index.js#L95) had a [false positive for Google Drive app file](https://github.com/PipedreamHQ/pipedream/actions/runs/4128588165/jobs/7133214611#step:4:1092).
@GTFalcao figured out that it searches for a string `"version:"` in the file, independently if it's in the app file.

**To Reproduce**
1. Create a new branch from `master`.
2. Change [this line](https://github.com/PipedreamHQ/pipedream/blob/master/.github/workflows/pull-request-checks.yaml#L8) to `on: push` and push the commit to the repo. Now the action will run on every commit pushed.
3. Go to Google Drive app file and ensure [these lines](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_drive/google_drive.app.mjs#L270) has exactly `version: "v3"`: 
```js
return drive.drive({
  version: "v3",
  auth,
});
```
4. Change anything in the `common/utils.mjs` file.
5. Push all commits and see the error in the Github Action.

**Expected behavior**
No errors regarding version bump in the app file.

**Screenshots**
<img width="692" alt="image" src="https://user-images.githubusercontent.com/15385076/217806547-b9d6f9bc-1dc8-4e98-9190-4862a876f2dc.png">

**Some possible fixes**
- Only iterate through component files when checking version changes
- Ensure that the file is a component before appending the version bump error
- Avoid matching `version:` if it's within `methods` or `propDefinitions` blocks

Hussain0520:
Hey is this issue still open? I would like to work

andrewjschuang:
@Hussain0520 I really appreciate your effort to help.
May I ask if you have knowledge of the GitHub Action Checks that are executed on component PRs?

Hussain0520:
Actually I'm not aware of it but with your help I guess I can contribute :)

andrewjschuang:
I think in this case it would be better to leave this task for a Pipedream Component Developer, they will have more experience and insights on what is necessary to adjust.

You are very welcome though to tackle another issue in the **Prioritized** or **To Do** columns of the [Project Board](https://github.com/PipedreamHQ/pipedream/projects/1)!

vunguyenhung:
[...]

PR #5481, opened 2023-02-14
[BUG] Github action script checking version of app file #5443

lcaresia:
I tried a new way to check, instead of just check for an string `version:`, now is finding for a `version:` string with two spaces before to have sure this version will be in "first level" of the exported module, not inside `propDefinitions` or other key


andrewjschuang (changes requested):
The error still persists: https://github.com/PipedreamHQ/pipedream/issues/5443#issuecomment-1431083004

andrewjschuang:
Can confirm `google_drive.app.mjs` is not in the list of components that need version bump. Approved!

https://github.com/PipedreamHQ/pipedream/actions/runs/4204398519/jobs/7295015332

andrewjschuang (approved):

PR #5519, opened 2023-02-18
Dropbox - Update Path prop

### LINKED ISSUE #5516: [BUG] Dropbox Path prop requires absolute path, else async options don't return matching folders
I have to search for the absolute path of my folder for it to appear in the options presented in the **Path** prop for Dropbox components.

A user reported this example: If I want the "YT Audio" folder, I have to type /YT Audio in full:

CleanShot 2023-02-17 at 11 47 42@2x

If the Dropbox search API supports this, I ideally want full substring path matches. e.g. if I had a folder `/hello/world/my_folder_here`, searches for “my_folder_here” should return that path and any other paths where any part of the path matches "my_folder_here".

michelle0927:
Resolves #5516


feyzullah (dismissed):
LGTM!

lcaresia (approved):

PR #5534, opened 2023-02-21
Adding tip about workspace CLI usage

ctrlaltdylan:
Adds a small warning about PD accounts with workspaces when logging into the PD CLI.

We've already updated the CLI profile reference documentation to replace organization IDs with workspace IDs.

However the default `pd login` behavior is currently not compatible with workspaces by default, so adding a tip about this:

https://linear.app/pipedream/issue/DRE-212/[docs]-update-cli-docs-to-include-organization-id-argument

### DIFF
--- a/docs/docs/cli/login/README.md
+++ b/docs/docs/cli/login/README.md
@@ -23,6 +23,14 @@ Once you're done, go back to your shell and you should see confirmation that you
 Logged in as dylburger (dylan@pipedream.com)
 ```
 
+::: warning
+
+If you are a member of a [workspace](/docs/workspaces/) then you'll also need to configure the `pd` CLI with your workspace ID.
+
+[Follow this guide](/cli/reference/#creating-a-profile-for-a-workspace) to learn how to find your workspace ID and associate it with a `pd` profile.
+
+:::
+
 ## Signing up for Pipedream via the CLI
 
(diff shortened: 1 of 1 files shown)


docs/docs/cli/login/README.md
```diff
@@ -23,6 +23,14 @@ Once you're done, go back to your shell and you should see confirmation that you
 Logged in as dylburger (dylan@pipedream.com)
 ```
 
+::: warning
+
+If you are a member of a [workspace](/docs/workspaces/) then you'll also need to configure the `pd` CLI with your workspace ID.
```
dylburger, line 28: I don't know if all users will know what this means, but since it's now the default, I would remove the warning section and just include it as a part of the standard instructions for logging into the CLI. Since we'll be migrating existing users to workspaces soon, this will also apply to them.

dylburger (changes requested):
See my one comment re: removing the warning and just including this in the general instructions

docs/docs/cli/login/README.md
ctrlaltdylan: Good point, moved it as a general instruction as part of logging in with the CLI.

Best case scenario, we add workspace configuration generation as part of `pd login` and `pd signup`.

Issue #5497, opened 2023-02-15
[FEATURE] `@pipedream/platform` axios should support the ability to create custom `axios` constructor, or pass all standard `axios` options


dylburger:
I need to set custom options on `axios`, which I'd normally do via `axios.create`. `@pipedream/platform` doesn't allow the creation of a custom constructor.

Whoever takes this should review the entire `axios` implementation in the `@pipedream/platform` package, and ensure users are able to use as much of the existing axios API as they can. We should maintain the `@pipedream/platform` capabilities for error handling / how it exports data in Pipedream steps (like it does today), but expose as much of the standard `axios` API to users as possible.

PR #5544, opened 2023-02-22
[FEATURE] @pipedream/platform axios should support the ability to create custom axios constructor, or pass all standard axios options #5497

lcaresia:


vunguyenhung:
Hello @lcaresia @andrewjschuang, would you mind advice how can I test this? Or is this needed testing?

andrewjschuang:
@vunguyenhung I think you can create two instances of each `axios`, e.g.
```js
import axios from "axios";
import { axios as pipedreamAxios } from "./platform/dist/index.js";
```
or something like that.

And then create a config to pass in to the constructor with most or all `axios` parameters, make the request and see if they were really used by `pipedreamAxios`.

vunguyenhung:
Hello @andrewjschuang, I couldn't test the action due to an error on the publish action step. 

Could you check the test report below and advice how can I test it?
[Axios_5544_1037.pdf](https://github.com/PipedreamHQ/pipedream/files/10835843/Axios_5544_1037.pdf)

andrewjschuang:
Hi @lcaresia, looking at the code, seems like the change doesn't reflect in anything in practice, since we don't remove any key from the `config` object.

Also, looking at [the thread](https://pipedream.com/community/t/can-i-use-axios-directly-with-pipedream/4774/2) that originated this issue, I think it's if best we implemente the [axios.create()](https://axios-http.com/docs/instance) method.

I also thought of adding a flag `fullResponse` with `default: false` in `AxiosRequestConfig` that enables the user to receive the whole [response schema](https://axios-http.com/docs/res_schema) from the request instead of just `data`, what do you think?

lcaresia:
@vunguyenhung @andrewjschuang 

You can test creating an `test.js` file on project root folder with this code:

```
const {
  axios, createAxiosInstance,
} = require("./platform/dist");

async function test1() {
  const createdAxios = createAxiosInstance({
    baseURL: "https://random-data-api.com/api/v2",
  });

  const response = await createdAxios.get("users");

  console.log("createdAxios:", response);
}

async function test2() {
  const data = await axios(this, {
    url: "https://random-data-api.com/api/v2/users",
    returnRawResponse: false,
  });

  console.log("onlyData:", data);
}

async function test3() {
  // const axios = createAxiosInstance({});

  const response = await axios(this, {
    url: "https://random-data-api.com/api/v2/users",
    returnRawResponse: true,
  });

  console.log("returnRawResponse:", response);
}

test1();
test2();
test3();
```

platform/lib/index.ts
```diff
@@ -1,10 +1,11 @@
 import * as t from "io-ts";
 
-import axios from "./axios";
+import axios, { createAxiosInstance } from "./axios";
 import { AxiosRequestConfig as AxiosConfig } from "axios";
 
 export {
   axios,
+  createAxiosInstance,
```
andrewjschuang, line 8:
```suggestion
```
See other comment about `axios.create`

platform/lib/index.ts
```diff
@@ -1,10 +1,11 @@
 import * as t from "io-ts";
 
-import axios from "./axios";
+import axios, { createAxiosInstance } from "./axios";
```
andrewjschuang, line 3:
```suggestion
```
See other comment about `axios.create`

andrewjschuang (changes requested):
Thanks @lcaresia!
I've left some comments about some improvements, but overall this is in the right track!

platform/lib/axios.ts
```diff
@@ -5,6 +5,10 @@ import * as querystring from "querystring";
 import { cloneSafe } from "./utils";
 import { ConfigurationError } from "./errors";
 
+interface PipedreamAxiosRequestConfig extends AxiosRequestConfig {
```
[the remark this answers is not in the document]
lcaresia: Oh, I didn't see that

andrewjschuang (approved):

PR #5550, opened 2023-02-22
[Typeform] Respond http only on end of execution

### LINKED ISSUE #5530: [BUG] Typeform source should issue custom HTTP response to retry failed executions
Currently, Typeform sources issue HTTP 200 OK responses as soon as Pipedream accepts the HTTP request. If the source `throw`s an error within the `run` method, the execution will fail, and Typeform won't retry the event.

We should set `customResponse: true` on the HTTP interface for Typeform sources, and issue a 200 OK response at the end of the `run` method when the run succeeds. With this setting, Pipedream will automatically issue a 400 OK response on Errors, so you don't need to `catch` every individual error and return a 4XX status code. Just returning a successful response after the execution should be enough.

Other sources are subject to this behavior, as well, so we should revisit the general question on the platform-level.

andrewjschuang:
Resolves #5530.

### DIFF
--- a/components/typeform/package.json
+++ b/components/typeform/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/typeform",
-  "version": "0.3.7",
+  "version": "0.3.8",
   "description": "Pipedream Typeform Components",
   "main": "typeform.app.mjs",
   "keywords": [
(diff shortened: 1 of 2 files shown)


lcaresia (approved):
LGTM!

vunguyenhung:
I tried to submit until the form closed but it seems like it needs special condition to produce the error. 
image

Since this is a minimal changes, and I have confirmed it works. So I will move this to ready for release

Issue #5560, opened 2023-02-23
[BUG] DM Received Slack Trigger emits on every DM *sent* as well as received


dannyroosevelt:
The DM Received Slack trigger appears to emit for every DM I receive **or send**.

This should only emit for DMs I receive.

PR #5563, opened 2023-02-24
[BUG] DM Received Slack Trigger emits on every DM *sent* as well as received #5560

lcaresia:
Is working as expect on tests, so I think the problem is with the version in production.
I just will increase the version to publish again to production.

Cc: @dannyroosevelt

### DIFF
--- a/components/slack/package.json
+++ b/components/slack/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/slack",
-  "version": "0.4.18",
+  "version": "0.4.19",
   "description": "Pipedream Slack Components",
   "main": "slack.app.mjs",
   "keywords": [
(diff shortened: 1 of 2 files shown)


vunguyenhung:
Hello @lcaresia, I tested with my test account but and still see there're still events when I send a message to a user
image
image
image

Could you check again?

lcaresia:
You set to not received `myselfs` messages?

dannyroosevelt:
> @vunguyenhung Do you set to not received `myselfs` messages?

Oh, I see that is indeed the issue... but that's strange and I still think is not intentional for this trigger. We should remove that prop and not emit on DMs I **send**, only receive.

vunguyenhung:
Hi @lcaresia @dannyroosevelt

> Do you set to not received myselfs messages?

Oh I actually set the `Ignore Myself: FALSE`. I think the prop `Ignore Myself` is indeed quite confusing, @lcaresia I think we could have the source only emit the events for receiving message and remove that prop?

lcaresia:
@vunguyenhung Can you test again?

vunguyenhung:

Test report
[Slack_5560_1012.pdf](https://github.com/PipedreamHQ/pipedream/files/10835745/Slack_5560_1012.pdf)

alysonturing (approved):

PR #5628, opened 2023-03-06
Shopify - new and updated actions

### LINKED ISSUE #5596: [IMPROVEMENT] Shopify - Add Products to Custom Collections and Search Custom Collection
Shopify
----

There're 2 suggested improvements for this ticket:

1. On the Add Product to Custom Collection action, change it to add Products to Custom Collections, and allow user to select multiple products and multiple collections. The end result is multiple products are added to multiple custom collections
2. Search Custom Collection (by name)
----
**Original message**
I would like to be be able to add a single product ID to multiple custom collections at once. for example I would like to use the "style" metafield on a product to either add to or create multiple collections for the same product.

image

image
image

On the same topic as can be seen above if possible I would like to add a product ID to a collection using the "Collection Title" rather than collection ID. It would make things much easier than having to search for a collection or only adding products to collections that are created in the same workflow. Alternatively some kind of "GET collection" method that would allow us to get the collection ID via a title search.

image

### LINKED ISSUE #5598: [ACTION] Update Shopify Product/Product Variant HS Code & SKU
Shopify

https://shopify.dev/docs/api/admin-rest/2023-01/resources/inventoryitem#resource-object
country_code_of_origin
harmonized_system_code
seem to be the 2 most important ones as a consequence of the new shopify rule on international shipping:
image

https://shopify.dev/docs/api/admin-graphql/2023-01/input-objects/ProductVariantInput#field-productvariantinput-harmonizedSystemCode

https://shopify.dev/docs/api/admin-graphql/2023-01/objects/Duty#field-duty-harmonizedsystemcode
https://shopify.dev/docs/api/admin-graphql/2023-01/objects/Duty#field-duty-countryCodeOfOrigin
(this seems to be for editing an item in an order rather than inventory items which is less useful)

These additional links I found but probably aren't super relevent.

While you're in there you may want to consider adding the ability to edit SKU numbers:
https://shopify.dev/docs/api/admin-graphql/2023-01/input-objects/ProductVariantInput#field-productvariantinput-sku

michelle0927:
Resolves #5596 
Resolves #5598


components/shopify/actions/add-products-to-custom-collections/add-products-to-custom-collections.mjs
```diff
@@ -0,0 +1,47 @@
+import shopify from "../../shopify.app.mjs";
+
+export default {
+  key: "shopify-add-products-to-custom-collections",
```
GTFalcao, line 4: Is it really needed to change the component key here, rather than just the name and description? Seeing as it's just an added `s` and, correct me if I'm wrong, I believe that would require any users using it to remove the step from their existing workflows and re-configure it with the new action.

I'm not sure where we draw the line between keeping the component key 100% faithful to its name, and avoiding this potential hassle for users that already have the component configured in their workflows.

BTW, if the key is indeed updated, I think the version should be `0.0.1`

components/shopify/actions/search-custom-collection-by-name/search-custom-collection-by-name.mjs
```diff
@@ -0,0 +1,47 @@
...
+    if (exactMatch) {
+      params.title = title;
+    }
+
+    let collections = await this.shopify.getObjects("customCollection", params);
+
+    if (!exactMatch) {
+      collections = collections.filter((collection) =>
+        collection.title.toLowerCase().includes(title.toLowerCase()));
+    }
```
GTFalcao, line 42: Minor suggestion: the title being compared only needs to be converted to lowercase once, instead of it being done for every item in the array
```suggestion
    if (!exactMatch) {
      const lowerCaseTitle = title.toLowerCase();
      collections = collections.filter(({ title }) =>
        title.toLowerCase().includes(lowerCaseTitle));
    }
```

GTFalcao (commented):
Hi @michelle0927 , thanks for your contribution! I left a couple of minor comments regarding the component key change and a small code suggestion.

Since these are minor I'm moving the issue forward to QA

vunguyenhung:

Test report
[Shopify_5628_1032.pdf](https://github.com/PipedreamHQ/pipedream/files/10905149/Shopify_5628_1032.pdf)

GTFalcao (approved):

zecul:
I'm having a little trouble adding products IDs when not selecting from the drop down menu. Is this my error or something else?

image
image

If I do select a product ID from the drop down I'm then having touble adding it to a collection that isn't from the dropdown:

image
image

Is there a format for entering these values that I'm not adhering to? I've tried a few basic ones, but last time it required stringifying etc and I'm not super confident with that without specific instruction.

michelle0927:
@zecul Product IDs and CollectionIds need to be entered as arrays. ;)
So if it’s a custom expression, it should be like this
`{{[steps.search_products.$return_value[0].id]}`
or
`{{[12345,67890]}}`

zecul:
@michelle0927 Thanks!

I tried double curly brackets and single square brackets but not both together!

PR #5633, opened 2023-03-07
New OpenAI actions (Chat, Whisper, etc.)

dylburger:


dannyroosevelt:
I think you need to update the reference to the app file in the create-image action
```
➜  actions git:(openai-new-actions) ✗ pd publish create-image/create-image.mjs
parse failure (Could not resolve "../../openai.app.mjs" at create-image/create-image.mjs:19)
```

components/openai/actions/common/common.mjs
```diff
@@ -0,0 +1,128 @@
...
+    topP: {
+      label: "Top P",
+      description: "An alternative to sampling with temperature, called nucleus sampling, where the model considers the results of the tokens with top_p probability mass. So 0.1 means only the tokens comprising the top 10% probability mass are considered. We generally recommend altering this or `temperature` but not both.",
+      type: "string",
+      optional: true,
+    },
+    n: {
+      label: "N",
+      description: "How many completions to generate for each prompt.",
+      type: "string",
```
dannyroosevelt, line 28: should be integer, right?

components/openai/app/openai.app.mjs
```diff
@@ -0,0 +1,160 @@
...
+      description: "The ID of the model to use for chat completions",
+      type: "string",
+      async options() {
+        return (await this.getChatCompletionModels({})).map((model) => model.id);
+      },
+      default: "gpt-3.5-turbo",
+    },
+    embeddingsModelId: {
+      label: "Model",
+      description: "The ID of the embeddings model to use. OpenAI recommends using `text-embedding-ada-002` for nearly all use cases: \"It's better, cheaper, and simpler to use. [Read the blog post announcement](https://openai.com/blog/new-and-improved-embedding-model)\".",
```
dannyroosevelt, line 27: Nit, but maybe the trailing " is in the wrong place? Should it read as below? (before linking to the blog post) Not positive though.
```
description: "The ID of the embeddings model to use. OpenAI recommends using `text-embedding-ada-002` for nearly all use cases: \"It's better, cheaper, and simpler to use." [Read the blog post announcement](https://openai.com/blog/new-and-improved-embedding-model)\.",
```

dylburger: that's actually part of the quote from their docs

ctrlaltdylan:
The only comment I have on the **Chat** action is that the **Assistant Response** prop is a bit confusing.

Is the intent that the user can provide a history of assistant messages specifically to this prop?

Like a subset of the **Prior Message History** prop?

If so, should the description say that? Currently the description leads the user to believe it's used for storing messages, not providing context to the chat prompt.

ctrlaltdylan:
The **Classify into categories** props might be a bit broken:


CleanShot 2023-03-08 at 14 37 51@2x

PR #5725, opened 2023-03-14
Postgresql - new row trigger bug fix

### LINKED ISSUE #5706: [BUG] Postgres New Row Trigger not emitting all events for tables with over 1000 records
The Postgres New Row Trigger retrieves all rows and relies on "dedupe" to check for new rows. It should only retrieve records where the specified column is greater than it was in the last row retrieved.

michelle0927:
Resolves #5706


luancazarine (approved):

PR #5727, opened 2023-03-14
[GitHub Actions] Fail step if any error in publishing

andrewjschuang:
Currently, the action step doesn't fail when there is an error when attempting to publish a component to the Pipedream registry. This PR enables the developer to be better notified of possible errors so they can fix them ASAP.


luancazarine (approved):
Hi @andrewjschuang, thanks for your contribution! LGTM!

PR #5734, opened 2023-03-15
Moved quote creation to one new action  - Wise

lcaresia:
Is need to test

**Action**

- Create Quote
- Create Transfer

Cc: @zalmanlew @vunguyenhung


components/wise/actions/create-quote/create-quote.mjs
```diff
@@ -0,0 +1,72 @@
...
+        wise,
+        "currency",
+        (c) => ({
+          sourceCurrency: c.sourceCurrency,
+        }),
+      ],
+    },
+    targetAmount: {
+      label: "Target Amount",
+      description: "Amount in target currency to be received by the recipient. E.g. `100`",
```
zalmanlew, line 39: Include float so the user knows they can send floating point numbers
```suggestion
      description: "Amount in source currency to be received by the recipient. E.g. `100.00`. Must specify either **Target Amount** or **Source Amount**",
```

components/wise/actions/create-quote/create-quote.mjs
```diff
@@ -0,0 +1,72 @@
...
+      type: "string",
```
zalmanlew, line 40: need to set it to optional since only one will be required
```suggestion
      type: "string",
      optional: true,
```

components/wise/actions/create-quote/create-quote.mjs
```diff
@@ -0,0 +1,72 @@
...
+    },
+    sourceAmount: {
+      label: "Target Amount",
+      description: "Amount in source currency to be received by the recipient. E.g. `100`",
+      type: "string",
```
zalmanlew, line 45: need to set it to optional since only one will be required
```suggestion
      type: "string",
      optional: true,
```

components/wise/actions/create-quote/create-quote.mjs
```diff
@@ -0,0 +1,72 @@
...
+      ],
+    },
+    targetAmount: {
+      label: "Target Amount",
+      description: "Amount in target currency to be received by the recipient. E.g. `100`",
+      type: "string",
+    },
+    sourceAmount: {
+      label: "Target Amount",
+      description: "Amount in source currency to be received by the recipient. E.g. `100`",
```
zalmanlew, line 44: Include float so the user knows they can send floating point numbers
```suggestion
      description: "Amount in source currency to be received by the recipient. E.g. `100.00`. Must specify either **Target Amount** or **Source Amount**",
```

components/wise/actions/create-quote/create-quote.mjs
```diff
@@ -0,0 +1,72 @@
...
+      type: "string",
+    },
+  },
+  async run({ $ }) {
+    if (this.targetAmount && this.sourceAmount) {
```
zalmanlew, line 49: add safety check if the user doesn't set either a source or target amount
```suggestion
    if ((this.targetAmount && this.sourceAmount) || (!this.targetAmount && !this.sourceAmount)) {
```

zalmanlew (commented):
@lcaresia have a few fixes and suggestions

components/wise/actions/create-quote/create-quote.mjs
```diff
@@ -46,6 +47,14 @@ export default {
       type: "string",
       optional: true,
     },
+    payoutType: {
+      label: "Payout Type",
+      description: "Amount in source currency to be received by the recipient. E.g. `100.00`. Must specify either **Target Amount** or **Source Amount**",
```
zalmanlew, line 52: correct prop description - added info from docs
```suggestion
     description: "Preferred payout method. Default value is `BANK_TRANSFER`. If you are funding the transfer from a Multi Currency Balance, you must set the `payOut` as `BALANCE` in order to get the correct pricing in the quote . By not doing so, it will default to `BANK_TRANSFER` and the fees will be inconsistent between quote and transfer.",
```

luancazarine (changes requested):
Hi @lcaresia, Thanks for your contribution. I just left a version fix suggestion.

luancazarine (approved):
LGTM!

vunguyenhung:

Test report
[wise_5734.pdf](https://github.com/PipedreamHQ/pipedream/files/11007894/wise_5734.pdf)

Issue #5755, opened 2023-03-16
[ACTION] Search Categories


SGZabrav:
**Steps to reproduce**
Connect to a Twitch account, then use the following configuration
![image](https://user-images.githubusercontent.com/18277920/225684565-c24e801b-4248-4bed-b859-3993f9e40250.png)

**Expected Behavior**
- The action should be executed successfully

vunguyenhung:
Hello @SGZabrav, thanks for raising this. I've found another bug for this. And I will add this to Pipedream prioritized backlog
![image](https://user-images.githubusercontent.com/18277920/225684044-f8bf1fb8-20a8-4711-802a-5cce61a606c6.png)

PR #5799, opened 2023-03-20
[ACTION] Search Categories #5755

lcaresia:


vunguyenhung:
I confirmed that the issue is fixed now
image
image

PR #5845, opened 2023-03-24
Pipedream Data Stores / Helper Action

### LINKED ISSUE #5640: [ACTION] Pipedream Data Stores / Helper Action
Pipedream Data Stores

I would like an ability to exclude keys from a Data Store return.  Basically Data Store A minus Data Store B. 

In my case Data Store A is my total list of products I need to process and Data Store B is the list of products already processed.

Right now the key/value pairs are indentical but I can think of a use where only the keys would match so Data Store A's Key(x1)/Value(y1), Key(x2)/Value(y2)  minus Data Store B's Key(x1)/Value(z1) should return Key(x2)/Value(y2).

michelle0927:
Resolves #5640

### DIFF
--- a/components/data_stores/actions/get-difference/get-difference.mjs
+++ b/components/data_stores/actions/get-difference/get-difference.mjs
@@ -1,0 +1,74 @@
+import app from "../../data_stores.app.mjs";
+import options from "../../common/options.mjs";
+
+export default {
+  key: "data_stores-get-difference",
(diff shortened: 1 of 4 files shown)


vunguyenhung:

Test report
[Datastore_5640_1756.pdf](https://github.com/PipedreamHQ/pipedream/files/11077818/Datastore_5640_1756.pdf)

PR #5862, opened 2023-03-27
#5489 - Twitter V2 API (refactory)

### LINKED ISSUE #5489: [Refactor] Support Twitter v2 API and OAuth 2.0
At some point in all of the recent Twitter API changes, the auth flow for connecting a new account stopped working. This appears to be constrained only to Twitter's OAuth 1.0 integration, but not OAuth 2.0.
https://developer.twitter.com/en/docs/twitter-api/getting-started/about-twitter-api

In order to move the auth implementation to use OAuth 2.0, we should also update all of the components to use their v2 API, which may require a significant refactor.

For now, let's use the new app we've published, `twitter_v2` for these components, but we may end up using the existing `twitter` app, which would require a `major` version update, since that'd be a breaking change.

GTFalcao:
### Actions
- Add User To List
- Create Tweet
- Delete Tweet
- Follow User
- Get Tweet
- Get User
- Like Tweet
- List Followers
- List Liked Tweets
- List Lists
- List Mentions
- List User Tweets
- Retweet
- Search Tweets
- Search Tweets in List
- Unfollow User
- Unlike Tweet

### Sources

- New Follower Received by User
- New List Followed by User
- ~~New Retweet Posted~~
- New Tweet Liked by User
- New Tweet Posted by User
- New Tweet Posted in List
- New Tweet Posted Matching Query
- New Unfollower of User
- New User Followed by User

Closes #5489


michelle0927 (dismissed):
Looks good, ready for QA!

vunguyenhung:

[Twitter_5862_1118.pdf](https://github.com/PipedreamHQ/pipedream/files/11118115/Twitter_5862_1118.pdf)

GTFalcao:
Hey @vunguyenhung , thanks for the feedback, I had some 401 responses before but I assumed they were due to bad input (i.e. trying to obtain private information from a private user) since they did not happen again when I changed the input afterwards.

With your results and further testing, it seems these 401 responses are due to something in the app auth. This could be something like not waiting for the new token before allowing the request to continue (with the old token). cc @dannyroosevelt 

So I've implemented retrying at the basic request level to avoid this (every request will be attempted up to three times before it can raise an exception). I've deployed several sources and you can see a log when a request fails, but assuming it succeeds when retried, no exception will be raised and the data will be obtained normally.

I fixed other issues as well such as the list async options mistake (forgot to change the method name only for lists). Let me know if you need anything!

vunguyenhung:
Hello @GTFalcao, I see that you committed [some API Keys here](https://github.com/PipedreamHQ/pipedream/pull/5862/files#diff-6a92e9ea13db69404665654214951ed63b14d135663639226a1539139e0744b6). Is is necessary?

vunguyenhung:

[Twitter_5862_1319.pdf](https://github.com/PipedreamHQ/pipedream/files/11135308/Twitter_5862_1319.pdf)

GTFalcao:
Hi @vunguyenhung , thanks for the note, that file was not meant to be committed indeed so I replaced that api key already.

Regarding the errors that persisted, it seems for these cases we were getting `undefined` instead of the expected empty array. I've added checks to prevent that, and validated the async options both with and without lists on my account.

Also, having fixed that issue which was also affecting "New Retweet Posted", it seems this functionality is not actually provided by the V2 API - the "referenced tweets" which includes "retweeted" are actually the other way around, i.e. when you specify a Tweet, you can see if it was a retweet of another tweet, instead of seeing retweets of it.

The [V2 migration endpoint map](https://developer.twitter.com/en/docs/twitter-api/migrate/twitter-api-endpoint-map) specifies [Retweets lookup](https://developer.twitter.com/en/docs/twitter-api/tweets/retweets/api-reference) as the replacement for the V1 endpoint of listing retweets, but that unfortunately only allows listing the users who retweeted the specified tweet.

So I'm removing the "New Retweet Posted" source for now, as I have not found another way to obtain this info on the V2 API at all. Interestingly however, this functionality is present for ["quote tweets"](https://developer.twitter.com/en/docs/twitter-api/tweets/quote-tweets/introduction) (which are not the same as retweets).

GTFalcao:
@dannyroosevelt I think we still have some pending auth stuff, so let me know if/when any further action is needed here - should we mark this as blocked?

dannyroosevelt:
Yes let's hold off on merging this for now, I'm working on some auth implementation changes to support Twitter's new API pricing tiers.

alysonturing (approved):
LGTM (I checked only what was changed from the previous review)

PR #5892, opened 2023-03-30
Shopify - Get Metafield action: Add Namespace and key prop for filtering

### LINKED ISSUE #5730: [IMPROVEMENT] Shopify - Get Metafield action: Add Namespace and key prop for filtering
Add Namespace & Key as prop fields on Shopify Get Metafields.

I would like to be abe to "get_metafields" by resource>namespace>key
Ideally it would allow multiple namespace/keys to be entered as an array to remove the need for multiple get_metafield steps to access multiple metafields.

Currently the get_metafields action returns all the metafields that a product has in the order that they appear (and skips them if empty) so its not possible to process multiple products in the same workflow (because they each have a different index value)

Product 1:
image
Product 2:
image

For the first product the labels.name metafield is in position 14 and in the 2nd product it's in position 15. I want to create a workflow that processes all my products in order but I need to know where the metafields are going to be for each product in order to update them correctly:

image

michelle0927:
Resolves #5730


andrewjschuang (approved):
Thanks, LGTM!

vunguyenhung:
I've verified this, as only metafields with specified namespace is added
image

PR #5917, opened 2023-04-03
Smartsheet - New Components

### LINKED ISSUE #3785: Smartsheet
### App
Smartsheet

### Triggers
* Updated Row
* New Comment Added
* New Row

### Actions
* Add Row to Sheet
* New Sheet From Template
* Update Row

michelle0927:
Resolves #3785


components/smartsheet/actions/update-row/update-row.mjs
```diff
@@ -0,0 +1,93 @@
...
+        smartsheet,
+        "rowId",
+        (c) => ({
+          sheetId: c.sheetId,
+        }),
+      ],
+      reloadProps: true,
+    },
+  },
+  async additionalProps() {
```
andrewjschuang, line 28: See comment above for `additionalProps`

components/smartsheet/actions/update-row/update-row.mjs
```diff
@@ -0,0 +1,93 @@
...
+        };
+      }
+      else if (column?.options) {
+        props[column.id].options = column.options;
+      }
+    }
+    return props;
+  },
+  async run({ $ }) {
+    /* eslint-disable no-unused-vars */
```
andrewjschuang, line 65: Is there a reason for `/* eslint-disable no-unused-vars */` to be here?

andrewjschuang (commented):
Thanks, looks good overall but we need to figure out a way to work around `additionalProps` for custom expressions. Can you think of a way?

components/smartsheet/actions/add-row-to-sheet/add-row-to-sheet.mjs
```diff
@@ -0,0 +1,88 @@
...
+    },
+    toTop: {
+      type: "boolean",
+      label: "To Top",
+      description: "Set to `true` to add new row to the top of the sheet",
+      optional: true,
+      default: false,
+    },
+  },
+  async additionalProps() {
```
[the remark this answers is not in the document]
michelle0927: I added a `cells` prop that allows the user to manually enter the column values in case `sheetId` is entered as a custom expression.

components/smartsheet/smartsheet.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+        return data?.map(({
+          id, name,
+        }) => ({
+          label: name,
+          value: id,
+        })) || [];
+      },
+    },
+    rowId: {
+      type: "string",
```
andrewjschuang, line 26:
```suggestion
      type: "integer",
```

Should this be of type integer?

components/smartsheet/smartsheet.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+      async options({ sheetId }) {
+        if (isNaN(sheetId)) {
+          return [];
+        }
+        const { rows } = await this.getSheet(sheetId);
+        return rows?.map(({ id }) => id.toString()) || [];
+      },
+    },
+    templateId: {
+      type: "string",
```
andrewjschuang, line 38:
```suggestion
      type: "integer",
```

Should this be of type integer?

components/smartsheet/smartsheet.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+        return templates?.map(({
+          id, name,
+        }) => ({
+          label: name,
+          value: id,
+        })) || [];
+      },
+    },
+    workspaceId: {
+      type: "string",
```
andrewjschuang, line 64:
```suggestion
      type: "integer",
```

Should this be of type integer?

components/smartsheet/smartsheet.app.mjs
```diff
@@ -1,11 +1,233 @@
...
+        return data?.map(({
+          id, name,
+        }) => ({
+          label: name,
+          value: id,
+        })) || [];
+      },
+    },
+    folderId: {
+      type: "string",
```
andrewjschuang, line 83:
```suggestion
      type: "integer",
```

Should this be of type integer?

andrewjschuang (changes requested):
Thanks for adding the cells prop!

Just a few questions regarding the use of `isNan()`:
- What does the API return if `sheetId` or `rowId` is not a number?
- Would it be better to `throw ConfigurationError` instead of returning a empty array?
- I think we could create a function `validateId` that receives an `id` and checks if it's a number or not. And this function is called in `getRow`, `getSheet`, `listColumns`, `createSheetInFolder`, etc. So the logic is contained inside the method that makes the request and not outside

michelle0927:
> Thanks for adding the cells prop!
> 
> Just a few questions regarding the use of `isNan()`:
> 
> * What does the API return if `sheetId` or `rowId` is not a number?
> * Would it be better to `throw ConfigurationError` instead of returning a empty array?
> * I think we could create a function `validateId` that receives an `id` and checks if it's a number or not. And this function is called in `getRow`, `getSheet`, `listColumns`, `createSheetInFolder`, etc. So the logic is contained inside the method that makes the request and not outside

I wanted to allow the user to enter custom expressions instead of selecting async options, as you suggested. If a custom expression was entered for `sheetId`, it would try to put the expression in as a path parameter and result in an error. Similarly, the `additionalProps` sections would get errors if `sheetId` or `rowId` were entered as custom expressions. 

My intent with `isNan()` was to use it as a way to detect if a custom expression has been entered, and if so, allow the user to use the components without getting errors. There's probably a better way to do that though, so if you have an idea, let me know and I'll update it.

andrewjschuang (dismissed):
Thanks, LGTM!

For some reason I can't create an account to see how the props work - @vunguyenhung can you please verify if it is suitable when QAing?

vunguyenhung:

Test report
[Smartsheet_5917_2206.pdf](https://github.com/PipedreamHQ/pipedream/files/11171941/Smartsheet_5917_2206.pdf)

andrewjschuang (approved):

PR #6109, opened 2023-04-20
RSS entries not in order given by the RSS feed

### LINKED ISSUE #6068: [BUG] RSS entries are not ordered like the order given by the RSS feed
**Describe the bug**
RSS feed entries are not ordered like the order given by the RSS feed URL but they are sent in a disorderly manner by the trigger.
For example, if my trigger fires every half hour and it finds 5 new events, these are not sorted like in the feed from which it takes them. So, by what value are they sorted by the trigger?

**Screenshots**
This is the original feed order (1-2-3-4):
FireShot Capture 010 -  - script googleusercontent com

This is the order given by the trigger (2-3-4-1):
Screenshot 2023-04-18 at 11 50 46 AM

michelle0927:
Resolves #6068

### DIFF
--- a/components/rss/package.json
+++ b/components/rss/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/rss",
-  "version": "0.5.1",
+  "version": "0.5.2",
   "description": "Pipedream RSS Components",
   "main": "dist/app/rss.app.mjs",
   "types": "dist/app/rss.app.d.ts",


feyzullah (approved):
LGTM!

vunguyenhung:

Test report
[RSS_6109.pdf](https://github.com/PipedreamHQ/pipedream/files/11297565/RSS_6109.pdf)

PR #6246, opened 2023-05-03
[Zoho CRM] Dynamic Modules + Refactor

### LINKED ISSUE #6228: Update Zoho CRM
Currently we are listing Modules options in the actions via a static list: https://github.com/PipedreamHQ/pipedream/blob/master/components/zoho_crm/zoho_crm.app.mjs#L55

That means that custom modules are not supported (but [they are fetched via async options in the sources](https://github.com/PipedreamHQ/pipedream/blob/master/components/zoho_crm/sources/new-event/new-event.mjs#L23)).

Let's use async options to fetch all modules for all Zoho CRM components, and let's make sure the [additionalProps](https://github.com/PipedreamHQ/pipedream/blob/master/components/zoho_crm/actions/create-object/create-object.mjs#L18) still work.

andrewjschuang:
- Makes all props list modules dynamically.
- Refactors legacy actions

Resolves #6228.

## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough


components/zoho_crm/actions/list-modules/list-modules.mjs
```diff
@@ -0,0 +1,17 @@
+import zohoCrm from "../../zoho_crm.app.mjs";
+
+export default {
+  key: "zoho_crm-list-modules",
+  name: "List Modules",
+  description: "Retrieves a list of all the modules available in your CRM account.",
+  version: "0.2.0",
```
GTFalcao, line 7: Since the key is changing, shouldn't this be version `0.0.1` of the "new" component?

Also remember to get the old key unpublished when this is merged

components/zoho_crm/zoho_crm.app.mjs
```diff
@@ -112,6 +80,29 @@ export default {
...
+        label = record.Account_Name;
+      } else if (moduleType === "Contacts") {
+        label = record.Full_Name;
+      } else if (moduleType === "users?type=ActiveUsers") {
+        label = record.full_name;
+      } else if (moduleType === "Attachments") {
+        label = record.File_Name;
+      } else {
+        label = value;
+      }
```
GTFalcao, line 99:
```suggestion
      const paramsObj = {
        Leads: "Full_Name",
        Accounts: "Account_Name",
        Contacts: "Full_Name",
        'users?type="ActiveUsers"': "full_name",
        Attachments: "File_Name",
      };
      const label = paramsObj[moduleType] ?? value;
```
Optional - another way to do it without repeating the assignments and comparisons.

GTFalcao (commented):
Hi @andrewjschuang , thanks for your contribution, nice refactor! I left a couple of small comments.

components/zoho_crm/zoho_crm.app.mjs
andrewjschuang: Thank you, this is MUCH better!

GTFalcao (approved):
Looks good to me! Moving it forward

vunguyenhung:

[Zoho_CRM_6228_1800.pdf](https://github.com/PipedreamHQ/pipedream/files/11441083/Zoho_CRM_6228_1800.pdf)

GTFalcao (approved):

PR #6391, opened 2023-05-10
[BUG] Airtable fix computed fields should not appear as editable

jcortes:
### <samp>🤖 Generated by Copilot at 5e4451d</samp>

Updated the Airtable component to handle computed fields and fix bugs. Added a utility function to exclude computed fields from the action props and a constant for the lookup field type. Bumped the versions of the actions and the `@pipedream/airtable` package.

### DIFF
--- a/components/airtable/actions/create-or-update-record/create-or-update-record.mjs
+++ b/components/airtable/actions/create-or-update-record/create-or-update-record.mjs
@@ -3,8 +3,8 @@ import common from "../common.mjs";
 import commonActions from "../../common/actions.mjs";
 
 export default {
-  key: "airtable-create-single-record-or-update",
-  name: "Create single record or update",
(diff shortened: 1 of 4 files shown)


components/airtable/common/utils.mjs
```diff
@@ -80,6 +80,26 @@ function fieldToProp(field) {
   };
 }
 
+function isComputedField(field) {
```
dylburger, line 83:
```suggestion
function isComputedField(field) {
```

dylburger: Changed comment to start with a verb

GTFalcao (approved):
Hi @jcortes , thanks for the fix, looks good to me!

vunguyenhung:
Hi, I can confirm that the computed fields are not shown anymore
image
image

PR #6558, opened 2023-05-19
possible fix for bingX Signature not match error

alysonturing:
## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough

### DIFF
--- a/components/bingx/actions/trade-new-order/trade-new-order.mjs
+++ b/components/bingx/actions/trade-new-order/trade-new-order.mjs
@@ -1,6 +1,7 @@
 import bingx from "../../bingx.app.mjs";
 
 export default {
+  type: "action",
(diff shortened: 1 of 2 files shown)


components/bingx/actions/trade-new-order/trade-new-order.mjs
```diff
@@ -56,12 +57,21 @@ export default {
       optional: true,
     },
   },
-  type: "action",
+  methods: {
+    // fix for error: "signature not match"
```
dylburger, line 61:
```suggestion
    // Fix for error: "Signature not match"
```

dylburger: Capitalized first letter of the comment, corrected grammar

PR #6562, opened 2023-05-19
[ACTIONS/SOURCES] Baselinker new actions sources

jcortes:
### <samp>🤖 Generated by Copilot at 6793696</samp>

This pull request adds a new app and several components for BaseLinker, an e-commerce platform, to the Pipedream repository. The app component defines the common methods and props for interacting with the BaseLinker API. The action components allow creating, updating, or deleting inventory products and orders. The source components emit events for new orders created and order status updated. The pull request also updates some dependencies and deletes some unused files.

### DIFF
--- a/components/baselinker/actions/create-order/create-order.mjs
+++ b/components/baselinker/actions/create-order/create-order.mjs
@@ -7,7 +7,7 @@ export default {
   name: "Create Order",
   description: "It allows adding a new order to the BaseLinker order manager. [See the Documentation](https://api.baselinker.com/index.php?method=addOrder).",
   type: "action",
-  version: "0.0.1685500561",
+  version: "0.0.1",
   props: {
     app,
     orderStatusId: {


michelle0927 (changes requested):
Looks good, other than needing to resolve some merge conflicts.
Suggestion about sources: They'll emit events on the first run, but I believe we're still supposed to be emitting "historical" events on deploy. What do you think about adding a deploy hook to polling.mjs?

michelle0927:
> > Looks good, other than needing to resolve some merge conflicts. Suggestion about sources: They'll emit events on the first run, but I believe we're still supposed to be emitting "historical" events on deploy. What do you think about adding a deploy hook to polling.mjs?
> 
> Hi @michelle0927 so I'm not doing that since it's a polling only strategy. It means the first run counts as a deploy polling the first resources right? IMO that's needed when we are dealling with webhooks.

Ok, interesting. I know that originally, we were required to create historical events for Polling sources ONLY, and then we started creating them for Webhook sources as well. Since these will emit events on the first run, I think that's fine. Maybe @dannyroosevelt can weigh in on when we should be emitting historical events in deploy().
The merge conflicts will still need to be resolved, but moving to QA!

vunguyenhung:

[BaseLinker_6456_1035.pdf](https://github.com/PipedreamHQ/pipedream/files/11595446/BaseLinker_6456_1035.pdf)

vunguyenhung:

Test report
[BaseLinker_6456_0946.pdf](https://github.com/PipedreamHQ/pipedream/files/11608098/BaseLinker_6456_0946.pdf)

Issue #6596, opened 2023-05-23
[BUG] Vestaboard Create Message - Characters doesn't seem to be the right type


johndbritton:
**Describe the bug**
The "Characters" option of "Vestaboard Create Message" is the wrong type.

**To Reproduce**
1. Use the Vestaboard Create Message action
2. Choose to send "Characters" not "Text"
3. Enter a 6x22 two dimensional array of integers
4. Error message:

```
Error - Request failed with status code 500
{"title":"Message was missing characters and text","status":500,"type":"https://javalin.io/documentation#internalservererrorresponse","details":[]}
```

**Expected behavior**
The message is created

**Additional context**
The issue seems to be that the parameter is being sent as a `String` rather than an array of arrays of integers.

Here's the relevant source: https://github.com/PipedreamHQ/pipedream/blob/87a461b9ac984b141311007c7d052ae8d3f9a64b/components/vestaboard/actions/create-message/create-message.mjs#L24-L29

Perhaps line 25 should be changed to something else.

cc/ @michelle0927 @malexanderlim  https://github.com/PipedreamHQ/pipedream/issues/6402 https://github.com/PipedreamHQ/pipedream/pull/6526

johndbritton:
@michelle0927 thanks for the quick reply, that indeed did the trick. I didn't realize the `{{ }}` were needed there.

dannyroosevelt:
@michelle0927 do we need to modify the prop type?

michelle0927:
> @michelle0927 do we need to modify the prop type?

@dannyroosevelt No, but I should probably update it to validate the input and try to convert it to an array if it's a string.

PR #6609, opened 2023-05-23
Vestaboard Create Message Update - allow characters prop to be string or array

michelle0927:
Updates the Vestaboard Create Message Action to allow the `characters` prop to be entered as either a string or an array.

Related issue https://github.com/PipedreamHQ/pipedream/issues/6596


vunguyenhung:
Hi, as I don't have a Vestaboard to test, I will skip the testing for this

PR #6616, opened 2023-05-24
Update links from /limits to /docs/limits

chipach:
## WHAT

Fix documentation links from /limits to /docs/limits

## WHY

Because I kept clicking the wrong link in the Pricing page on the website.

## HOW

Mass change of /limits to /docs/limits

### DIFF
--- a/docs/docs/code/bash/README.md
+++ b/docs/docs/code/bash/README.md
@@ -146,5 +146,5 @@ ls /tmp
 ```
 
 :::warning
-The `/tmp` directory does not have unlimited storage. Please refer to the [disk limits](/limits/#disk) for details.
+The `/tmp` directory does not have unlimited storage. Please refer to the [disk limits](/docs/limits/#disk) for details.
 :::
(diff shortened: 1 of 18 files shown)


dylburger:
@chipach `/docs` is actually assumed because it's set as the base path in the VuePress / Vercel app that serves docs. If you visit the preview links above and visit e.g. the Bash page where you made the change, you'll note that your link now direct to `/docs/docs/limits/#disk`, which has an extra `/docs`. 

Did a specific link not work for you before?

dylburger (changes requested):
See my comment in the main PR. These links are all now broken.

chipach:
https://pipedream.com/docs/pricing has at least 1 broken link.  Feel free to close this PR; I have no way of really testing this through however you serve your website, but the links from an HTML point of view relative to the directory structure seem wrong.

```
<p>But Free account have a <a href="/limits#daily-credits-limit">daily limit of free credits</a> that cannot be exceed. Standard <a href="/docs/limits/" class="">Pipedream platform limits</a> apply to Free Accounts as well.</p>
But Free account have a 
<a href="/limits#daily-credits-limit">daily limit of free credits</a>
 that cannot be exceed. Standard 
<a href="/docs/limits/" class="">Pipedream platform limits</a>
 apply to Free Accounts as well.
```

dylburger:
You can test locally using the instructions [here](https://github.com/PipedreamHQ/pipedream/tree/master/docs). I know this setup might seem odd, but it's the way we have to manage given the way we serve the site. Thanks for the PR regardless! If you see anything else awry please let us know.

PR #6693, opened 2023-05-30
Returning file path instead of file contents

dylburger:
## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough


components/microsoft_onedrive/actions/download-file/download-file.mjs
```diff
@@ -7,7 +7,7 @@ export default {
   name: "Download File",
   description: "Download a file stored in OneDrive. [See the documentation](https://learn.microsoft.com/en-us/onedrive/developer/rest-api/api/driveitem_get_content?view=odsp-graph-online)",
   key: "microsoft_onedrive-download-file",
-  version: "0.0.1",
+  version: "0.0.2",
```
dylburger, line 10:
```suggestion
  version: "0.0.2",
```

dylburger: Removed unnecessary double quotes around the version number.

PR #6698, opened 2023-05-30
Slack - Update icon emoji prop to list all available emojis

### LINKED ISSUE #6642: [BUG] Slack - New Reaction Added - Icon Emoji prop only loads 14 emojis
Slack.

Seems like currently slack New Reaction Added (Instant) only has 14 emojis anyway to add more and add a custom one also? 
image

For example, I want to use :round_pushpin: when a user reacts to that in any channel it sends the current message to a Chanel.

**Expected behavior:**
- I should be able to load more emoji when clicking Load more button. 
- I should be able to search for an emoji

https://api.slack.com/docs

michelle0927:
Resolves #6642


mendelsphotography (commented):
Looks good to me.

vellames (approved):
lgtm

PR #6731, opened 2023-06-01
RSS - store ids in db to filter against

michelle0927:
Updates actions "New Item in Feed" and "New Item From Multiple RSS Feeds" to store ids in the database and filter out any incoming events that have already been processed.

### DIFF
--- a/components/rss/package.json
+++ b/components/rss/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/rss",
-  "version": "0.5.2",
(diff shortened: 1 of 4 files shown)


vunguyenhung:
Hi @michelle0927, as Dylan has fixed [the issue here](https://pipedream-users.slack.com/archives/G01J50YGJGP/p1685634589351939?thread_ts=1685627193.799759&cid=G01J50YGJGP), do you think this is still needed?

PR #6739, opened 2023-06-01
Smartsheet - Update Triggers + Add New Row Deleted

### LINKED ISSUE #6733: [TRIGGER] Smartsheet - Update Triggers + Add New
Currently, the Smartsheet triggers are lean payloads - from their documentation,
> The callback payload is a "skinny" payload -- it indicates which objects changed and the type of events that occurred, but does not contain any data from the objects themselves.

That being said, the sources would be more useful if they contained the actual change as well.
> Your application receives the HTTP POST, acknowledges the callback by responding with HTTP status 200, and processes the callback. Because the callbacks are "skinny", make one or more requests to the Smartsheet API to retrieve the actual data associated with the change.

The request here is twofold:
1. Modify the existing triggers such that it processes the callback, and makes an additional call associated with the actual event to retrieve the actual change. For example, for New Row Added, show the value of the new row.
2. Create a new triggers (that retrieves the actual change) for the `row.deleted` event.
App: Smartsheet
Event: "New Row Deleted"

https://smartsheet.redoc.ly/tag/webhooksDescription#section/Intro-to-Webhooks/Webhook-Ownership-and-Access

michelle0927:
Resolves #6733

### DIFF
--- a/components/smartsheet/sources/new-row-deleted/new-row-deleted.mjs
+++ b/components/smartsheet/sources/new-row-deleted/new-row-deleted.mjs
@@ -21,7 +21,7 @@ export default {
     generateMeta(event) {
       return {
         id: event.id,
-        summary: `Row ${event.id} added`,
+        summary: `Row ${event.id} deleted`,
         ts: Date.parse(event.timestamp),
       };
     },


vunguyenhung:

Test report
[Smartsheet_6733_1718.pdf](https://github.com/PipedreamHQ/pipedream/files/11663275/Smartsheet_6733_1718.pdf)

PR #6972, opened 2023-06-21
Creating async options error messaging for Facebook Groups

malexanderlim:
### <samp>🤖 Generated by Copilot at d38e31b</samp>

Improved error handling for Facebook Groups source. The source now checks if the selected group has the Pipedream app installed and shows a helpful message if not.

### <samp>🤖 Generated by Copilot at d38e31b</samp>

> _Oh we are the coders of the Pipedream app_
> _And we work on the `postId` prop_
> _We handle the errors when the group is wrong_
> _And we tell the user how to get along_

## WHY

Screenshot 2023-06-21 at 12 16 43 PM

### <samp>🤖 Generated by Copilot at d38e31b</samp>

* Import and throw `ConfigurationError` for groups without Pipedream app ([link](https://github.com/PipedreamHQ/pipedream/pull/6972/files?diff=unified&w=0#diff-8e811b19be1470e1e31109e1a0191d72054a0c1e1466d194ee24e7be45bb58ddL1-R4), [link](https://github.com/PipedreamHQ/pipedream/pull/6972/files?diff=unified&w=0#diff-8e811b19be1470e1e31109e1a0191d72054a0c1e1466d194ee24e7be45bb58ddL30-R47))

### DIFF
--- a/components/facebook_groups/facebook_groups.app.mjs
+++ b/components/facebook_groups/facebook_groups.app.mjs
@@ -43,7 +43,7 @@ export default {
           })) || [];
         } catch (error) {
           console.error(error);
-          throw new ConfigurationError("Please doublecheck that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups).");
+          throw new ConfigurationError("Please double-check that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups).");
         }
       },
     },
--- a/components/facebook_groups/facebook_groups.app.mjs
+++ b/components/facebook_groups/facebook_groups.app.mjs
@@ -1,4 +1,7 @@
-import { axios } from "@pipedream/platform";
+import {
+  axios,
+  ConfigurationError,
+} from "@pipedream/platform";
 
 export default {
   type: "app",
@@ -27,16 +30,21 @@ export default {
       label: "Post",
       description: "Identifier of a post",
       async options({ groupId }) {
-        const opts = {
-          groupId,
-        };
-        const { data } = await this.listPosts(opts);
-        return data?.map(({
-          id, message,
-        }) => ({
-          label: message || id,
-          value: id,
-        })) || [];
+        try {
+          const opts = {
+            groupId,
+          };
+          const { data } = await this.listPosts(opts);
+          return data?.map(({
+            id, message,
+          }) => ({
+            label: message || id,
+            value: id,
+          })) || [];
+        } catch (error) {
+          console.error(error);
+          throw new ConfigurationError("Please doublecheck that the Pipedream app has been [installed](https://pipedream.com/apps/facebook-groups) in the group you have selected above.");
+        }
       },
     },
     maxResults: {
--- a/components/facebook_groups/facebook_groups.app.mjs
+++ b/components/facebook_groups/facebook_groups.app.mjs
@@ -43,7 +43,7 @@ export default {
           })) || [];
         } catch (error) {
           console.error(error);
-          throw new ConfigurationError("Please doublecheck that the Pipedream app has been installed in the group you have selected above. Learn more [here].(https://pipedream.com/apps/facebook-groups)");
+          throw new ConfigurationError("Please doublecheck that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups).");
         }
       },
     },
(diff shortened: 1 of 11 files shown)


components/facebook_groups/facebook_groups.app.mjs
```diff
@@ -27,16 +30,21 @@ export default {
...
+          const { data } = await this.listPosts(opts);
+          return data?.map(({
+            id, message,
+          }) => ({
+            label: message || id,
+            value: id,
+          })) || [];
+        } catch (error) {
+          console.error(error);
+          throw new ConfigurationError("Please doublecheck that the Pipedream app has been installed in the group you have selected above. Learn more [here].(https://pipedream.com/apps/facebook-groups)");
```
dylburger, line 46:
```suggestion
throw new ConfigurationError("Please double-check that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups).");
```

dylburger: Changed 'doublecheck' to 'double-check' according to Google's Developer docs style guide.

components/facebook_groups/facebook_groups.app.mjs
```diff
@@ -27,16 +30,21 @@ export default {
...
+          const { data } = await this.listPosts(opts);
+          return data?.map(({
+            id, message,
+          }) => ({
+            label: message || id,
+            value: id,
+          })) || [];
+        } catch (error) {
+          console.error(error);
+          throw new ConfigurationError("Please doublecheck that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups).");
```
dylburger, line 46:
```suggestion
          throw new ConfigurationError("Please double-check that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups).");
```

dylburger: Changed 'doublecheck' to 'double-check' to follow Google's Developer docs style guide (hyphenation).

components/facebook_groups/facebook_groups.app.mjs
```diff
@@ -27,16 +30,21 @@ export default {
...
+          const { data } = await this.listPosts(opts);
+          return data?.map(({
+            id, message,
+          }) => ({
+            label: message || id,
+            value: id,
+          })) || [];
+        } catch (error) {
+          console.error(error);
+          throw new ConfigurationError("Please double-check that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups#getting-started).");
```
dylburger, line 46:
```suggestion
          throw new ConfigurationError("Please double-check that the Pipedream app has been installed in the group you have selected above. Learn more [here](https://pipedream.com/apps/facebook-groups#getting-started).");
```

dylburger: Changed 'double-check' to 'double-check' to follow Google's Developer docs style guide (https://developers.google.com/style/hyphens)

dannyroosevelt (commented):
is that `post` prop common in all components? is that the only place you need to catch that?

malexanderlim:
I've added error messaging for all actions.

malexanderlim:
I've added error messaging for all actions.

components/facebook_groups/facebook_groups.app.mjs
```diff
@@ -7,7 +10,7 @@ export default {
     group: {
       type: "string",
       label: "Group",
-      description: "Identifier of a group",
+      description: "Identifier of a group. The Pipedream app must be installed in the Facebook group in order to make any API calls or receive data.",
       async options() {
         const userId = await this.getUserId();
         const opts = {
```
michelle0927, line 16: You could add a try/catch block in the `group` async options, just like in the `post` async options. Both of the sources have a `group` prop, so if they received an error, the user would see the error message & link while setting up the props, and wouldn't be able to deploy.

michelle0927 (approved):
Looks good. I left a comment/suggestion about displaying the error msg & link for sources.

components/facebook_groups/facebook_groups.app.mjs
malexanderlim: I think the issue with this is that the groups async option which uses `listGroups` wouldn't actually fail if the Pipedream app wasn't installed in the group, so it differs as compared to the `post` async option where `listPosts` actually requires the prior installation. 

I will merge this PR for now and come back to error message for sources, but I think that having the similar try/catch block on `groups` would not actually flag the lack of installation of the app given what the `groups` async options actually calls.

PR #7038, opened 2023-06-26
SellerCloud - New Components

### LINKED ISSUE #6561: [Components] SellerCloud
### App
SellerCloud
https://sellercloud.com/

Triggers:
1. Order Created: Triggered when a new order is added to SellerCloud.
2. Inventory Updated: Triggered when an item's inventory level changes.

Actions:
1. Create Product: Add a new product to the SellerCloud catalog.
2. Update Order Status: Modify the status of an existing order.
3. Adjust Inventory: Change the inventory level of a specific product.

_This list of suggested components was generated using ChatGPT_

michelle0927:
Resolves #6561 

Triggers

- New Order Created
- New Inventory Updated

Actions

- Create Product
- Update Order Status
- Adjust Inventory


components/sellercloud/sources/inventory-updated/inventory-updated.mjs
```diff
@@ -0,0 +1,71 @@
+import common from "../common/common.mjs";
+
+export default {
+  ...common,
+  key: "sellercloud-inventory-updated",
+  name: "New Inventory Updated",
```
GTFalcao, line 6:
```suggestion
  name: "Inventory Updated",
```
I think we can omit the "new" keyword here, since it's not actually a new item but rather an update.

I know our ESLint plugin warns source names should start with "New", but I understand it's a guideline, not a rule, and some names (such as this one) are more straightforward without it. What do you think?

GTFalcao (commented):
Hi @michelle0927 , thanks for your contribution!

I left one comment regarding a source name - but this looks good to go.

GTFalcao (approved):

PR #7047, opened 2023-06-27
Adding Sample Test Event for discord_bot-new-message-in-channel

dannyroosevelt:
Adding Sample Test Event for discord_bot-new-message-in-channel

Created via Pipedream


GTFalcao (approved):
@dannyroosevelt wasn't sure if this was just a test, but since you bumped the versions I'm approving it and moving it forward

PR #6552, opened 2023-05-19
feat: add new update transaction action to you_need_a_budget

Maxim-Filimonov:
## WHAT

Adds new update transaction action

## WHY

To be able to not only create but update existing transactions.

## HOW

?

PR #7099, opened 2023-06-30
Pr/6552

jcortes:
Fixes #6552


GTFalcao (commented):
Hi @jcortes , thanks for the contribution, I left just one comment regarding usability which can probably be better verified in the QA step, so I'm moving it forward.

PR #8157, opened 2023-09-20
Wise Payment Completed Triggers not receiving events

michelle0927:
According to the [documentation](https://docs.wise.com/api-docs/features/webhooks-notifications), Wise webhooks should return a `200 response`. Note: I haven't fully tested this, as I'm not sure how to do so without creating a monetary transaction.

## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough

### DIFF
--- a/components/wise/package.json
+++ b/components/wise/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/wise",
-  "version": "0.0.6",
+  "version": "0.0.7",
   "description": "Pipedream Wise Components",
   "main": "wise.app.mjs",
   "keywords": [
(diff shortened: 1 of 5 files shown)


zalmanlew:
@michelle0927 I think this issue is on Wise end, I've been emailing their support - haven't gotten anywhere yet.

I'm not certain changing to a 200 response would solve it.

As for testing it, you can create transactions (without funding them) which will create webhook events (it should if it was working).

vellames (approved):

zalmanlew:
This is definitely an issue on Wise - I don't think we need to change anything.

I received the webhooks for automatic pre-scheduled payments.

PR #8981, opened 2023-11-15
New Components - loop_returns

### LINKED ISSUE #8801: [Components] loopreturns
# loopreturns
#### Prompt
Emit new event when a new return is created. There are no required or optional props.
#### Prompt
Emit new event when a label is updated. It needs to be configured with the label ID as a prop.
#### Prompt
Emit new event when the status of a return has been updated. The updated return's ID is needed as an essential prop.
#### Prompt
Cancels a pending return request inside Loop. Requires return ID as a prop.
#### Prompt
Flags a particular return as important inside Loop. Requires return ID as a mandatory prop. Optionally, flag-reason can be included to clarify the cause.
#### Prompt
Starts the processing of a return inside Loop. Return ID is a required prop to initiate the process.

jcortes:
Resolves #8801.

### DIFF
--- a/components/loop_returns/loop_returns.app.mjs
+++ b/components/loop_returns/loop_returns.app.mjs
@@ -48,7 +48,7 @@ export default {
     getUrl(path) {
       return `${constants.BASE_URL}${constants.VERSION_PATH}${path}`;
     },
-    _makeRequest({
+    async _makeRequest({
       $ = this, path, headers, ...args
     } = {}) {
       const config = {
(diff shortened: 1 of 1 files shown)


components/loop_returns/actions/flag-return/flag-return.mjs
```diff
@@ -0,0 +1,41 @@
+import app from "../../loop_returns.app.mjs";
+
+export default {
+  key: "loop_returns-flag-return",
+  name: "Flag Return",
+  description: "Flags a particular return as important inside Loop. Requires return ID as a mandatory prop.",
```
michelle0927, line 6: Should include a doc link.
```suggestion
  description: "Flags a particular return as important inside Loop. Requires return ID as a mandatory prop. [See the documentation](https://docs.loopreturns.com/reference/flagreturn)",
```

michelle0927 (commented):
Looks good! Ready for QA!

michelle0927 (approved):

michelle0927 (approved):

Issue #9263, opened 2023-12-12
[Components] urlbae


pipedream-component-development:
# urlbae
## URLs
- https://urlbae.com/developers
#### Prompt
Generates a QR Code. The required prop is 'qr-content', which would be the desired content to be encoded in the QR code. The returned data will include a generated QR code.
#### Prompt
Returns a shortened URL. The required prop is 'url', being the URL that needs to be shortened. The output will be a shortened version of the input URL.

lcaresia:
@sergioeliot2039 @malexanderlim I cant' find the app to this task. Can you tell me when it's ready to develop?

Thanks

malexanderlim:
Hey @lcaresia , is this not showing up for you when you search for the app? I see this as `beta`:

https://github.com/PipedreamHQ/pipedream/blob/11723e0a24e83dae3f2a636fe2c6df7859ca8ee2/components/urlbae/urlbae.app.mjs#L4

https://pipedream.com/apps/urlbae

lcaresia:
@malexanderlim Sorry, I make a wrong comment and forgot to delete, everything is ok.

PR #9342, opened 2023-12-18
[Components] urlbae #9263

lcaresia:
## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough

### DIFF
--- a/components/urlbae/urlbae.app.mjs
+++ b/components/urlbae/urlbae.app.mjs
@@ -1,4 +1,4 @@
-import { axios } from "@pipedreamhq/platform";
+import { axios } from "@pipedream/platform";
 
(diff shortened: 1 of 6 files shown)


components/urlbae/urlbae.app.mjs
```diff
@@ -1,11 +1,40 @@
+import { axios } from "@pipedreamhq/platform";
```
GTFalcao, line 1: Looks like the package name is wrong, from the automated checks

GTFalcao (changes requested):

GTFalcao (approved):
I fixed the package name and moved the issue forward to QA.

PR #9474, opened 2023-12-26
[Trigger] l3mbda - New Sources

jcortes:
## WHAT

copilot:summary

## HOW

copilot:walkthrough

### DIFF
--- a/components/l3mbda/sources/common/webhook.mjs
+++ b/components/l3mbda/sources/common/webhook.mjs
@@ -1,0 +1,117 @@
+import app from "../../l3mbda.app.mjs";
+import { ConfigurationError } from "@pipedream/platform";
+
+export default {
+  props: {
+    app,
+    db: "$.service.db",
+    http: "$.interface.http",
+    from: {
+      propDefinition: [
+        app,
+        "from",
+      ],
+    },
+    to: {
+      propDefinition: [
+        app,
+        "to",
+      ],
+    },
+  },
+  methods: {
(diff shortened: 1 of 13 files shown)


components/l3mbda/sources/common/webhook.mjs
```diff
@@ -0,0 +1,126 @@
...
+import { ConfigurationError } from "@pipedream/platform";
+
+export default {
+  props: {
+    app,
+    db: "$.service.db",
+    http: {
+      type: "$.interface.http",
+      customResponse: true,
+    },
```
michelle0927, line 11: Since no custom response is being made, this can be shortened:
```suggestion
    http: "$.interface.http",
```

components/l3mbda/sources/common/webhook.mjs
```diff
@@ -0,0 +1,126 @@
...
+        app,
+        "to",
+      ],
+    },
+    token: {
+      propDefinition: [
+        app,
+        "token",
+      ],
+    },
```
michelle0927, line 29: The prop `token` is declared in `webhook.mjs`, and also in `new-erc-1115-transfer-instant`, `new-erc-20-transfer-instant`, and `new-erc-721-transfer-instant`. It looks like maybe it can be removed from `webhook.mjs`?
```suggestion
```

michelle0927 (changes requested):

michelle0927 (approved):

PR #9612, opened 2024-01-01
Espo CRM - New Components

### LINKED ISSUE #9490: [Components] espo-crm
# espo-crm
#### Prompt
Emit new event upon the creation, update, deletion, or changes of any field in various entity types. The user needs to specify the entity type and field to be tracked as required props.
#### Prompt
Creates a new contact in espo-crm. Required props include the contact's name, email, and phone number. Optional props include the contact's address, job title, and company.
#### Prompt
Creates a new task in espo-crm. The task name and due date are required props. Optional props might be the task description, assigned user, or priority level.
#### Prompt
Retrieves a list of accounts from espo-crm. This action does not require any props but you can optionally specify filters like name or industry to narrow down the results.

michelle0927:
Resolves #9490 

Note to QA: You may need to change your "assignment permissions" to create tasks. (https://forum.espocrm.com/forum/general/77547-in-case-your-api-calls-fail-with-http-status-403). Also make sure the api user has webhook permissions. Allow time for webhook events to arrive, as they can take a few minutes. 

## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough


components/espocrm/actions/list-accounts/list-accounts.mjs
```diff
@@ -0,0 +1,28 @@
...
+    const items = this.espoCrm.paginate({
+      resourceFn: this.espoCrm.getAccounts,
+      args: {
+        $,
+      },
+    });
+    const accounts = [];
+    for await (const item of items) {
+      accounts.push(item);
+    }
```
lcaresia, line 22:
```suggestion
    const accounts = Object.values(items);
```

lcaresia (approved):

components/espocrm/actions/list-accounts/list-accounts.mjs
michelle0927: Using `Object.values` doesn't work in this situation because `items` is coming from a generator function. We have to `await` the yield of each item from `*paginate`.

lcaresia (dismissed):
LGTM!

lcaresia (approved):

PR #9691, opened 2024-01-04
Use different request parameters when using a shared drive

andrewjschuang:
## WHAT

This PR fixes and updates listing function calls (`drive.files.list`) to support both My Drive and Shared Drives correctly.

## WHY

Previous code was listing files from My Drive only, therefore new files in Shared Drives were not being noticed.

## HOW

Refactored all `drive().files.list()` calls to use the existing functions `googleDrive.listFilesInPage()` and `utils.getListFilesOpts()`.

### DIFF
--- a/components/google_drive/package.json
+++ b/components/google_drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/google_drive",
-  "version": "0.6.13",
+  "version": "0.6.14",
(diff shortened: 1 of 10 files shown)


andrewjschuang:
There's no outstanding code changes, so skipping PR Review.
@vunguyenhung could you help QA the affected sources for regressions?

PR #9800, opened 2024-01-10
Including common file exception in key check

GTFalcao:
## WHAT

copilot:summary

copilot:poem

## WHY

Common files should be ignored when checking for component keys, just like test event files

## HOW

copilot:walkthrough


michelle0927 (approved):
@GTFalcao Thanks for the fix!

PR #9883, opened 2024-01-16
Microsoft Power BI - Triggers - Add option for custom id for dataset

michelle0927:
Updates the `datasetId` description in action Add Rows to Dataset Table.
Adds optional `customDatasetId` prop for sources.


luancazarine (approved):

PR #9925, opened 2024-01-17
Gmail Send Email - attachments prop refactor

### LINKED ISSUE #4784: [BUG] Gmail send-email: expression in attachment 'filename' not parsed
**Describe the bug**
I have a step where I want to send an email with a specific file, and with a filename based on a reference from the object explorer. However, the email is send with `"{{event.file.name}}"` instead of `"article.pdf"` for example.

**To Reproduce**
1. Create a Gmail send-email action
2. Enter a reference {{}} in the 'key' part of the attachment
3. Notice the send email has {{}} instead of the parsed reference value.

**Expected behavior**
The expression, e.g., `{{event.file.name}}`, is parsed before email is sent.

**Screenshots**
Screenshot 2022-11-19 at 15 58 38

**Desktop (please complete the following information):**
 - OS: macOS
 - Browser Safari
 - Version 16.1

andrewjschuang:
Resolves #4784.

### DIFF
--- a/components/gmail/actions/send-email/send-email.mjs
+++ b/components/gmail/actions/send-email/send-email.mjs
@@ -5,7 +5,7 @@ export default {
   key: "gmail-send-email",
   name: "Send Email",
   description: "Send an email from your Google Workspace email account",
-  version: "0.0.12",
+  version: "0.1.0",
   type: "action",
   props: {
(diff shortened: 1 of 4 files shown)


malexanderlim (commented):
lgtm!

PR #10006, opened 2024-01-23
[Docs] Add guide to export data store data

andrewjschuang:
## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough

### DIFF
--- a/docs/docs/data-stores/README.md
+++ b/docs/docs/data-stores/README.md
@@ -169,7 +169,7 @@ The following Node.js example action will export one batch of data via an HTTP P
 If the data contained in each key is large, consider lowering the number of `batch_quantity`.
 :::
 
-- Adjust your workflow memory and timeout settings according to the size of the data. A starting recommendation is to set memory at 512 MB and timeout to 60 seconds.
+- Adjust your [workflow memory and timeout settings](/workflows/settings/) according to the size of the data in your data store. Set the memory at 512 MB and timeout to 60 seconds and adjust higher if needed.
 
 - Monitor the exporting logs after each execution to make sure there weren't errors. Run the step as many times as needed until all your data is exported.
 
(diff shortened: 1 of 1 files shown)


docs/docs/data-stores/README.md
```diff
@@ -158,3 +158,70 @@ But you cannot serialize functions, classes, sets, maps, or other complex object
...
+:::tip
+If the data contained in each key is large, consider lowering the number of `batch_quantity`.
+:::
+
+- Adjust your workflow memory and timeout settings according to the size of the data. A starting recommendation is to set memory at 512 MB and timeout to 60 seconds.
+
+- Monitor the exporting logs after each execution to make sure there weren't errors. Run the step as many times as needed until all your data is exported.
+
+:::warning
+This action deletes the keys that were successfully exported. It is advisable to first run a test without deleting the keys. In case of any unforeseen errors, your data will still be safe.
```
ctrlaltdylan, line 177: @andrewjschuang I'm a bit confused on what this step is recommending.

I don't see a prop option to perform a dry run without deleting keys. How are users supposed to follow this advice without an option to control that behavior?

Maybe adding a `boolean` prop that controls deletion is the best way to handle this? I can see a situation where the user wants to delete the data after it's read, vs a simple back up first, then delete after. Just in case the consumer of the data fails to save it.

docs/docs/data-stores/README.md
```diff
@@ -158,3 +158,70 @@ But you cannot serialize functions, classes, sets, maps, or other complex object
 You can retrieve up to {{$site.themeConfig.DATA_STORES_MAX_KEYS}} keys from a data store in a single query.
 
 If you're using a pre-built action or code to retrieve all records or keys, and your data store contains more than {{$site.themeConfig.DATA_STORES_MAX_KEYS}} records, you'll receive a 426 error.
+
+## Exporting data to an external service
+
+In order to stay within the [data store limits](#data-store-limits), you may need to export the data in your data store to an external service.
+
+The following Node.js example action will export one batch of data via an HTTP POST request. You may need to adapt the code to your needs.
```
ctrlaltdylan, line 166: Perhaps adding a share workflow link that exposes this step as an HTTP endpoint?

Or a workflow that saves this data to a MySQL/Mongo/Supabase table?

ctrlaltdylan (changes requested):

docs/docs/data-stores/README.md
andrewjschuang: The user would have to comment out the code that deletes the data, but good suggestion, will add that

docs/docs/data-stores/README.md
andrewjschuang: Good idea!

ctrlaltdylan (approved):

PR #10029, opened 2024-01-24
Update Monday.com integration to API Version 2024-01

### LINKED ISSUE #10017: [FEATURE] Update Monday.com integration to API Version `2024-01`
Monday.com recently released a new version of their API, which contains many breaking changes; please see the summary of changes for all of the updates:
https://developer.monday.com/api-reference/docs/migrating-to-v-2023-10#summary-of-changes

As a result of their changes, the majority of our triggers and actions are currently broken.

I've tested a handful of components, and specified which errors are being returned.

This list below is not conclusive, but the request is to update the integration to the version `2024-01` to bring our integration up to the latest version so that our registry components are functional again.

**Triggers**
- All triggers are failing with the async option to list boards due to the GraphQL changes.

1. New Item (Instant)
`TypeError: Cannot read properties of undefined (reading 'boards')`

2/3/4. New Column Value Updated (Instant), New Name Updated (Instant), New Sub-Item Update (Instant), New Specific Column Updated (Instant)...
Async options call to List Boards:
`Error fetching boards: undefined`

5. New Boards 
Does not return new boards when testing.

6. New User
Does not return new users.
...

**Actions**
1. Create board
`Failed to create board: undefined`

2. Create item
`Cannot read properties of undefined (reading 'filter')`

...

michelle0927:
Resolves #10017

### DIFF
--- a/components/monday/actions/create-board/create-board.mjs
+++ b/components/monday/actions/create-board/create-board.mjs
@@ -6,7 +6,7 @@ export default {
   name: "Create Board",
   description: "Creates a new board. [See the documentation](https://api.developer.monday.com/docs/boards#create-a-board)",
   type: "action",
-  version: "0.0.4",
+  version: "0.0.5",
   props: {
     monday,
     boardName: {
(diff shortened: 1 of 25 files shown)


GTFalcao (dismissed):
LGTM!

PR #10000, opened 2024-01-22
Apollo.io - New Actions

michelle0927:
Resolves #9939

PR #10072, opened 2024-01-25
Apollo.io - Set accountId as optional in Create Customer & Update Customer actions

michelle0927:
Requested by customer [here](https://github.com/PipedreamHQ/pipedream/pull/10000#issuecomment-1910249080).


lcaresia (approved):
LGTM!

PR #10190, opened 2024-02-01
Fix filter array case sensitiveness

### LINKED ISSUE #10189: [BUG] Filter app - Text arrays are not case sensitive in arrays
**Describe the bug**
For Filter components - when comparing values in text arrays with `caseSensitive = false`, the comparison is not correctly ignoring case.

**To Reproduce**
1. Create a **Filter - Continue Execution** step
2. Input `ANYTHING` as the initial value
3. Select `IN_ARRAY` for the condition and `TEXT` for the initial value type
4. Type `[ "ANYTHING" ]` for the second value
5. Set case sensitive to `false`
6. Test the step -- should continue but it exits.

**Additional context**
Reported by a user

andrewjschuang:
Resolves #10189.


GTFalcao (approved):
LGTM!

PR #10213, opened 2024-02-02
Node 20 Playwright and Puppeteer support

js07:
Fix the `@pipedream/browsers` package, Puppeteer components, and Playwright components.

Pin the `playwright-core`, `puppeteer-core`, and `@sparticuz/chromium` packages to the latest versions, which are compatible with Node 20.


ctrlaltdylan (dismissed):
👍  LGTM.

I assume we won't merge this until node v20 is fully out

malexanderlim (approved):
LGTM, tested all actions successfully on the new version!

PR #10262, opened 2024-02-07
Attempt to fix issue with escaped entities in RSS

cfjedimaster:
## WHAT

I ran into an issue where a workflow based on my RSS feed, https://www.raymondcamden.com/feed.xml, wasn't working correctly. The escape HTML entities in the title were being stripped out. I determined it came from the trigger. While looking at feedparser's github repo, I found that it was intentional, https://github.com/danmactough/node-feedparser/issues/243, and that a workaround was to check for the meta values. My fix does this and overwrites the item.title ortiginally set. I did *not* test this as it was small, but, it was pretty minor.

copilot:summary

copilot:poem

## WHY

I think I kinda covered it above.

## HOW

copilot:walkthrough


michelle0927 (approved):
LGTM!

components/rss/app/rss.app.ts
```diff
@@ -93,6 +93,16 @@ export default defineApp({
...
           let item: any = this.read();
+          /* 
+          Valid escaped entries in RSS are being stripped out, see issue: https://github.com/danmactough/node-feedparser/issues/243
+          Author suggests using the values below, so I check for them, if they exist, overwrite title.
+          */
+          if(item['atom:title'] && item['atom:title']['#']) {
+            item.title = item['atom:title']['#'];
+          } else if(item['rss:title'] && item['rss:title']['#']) {
+            item.title = item['rss:title']['#'];
+          }
```
michelle0927, line 104: Hi, @cfjedimaster. I haven't tested, but it looks like you need to check that `item` exists here.
```suggestion
          if(item && item['atom:title'] && item['atom:title']['#']) {
            item.title = item['atom:title']['#'];
          } else if(item && item['rss:title'] && item['rss:title']['#']) {
            item.title = item['rss:title']['#'];
          }
```

cfjedimaster: I'm probably a few days from being able to do this, but your mod looks good and if you were to edit my PR, I wouldn't complain (not even sure if that's a thing).

cfjedimaster:
So I made the code change to check for item, and I have a failure: " could not find app name slug in directory 'formpress'"  I have -no- idea what that means in this case.

michelle0927 (approved):

PR #10263, opened 2024-02-07
New Components - btcpay-server

### LINKED ISSUE #9662: [Components] btcpay-server
# btcpay-server
#### Prompt
Generates a new payment request for a user. Required props include the amount and currency, while optional props can include an expiry date or custom metadata.
#### Prompt
Fetches the balance of your on-chain store wallet. It only requires the store wallet ID as a prop.

michelle0927:
Resolves #9662.


lcaresia (approved):
LGTM!

PR #10319, opened 2024-02-09
Elevenlabs new action

### LINKED ISSUE #9089: [Action] Elevenlabs - Add Voice
Creating a new instant clone

https://elevenlabs.io/docs/api-reference/add-voice

GTFalcao:
Closes #9089 

I was not able to fully validate this as it seems to be a premium feature:
```
{"status":"can_not_use_instant_voice_cloning","message":"Your subscription has no access to use instant voice cloning, please upgrade."}
```
However, I was receiving an error related to the file format beforehand, and after adjusting the format I started receiving this error instead - so it's very likely that the format is correct now and only needs the proper subscription level.


components/elevenlabs/actions/add-voice/add-voice.mjs
```diff
@@ -0,0 +1,65 @@
...
+    if (description) data.append("description", description);
+
+    files.forEach((file) => {
+      data.append("files", fs.createReadStream(file.includes("tmp/")
+        ? file
+        : `/tmp/${file}`));});
+
+    if (labels) data.append("labels", labels);
+    data.append("name", name);
+
```
luancazarine, line 55: for (const [
      key,
      value,
    ] of Object.entries(data)) {
      formData.append(key, value);
    }
you can do it like this instead of adding one by one, just a suggestion

luancazarine (approved):
Hi @GTFalcao, thanks for your contribution! I just added a suggestion but I'm moving it to Ready For QA anyway.

components/elevenlabs/actions/add-voice/add-voice.mjs
GTFalcao: Thanks for the suggestion, in this case however I have to check if the variable is truthy before appending it, which is why I left it this way.

luancazarine (approved):

PR #10323, opened 2024-02-09
Shopify Partner improvement

### LINKED ISSUE #10316: [FEATURE] Shopify Partner - Async additionalProps to validate Shopify AppID prop

There are two primary issues here:
1. The description for Shopify App ID could potentially be misleading. 
Currently: gid://partners/App/5814709241
New: Open your app in the partner portal, and look at the URL to find its ID. If your URL is [https://partners.shopify.com/3027494/apps/51358007297/overview](https://partners.shopify.com/3027494/apps/51358007297/overview), enter `51358007297` here.

2. For each of the triggers that utilize Shopify App ID prop, you are able to create a source with any appID submitted without validation or feedback that the App ID is either invalid, or not linked to your partner account.

This request is three parts: 
1. To fix the description so that it is more clear as to which ID the user will input here.
2. Utilize async additionalProps to validate the appID, using this graphQL query:

```query {
  app(id: "gid://partners/App/<app_id>") {
    name
  }
}
```

3. Please use markdown notes to give this feedback to the user, e.g.
Shopify App: [Name of App Returned]

If there is an error, throw an error alert prop that states something like...:
"Invalid App ID. Please double check the app ID and ensure that it is correct, and visible within your organization.

GTFalcao:
Closes #10316


components/shopify_partner/sources/new-app-charges/new-app-charges.mjs
```diff
@@ -5,7 +5,7 @@ export default {
   key: "shopify_partner-new-app-charges",
   name: "New App Charges",
   type: "source",
-  version: "0.0.14",
+  version: "0.0.15",
```
lcaresia, line 8:
```suggestion
  version: "0.1.0",
```

lcaresia (approved):
LGTM!

Just one fix

components/shopify_partner/sources/new-app-charges/new-app-charges.mjs
GTFalcao: This source does not use the appId prop which is what was updated on this PR, hence why it only receives a minor version bump

lcaresia (approved):

PR #10343, opened 2024-02-12
YouTube (Data API) - replace forUsername with forHandle

### LINKED ISSUE #10338: [FEATURE] YouTube (Data API) - replace forUsername with forHandle
YouTube (Data API) - New Videos by Username is not easy to use, as usernames are no longer widely used on YouTube and have been replaced by @handles instead. After doing a bit of research, we should likely deprecate our support for usernames, and support handles instead here.

Screenshot 2024-02-12 at 11 26 22 AM

This ask is to replace the YouTube (Data API) - New Videos by Username with: YouTube (Data API) - New Videos by Handle instead. 

Please also include some validation to ensure that if the end-user accidentally includes the @ in the handle, that we do not pass this in the call.
- User provides a handle (and also handle the case where a user accidentally includes the @)
- Include instructions as to which part of the YouTube URL encompasses a handle. 
- Use async additionalProps with markdown notes to validate that the handle is valid, so the user has earlier feedback before clicking to create source.

michelle0927:
Resolves #10338


GTFalcao (approved):
LGTM!

The old components should probably be redirected to the new ones when this is merged

PR #10355, opened 2024-02-13
Snowflake - Use full database name to list columns

### LINKED ISSUE #10337: Snowflake - `columns` prop async handling broken
### App
Snowflake

### Summary:
Operation error: does not exist or not authorized

### Details:
We are attempting to use the Snowflake App and the Insert_Rows / Insert_Row action but we continue to receive an error when the async option Columns is being configured.  The error message received is as follows:

Operation Failed Error
SQL compilation error:
Table 'SANDBOX_WORKFLOW_LOGS' does not exist or not authorized.

Link to workflow:
https://pipedream.com/@onecare-data/projects/proj_Oesdbz/snowflake-acct-insert-example-p_3nCybaB/build

We tested the user credentials and now it is not an issue with the configuration in Snowflake.  

However, it looks like using a custom expression to specify the full database name (Database, Schema, Table) works.

Can component dev team investigate, and verify if we should always be passing the full database name? Interestingly, the action seems to work with just the table name for our Pipedream account, but if passing the full database name is more fail-proof, we should make the switch to that for the component.

### Screenshots:

**Prop using async options is not working**
Screenshot

**Using custom expressions works**
Screenshot 2024-02-12 at 3 07 52 PM

michelle0927:
Resolves #10337

### DIFF
--- a/components/snowflake/actions/insert-multiple-rows/insert-multiple-rows.mjs
+++ b/components/snowflake/actions/insert-multiple-rows/insert-multiple-rows.mjs
@@ -6,7 +6,7 @@ export default {
   key: "snowflake-insert-multiple-rows",
   name: "Insert Multiple Rows",
   description: "Insert multiple rows into a table",
-  version: "0.0.6",
+  version: "0.0.7",
   props: {
     snowflake,
     tableName: {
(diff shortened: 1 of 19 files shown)


lcaresia (approved):
LGTM!

PR #10370, opened 2024-02-14
Airtable - Return fieldId for Airtable triggers

### LINKED ISSUE #10362: [FEATURE] Return fieldId for Airtable triggers
Similar to the request in https://github.com/PipedreamHQ/pipedream/issues/9803, the request here would be to enhance the usability of the Airtable triggers, by not only returning the field name where applicable, but the fieldId as well. 

> The challenge I was facing is that the JSON data that’s passed through from the new-airtable-row trigger doesn’t include the column IDs. So if I want to trigger an email that uses, say someone’s first name from the column marked “first name” and it gets changed to “first_name” then that variable isn’t going to be available to the email. 

> What I’d hope for is a list of properties for the “new-row” trigger which include both the column id as well as the column title. That way I can securely pass the first name through to the email however it gets tweaked in the original airtable column.

This ask is to return fieldId along with field names where applicable for our Airtable triggers.

michelle0927:
Resolves #10362


lcaresia (approved):
LGTM!

lcaresia (approved):

PR #10458, opened 2024-02-19
docs: File Stores, improve $.files.dir() example

NReilingh:
## WHAT

copilot:summary

copilot:poem

## WHY

const dirs = $.files.dir() isn't good semantically. Nodes is more accurate.

## HOW

copilot:walkthrough


michelle0927 (approved):
LGTM!

PR #10463, opened 2024-02-19
Notion - add includeNewPages prop to Update Page in Database

### LINKED ISSUE #10223: [Notion] Update Page in Database Trigger also triggers for created page
**A clear and concise description of what the problem is. Ex. I'm always frustrated when [...]**
So I have a question on this trigger. Is there a way to have it to where it does not send the trigger when the database entry is created, but only when the specified fields are updated. Testing my workflow to notify users on a specific day when a formula field is updated, but it is also sending out a notification when the page is created as well.
This could be a selectable option in the trigger deployment, something like "include page creation" with a true or false option.

**Do you have a workaround?**
No

No

**Additional context**
No

michelle0927:
Resolves #10223


GTFalcao (approved):
LGTM!

GTFalcao (approved):

PR #10503, opened 2024-02-21
Snowflake - Add database and schema props

### LINKED ISSUE #10485: [FEATURE] Snowflake - Add database and schema props
Instead of having to create multiple separate account connections to work with different databases and schemas, we should have props to specify these. This will allow the end-user to use our components more seamlessly, as they will be able to specify the Database - Schema - Table, which is both more flexible and makes the List Columns call more fail-proof.

michelle0927:
Resolves #10485


lcaresia (approved):
LGTM!

PR #10612, opened 2024-02-26
New Components - zixflow

### LINKED ISSUE #10573: [Components] zixflow
# zixflow
#### Prompt
Creates a new activity or task within Zixflow. Required props are the task title and task description. Optional props include due date and assigned members.
#### Prompt
Updates an existing activity or task in Zixflow. The activity ID is a required prop. Optional props could be changes to the task title, task description, due date, or assigned members.
#### Prompt
Removes an existing activity or task from Zixflow. The only required prop is the activity ID.

lcaresia:
Resolves #10573.


components/zixflow/package.json
```diff
@@ -12,4 +12,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
\ No newline at end of file
+}
```
GTFalcao, line 15: Package version should be updated and the `@pipedream/platform` should be installed

GTFalcao (commented):
Only the package needs to be updated, otherwise good to go!

components/zixflow/package.json
```diff
@@ -12,4 +12,4 @@
   "publishConfig": {
     "access": "public"
   }
```
GTFalcao, line 14: Still needs `@pipedream/platform` added as a dependency

PR #10620, opened 2024-02-27
New Components - keysender

### LINKED ISSUE #10516: [Components] keysender
# keysender
## URLs
- http://panel.keysender.co.uk/api
#### Prompt
Create a new order dispatch within keysender. Required props for this action include the order details, such as the product ID, quantity, and recipient information. Optional props could be additional information about the order like the order type or priority status.

lcaresia:
Resolves #10516.


components/keysender/actions/create-transaction/create-transaction.mjs
```diff
@@ -0,0 +1,65 @@
...
+      description: "Email address of the payer",
+    },
+    databaseId: {
+      propDefinition: [
+        keysender,
+        "databaseId",
+      ],
+    },
+    amount: {
+      type: "string",
```
GTFalcao, line 24: Is this supposed to be a string? API documentation lists it as a float, so maybe it should be a string prop but converted to a number (and checked to be a valid number) before being sent

GTFalcao (commented):
Just one thing to be checked, but moving it to QA!

GTFalcao (approved):

PR #10630, opened 2024-02-28
Asana - updated pagination of props

michelle0927:
Adds a default limit of 25 results per page for props that use pagination.


GTFalcao (approved):
LGTM!

PR #9281, opened 2023-12-13
[FIX] salesforce terminology misalignment replaced object to record

jcortes:
### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 1e659d1</samp>

This pull request adds new components to the `@pipedreamhq/platform` package that emit events from various web services and platforms. It also updates the `@pipedream/salesforce_rest_api` package to version `1.0.0` and renames the sources and files to use the term `record` instead of `object` to match the Salesforce terminology.

### <samp>🤖[[deprecated]](https://githubnext.com/copilot-for-prs-sunset) Generated by Copilot at 1e659d1</samp>

*  Rename and update sources to use `record` instead of `object` in name, key, description, and version to match Salesforce terminology ([link](https://github.com/PipedreamHQ/pipedream/pull/9281/files?diff=unified&w=0#diff-4d33244013f381da4e72ac33b8d2a4f509243d27a17958877b5bba27382da111L7-R10),[link](https://github.com/PipedreamHQ/pipedream/pull/9281/files?diff=unified&w=0#diff-ba5234637ae917b7e6b98fb8daca5fe1039b8dc241f98fc19489c725f8087399L8-R11),[link](https://github.com/PipedreamHQ/pipedream/pull/9281/files?diff=unified&w=0#diff-38a0d767f3998125ff06af9ffc6fe61e67ef1ce6cffc2f8decdd55eae1b88a01L8-R11),[link](https://github.com/PipedreamHQ/pipedream/pull/9281/files?diff=unified&w=0#diff-fc094cb6d2b46787ac9fdd7d215e2de85b517cf793aa3716d86f15dd2748768aL8-R11))
* Bump up the version of the `@pipedream/salesforce_rest_api` package to `1.0.0` to reflect the breaking changes in the source names and keys ([link](https://github.com/PipedreamHQ/pipedream/pull/9281/files?diff=unified&w=0#diff-69edd3d7bcf28d5b76b6be1ff8f0765e83f1a3027b155032457ba767c04df46eL3-R3))
  - `diffy` for emitting events when a web page changes ([link](https://github.com/PipedreamHQ/pipedream/pull/9281/files?diff=unified&w=0#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bbR1536-R1538))
[...]

PR #10640, opened 2024-02-29
Fix names of Salesforce New Object Updated sources

js07:
Replace references to "object" with "record" in the names, keys, and descriptions of the Salesforce New Object Updated sources.

In Salesforce the Object is the table and Record is a row within a Table/Object.

Related: #9281

### DIFF
--- a/components/salesforce_rest_api/package.json
+++ b/components/salesforce_rest_api/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/salesforce_rest_api",
-  "version": "1.2.0",
+  "version": "1.2.1",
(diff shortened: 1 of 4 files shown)


GTFalcao (commented):
Looks good to me, though I'm not sure we should update the component `key`s (instead of just their names), as I believe that would cause users to have to reconfigure the component in existing workflows - @malexanderlim please check this out when you're able, as if the proper move is indeed to rename the keys, we'd need to redirect the old component to the new one.

GTFalcao (approved):

malexanderlim:
Unpublished and redirected to new version.

PR #10654, opened 2024-02-29
RSS - Set the Published After to default to CURRENT_DATE - 1 Day

### LINKED ISSUE #10626: [BUG] RSS: New Item in Feed source does not emit new event when Published After Than is NOT specified
**Describe the bug**
RSS feed does not emit new event when Published After Than is NOT specified

**To Reproduce**
1. Create a new RSS: New Item in Feed source with the Feed URL: `https://tsn.ua/rss/full.rss` without setting Published After Than prop
image
2. Wait until there's a new rss item, then click Run Now
3. Observe that there's no event emitted

When I specify Published After Than prop, I can see new event emitted normally
image

**Expected behavior**
The RSS source should use the source creation time as Published After Than prop and emit new event correctly

michelle0927:
Resolves #10626


lcaresia (approved):
LGTM!

PR #10764, opened 2024-03-05
Google Drive - add filterType prop to List Files action

### LINKED ISSUE #10651: [BUG] Google Drive - List Files action's `Filter Text` prop does not works for mp4 extension
**Describe the bug**
When the Google Drive has multiple mp4 files with the file name pattern {number}.mp4, the Google Drive - List Files action's  `Filter Text` prop will not work correctly

**To Reproduce**
1. Upload multiple mp4 files to your Google Drive folder and name them follow the pattern {number}.mp4
image
2. Create the Google Drive - List Files action to list files in the above Google Drive folder, then set the `Filter Text` prop to `4.mp4`
image
3. Observe that the result contains **all** mp4 files -- which is incorrect. The result should only contains `4.mp4` file

**Expected behavior**
In the Google Drive - List File code, I see that we [always uses `contains` in the search query string](https://github.com/PipedreamHQ/pipedream/blob/61eb41429854cfe7bade98c6e0cdfce2d1e42caf/components/google_drive/actions/list-files/list-files.mjs#L57-L61). We need to add an optional prop `Filter Type` with options: `CONTAINS` or `EXACT MATCH` so user can control how the action filter files

**Additional context**
First raised by a customer here: https://pipedream-users.slack.com/archives/CPTJYRY5A/p1709215450193089

michelle0927:
Resolves #10651


lcaresia (dismissed):
LGTM!

lcaresia (approved):

PR #10823, opened 2024-03-07
Twitter sources improvement/bugfix

### LINKED ISSUE #10698: Twitter
### App
Twitter

### Summary:
The media fields are not returned

### Details:
The twitter get DM source is not returning everything correctly. When I run the same command that the source is, my response has the media expansions which includes the URL. In the log I can see the command being run includes the media expansions and url field I highlighted in my test code, but the resulting object coming back from the source does not. I think this is something todo with how totalDatais built within this file. But I cannot see why some of the response payload is left out.  Without the URL, the image cannot be retrieved. The downside is the workaround is to make an added call to dm_events which has an incredibly restrictive rate limit.

### Screenshots:
https://files.slack.com/files-pri/TMZFXGSBF-F06MLBY34EA/image.png
Screenshot

GTFalcao:
Closes #10698 

Note: this is difficult to validate due to the restrictions on the Twitter API. I validated the request structures with the basic 'Get User' action which works for the authenticated user, and I tested the implemented functions with sample responses from the Twitter API documentation.

Explaining the original issue and this solution:

Multi-item requests usually return an object containing `data` (array of primary results) and `includes` (array of secondary objects related to or referenced by the primary ones). Paginated requests then combine all of the individually performed requests' `data` and `includes`.

Previously, sources looked only at the `data`, to emit an event for each item received, and ignored the received `includes`. Now, they will perform a function (`getObjIncludes`) on each item in `data` before emitting it. This function scans the item for references to secondary objects (e.g. `pinned_tweet_id` in a User item, referencing a Tweet; or `owner_id` in a List item, referencing a User), then uses the assembled list of references to scan `includes` for these objects, and finally adds them to an `includes` array on the original item within `data`.

Thus, each emitted event will include the item and all referenced objects' data within it.


lcaresia (approved):
LGTM!

PR #10882, opened 2024-03-12
New Components - sendloop

### LINKED ISSUE #10801: [Components] sendloop
# sendloop
#### Prompt
Emit new event when a subscriber's status changes to 'hard bounce'. Required props: subscriber's email. Optional props: bounce status.
#### Prompt
Emit new event when a new subscriber is added to the list. Required props: subscriber's email. Optional props: added time.
#### Prompt
Emit new event when a subscriber opens an email. Required props: subscriber's email. Optional props: email's subject, open time.
#### Prompt
Adds a new subscriber to a specified list. In this action, you need to provide an 'email address' prop which will be used to identify the subscriber. You also need to provide a 'list ID' prop to specify which list the subscriber will be added to.
#### Prompt
Removes an email address from a specified list. The 'email address' prop is required to identify which subscriber should be removed. The 'list ID' prop is also necessary to indicate from which list the subscriber will be removed from.

michelle0927:
Resolves #10801.


components/sendloop/actions/add-subscriber/add-subscriber.mjs
```diff
@@ -0,0 +1,42 @@
+import sendloop from "../../sendloop.app.mjs";
+
+export default {
+  key: "sendloop-add-subscriber",
+  name: "Add Subscriber",
+  description: "Adds a new subscriber to a specified list.",
```
GTFalcao, line 6: Do we have API documentation links for these components?

components/sendloop/sources/new-email-open/new-email-open.mjs
```diff
@@ -0,0 +1,43 @@
+import common from "../common/base.mjs";
+import sampleEmit from "./test-event.mjs";
+
+export default {
+  ...common,
+  key: "sendloop-new-email-open",
+  name: "New Email Open",
```
GTFalcao, line 7:
```suggestion
  name: "New Email Opened",
```
Seems like this would be more clear (file/folder names and component key would be changed as well) - unless this is a very specific business name? ("email open" as a noun/entity)

components/sendloop/sources/new-subscriber/new-subscriber.mjs
```diff
@@ -0,0 +1,43 @@
+import common from "../common/base.mjs";
+import sampleEmit from "./test-event.mjs";
+
+export default {
+  ...common,
+  key: "sendloop-new-subscriber",
+  name: "New Subscriber",
```
GTFalcao, line 7:
```suggestion
  name: "New Subscriber Added",
```
I think this would be more in accordance with the source naming guidelines as well, to avoid mistaking this for an action which creates a new subscriber

GTFalcao (commented):
Code looks good, I suggested some adjustments on the components' names, but moving it to QA

components/sendloop/actions/add-subscriber/add-subscriber.mjs
michelle0927: You have to be logged into the UI to see the docs. 😦

GTFalcao (approved):

PR #10884, opened 2024-03-12
New Components - shorten_rest

### LINKED ISSUE #10863: [Components] shorten-rest
# shorten-rest
#### Prompt
Gets the click data for a specific URL. Requires the URL as a prop.
#### Prompt
Shortens a given long URL into an alias. Requires the long URL and an optional alias name as props. If the alias name is not provided, the system generates one. If the domain input is not provided, it defaults to short.fyi.

jcortes:
Resolves #10863.

### DIFF
--- a/components/shorten_rest/actions/get-clicks/get-clicks.mjs
+++ b/components/shorten_rest/actions/get-clicks/get-clicks.mjs
@@ -1,26 +1,29 @@
-import shortenRest from "../../shorten_rest.app.mjs";
-import { axios } from "@pipedream/platform";
+import app from "../../shorten_rest.app.mjs";
 
 export default {
   key: "shorten_rest-get-clicks",
-  name: "Get Click Data for a URL",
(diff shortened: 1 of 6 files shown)


components/shorten_rest/actions/get-clicks/get-clicks.mjs
```diff
@@ -0,0 +1,29 @@
...
+  name: "Get Clicks",
+  description: "Gets the click data. [See the documentation](https://docs.shorten.rest/#tag/Click/operation/GetClicks)",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    app,
+  },
+  methods: {
+    getClicks(args = {}) {
+      return this._makeRequest({
```
GTFalcao, line 14: I think the intention here was to call the app's request method
```suggestion
      return this.app._makeRequest({
```

GTFalcao (changes requested):
Just one change needed!

GTFalcao (dismissed):
LGTM!

GTFalcao (approved):

PR #11033, opened 2024-03-18
New Components - clicksend

### LINKED ISSUE #10991: [Components] clicksend
# clicksend
#### Prompt
Emit new event when an incoming SMS is received. Required props would be the receiver's phone number and the content of message.
#### Prompt
Emit new event as soon as a new SMS message is received. Required props could be the receiver's contact number and message details.
#### Prompt
Emit new event when a new voice message is received or sent. Required props would be the recipient's number, length of voice message and the audio message file.
#### Prompt
Creates a new contact in a specific list. Required prop is 'list_id'. Optional props are 'name', 'phone_number', and 'email'.
#### Prompt
Sends a new SMS to one or several recipients. Required prop is 'recipient_number'. Optional prop is 'message'.
#### Prompt
Sends a new MMS to one or multiple recipients. Required props are 'recipient_number' and 'file_url'. Optional prop is 'message'.

jcortes:
Resolves #10991.


lcaresia (approved):
LGTM!

PR #11097, opened 2024-03-20
Adding language about polling vs. webhook triggers for Github.

malexanderlim:
## WHAT

copilot:summary

copilot:poem

## HOW

copilot:walkthrough


components/github/README.md
```diff
@@ -1,257 +1,12 @@
-# Github API Integration Platform
+# Getting Started
 
-### Connect Github + 1000s of apps, remarkably fast.
+## Github Triggers: Webhooks vs. Polling
+The Github triggers are built in way that configures the trigger as a webhook if you are the admin of a repo, and falls back to polling if you are not the admin of a repo.
```
dannyroosevelt, line 4:
```suggestion
The Github triggers in Pipedream enable you to get notified immediately via a webhook if you have `admin` rights on the repo you're watching. Otherwise you can still poll for updates at a regular interval for any other repo where you might not have `admin` rights.
```

dannyroosevelt: Doesn't need to be this language exactly, but I think yours could use a little more context for users who might not be familiar with Pipedream.

components/github/README.md
```diff
@@ -1,257 +1,12 @@
...
 
----
+**Example: New or Updated Issue (Admin)**
```
dannyroosevelt, line 6: What does (Admin) mean here? That's a little confusing to me

components/github/README.md
```diff
@@ -1,257 +1,12 @@
...
+The user is an administrator of the repo, and this trigger will be configured as a webhook - so any time there is a new or updated issue in the repo, an event will automatically be emitted.
```
dannyroosevelt, line 7:
```suggestion
If you are an admin on the repo, this trigger will be configured as a webhook — so any time there is a new or updated issue in the repo, an event will immediately get emitted.
```

components/github/README.md
```diff
@@ -1,257 +1,12 @@
...
-You can read about our platform security and privacy [here](https://pipedream.com/docs/privacy-and-security/).
-
-If you'd like to report a suspected vulnerability or security issue, or have any questions about the security of the product, please contact our security team at **security@pipedream.com**.
-
-## Troubleshooting
-
-Note: Event Source [New Card in Column](https://github.com/PipedreamHQ/pipedream/blob/master/components/github/sources/new-card-in-column/new-card-in-column.mjs) only supports legacy (classic) projects.
-
-Please [reach out](https://pipedream.com/support/) to the Pipedream team with any technical issues or questions about the Github integration. We're happy to help!
+**Example: New or Updated Issue (Non-Admin)**
```
dannyroosevelt, line 10: The `(Admin)` vs `(Non-Admin)` terminology is a bit confusing to me, as an uninformed user, especially since we're talking about repo-specific access, and not user-specific.

components/github/README.md
```diff
@@ -1,257 +1,12 @@
...
+The user is not an administator of the repo, and has an option to the cadence by which to poll for new or updated issues.
```
dannyroosevelt, line 11:
```suggestion
If you do not have `admin` rights on the repo you're watching, you can configure the Pipedream trigger to poll for updates on a regular interval.
```

dannyroosevelt: Again, does not need this exact language, but I think your wording could be improved a little (and was maybe missing a word?).

dannyroosevelt (changes requested):
Added some comments, and we're chatting in Slack, but I don't think we want or need to delete all the older content (though it should probably be updated).

components/github/README.md
malexanderlim: Makes sense - dropped this.

components/github/README.md
malexanderlim: Dropped this part, and just explained with the two examples.

dannyroosevelt:
Since you're in there, can you update all of these content in that readme to make sure it's all accurate? I think a lot is stale.

malexanderlim:
> Since you're in there, can you update all of these content in that readme to make sure it's all accurate? I think a lot is stale.
Updated all stale content and included new links to all of our current triggers and actions in the original readme.

dannyroosevelt (approved):

PR #11166, opened 2024-03-26
SQL Proxy Methods and Types

jverce:
# Description
Add new lib to `@pipedream/platform` to document the methods that have to be implemented for a component to be compatible with the SQL proxy

### DIFF
--- a/platform/lib/sql-proxy.ts
+++ b/platform/lib/sql-proxy.ts
@@ -1,0 +1,51 @@
+import { ConfigurationError } from "@pipedream/platform";
+
+export type ClientConfiguration = object;
+export type ProxyArgs = {
+  query: string;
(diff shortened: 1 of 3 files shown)


platform/lib/sql-proxy.ts
```diff
@@ -0,0 +1,51 @@
+import { ConfigurationError } from "@pipedream/platform";
```
andrewjschuang, line 1: Import locally instead of from package

```suggestion
import { ConfigurationError } from "./errors";
```

andrewjschuang (changes requested):
One more thing, need to run `npm run build` and commit the `.js` file

andrewjschuang (approved):

PR #11206, opened 2024-03-28
Monday - Update Column Values action - Support Files type column

### LINKED ISSUE #11072: Monday.com: Update Column Values action - Support Files type column
## App
Monday.com

## Requirements
- Improve Monday.com - Update Column Values to support File type. The user can specify the path to temp dir, the action will upload the file using the [API Doc](https://developer.monday.com/api-reference/reference/assets-1#add-file-to-the-file-column)
image
image

## Context
Is there a way to upload a document to a monday.com column?

When using the existing "Update Column Values" step, if I select a "files" column from monday.com that allows files to be stored, then an error is returned. Updating columns with text works fine, but not updating a file.

michelle0927:
Resolves #11072


lcaresia (dismissed):
LGTM!

lcaresia (approved):
LGTM!

PR #11210, opened 2024-03-28
Figma Team ID auth support

### LINKED ISSUE #11193: [FEATURE] Add support for team_id custom field for Figma
`Team ID` is used in many of our components. I've created a new custom field, `team_id` so that the user can input this upon account connection, and it would be easier to use than having to provide Team ID manually for each action. This request is to use `team_id` from the custom field when available.

Please use additionalProps to use the `team_id` from auth if provided, otherwise render Team ID as a required prop.

GTFalcao:
Closes #11193


luancazarine (approved):

PR #11219, opened 2024-03-29
New Components - loopify

### LINKED ISSUE #11183: [Components] loopify
# loopify
## URLs
- https://api.loopify.com/docs
#### Prompt
Emit new event when a Connect block occurs in Flow. Required props: Connect block information.
#### Prompt
Emit new event when a Webhook occurs in Flow. Required props: Webhook source and target.
#### Prompt
Adds a contact to an 'API Entry' block in a loopify flow. Requires a 'contact' prop for the contact details and an 'API Entry' prop for the block to which it should be added.
#### Prompt
Adds a contact to a 'New Entry' block in a loopify flow. Requires 'contact' prop for the contact details and 'New Entry' prop for the block to which it should be added.
#### Prompt
Creates or updates a contact in loopify. Requires a 'contact' prop with contact details. If the contact exists, it will be updated, otherwise, a new contact will be created.

jcortes:
Resolves #11183.

### DIFF
--- a/components/loopify/actions/create-update-contact/create-update-contact.mjs
+++ b/components/loopify/actions/create-update-contact/create-update-contact.mjs
@@ -132,7 +132,7 @@ export default {
         },
       });
 
-      $.export("$summary", `Successfully updated contact with ID \`${response._id}\``);
+      $.export("$summary", `Successfully updated contact with ID \`${contact._id}\``);
       return response;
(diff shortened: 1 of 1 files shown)


michelle0927 (approved):
LGTM!

michelle0927 (approved):

PR #11376, opened 2024-04-09
New Components - diffbot

### LINKED ISSUE #10795: [Components] diffbot
# diffbot
#### Prompt
Finds and returns detailed information about an organization using its name or URL. The user needs to configure 'name' or 'url' props.
#### Prompt
Searches for a person using their name and returns collected information. The user must set the 'name' prop.

lcaresia:
Resolves #10795.


components/diffbot/actions/enhance-entity/enhance-entity.mjs
```diff
@@ -0,0 +1,51 @@
...
+  },
+  async run({ $ }) {
+    const response = await this.app.enhanceEntity({
+      $,
+      data: {
+        type: this.type,
+        name: this.name,
+        location: this.location,
+        url: this.url,
+      },
```
GTFalcao, line 44: When the prop names match the parameter names, we can simplify this greatly:
```suggestion
    const { app, ...data } = this;
    const response = await app.enhanceEntity({
      $,
      data,
```

components/diffbot/diffbot.app.mjs
```diff
@@ -1,11 +1,84 @@
...
+      return this._makeRequest({
+        path: "/analyse",
+        ...args,
+      });
+    },
+    async listAccounts(args = {}) {
+      return this._makeRequest({
+        path: "/accounts",
+        ...args,
+      });
```
GTFalcao, line 81: Are these two methods being used? Looks like they aren't

GTFalcao (commented):
I left a suggestion and there's some unused code to be removed, other than that, moving it to QA!

GTFalcao (approved):

PR #11396, opened 2024-04-10
New Components - rocketchat

### LINKED ISSUE #11387: [Components] rocketchat
# rocketchat
#### Prompt
Emit new event when a new channel is created in RocketChat. No props are required.
#### Prompt
Emit new event when a new message is posted to a specific public channel. Required props: target channel. 
#### Prompt
Emit new event when a new user is created in RocketChat. No props are required.
#### Prompt
Creates a new room based on the props: room name (required), members (optional, list of usernames), and type (optional, public or private). Make sure the room's name is unique.
#### Prompt
Sends a new direct message to a specific user using the recipient's username (required prop) and a text (required prop). Make sure the username is correctly spelled.
#### Prompt
Updates the user's status using two props: status text and status type (both required). The status type can be online, away, busy or offline. Be aware of the character limit on status text.

michelle0927:
Resolves #11387.


luancazarine (approved):

PR #11603, opened 2024-04-21
Reusing Gitlab components for Developer App

### LINKED ISSUE #11547: [FEATURE] Share components between GitLab --> Gitlab (Developer App)
We've added a new GitLab (Developer App) that enables users who are using self-hosted GitLab to be able to connect using their own application. 

Can the components that are currently available within GitLab be shared with the new GitLab (Developer App)?

GTFalcao:
Closes #11547 

Components are importing directly from their Gitlab counterparts, with all propDefinitions mapped to point to `gitlab_developer_app` instead of the original.


luancazarine (dismissed):

luancazarine (approved):

PR #11652, opened 2024-04-26
Google Drive Triggers text/usability improvements

GTFalcao:
- Adjusted several text descriptions;
- Removed the 'Watch for Properties Changes' prop from several sources, except 'New or Modified Files';
- Added a 'Minimum Interval' prop to the sources that trigger on file modifications. This allows the user to set a minimum interval for events to be emitted for the same file, defaulting to 1 minute, and can be disabled using 0.


luancazarine (dismissed):

GTFalcao:
@vunguyenhung , as I understand it, these sources are also triggered for new files/folders, or in other situations that do not necessarily have 'changes' to be emitted - do you think we should emit an empty 'changes' object to maintain the schema in this case?

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.mjs
```diff
@@ -8,10 +14,10 @@ import googleDrive from "../../google_drive.app.mjs";
  */
 export default {
   key: "google_drive-add-file-sharing-preference",
-  name: "Add File Sharing Preference",
+  name: "Share File",
   description:
-    "Add a [sharing](https://support.google.com/drive/answer/7166529) permission to the sharing preferences of a file or folder and provide a sharing URL. [See the docs](https://developers.google.com/drive/api/v3/reference/permissions/create) for more information",
-  version: "0.1.4",
+    "Add a [sharing permission](https://support.google.com/drive/answer/7166529) to the sharing preferences of a file or folder and provide a sharing URL. [See the documentation](https://developers.google.com/drive/api/v3/reference/permissions/create)",
+  version: "0.1.{{ts}}",
```
luancazarine, line 20: please set the version.

luancazarine (changes requested):
Hi @GTFalcao, I just added a minor change request.

components/google_drive/actions/add-file-sharing-preference/add-file-sharing-preference.mjs
GTFalcao: Thanks for catching that! Fixed

PR #11678, opened 2024-04-26
Gmail - Update to New Labeled Email Source

### LINKED ISSUE #11394: [Improvement] Gmail - New Labeled Email does not emit event for auto-labeled email
### App
Gmail (Developer App)

### Summary:
Gmail label trigger doesn't always work

### Details:
If the user manually applies a label to my Gmail inbox, then the email gets picked up when polled. However, if the user usess the functionality of the Gmail app to apply the label automatically as it lands in my inbox (via a filter), there's no event emitted. 

I think this is an important user requirement, as the whole point is automation (and not having to manually apply a label).
I can think of a few work arounds, but it would be useful to have the trigger fixed to accommodate this if possible? Thank you :)

### Screenshots:
No screenshots included

michelle0927:
Resolves #11394

### DIFF
--- a/components/gmail_custom_oauth/package.json
+++ b/components/gmail_custom_oauth/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/gmail_custom_oauth",
-  "version": "0.1.6",
+  "version": "0.1.7",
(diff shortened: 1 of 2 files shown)


luancazarine (approved):

PR #11698, opened 2024-04-29
Google postmaster improvement

GTFalcao:
Adding an info box and an optional prop to control the "and/or" behavior of the Google Postmaster Tools API source.

Default behavior is being changed to 'or' instead of 'and', that is, the source will emit events if any of the configured props match (as opposed to requiring all of them).


luancazarine (approved):

PR #11711, opened 2024-04-30
Google Sheet Sources/Triggers Improvements

michelle0927:
- Added test events to all sources
- Move constants.mjs to /common folder
- Updated `new-row-added` sources to store row hash in db to check against instead of setting it as meta id for deduping (so that new rows aren't skipped if the id > 64 characters)
- Updated `new-updates` sources to shorten the meta id so that items aren't skipped if the id > 64 characters.
- Combine shared drive & non-shared drive sources
- New Source - "New Comment"


lcaresia (approved):
LGTM!

Issue #11743, opened 2024-05-01
[Components] perplexity


pipedream-component-development:
# perplexity
## URLs
- https://docs.perplexity.ai/

GTFalcao:
@vunguyenhung if possible, please check [the comment on the PR](https://github.com/PipedreamHQ/pipedream/pull/11808#pullrequestreview-2037116350) regarding an usability concern/suggestion for this component

PR #11808, opened 2024-05-02
[Components] perplexity #11743

lcaresia:

### DIFF
--- a/components/perplexity/actions/chat-completions/chat-completions.mjs
+++ b/components/perplexity/actions/chat-completions/chat-completions.mjs
@@ -1,0 +1,48 @@
+import app from "../../perplexity.app.mjs";
+
+export default {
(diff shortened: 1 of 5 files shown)


GTFalcao (commented):
Shouldn't the user be able to pass more than one message? The way this is coded only allows for one.

Perhaps this could be done in a similar fashion as Loyverse's Update Inventory Levels (`components\loyverse\actions\update-inventory-levels\update-inventory-levels.mjs`), which takes an array of strings and generates a JSON-stringified object each time the user fills these props, for them to paste into the array.

In this case, `content` and `role` would be the props used to build each object, and a `messages` prop would be an array of strings that is actually sent in the request.

Since this is a usability / QA concern, I'll move it forward.

GTFalcao (approved):

Issue #11789, opened 2024-05-01
[Components] zip_archive_api


pipedream-component-development:
# zip_archive_api
## URLs
- https://archiveapi.com/rest-api/file-compression/

GTFalcao:
@vunguyenhung @malexanderlim I've commented on a usability concern on this issue's PR - it seems this API supports file uploading (via form-data), if that's the case, our components should probably support that as well.

malexanderlim:
Thanks @GTFalcao . @lcaresia , could you implement to accept files both via File URL / using Form Data / application/octet-stream as well, to maintain maximum flexibility here?

PR #11827, opened 2024-05-06
[Components] zip_archive_api #11789

lcaresia:

### DIFF
--- a/components/zip_archive_api/actions/compress-files/compress-files.mjs
+++ b/components/zip_archive_api/actions/compress-files/compress-files.mjs
@@ -1,0 +1,51 @@
+import app from "../../zip_archive_api.app.mjs";
+
+export default {
+  key: "zip_archive_api-compress-files",
+  name: "Compress Files",
+  description: "Compress files provided through URLs into a zip folder. [See the documentation](https://archiveapi.com/rest-api/file-compression/)",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    app,
+    archiveName: {
+      propDefinition: [
+        app,
+        "archiveName",
+      ],
+    },
+    compressionLevel: {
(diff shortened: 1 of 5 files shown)


components/zip_archive_api/actions/compress-files/compress-files.mjs
```diff
@@ -0,0 +1,51 @@
...
+  },
+  async run({ $ }) {
+    const response = await this.app.compressFiles({
+      $,
+      data: {
+        archiveName: this.archiveName,
+        compressionLevel: this.compressionLevel,
+        password: this.password,
+        files: this.files,
+      },
```
GTFalcao, line 44:
```suggestion
    const { app, ...data } = this;
    const response = await app.compressFiles({
      $,
      data,
```

components/zip_archive_api/actions/extract-files/extract-files.mjs
```diff
@@ -0,0 +1,37 @@
...
+    },
+    password: {
+      propDefinition: [
+        app,
+        "password",
+      ],
+    },
+  },
+  async run({ $ }) {
+    const response = await this.app.extractFiles({
```
GTFalcao, line 25: Can be simplified as with the previous one

GTFalcao (commented):
Seems like both components here should support file uploading (`/tmp` folder) as well as the implemented method of passing URLs.

The API docs show support for uploading with form-data.

luancazarine (dismissed):
Hi @lcaresia, thanks for your contribution! LGTM!

lcaresia:
@vunguyenhung I didn't find a way to make `extract-files` action work, so I removed it.

luancazarine (approved):

PR #11828, opened 2024-05-06
[BUG] Snowflake Query Results trigger does not emit on prop configuration

### LINKED ISSUE #11811: [BUG] Snowflake Query Results trigger does not emit on prop configuration
**Describe the bug**
[Snowflake - Query Results trigger](https://github.com/PipedreamHQ/pipedream/tree/master/components/snowflake/sources/query-results)

When the `Emit Individual` prop is set to `false`, the trigger logs show the warning:
> Pipedream detected that component was still running code. Make sure you await all promises.

Not sure what’s the exact issue, but I suspect it's when [it emits the rowStream object](https://github.com/PipedreamHQ/pipedream/blob/master/components/snowflake/sources/common.mjs#L27).

Sample SQL Query:
```
select count(*) as workflow_run_count from collection where date_trunc(hour, execution_timestamp) >= DATEADD(HOUR,-3,sysdate());
```

May be related to the specific SQL query above, other queries should be tested.

**Additional context**
Reported by a user.

jcortes:
## WHY

Resolves #11811

### DIFF
--- a/components/snowflake/package.json
+++ b/components/snowflake/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/snowflake",
-  "version": "0.9.6",
+  "version": "0.9.7",
   "description": "Pipedream Snowflake Components",
   "main": "snowflake.app.mjs",
   "keywords": [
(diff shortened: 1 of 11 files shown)


luancazarine (approved):
Hi @jcortes, thanks for your contribution! LGTM!

PR #11895, opened 2024-05-09
docs: add tap info for brew installation

joscha:
adds brew tap instructions.
brew tap is idempotent, so adding it to the instructions is a no-op for people that have it tapped already but for new users:

## WHY

Without tapping it first you get the following error:
```
Warning: 'pipedreamhq/pd-cli/pipedream' formula is unreadable: No available formula with the name "pipedreamhq/pd-cli/pipedream".
Please tap it and then try again: brew tap pipedreamhq/pd-cli
```


luancazarine (approved):
LGTM

PR #11904, opened 2024-05-09
Updating Perplexity to latest models.

malexanderlim:
## WHY

Perplexity is deprecating old models in favor of new Llama-3 models.


sergio-eliot-rodriguez (approved):
LGTM!

PR #11958, opened 2024-05-16
[Components] insighto_ai

### LINKED ISSUE #11937: [Components] insighto_ai
# insighto_ai
## URLs
- https://api.insighto.ai/docs#
#### Prompt
Emit new event when a new form is captured. Required props are form_id. Optional props could be user_id to fetch form by specific user.
#### Prompt
Emit new event when a new conversation is created. Conversation_id could be a required prop. Optional props could be user_id to initiate the conversation with a specific user.
#### Prompt
Emit new event when a new contact is created. Required props include contact_id. An optional prop could be user_id, to create contact under a specific user.
#### Prompt
Adds a text blob into an existing data source. Required prop: 'source-id' to specify the target data source; 'text-blob', a string value representing the text to be added.
#### Prompt
Creates a new contact within the system. Required props: 'contact-info', containing details like name, email, phone. Optional prop: 'notes' for any additional information about the contact.
#### Prompt
Executes an outbound call to a specified phone number. Required prop: 'phone-number' where the call is intended. Optional prop: 'call-script' for a predefined conversation script.

jcortes:
## WHY

Resolves #11937

### DIFF
--- a/components/insighto_ai/actions/add-text-blob/add-text-blob.mjs
+++ b/components/insighto_ai/actions/add-text-blob/add-text-blob.mjs
@@ -1,0 +1,88 @@
+import app from "../../insighto_ai.app.mjs";
+import utils from "../../common/utils.mjs";
+
+export default {
+  key: "insighto_ai-add-text-blob",
+  name: "Add Text Blob",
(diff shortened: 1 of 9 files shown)


GTFalcao (approved):
LGTM!

PR #12121, opened 2024-05-24
SQL Prop Type Improvements

jverce:
# Changelog
* Split DB schema info into 2 separate parts (i.e. `metadata` and `schema`)
* Define the types for each of the new parts of the DB schema
* Export all the types for the SQL prop so that they can be consumed by downstream dependencies
* Bump `@pipedream/platform` major version


andrewjschuang (approved):
lgtm!

PR #11837, opened 2024-05-06
Google Drive actions improvements

GTFalcao:
Several text, consistency and code improvements on Google Drive actions

PR #12151, opened 2024-05-29
Google Drive Source adjustment

### LINKED ISSUE #12131: [FEATURE] Refactor Google Drive - New or Modified Comment
[Google Drive - New or Modified Comment](https://github.com/PipedreamHQ/pipedream/blob/master/components/google_drive/sources/new-or-modified-comments/new-or-modified-comments.mjs) trigger is causing issues related to Google Drive quota, as it is making far too many calls to grab changes for all files then list comments.

This should be refactored to have Folder and File as required props, so that the user _must_ select a File to watch. If not, if a folder has thousands of files, we will run into the same problem which can cause the user to exceed their Google per-user-per-minute / day quota, and cascade into Pipedream hitting the GCP project limit quota as well.

GTFalcao:
Closes #12131 

~~Note: bumping package to `0.8.1` as #11837 should be merged first and will bump it to `0.8.0`~~


michelle0927 (approved):
LGTM!

PR #12184, opened 2024-05-31
Slack updates to send-message action

michelle0927:
- Have Channel Type just have human readable values (User / Direct Message, Group, Private Channel, Public Channel)
- For some reason, channel ID prop isn't reloading for me when I select different channel types.


GTFalcao (dismissed):
LGTM!

Not sure if this needs to be QA'd or can be merged straight away, moving it to QA as is standard

Issue #11938, opened 2024-05-13
[Components] kodagpt


pipedream-component-development:
# kodagpt
## URLs
- https://kodagpt.readme.io/reference/iniciando-com-api-kodagpt
#### Prompt
Queries the KodaGPT with a specifically formulated question. Required props: 'question', a string that defines the question to ask. Optional props: 'model', a string defining which version of the AI model to use for fetching the answer.
#### URLs
- https://kodagpt.readme.io/reference/iniciando-com-api-kodagpt
- https://kodagpt.readme.io/reference

PR #12369, opened 2024-06-10
[Components] kodagpt #11938

lcaresia:
## Summary by CodeRabbit

- **New Features**
  - Introduced semantic search functionality for the Kodagpt chatbot, allowing users to perform semantic searches within chatbot data.

- **Improvements**
  - Added new prop definitions and methods for interacting with the Kodagpt API to support the new semantic search feature.


components/kodagpt/kodagpt.app.mjs
```diff
@@ -1,11 +1,43 @@
...
+          "Content-Type": "application/json",
+          "API_KEY": `${this.$auth.api_key}`,
+        },
+      });
+    },
+    async semanticSearch({
+      question, ...args
+    }) {
+      return this._makeRequest({
+        path: `/embed/${this.$auth.chatbot_id}/${question}`,
```
GTFalcao, line 38: Shouldn't `question` be URL-encoded here?

GTFalcao (commented):
I'll move this forward to QA and I suggest testing with special characters in the question, as I believe it should be URL-encoded since it's being used directly in the path

luancazarine (approved):
Hi @lcaresia, thanks for your contribution! LGTM!

luancazarine (approved):

PR #12457, opened 2024-06-17
Zoho Survey - Update per feedback from their team

### LINKED ISSUE #12438: [FEATURE] Update Zoho Survey per feedback from their team
Our development team has requested make following changes on the "New survey response" trigger. 
1. An extra field -  "Others" is displayed for every survey event. This is not a survey field. Please refer to the attached screenshot (Select_event_trigger.png). 
2. By default for all API responses we send the responses in an encoded HTML entity format. Please decode the HTML entities in all API responses such as portal list, survey list, department list, collector list, distribution list. 
Sample responses "&amp;"  should be converted to  "&", "&apos;" should be converted to ' , &#39; should be converted to '. 

3. Please process the Events in the "New survey response" trigger based on the "key" using the "Get survey fields/variables" API. As per the current implementation, when a survey field is modified in the survey, the flow is getting failed. The label should be used for user readability and the other logic should be implemented using the survey field "key". Please note that "key" is the unique identifier for a survey field. We will send "key" in the webhook responses and in the "Get survey fields/variables" API .  The survey fields label in the "Get survey field/variables" API may contain duplicate values and it can be edited by the end user.

michelle0927:
Resolves #12438 

## Summary by CodeRabbit

- **New Features**
  - Enhanced survey response processing to decode HTML entities in labels, improving the readability of survey results.
  
- **Bug Fixes**
  - Improved handling of response IDs for better accuracy in tracking survey responses.

- **Chores**
  - Updated versions for components and added a new dependency (`html-entities-decoder`) for enhanced label processing.


luancazarine (dismissed):

luancazarine (approved):

PR #12519, opened 2024-06-21
Google Ads - Create Report improvement

GTFalcao:
- Added the option to search for and select specific objects to generate a report for, instead of always querying all objects,
- Added option to filter by date range (predefined and custom).

## Summary by CodeRabbit

- **New Features**
  - Added object filtering and date range selection in report generation for Google Ads.
  - Enhanced resource listing with dynamic SQL query generation.

- **Bug Fixes**
  - Adjusted data structures by removing outdated segments (e.g., day_of_week, month) from various modules.

- **Improvements**
  - Streamlined data processing logic by removing redundant date-related segments.
  - Updated event emission for new Google Ads campaigns with a new version.


lcaresia (dismissed):

GTFalcao:
@vunguyenhung I cannot reproduce that error, but I believe it's related to the previous default value ("date") remaining in the optional prop after updating the component (the new default value is "segments.date"). If you reload the additionalProps and/or re-select that prop's value (or simply configure the action from scratch), that error won't occur.

Note: that default value is not present in the live version, so this wouldn't occur to users updating their components after releasing.

Btw, your setup gives me a business rule error where "metrics cannot be requested for a manager account", although this is a dynamic business rule on Google's end and not an error in the component (you'll have to either select a managed account, or not request metrics).

lcaresia (approved):

PR #12636, opened 2024-06-28
New Components - spiritme

### LINKED ISSUE #12614: [Components] spiritme
# spiritme
## URLs
- https://api.spiritme.tech/api/swagger/
#### Prompt
Emits a new event when an avatar video completes rendering. Required props: avatar ID. Optional props: system status notifications (success, failure, insufficient balance).
#### Prompt
Generates a new video using specific voice and avatar props. The voice prop must describe the characteristics of the voice to be used, such as pitch, speed, and accent. The avatar prop should provide identifiers for the avatar to be used. In case the avatar prop is not provided, a default avatar is used. The optional props could include video length, script, and desired animation.

michelle0927:
Resolves #12614.


components/spiritme/actions/generate-video/generate-video.mjs
```diff
@@ -0,0 +1,160 @@
...
+            ],
+          },
+        ],
+        auto_emotions_markup: autoEmotionsMarkup,
+      },
+    });
+
+    if (waitForCompletion) {
+      const timer = (ms) => new Promise((res) => setTimeout(res, ms));
+      while (response.status !== "success") {
```
GTFalcao, line 149: Any chance we should set a limit on retries here? Or is that handled by a default component execution timeout?

GTFalcao (commented):
Probably good to go, I left one comment which may not need any action

components/spiritme/actions/generate-video/generate-video.mjs
michelle0927: It does fail if the workflow times out, so I think it's ok.

GTFalcao (approved):

PR #12814, opened 2024-07-11
Types - increase package version

michelle0927:
## Summary by CodeRabbit

- **New Features**
  - Enhanced deduplication options by adding `"last"` to the existing `"greatest"` and `"unique"` options.
  - Improved error handling with a new method that detects missing `id` properties.

- **Bug Fixes**
  - Updated version from "0.3.0" to "0.3.1" for better compatibility and minor improvements.


GTFalcao (approved):
LGTM!

types/src/index.ts
```diff
@@ -349,34 +349,21 @@ interface BaseSource<
...
-
 export type DedupedSource<
   Methods,
   SourcePropDefinitions
 > = Modify<BaseSource<
   Methods,
   SourcePropDefinitions
 >, {
-  dedupe: "greatest" | "unique";
+  dedupe: "last" | "greatest" | "unique";
```
joscha, line 359: @michelle0927 aren't the semantics different here, assuming that a deduping strategy is always given? The default is `last`, which is the same as not specifying any, which is a valid source definition. The changes here means that with these types there is no default anymore and a strategy has to be always defined. That's not how the actual code works though, I believe, as it would be a backwards-incompatible change to the runtime.

PR #12854, opened 2024-07-16
Types - allow sources to omit dedupe property

js07:
## WHY

The only supported `Source` type was `DedupedSource`, which required the `dedupe` property and for the emit metadata to include an `id`. Since [`dedupe` and `id` are optional](https://pipedream.com/docs/components/api#component-structure),  the `Source` type should all include sources without the `dedupe` property.

For example, the following should pass TypeScript type checks:
```ts
import { defineSource } from "@pipedream/types";

export default defineSource({
  key: "source",
  name: "Test Source",
  description: "hello, world",
  version: "0.0.1",
  type: "source",
  async run() {
    this.$emit({
      foo: "bar ",
    }, {
      name: "channel",
      summary: "Summary",
      ts: 123,
    });
  },
});
```

## Summary by CodeRabbit

- **New Features**
  - Introduced support for non-deduplicated sources, expanding source definitions without a deduplication property.

- **Tests**
  - Added tests for non-deduplicated source functionality to ensure proper implementation.

### DIFF
--- a/types/package-lock.json
+++ b/types/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@pipedream/types",
-  "version": "0.3.0",
+  "version": "0.3.2",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@pipedream/types",
-      "version": "0.3.0",
+      "version": "0.3.2",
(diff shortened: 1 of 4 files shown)


js07:
Closed in favor of #12921

PR #12859, opened 2024-07-16
Adding End Workflow action.

malexanderlim:
## Summary by CodeRabbit

- **New Features**
  - Introduced an "End Workflow" action to terminate workflow execution, allowing users to specify a reason for ending the workflow.
- **Improvements**
  - Added a new property to define a reason for ending workflow execution.
- **Chores**
  - Updated the package version from 0.0.1 to 0.0.2.


components/end/actions/end-workflow/end-workflow.mjs
```diff
@@ -0,0 +1,18 @@
+export default ({
+  name: "End Workflow",
+  description: "Terminate the workflow execution.",
+  version: "0.0.1",
+  type: "action",
+  key: "end_workflow",
+  props: {
```
dannyroosevelt, line 7: missing the app prop; I believe we won't show the app's icon without it

components/end/actions/end-workflow/end-workflow.mjs
```diff
@@ -0,0 +1,18 @@
+export default ({
+  name: "End Workflow",
+  description: "Terminate the workflow execution.",
```
dannyroosevelt, line 3: should this have a `.` at the end? it's not a complete sentence.

components/end/actions/end-workflow/end-workflow.mjs
```diff
@@ -0,0 +1,18 @@
...
+  version: "0.0.1",
+  type: "action",
+  key: "end_workflow",
+  props: {
+    reason: {
+      type: "string",
+      label: "Reason",
+      description: "An optional reason why the workflow execution was ended.",
```
dannyroosevelt, line 11: same feedback -- not a complete sentence.

components/end/actions/end-workflow/end-workflow.mjs
```diff
@@ -0,0 +1,18 @@
+export default ({
+  name: "End Workflow",
+  description: "Terminate the workflow execution.",
+  version: "0.0.1",
+  type: "action",
+  key: "end_workflow",
```
dannyroosevelt, line 6: the component key should include the app slug and should use `-` as a delimiter instead of `_`.

dannyroosevelt (changes requested):

components/end/end.app.mjs
```diff
@@ -1,11 +1,13 @@
...
-  propDefinitions: {},
-  methods: {
-    // this.$auth contains connected account data
-    authKeys() {
-      console.log(Object.keys(this.$auth));
+  propDefinitions: {
+    reason: {
+      type: "string",
+      label: "Reason",
+      description: "A reason why the workflow execution was ended",
```
dannyroosevelt, line 8: nitpicking, but this wording is awkward to. does it need to be in passive voice? maybe something like, "Reason for terminating / exiting / ending workflow execution", for example

components/end/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/end",
-  "version": "0.0.1",
+  "version": "0.0.2",
```
dannyroosevelt, line 3: shouldn't it be 0.0.1?

dannyroosevelt (approved):

PR #12903, opened 2024-07-17
John/pipedream sdk

celador:


packages/sdk/examples/next-app/app/page.tsx
```diff
@@ -0,0 +1,108 @@
...
+            </p>
+            <p>
+              <button type="submit" onClick={signIn} style={{ all: "revert" }}>Sign in</button>
+            </p>
+          </form>
+
+      }
+      {
+        initialized && <div>
+          {token && <p>Token: {token}</p>}
```
bizyb, line 86: This is a bit awkward. A better UX would be to render `Token: <message>` to indicate that it's missing. The message could be something informative. The alternative just had me waiting for a while, not knowing whether i should click on the connect button.

packages/sdk/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  externalId: string;
+  clientName?: string
+};
+
+// type AccountId = string;
+// type AccountKeyFields = {
+//   externalId: string;
+//   app: string;
+// };
+// type AccountKey = AccountId | AccountKeyFields;
```
bizyb, line 18: Will we need this?

bizyb (commented):
Just some minor suggestions so far. Another improvement could be caching the state in localStroge so you don't lose your state whenever you fresh the page, which i did often.

PR #13015, opened 2024-07-26
Share components from Gmail (Developer App) to Gmail app

### LINKED ISSUE #13013: [FEATURE] Share components from Gmail (Developer App) to Gmail app
In preparation of applying for Gmail Restricted Scopes, we will need all triggers and actions from the Gmail (Developer App) shared with the primary Gmail app.

Please note that you will encounter a scopes error when testing, please ping @malexanderlim and I will temporarily add the necessary scopes to the Gmail app for testing.

michelle0927:
Resolves #13013 

## Summary by CodeRabbit

- **New Features**
	- Introduced functionalities to add labels to emails, create draft emails, find emails, and update email signatures in Gmail.
	- Launched a new source component to emit events for new labeled emails in Gmail, enhancing event tracking for users.
	- Implemented polling and event-handling modules for new email and attachment events, improving real-time email processing and user interaction.

- **Enhancements**
	- Improved documentation links for sending emails for better user guidance.
	- Added structured metadata for emitted events related to labeled and sent emails.

- **Version Update**
	- Updated version numbers for various modules, indicating improvements and adjustments.


components/gmail/actions/add-label-to-email/add-label-to-email.mjs
```diff
@@ -0,0 +1,34 @@
...
+    },
+  },
+  async run({ $ }) {
+    const response = await this.gmail.addLabelToEmail({
+      message: this.message,
+      label: this.label.value,
+    });
+    $.export("$summary", `Successfully added ${this.label.label} label to email`);
+    return response;
+  },
```
GTFalcao, line 33: Can at least the `run` method be imported directly from the other app's original action, to avoid duplicating the code?

GTFalcao (commented):
@michelle0927 I'm assuming the intention is to maintain both apps with the same components (besides the additional auth check). If so, you could also reuse the props the same way `gitlab_developer_app` did, with a function that imports the propDefinitions from the original app and converts them to use the new app.

Example of this in `components\gitlab_developer_app\actions\create-branch\create-branch.mjs`

michelle0927:
> @michelle0927 I'm assuming the intention is to maintain both apps with the same components (besides the additional auth check). If so, you could also reuse the props the same way `gitlab_developer_app` did, with a function that imports the propDefinitions from the original app and converts them to use the new app.
> 
> Example of this in `components\gitlab_developer_app\actions\create-branch\create-branch.mjs`
> 
> Would be helpful so that any changes made to one app would automatically be imported by the other, without having to apply the change twice (can be easily overlooked when maintaining these components in the future)

@GTFalcao I hadn't updated the Gmail Developer App components because they're supposedly being deprecated. But in the mean time, I agree that there's no point in duplicating code.

lcaresia (approved):

michelle0927:
@vunguyenhung I added some checks to help avoid the error with "New Labeled Email", but I haven't been able to reproduce it myself. If you encounter it again, could you dropdown on the Error message for your screenshot? Sometimes that helps to figure out where the error is coming from.

lcaresia (approved):

PR #13050, opened 2024-07-31
fix(types): allow Date

joscha:
This allows `Date` to be exported. It automatically gets serialized into it's string value.

### DIFF
--- a/types/src/index.ts
+++ b/types/src/index.ts
@@ -5,6 +5,7 @@ export type JSONValue =
   | number
   | boolean
   | null
+  | Date
   | JSONValue[]
   | { [key: string]: JSONValue; };
 


luancazarine (approved):
Hi @joscha, LGTM! Ready for QA!

joscha:
> Hey @joscha @luancazarine @GTFalcao, may I ask how can I test this?

If we create a new object in the types tests that looks like this:

```ts
const x: JSONValue = {
  myDate: new Date()
}
```

It should pass the type check. Do you have runtime tests for this part as well? Somewhere inside your closed source? Then you could add the same structure in the unit test where your serializer is tested.

GTFalcao (commented):
Hi @joscha , thanks for your contribution!

I've verified all other types where `JSONValue` is used, and I don't see any negative impact that could be caused by the addition of `Date`.

Note: this needs a version increase before it can be merged and published.

joscha:
> I've verified all other types where `JSONValue` is used, and I don't see any negative impact that could be caused by the addition of `Date`.

Great! Should I add some rudimentary tests for this?


> Note: this needs a version increase before it can be merged and published.

Is that something the releaser does or do you want me to bump the version in this pull request?

GTFalcao:
@joscha the version increase should be part of the pull request, yes.

Can you provide some more context on the usefulness of being able to export/return a Date object directly, instead of converting it to a ISO string first? Is that just for the shortcut, or is there a situation where this would be essential?

The reason I ask is, since Date is not a universally serializable JSON value (like a string or number), it may not make sense to be added as a JSONValue.

joscha:
> or is there a situation where this would be essential?

Generally I suppose I could always transform from a Date into an ISO string. I did notice that currently Pipedream does serialize the  `Date` type, but the type definitions suggest it doesn't, so this PR is mainly bringing the type definitions up to par with what the Pipedream implementation actually does. An alternative would be to throw an error in the serialization logic of Pipedream (when using `$.export` or `return`), however that would be a backwards-incompatible change, thus I assumed it would make more sense to update the types.

malexanderlim:
Hi @joscha, thanks for identifying this behavior with Date objects in our system. I'm going to close this PR for now, as we would like to keep our options open for a more robust date/time implementation in the future. Keeping the types more conservative now will give us more flexibility for a proper native date handling in the future.

Thanks for your contribution and we'd be happy to revisit this down the line.

PR #13186, opened 2024-08-07
13133 GitHub adjustments

### LINKED ISSUE #13133: [Refactor] Github components
To be able to support the migration of Github Projects (legacy) over to Github Projects v2, we need to update two sources and create a new action to support Projects v2.

**Source: [New Issue with Status (Projects V2)]**(https://github.com/PipedreamHQ/pipedream/blob/master/components/github/sources/new-issue-with-status/new-issue-with-status.mjs)
- The "Item Status" config should be multi-select, and allow nothing to be selected (in which case all statuses get emitted
- The item and issue are retrieved, which is good, but then the issue is hoisted to the top as the entire emitted object.  Due to this, we lose a lot of potentially important context e.g.:
`action` - what type of change caused this event
`changes` - what data fields actually changed
`sender` - what user made this change
other context like node ID's and such that users might want to use
- Assuming there's no data security or size issues, instead of hoisting issue, I believe it would be better to just add item and issue to the original payload, and emit that instead.

**Source: [New or Updated Issue from GitHub]**(https://github.com/PipedreamHQ/pipedream/blob/master/components/github/sources/new-or-updated-issue/new-or-updated-issue.mjs)
- In this case issue, sender, and action are hoisted, but we still lose other context from the original payload.
e.g. for a labeled action, there is a label field that gives the user context of which label was added.
The hoisting of issue has a bug where the object is placed under key [object Object].
There are probably other fields for other action types that are being omitted as well.
Why not just emit the original payload?

**New Action**: Update Projects V2 Item Status

```
import { graphql } from "@octokit/graphql";

  props: {
    github: {
      type: "app",
      app: "github",
    }
  },
  async run({steps, $}) {
    const projectId = "PVT_*********"
    const itemId = "PVTI_********"
    const fieldId = "PVTSSF_********" // status field
    const blockedStateOptionId = "**********" // status id
    const graphqlWithAuth = graphql.defaults({
      headers: {
        authorization: `token ${this.github.$auth.oauth_access_token}`,
      },
    })
    return await graphqlWithAuth({
      query: `mutation moveItemToColumn(\$projectId: ID!, \$itemId: ID!, \$fieldId: ID!, \$value: ProjectV2FieldValue!) {
        updateProjectV2ItemFieldValue(input: {projectId: \$projectId, itemId: \$itemId, fieldId: \$fieldId, value: \$value}){
          projectV2Item {
            id
          }
        }
      }`,
      projectId,
      itemId,
      fieldId,
      value: { singleSelectOptionId: blockedStateOptionId },
    })
  },
})
```

GTFalcao:
Closes #13133 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new action for updating the status of an item within GitHub Projects (V2).
	- Added GraphQL mutations for updating project items.

- **Improvements**
	- Enhanced event handling and clarity in the `new-issue-with-status` module, including better item summary construction and streamlined historical event loading.
	- Modified the `$emit` function to simplify event emissions in the common module.

- **Version Updates**
	- Incremented version numbers for various modules, including a significant update for `new-issue-with-status` from 0.0.17 to 0.1.0, indicating potential bug fixes or enhancements.
	- Minor version updates for multiple other GitHub components, reflecting incremental improvements or bug fixes.


luancazarine (dismissed):

luancazarine (approved):

PR #13715, opened 2024-08-27
Pandadoc new actions

### LINKED ISSUE #7639: [ACTION] PandaDoc: Download Document
**Download Document**
Required prop of document ID

Optional prop
- watermark_text string
- watermark_color string
- watermark_font_size int
- watermark_opacity float
- separate_files bool

The API Call returns a PDF file - which could just be written to the `/tmp` directory.

The return value would be the path to the file.

A nice feature to add would be an optional prop to allow removing the "Certificate" page from the document (the last page of the PDF). There is a use case in just wanting the original filled document without the certificate. If the user toggled the prop to remove the certificate page it just needs to delete the _last_ page of the PDF.
___
Would also be neat to have a **[Download Protected Document](https://developers.pandadoc.com/reference/download-protected-document)** action which would work the same (slightly different API call) have the same required document ID prop and optional separate_files prop.

https://developers.pandadoc.com/reference/download-document

GTFalcao:
Closes #7639 

I have not fully validated these actions as I am waiting for a response regarding free trial extension on my PandaDoc account.

## Summary by CodeRabbit

- **New Features**
	- Introduced functionality for downloading documents as PDFs, including options for watermark customization.
	- Added capability to download protected documents with enhanced retrieval options.
	- New properties allow users to specify completed document IDs and manage file outputs, including separate files and custom filenames.

- **Bug Fixes**
	- Improved document management and retrieval processes for better user experience.

- **Chores**
	- Updated package version to reflect new features and improvements.
	- Minor updates to action descriptions for clarity and professionalism.


michelle0927 (changes requested):
Looks good, left one comment. Also, you need to update the component versions since the app file was updated.

components/pandadoc/pandadoc.app.mjs
```diff
@@ -23,6 +24,26 @@ export default {
...
+      type: "string",
+      label: "Completed Document Id",
+      description: "Select a completed document or provide its ID",
+      async options() {
+        const response = await this.listDocuments({
+          params: {
+            deleted: false,
+            status: DOCUMENT_STATUS_COMPLETED,
+          },
+        });
```
michelle0927, line 37: Should include pagination:
```suggestion
      async options({ page }) {
        const response = await this.listDocuments({
          params: {
            deleted: false,
            status: DOCUMENT_STATUS_COMPLETED,
            page: page + 1,
          },
        });
```

michelle0927 (changes requested):

components/pandadoc/pandadoc.app.mjs
GTFalcao: Thanks, good catch - I've also added it to all other async options props in this app as an improvement

michelle0927 (dismissed):
LGTM!

michelle0927 (approved):

PR #13767, opened 2024-08-29
New Components - crunchbase

### LINKED ISSUE #13188: [Components] crunchbase
# crunchbase
#### Prompt
Emit new event when a user creates a list. Props required: user 
#### Prompt
Emit new event when a user creates a saved search. Props required: user. Optional prop: search details
#### Prompt
Emit new event when a company is added to a list. Required props: company id, list id

luancazarine:
Resolves #13188.

### DIFF
--- a/components/crunchbase/actions/search-organizations/search-organizations.mjs
+++ b/components/crunchbase/actions/search-organizations/search-organizations.mjs
@@ -1,0 +1,40 @@
+import crunchbase from "../../crunchbase.app.mjs";
+import { axios } from "@pipedream/platform";
+
+export default {
+  key: "crunchbase-search-organizations",
+  name: "Search Organizations",
+  description: "Search for organizations based on specified criteria. [See the documentation](https://data.crunchbase.com/reference/post_data-searches-organizations)",
+  version: "0.0.{{ts}}",
+  type: "action",
+  props: {
+    crunchbase,
+    fieldIds: {
+      propDefinition: [
+        crunchbase,
+        "fieldIds",
+      ],
+    },
+    order: {
+      propDefinition: [
+        crunchbase,
+        "order",
+      ],
+    },
+    query: {
+      propDefinition: [
+        crunchbase,
+        "query",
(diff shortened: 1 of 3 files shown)


components/crunchbase/crunchbase.app.mjs
```diff
@@ -1,11 +1,68 @@
...
+        url: this._baseUrl() + path,
+        headers: this._headers(),
+        ...opts,
+      });
+    },
+    getOrganizationDetails({ entityId }) {
+      return this._makeRequest({
+        path: `/entities/organizations/${entityId}`,
+      });
+    },
```
michelle0927, line 59: Need to allow for the `$` variable.
```suggestion
    getOrganizationDetails({ entityId, ...opts }) {
      return this._makeRequest({
        path: `/entities/organizations/${entityId}`,
        ...opts,
      });
    },
```

components/crunchbase/actions/search-organizations/search-organizations.mjs
```diff
@@ -0,0 +1,37 @@
...
+      label: "Field IDs",
+      description: "Fields to include on the resulting entity",
+      options: FIELDS_OPTIONS,
+    },
+    query: {
+      type: "string[]",
+      label: "Query",
+      description: "Array of stringified objects for searching organizations",
+      optional: true,
+    },
```
michelle0927, line 24: I think the `query` prop might be confusing for some users. What if instead of the `query` prop, there were `fieldId`, `operatorId`, `type`, and `values` props. There would only be one query object allowed in the request, but we could create it for the user based on those other props. What do you think?

michelle0927 (changes requested):

components/crunchbase/actions/search-organizations/search-organizations.mjs
luancazarine: I think only one object is not good. I'll add an example and a link on the description. I think it's better.

PR #13868, opened 2024-09-06
Gmail usability pt.2

### LINKED ISSUE #5663: [ACTION] Enhancement: Support send from email for Gmail action

Gmail (Send message)

It would be great if it was possible to set the `from` address when sending messages with the Gmail app.

### LINKED ISSUE #13860: [FEATURE] Gmail usability pt.2
## Implement New Trigger Received (webhook and polling) as a single trigger.

New Prop: Trigger Type: webhook | polling
1. If user selects webhook, show the alert prop (or would this be better as a description?) ("Configuring this trigger as a **webhook** requires users to set up a Custom OAuth client for Gmail, which is available on Pipedream's Advanced plan or higher. Please see the instructions [here](https://pipedream.com/apps/gmail/triggers/new-email-received-instant).")
2. If the user has selected an account that is using the Pipedream client, throw a configurationError with ("You must use a custom OAuth client to use this component. Please see [here](https://pipedream.com/docs/connected-accounts/oauth-clients) for more details.") that blocks them from further configuration. The current state allows them to continue to configure when it will never succeed.
3. Instead of having the Service Account Key JSON in $.auth, given that this is only required for this one component, we should just make it a prop with the following description:
```
1) [Create a service account in GCP](https://cloud.google.com/iam/docs/creating-managing-service-accounts) and set the following permission: **Pub/Sub Admin**

2) [Generate a service account key](https://cloud.google.com/iam/docs/creating-managing-service-account-keys)

3) Download the key details in JSON format

4) Open the JSON in a text editor, and **copy and paste its contents here**.
If user selects polling, have the polling trigger set up.
```

4. Update path to the README once the webhook/polling trigger has been collapsed into one.

## New prop "From" for sending mail
Related to https://github.com/PipedreamHQ/pipedream/issues/5663

It looks like there is an endpoint to List Verified Gmail Aliases; doing a little research I found this:
https://developers.google.com/gmail/api/reference/rest/v1/users.settings.sendAs/list

Could we add a new optional prop for From as a new optional prop for the relevant actions?

## Attachment prop
This could be easier to use if we made two type: "string" props, filename and filepath

michelle0927:
Resolves #13860
Resolves #5663 

## Summary by CodeRabbit

- **New Features**
  - Enhanced email draft creation with explicit attachment handling.
  - New properties for email sending, including `fromEmail` and refined attachment options.
  - Major update to the "New Email Received" component, allowing configuration for polling or webhook modes.
  - Improved instructions and flexibility in the "New Email Received" setup guide.

- **Bug Fixes**
  - Improved validation checks for email components to prevent misconfigurations.

- **Version Updates**
  - Incremented version numbers for multiple components, indicating potential enhancements or bug fixes.


luancazarine (approved):

luancazarine (approved):

PR #13869, opened 2024-09-06
normalize url -- draft

dannyroosevelt:

### DIFF
--- a/components/puppeteer/actions/common/common.mjs
+++ b/components/puppeteer/actions/common/common.mjs
@@ -1,0 +1,19 @@
+export default {
+  props: {
+    url: {
+      type: "string",
+      label: "URL",
+      description:
+        "The URL of the page to scrape. For example, `https://example.com`.",
+    },
+  },
+  methods: {
+    async run({ $ }) {
+      let url = this.url;
+      if (!url.startsWith("http")) {
+        url = `http://${url}`;
+      }
+      return url;
+    },
+  },
+};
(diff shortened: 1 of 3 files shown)


components/puppeteer/actions/get-html/get-html.mjs
```diff
@@ -1,23 +1,21 @@
...
-      label: "URL",
-      description: "The URL of the page to scrape.",
-    },
+    ...common.props,
   },
   async run({ $ }) {
     const browser = await this.puppeteer.launch();
     const page = await browser.newPage();
-    await page.goto(this.url);
+    await page.goto(common.props.url);
```
dannyroosevelt, line 18: I think this is not right but I'm not sure how to properly reference it

components/puppeteer/actions/common/common.mjs
```diff
@@ -0,0 +1,19 @@
...
+  props: {
+    url: {
+      type: "string",
+      label: "URL",
+      description:
+        "The URL of the page to scrape. For example, `https://example.com`.",
+    },
+  },
+  methods: {
+    async run({ $ }) {
```
GTFalcao, line 11: I think this can be a synchronous method and named something else to avoid confusion, also $ is not used at all so it doesn't have to be passed in
```suggestion
    getFilteredUrl() {
```

components/puppeteer/actions/get-html/get-html.mjs
```diff
@@ -1,28 +1,27 @@
 import puppeteer from "../../puppeteer.app.mjs";
+import common from "../common/common.mjs";
 
 export default {
```
GTFalcao, line 4: All components could import the common method and call it as their own instead of having to use bind, e.g.
```suggestion
export default {
  ...common,
```
or
```suggestion
export default {
  methods: common.methods,
```
and then, when running, calling the method as
`const url = this.getFilteredUrl();`

GTFalcao (approved):
Looks good to me!

PR #13889, opened 2024-09-10
[ACTION] Improve Notion actions to loop block-creation, comply with request limits

### LINKED ISSUE #13857: [ACTION] Improve Notion actions to loop block-creation, comply with request limits
I'd like the Notion actions that can create blocks to execute a loop that will allow for an arbitrary number of blocks to be created, beyond the limits of a single API request.

The relevant limits are:

- A request can contain only 1,000 total blocks
- A request that allows for a children array can handle nesting to 2 levels (e.g. you can do a Toggle block, with Paragraph children, which each have a Rich Text Object array). I'm not sure if the Rich Text Object array on a block counts as a level, so I'm not sure if you could do the above example at the same time you create a page (where the Toggles would be in the children array of the page).
- Any array can have up to 100 elements
- A Rich Text Object can have up to 2,000 characters

Notion - specifically these actions:

- Create Page (will need to create the page, then use the Append Block Children endpoint for additional blocks)
- Create Page in Database (will need to create the page, then use the Append Block Children endpoint for additional blocks)
- Append Blocks

https://developers.notion.com/reference/request-limits

jcortes:
… blocks each

## WHY

Added `chunkArray` function to get chunks of 1000 blocks each

Resolves #13857

## Summary by CodeRabbit

- **New Features**
	- Introduced a method for processing large arrays in manageable chunks, improving performance when appending blocks.
	- Updated user feedback to reflect the total number of blocks appended after processing.

- **Chores**
	- Updated the version of the `@pipedream/notion` package to 0.1.20.

### DIFF
--- a/components/notion/actions/append-block/append-block.mjs
+++ b/components/notion/actions/append-block/append-block.mjs
@@ -117,6 +117,6 @@ export default {
     const totalAppended = results.reduce((sum, res) => sum + res.length, 0);
 
     $.export("$summary", `Appended ${totalAppended} block(s) successfully`);
-    return results;
+    return results.flat();
   },
 };


michelle0927 (approved):
LGTM!

michelle0927 (approved):

PR #13949, opened 2024-09-13
Remove docs v3 redirects

andrewjschuang:
## Summary by CodeRabbit

- **New Features**
	- Streamlined routing configuration by removing outdated paths and redirects related to version 3 of the documentation.

- **Bug Fixes**
	- Removed deprecated rewrite and redirect rules to enhance clarity in documentation navigation.


docs-v2/vercel.json
```diff
@@ -1,14 +1,4 @@
 {
-  "rewrites": [
```
dylburger, line 2: You'll want to move these to the `redirects` section instead of removing them entirely, since we still have some `/docs/v3` links out there, so we want to redirect `/docs/v3` to `/docs`

docs-v2/vercel.json
```diff
@@ -226,10 +216,6 @@
       "source": "/docs/components/quickstart/nodejs/sources",
       "destination": "/docs/components/sources-quickstart"
     },
-    {
```
dylburger, line 229: These are fine to remove once you have the redirect to handle the root /v3 path

docs-v2/vercel.json
```diff
@@ -289,10 +275,6 @@
     {
       "source": "/docs/errors/",
       "destination": "/docs/workflows/errors"
-    },
```
dylburger, line 292: Also fine to remove

dylburger (approved):

PR #13983, opened 2024-09-18
Abstract OAuth implementation in the SDK

jverce:
# Changelog
* Install `simple-oauth2` to handle OAuth client requests
* Rename some methods and types to reuse them
* Abstract the HTTP requests so that it can be used for Connect and for normal API requests
* Implement logic to fetch OAuth client credentials
* Bump minor version of the SDK

## WHY

So that we can make API requests using the OAuth credentials.

## Summary by CodeRabbit

- **New Features**
	- Introduced support for OAuth 2.0 authentication in the SDK.
	- Enhanced `ServerClient` functionality with new methods for handling API requests securely.
- **Improvements**
	- Updated SDK version to 0.1.0, reflecting new features and enhancements.
	- Added new dependencies to support OAuth 2.0 functionality, including TypeScript definitions.

### DIFF
--- a/packages/sdk/src/server/index.ts
+++ b/packages/sdk/src/server/index.ts
@@ -30,12 +30,12 @@ export type CreateServerClientOpts = {
   /**
    * The client ID of your workspace's OAuth application.
    */
-  oauthClientId: string;
+  oauthClientId?: string;
 
   /**
    * The client secret of your workspace's OAuth application.
    */
-  oauthClientSecret: string;
+  oauthClientSecret?: string;
 
   /**
    * The API host URL. Used by Pipedream employees. Defaults to "api.pipedream.com" if not provided.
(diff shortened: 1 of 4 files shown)


packages/sdk/src/server/index.ts
```diff
@@ -22,6 +27,16 @@ export type CreateServerClientOpts = {
    */
   secretKey: string;
 
+  /**
+   * The client ID of your workspace's OAuth application.
+   */
+  oauthClientId: string;
```
dylburger, line 33: We should make these optional for now, and handle that optionality in the rest of the code, since project keys alone can authorize Connect API requests and we don't yet want to force those customers to use OAuth clients. We may end up consolidating on OAuth for the Connect API, too, and just make people pass the dev / prod project ID (removing project secret entirely), but for now it's an either-or.

packages/sdk/src/server/index.ts
```diff
@@ -358,6 +394,52 @@ class ServerClient {
     return result;
   }
 
+  /**
+   * Makes a request to the Connect API.
```
dylburger, line 398: replace "Connect" with "Pipedream"

dylburger (approved):

PR #13992, opened 2024-09-18
Gmail - Improve New Email Received event shape

### LINKED ISSUE #13991: [TRIGGER] Improve / normalize Gmail New Email Received event shape
Currently, the event shape returned from Gmail for new email received is not easily usable in a workflow. Users will need to iterate through the array of objects, search for what they need, and I believe the specific location of each header is not consistent.

Can we normalize the output for some of the most common headers, like subject, from, reply-to, to, etc?

I'd also like to keep the raw event, then add another array for "parsed-headers" or something.

I'm not sure if this is also the case with the other Gmail sources, but let's make sure to apply consistent structure across all of them where relevant.

michelle0927:
Resolves #13991 

## Summary by CodeRabbit

- **New Features**
	- Updated the `@pipedream/gmail` package to version 0.1.6, which may include new features and improvements.
	- Enhanced email reception with a new `"reply-to"` field in the emitted email data.
	- Introduced a structured representation of email headers in the emitted message data.

- **Bug Fixes**
	- Improved handling of message details to prevent processing when no messages are available.
  
- **Refactor**
	- Adjusted event emission to handle emails individually for better responsiveness.
	- Modified timestamp handling for emitted events to ensure consistency and accuracy.


luancazarine (dismissed):

dannyroosevelt (approved):

PR #14045, opened 2024-09-20
Notion - updated-page source improvements

michelle0927:


components/notion/sources/updated-page/updated-page.mjs
```diff
@@ -114,8 +129,11 @@ export default {
         continue;
       }
 
-      const meta = this.generateMeta(page, constants.summaries.PAGE_UPDATED);
-      this.$emit(page, meta);
+      this.emitEvent(page);
+
+      if (Date.parse(page?.last_edited_time) < lastCheckedTimestamp) {
```
andrewjschuang, line 134: Does it make sense to add back the HTTP request param to filter out pages where `last_edited_time < lastCheckedTimestamp`:

```
{
  filter: {
    timestamp: "last_edited_time",
    last_edited_time: {
      on_or_after: new Date(lastCheckedTimestamp).toISOString()
    }
  }
}
```

michelle0927: @andrewjschuang Ah, I didn't notice there's an `on-or-after` filter in addition to `after`. I added the filter back in, but I'm still hesitant to be using `last_edited_time` (either for filtering or for exitng the loop) since that was the reason the source was skipping events to begin with. The `last_edited_time` is not necessarily updated when a page is updated. I guess in this case it's a choice between possibly skipping events and having to loop through a large amount of pages.

andrewjschuang (approved):
Nice improvements!

components/notion/sources/updated-page/updated-page.mjs
andrewjschuang: Thanks, based on my tests the Notion API was reliably updating the `last_edited_time`. The issues the user was seeing were most likely due to performance issues due to a huge number of pages in the database. I tested with a fairly large database, and with the improvements from this PR, I did not have any issues

andrewjschuang (approved):

andrewjschuang:
Summarizing our conversation at Slack:
- Issues were caused because the Notion `last_updated_time` timestamp is truncated to the minute, e.g. `2024-09-23T18:12:00.000Z` -- any other updates made within the same minute wouldn't be caught by Pipedream, since we dedupe on the timestamp.
- Previously, the HTTP request params included an `after` filter. But with the issue above, if an update was made in the same minute, the page wouldn't be returned because the filter is not inclusive.

Solution:
- not dedupe by timestamp
- use `on_or_after` filter to fetch pages that were updated on the same minute or after.

PR #14054, opened 2024-09-22
Small bug fixes, refactors in workflow invocation based on Danny's tests

dylburger:
## Summary by CodeRabbit

- **New Features**
	- Streamlined API client configuration by making `publicKey` and `secretKey` optional.
	- Enhanced request handling to support various body types, improving flexibility in API requests.
	- Introduced a mock implementation for the `simple-oauth2` library for easier testing.
	- Added Jest configuration and setup files to facilitate testing.
	- Comprehensive unit tests for the `ServerClient` class to ensure expected functionality.

- **Bug Fixes**
	- Improved error handling in request processing, allowing for better response parsing.

- **Documentation**
	- Updated comments for clarity and consistency across methods.


packages/sdk/src/server/index.ts
```diff
@@ -13,19 +13,14 @@ import {
  */
 export type CreateServerClientOpts = {
   /**
-   * The environment in which the server is running (e.g., "production", "development").
+   * The public API key for accessing the service.
    */
-  environment?: string;
```
dylburger, line 18: This is a weird diff but I was removing `environment`

packages/sdk/src/server/__tests__/server.test.ts
```diff
@@ -0,0 +1,646 @@
...
+          },
+        },
+      );
+
+      client = new ServerClient({
+        publicKey: "test-public-key",
+        secretKey: "test-secret-key",
+      });
+
+      const result = await client["makeRequest"]("/test-path", {
```
jverce, line 44: The `makeRequest` method does not exist, right? There's the `_makeRequest` method, which is private so I'd suggest not testing it directly and just focus on the public interface of `ServerClient`.

Also, why do we have to do `client["makeRequest"]` instead of `client.makeRequest`? Is it because of mocking? Or because the method is private?

packages/sdk/src/server/index.ts
```diff
@@ -404,64 +450,86 @@ class ServerClient {
...
+      "Content-Type": "application/json",
+      ...opts.headers,
     };
-    return this._makeRequest<T>(path, {
+
+    if (authType === "oauth") {
+      headers["Authorization"] = await this.oauthAuthorizationHeader();
+    } else {
+      headers["Authorization"] = this.connectAuthorizationHeader();
+    }
```
jverce, line 497: I'm now more convinced that we should split these clients into `ConnectClient` and `ApiClient`, WDYT?
We could also rename `ServerClient` (too verbose IMHO) to just `Client`, make it abstract, and implement the HTTP calling logic only.

```typescript
abstract class Client {
  private baseUrl: string;

  constructor(baseUrl: string) {
    this.baseUrl = baseUrl;
  }

  makeRequest(
    method: "GET" | "POST",
    path: string,
    headers: object,
    payload?: unknown,
  ) {
    // Do all the stuff that it does now
  }
}

class ConnectClient extends Client {
  private publicKey: string;
  private secretKey: string;

  constructor(publicKey: string, secretKey: string) {
    super("https://api.pipedream.com/v1/connect");
    this.publicKey = publicKey;
    this.privateKey = privateKey;
  }

  private makeRequest(
    method: "GET" | "POST",
    path: string,
    headers: object,
    payload?: unknown,
  ) {
    headers["Authorization"] = this.authorizationHeader();
    return super(method, path, headers, payload);
  }
}
```

jverce: Not now of course.

packages/sdk/src/server/__tests__/server.test.ts
jverce: I'd rather focus the tests on the public interface instead.

packages/sdk/src/server/index.ts
dylburger: 100%, maybe after we ship all the JWT validation in catcher and Rails we can work on this next sprint

PR #14152, opened 2024-10-01
[MS Azure SQL] request timeout

andrewjschuang:


andrewjschuang:
We should probably make `requestTimeout` configurable. cc @malexanderlim

PR #14235, opened 2024-10-08
New Components - bubble

### LINKED ISSUE #13187: [Components] bubble
# bubble
#### Prompt
Emits new event when a Bubble workflow that incorporates the plugin action is initiated. Necessary prop is a specified Bubble workflow. This allows for immediate integration with other platforms.
#### Prompt
Modifies an existing thing in the Bubble database by providing the Thing's ID. The user needs to configure the following prop: 'Thing ID' (required). Optionally, 'Attribute to Modify' and 'New Value' props can be added to specify what and how to change.
#### Prompt
Searches for a Thing in the Bubble database. This operation requires the 'Search Query' prop configured by the user. Add 'Search Parameters' as an optional prop to narrow down the search.
#### Prompt
Creates a new Thing in the Bubble database. The action requires 'Thing Attributes' prop configured by the user which defines the properties of the new Thing.

luancazarine:
Resolves #13187.

### DIFF
--- a/components/bubble/actions/create-thing/create-thing.mjs
+++ b/components/bubble/actions/create-thing/create-thing.mjs
@@ -1,0 +1,30 @@
+import bubble from "../../bubble.app.mjs";
+import { axios } from "@pipedream/platform";
+
+export default {
+  key: "bubble-create-thing",
+  name: "Create Thing",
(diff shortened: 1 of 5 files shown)


GTFalcao:
Closed as outdated and issue is blocked - feel free to reopen if there is new context

PR #14332, opened 2024-10-16
Gmail New Email Received - add error handling

michelle0927:
Adds error handling while retrieving messages, so messages that are deleted or not found will be skipped over instead of throwing an error.


GTFalcao (approved):
LGTM!

PR #14344, opened 2024-10-17
Adding new prop for requestTimeout for Azure SQL

malexanderlim:
See example workflow (Test Request Timeout [step](https://pipedream.com/@pd-component-dev/projects/proj_W7seoE9/p_aNCV2nZ/build)) for testing the prop.

## WHY

Per the feedback [here](https://github.com/PipedreamHQ/pipedream/pull/14152#issuecomment-2386623256), we should create a requestTimeout prop to make the query timeout window configurable. It used to default to 30s, then we bumped to 60s, but if a user needs to configure this to be longer than 60s, they should be able to for maximum flexibility.


malexanderlim:
@vunguyenhung , I've added the IP address to the allowlist. Let me know if this works now for you!

PR #14461, opened 2024-10-29
Slack - Send Block Kit Message

### LINKED ISSUE #14459: [ACTION] Slack - Send Block Kit Message
We've had a few support inquiries and users are running into discoverability issues when we condensed the Slack message functionality into a few actions. 

Our search does not cover component description, so even if we had it in the Slack - Send Message (Advanced) description, users would not be able to search and find it.

It makes sense to break out "Send Block Kit Message" as it's own distinct action here given the number of questions we've gotten since we've condensed the actions.

michelle0927:
Resolves #14459 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new module for sending customizable Block Kit messages to Slack, enhancing integration capabilities.
	- Added properties for conversation selection and optional notification text.

- **Version Updates**
	- Updated the Slack component version from 0.6.7 to 0.7.0.

### DIFF
--- a/components/slack/actions/send-block-kit-message/send-block-kit-message.mjs
+++ b/components/slack/actions/send-block-kit-message/send-block-kit-message.mjs
@@ -1,0 +1,43 @@
+import buildBlocks from "../common/build-blocks.mjs";
+import common from "../common/send-message.mjs";
+
+export default {
+  ...common,
+  ...buildBlocks,
+  key: "slack-send-block-kit-message",
+  name: "Build and Send a Block Kit Message",
+  description: "Configure custom blocks and send to a channel, group, or user. [See the documentation](https://api.slack.com/tools/block-kit-builder).",
+  version: "0.4.0",
+  type: "action",
+  props: {
+    slack: common.props.slack,
+    conversation: {
+      propDefinition: [
+        common.props.slack,
+        "conversation",
+      ],
+    },
+    text: {
+      type: "string",
+      label: "Notification Text",
+      description: "Optionally provide a string for Slack to display as the new message notification (if you do not provide this, notification will be blank).",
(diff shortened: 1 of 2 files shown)


components/slack/actions/send-block-kit-message/send-block-kit-message.mjs
```diff
@@ -0,0 +1,43 @@
+import buildBlocks from "../common/build-blocks.mjs";
+import common from "../common/send-message.mjs";
+
+export default {
+  ...common,
+  ...buildBlocks,
+  key: "slack-send-block-kit-message",
+  name: "Build and Send a Block Kit Message",
+  description: "Configure custom blocks and send to a channel, group, or user. [See the documentation](https://api.slack.com/tools/block-kit-builder).",
+  version: "0.4.0",
```
GTFalcao, line 10: Should probably be `0.0.1` if it's new

GTFalcao (commented):
Just left one minor comment regarding the version

components/slack/actions/send-block-kit-message/send-block-kit-message.mjs
michelle0927: It's an action that existed previously, but was removed. I've encountered issues with republishing components at version `0.0.1` if they were previously published at a higher version. Since this one was at `0.3.3` when it was removed, I decided to use `0.4.0`.

GTFalcao (approved):

components/slack/actions/send-block-kit-message/send-block-kit-message.mjs
GTFalcao: Gotcha!

PR #14679, opened 2024-11-18
Require project environment in SDK

bzwrk:
## Release Notes

- **New Features**
	- Introduced a new type for environment settings, ensuring only valid environments ("development" or "production") are accepted.
	- Updated client creation to require explicit environment specification.

- **Bug Fixes**
	- Enhanced error handling for OAuth token retrieval and improved messaging for failures.
	- Added validation for the workflow URL construction to handle invalid inputs.

- **Chores**
	- Updated version number to 1.0.5 in the SDK package.
	- Modified `.gitignore` to exclude specific directories from version control.
	- Removed `eslint` from development dependencies.


packages/sdk/src/server/index.ts
```diff
@@ -353,7 +353,10 @@ export class BackendClient {
    * @param opts - The options for configuring the server client.
    */
   constructor(opts: BackendClientOpts) {
-    this.environment = opts.environment ?? "production";
+    if (!opts.environment || !["development", "production"].includes(opts.environment)) {
```
dylburger, line 356: Small but I'd suggest moving this check to its own private method since we may want to validate this elsewhere in the SDK, and since the env validation logic may be more complex later

jverce (approved):

PR #14699, opened 2024-11-20
feat(sdk-package): call apis directly from client sdk

TheBestMoshe:
## WHY

Currently, the Pipedream Connect SDK only supports calling API endpoints from the server. This means that all client side calls need to be proxied by a trusted server.

This PR adds support for calling most API endpoints directly from the client.

## Release Notes

- **New Features**
	- Introduced a command-line interface (CLI) for interacting with the Pipedream Connect API, including commands for managing projects, accounts, and components.
	- Added `BrowserAsyncResponseManager` and `ServerAsyncResponseManager` classes for improved asynchronous response handling in browser and server contexts.
	- Updated the SDK to version 1.0.6, allowing the use of client Connect tokens for API calls directly from the client.

- **Improvements**
	- Enhanced token management in the `BrowserClient` with new methods for token retrieval and caching.
	- Updated configuration options for clients, allowing for more flexibility in API interactions.
	- Deprecated the `environments` property on the `createFrontendClient` method, streamlining client configuration.

- **Bug Fixes**
	- Improved error handling and response validation in test cases and client methods.

- **Documentation**
	- Updated type definitions and comments for better clarity on new functionalities and configurations.

### DIFF
--- a/packages/sdk/src/browser/index.ts
+++ b/packages/sdk/src/browser/index.ts
@@ -19,8 +19,7 @@ export type * from "../shared";
  */
 type CreateBrowserClientOpts = {
   /**
-   * The environment in which the browser client is running (e.g., "production",
-   * "development").
+   * @deprecated environment is set on the server when generating the client token
    */
   environment?: string;
 
@@ -119,7 +118,6 @@ type StartConnectOpts = {
  * @example
  * ```typescript
     const client = createFrontendClient({
-      environment: "development",
       tokenCallback,
       externalUserId,
     });
@@ -278,10 +276,6 @@ export class BrowserClient extends BaseClient {
     const qp = new URLSearchParams({
       token,
     });
-
-    if (this.environment) {
-      qp.set("environment", this.environment);
-    }
 
     if (typeof opts.app === "string") {
       qp.set("app", opts.app);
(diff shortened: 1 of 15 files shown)


packages/sdk/src/browser/index.ts
```diff
@@ -51,8 +83,10 @@ class ConnectError extends Error {}
 type StartConnectOpts = {
   /**
    * The token used for authenticating the connection.
+   *
+   * Optional if client already initialized with token
    */
-  token: string;
+  token?: string;
```
TheBestMoshe, line 89: Do we want to remove this and require the token be configured when the client is initialized?

This would be a breaking change.

packages/sdk/src/browser/index.ts
```diff
@@ -157,10 +247,11 @@ class BrowserClient {
     window.addEventListener("message", onMessage);
 
     try {
-      this.createIframe(opts);
+      await this.createIframe(opts);
     } catch (err) {
       opts.onError?.(err as ConnectError);
     }
+    this.refreshToken(); // token expires once it's used to create a connected account. We need to get a new token for the next requests.
```
TheBestMoshe, line 254: This hack of refreshing tokens isn't the best but it's needed based on the existing implementation. Ideally, we could stop deleting the tokens.

dylburger (approved):

PR #14831, opened 2024-12-04
Docusign - Download Documents

### LINKED ISSUE #14805: [Action] Docusign - Download Envelope Documents
This request is to add a new action called `Download Envelope Documents`.

The prop should enable the end-user to download envelope documents to /tmp by returning a documents object with a fileUrl path to /tmp. 
https://developers.docusign.com/docs/esign-rest-api/how-to/download-envelope-documents/

Download options:
- All Documents (PDF)
- All Documents (Zip)
- Certificate (PDF)
- Portfolio (PDF)

> The document ID specifies the document or documents that you want to download. You can use the document ID itself or one of the following keywords.
combined: Retrieves a PDF file that contains the combined content of all of the documents. If the account option Attach certification of completion to envelope is on, then the Certificate of Completion is also included in the PDF file. You set this account option in the Admin tool on the Signing Settings screen, or by setting the attachCompletedEnvelope property in the accountSettings object to true.
archive: Retrieves a ZIP archive that contains all of the PDF documents and the Certificate of Completion.
certificate: Retrieves the Certificate of Completion as a PDF file.
portfolio: Retrieves the envelope documents as a [PDF portfolio](https://helpx.adobe.com/acrobat/using/overview-pdf-portfolios.html).

Here is an example of how another competitive product handles this for reference - I think that the way we present this will offer additional functionality over this.
Screenshot 2024-12-02 at 11 58 39 AM

michelle0927:
Created new action "Download Documents" instead of adding a prop to New Change in Envelope Status because sources/triggers do not have access to the `/tmp` directory for a workflow.

Resolves #14805 

## Summary by CodeRabbit

- **New Features**
	- Introduced a module for downloading documents from DocuSign envelopes, allowing users to download documents in various formats.
	- Users can dynamically select envelope IDs and specify filenames for downloaded documents.

- **Bug Fixes**
	- Improved error handling during document downloads.

- **Chores**
	- Updated the version of the `@pipedream/docusign` package from 0.2.0 to 0.2.1.

### DIFF
--- a/components/docusign/actions/download-documents/download-documents.mjs
+++ b/components/docusign/actions/download-documents/download-documents.mjs
@@ -1,0 +1,110 @@
+import docusign from "../../docusign.app.mjs";
+import fs from "fs";
+
+export default {
+  key: "docusign-download-documents",
+  name: "Download Documents",
+  description: "Download the documents of an envelope to the /tmp directory. [See the documentation here](https://developers.docusign.com/docs/esign-rest-api/how-to/download-envelope-documents/)",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    docusign,
+    account: {
+      propDefinition: [
+        docusign,
+        "account",
+      ],
+    },
+    envelopeId: {
+      type: "string",
+      label: "Envelope ID",
+      description: "Identifier of the envelope to download documents from",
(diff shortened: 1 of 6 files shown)


luancazarine (approved):

PR #14958, opened 2024-12-12
Hubspot actions - support end user auth

### LINKED ISSUE #14913: [BUG] Hubspot actions can't be configured with end user auth
**Describe the bug**
Hubspot actions that make a Hubspot API call in `additionalProps` can't be configured when "Use end user auth via Connect" is enabled.

_Previously, this was not an issue since Pipedream-managed auth was the only auth type._

**To Reproduce**
1. Add the Hubspot **Search CRM** action to a workflow
2. Toggle "Use end user's auth via Connect"
3. Select "Contacts" in the dropdown for **Object Type**
4. See axios error and no additional fields for searching or creating an object

**Expected behavior**
I should be able to successfully configure the action when I have "Use end user auth via Connect" enabled:
1. The **Search Property**, **Search Value**, and **Additional properties to retrieve** props should be shown (without options) even if the Hubspot API call in `additionalProps` fails.
2. If **Create if not found?** is `true`, I should be able to enter an expression for a JSON object if properties can't be fetched. _This applies to all of the "Create *" and "Update *" Hubspot actions._

(This would be similar to the [fallback](https://github.com/PipedreamHQ/pipedream/blob/46e7530d240251261c8f64dd8e38b104e4b3c3a3/components/google_sheets/actions/add-single-row/add-single-row.mjs#L100) used in the Google Sheets Add Single Row action.)

**Screenshots**

Current behavior:
ac6dc9e

These props should be shown:
fc39843

michelle0927:
Resolves #14913 

## Release Notes

- **New Features**
  - Introduced `objectProperties` for structured input in various actions, enhancing object and engagement creation.
  - Added `updateIfExists` property to allow updates during contact creation errors.

- **Version Updates**
  - Incremented version numbers across multiple components for tracking changes.

- **Bug Fixes**
  - Improved error handling in `additionalProps` and `creationProps` methods for better robustness.

- **Refactor**
  - Updated import structure to utilize `appProp` for property management across multiple actions.


luancazarine (approved):

luancazarine (approved):

PR #14974, opened 2024-12-13
Zoho CRM - support picklist fields

### LINKED ISSUE #14939: [FEATURE] Investigate Zoho CRM filtering for picklist
We currently filter out `picklist` field types for Zoho CRM. 
https://github.com/PipedreamHQ/pipedream/blob/master/components/zoho_crm/actions/common/common-objects.mjs#L31C1-L32C1

A customer reached out as they're trying to use the **Update Object** action, but not all of the fields are coming through - specifically, `picklist` fields are missing.

We should investigate why this logic is in place, and if it make sense be able to pull in `picklist` field as well instead of filtering.

michelle0927:
Resolves #14939

### DIFF
--- a/components/zoho_crm/actions/common/common-objects.mjs
+++ b/components/zoho_crm/actions/common/common-objects.mjs
@@ -28,23 +28,33 @@ export default {
         && !field.system_mandatory
         && field.operation_type[`api_${type}`]
         && ![
-          "picklist",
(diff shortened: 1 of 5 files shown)


GTFalcao (dismissed):
LGTM!

GTFalcao (approved):

PR #14977, opened 2024-12-13
Hubspot - New Custom Object Property Change

### LINKED ISSUE #14945: HubSpot
### App
HubSpot
https://developers.hubspot.com/beta-docs/reference/api

### Sources
New or Updated Custom Object. This is helpful, but it would be more helpful if there's a way to select a specific HubSpot field within this action. For example, our use case is that we want to ONLY trigger the workflow if "Go Live Date" property has a value on our custom object (Site). Currently, the new or updated custom object trigger will enroll any record at all that has an update, but it would be great if there was a way we could specify further which specific property we want it to monitor for updates.

michelle0927:
Resolves #14945 

## Summary by CodeRabbit

- **New Features**
	- Introduced a new source component for HubSpot that emits events when a specified property of a custom object is created or updated.
	- Users can monitor custom object properties by specifying the `objectSchema` and the `property`.
  
- **Chores**
	- Updated the version of the `@pipedream/hubspot` package from `0.11.0` to `0.12.0`.

### DIFF
--- a/components/hubspot/package.json
+++ b/components/hubspot/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/hubspot",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "description": "Pipedream Hubspot Components",
   "main": "hubspot.app.mjs",
   "keywords": [
--- a/components/hubspot/sources/new-custom-object-property-change/new-custom-object-property-change.mjs
+++ b/components/hubspot/sources/new-custom-object-property-change/new-custom-object-property-change.mjs
@@ -20,7 +20,7 @@ export default {
     property: {
       type: "string",
       label: "Property",
-      description: "The company property to watch for changes",
+      description: "The custom object property to watch for changes",
       async options() {
         const properties = await this.getWriteOnlyProperties(this.objectSchema);
         return properties.map((property) => property.name);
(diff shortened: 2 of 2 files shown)


elvnsbmtr:
The agency we are working with just sent us the following in reference to the "New Or Updated Custom Object Property" trigger: "this is specifically for new site objects where the live date has changed. This isn't checking for existing site records. I also ran a test with a site to make sure it was specific to new objects versus existing"

Goal: they're attempting to build a trigger that will fire anytime a site_live property is updated--- regardless of whether it's a newly created custom object record or an existing record where site_live property is updating.

andrewjschuang:
The code is [filtering for the `hs_lastmodifieddate` property](https://github.com/PipedreamHQ/pipedream/blob/master/components/hubspot/sources/new-custom-object-property-change/new-custom-object-property-change.mjs#L76).
It should be emitting events regardless of a new object or existing one.

Are they [selecting the `site_live` property](https://github.com/PipedreamHQ/pipedream/blob/master/components/hubspot/sources/new-custom-object-property-change/new-custom-object-property-change.mjs#L72) correctly?

michelle0927:
@andrewjschuang @vunguyenhung I tested this with a few different properties, and it's working correctly for me. @elvnsbmtr Are there any errors being displayed in the source logs?

PR #14691, opened 2024-11-20
Airtable usability improvements

GTFalcao:
## Release Notes

- **New Features**
	- Introduced `warningAlert` to guide users on using custom expressions for `Base` and `Table` properties.
	- Added `customExpressionInfo` across various actions to provide instructions on custom expressions.
	- Enhanced `accountTierAlert` to inform users that views are available only for Airtable Enterprise accounts.
	- Updated `name` and `description` for the `New Modified or Deleted Records (Instant)` component to clarify its functionality.

- **Improvements**
	- Updated descriptions for clarity in multiple actions, including `create-comment`, `create-field`, and `delete-record`.
	- Revised version numbers for several components to reflect updates, including `0.4.0` for the package.
	- Improved error handling and clarity in the `get-record-or-create` action.

- **Removals**
	- Removed deprecated actions such as `create-single-record`, `update-record`, and various sources related to record changes.

PR #15015, opened 2024-12-17
Re-applying Airtable usability changes

GTFalcao:
Replaces #14691 

## Summary by CodeRabbit

- **New Features**
	- Added `warningAlert` property to provide guidance on using custom expressions.
	- Introduced `customExpressionInfo` property for enhanced instructions on custom expressions.
	- Updated descriptions and properties across various actions for clarity and improved user guidance.
	- New properties `changeTypes`, `infoBox1`, and `infoBox2` added to enhance event emission functionality.
	- Updated component names to reflect their functionality more accurately, such as "New Record Created, Updated or Deleted (Instant)".
	- Introduced `getDataTypes` method in multiple components for improved data handling.
	- Enhanced webhook functionality with properties for monitoring field schemas and including previous definitions.
	- Added `dedupe` property to several components to manage unique event emissions.

- **Bug Fixes**
	- Improved error handling and validation for record processing.

- **Version Updates**
	- Incremented version numbers across multiple components to reflect updates.

- **Documentation**
	- Enhanced descriptions and links to documentation for better user understanding.


GTFalcao:
This was already reviewed and QA'd on #14691 , but we needed a new branch due to merge conflicts.

Some sources were changed to webhook-based from polling-based, so I'll ask for another round of code review and QA only on the sources (the actions are unchanged from the previous PR)

GTFalcao:
I've reworked the event format so that the original payload is inside the "originalPayload" object, and the current field or record data (fetched in a separate request) in the "field" or "record" object, respectively. The update information (previous and new values) is also in another object. This should make the event structure more clear and easier to use.

I've also removed some props and turned them always on by default (e.g. always emit the previous value). And, finally, while I added event deduping which was previously missing, some webhook events will emit several separate changes, e.g. when changing the value of a field in the Airtable UI, it seems to be saved multiple times while typing the value. To minimize this, I've added a minimum threshold of 5 seconds before emitting more events for the same object (field or record).

GTFalcao:
- Fixed "New Field" that was falling back to the old event emission, now correctly emits the new formatted event;
- Fixed "New or Modified Field" syntax bug which I slipped while testing the other field source;
- Fixed "New Record Created, Updated or Deleted (Instant)" deleted events not being emitted properly. They are now being handled accordingly and have a separate summary ("x records deleted", or "record deleted: [id]").

For the other sources which were requested to be removed: I've discussed this with @malexanderlim , originally we were going to reduce the number of sources due to their functionality being superseded by ohers, but we decided to keep the previous separate sources for minor use cases. Therefore, sources such as "New, Modified or Deleted Records" (polling, non-webhook version) are thus being kept exactly as they were before, only with minor text improvements.

Let me know if this sounds good to you @vunguyenhung and let me know if the above sources are good to go now. Note that in some cases (e.g. deleted records) we can't fetch the info available in other event types, so the event body may vary. I've tried to strike the best balance between straightforwardness and completeness in the event formats, but if you feel any still need to be changed, let me know.

PR #15055, opened 2024-12-20
Postgres - remove ssl object when unnecessary

andrewjschuang:
## WHY

A user was facing connection issues when the SSL object was being passed in the PG client configuration even with just `rejectUnauthorized = false`.

## Summary by CodeRabbit

- **New Features**
	- Enhanced SSL configuration logic for PostgreSQL component.
- **Bug Fixes**
	- Improved error messaging for SSL verification issues.
- **Chores**
	- Updated package versions for various PostgreSQL actions and sources from 2.0.6 to 2.0.7 and from 3.0.6 to 3.0.7.

### DIFF
--- a/components/postgresql/postgresql.app.mjs
+++ b/components/postgresql/postgresql.app.mjs
@@ -93,8 +93,11 @@ export default {
         ...(cert && {
           cert,
         }),
-        rejectUnauthorized: mode !== "skip_verification",
       };
+
+      if (mode === "verify_identity") {
+        ssl.rejectUnauthorized = true;
+      }
 
       return Object.keys(ssl).length > 0
         ? ssl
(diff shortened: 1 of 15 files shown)


jverce (approved):

components/postgresql/postgresql.app.mjs
```diff
@@ -93,9 +93,12 @@ export default {
         ...(cert && {
           cert,
         }),
-        rejectUnauthorized: mode !== "skip_verification",
       };
 
+      if (mode === "verify_identity") {
+        ssl.rejectUnauthorized = true;
+      }
```
js07, line 100: I think this wouldn't allow setting `rejectUnauthorized` to `false`, which someone may want to do to connect without CA verification to a server that requires SSL. (It looks like the `pg` package [sets `rejectUnauthorized` to `true` by default](https://github.com/brianc/node-postgres/blob/373093d176f0ceb0266333680b20d0acf38004e3/docs/pages/announcements.mdx#pg80-release) for SSL connections.)

Maybe set `ssl.rejectUnauthorized` if `mode` is not null, and omit it otherwise? For example:
```js
      if (mode) {
        ssl.rejectUnauthorized = mode === "verify_identity";
      }
```

michelle0927:
Hi @andrewjschuang. Is this still waiting for review?

andrewjschuang:
@michelle0927 this got lost and I need to check the status for it, so feel free to skip for now. Thanks!

GTFalcao:
Closing as outdated and blocked

PR #15110, opened 2024-12-30
Update app files that use relative references to point to packages instead

### LINKED ISSUE #15080: Update app files that use relative references to point to packages instead
Can we update the Linear app file and any other app files that use relative references to instead point to the package, like [here](https://github.com/PipedreamHQ/pipedream/blob/e5390d0d5005c9b6e7b3633d30ef19552a09266f/components/linear/linear.app.mjs#L1)?
So we’d want to update

```import linearApp from "../linear_app/linear_app.app.mjs";
```
to

```import linearApp from "@pipedream/linear_app";
```

This should be tested to ensure that it works as expected still.

### App Slug list

1. [linear](https://github.com/PipedreamHQ/pipedream/blob/326638182ab9b5282408ed0cf5135ca820d55985/components/linear/linear.app.mjs#L4)
2. [monday_oauth](https://github.com/PipedreamHQ/pipedream/blob/326638182ab9b5282408ed0cf5135ca820d55985/components/monday_oauth/monday_oauth.app.mjs)
3. [canva_enterprise](https://github.com/PipedreamHQ/pipedream/blob/326638182ab9b5282408ed0cf5135ca820d55985/components/canva_enterprise/canva_enterprise.app.mjs)

jcortes:
## WHY
Resolves #15080

### DIFF
--- a/components/brex_staging/actions/create-card/create-card.mjs
+++ b/components/brex_staging/actions/create-card/create-card.mjs
@@ -1,4 +1,4 @@
-import common from "../../../brex/actions/create-card/common.mjs";
+import common from "@pipedream/brex/actions/create-card/common.mjs";
 import brexApp from "../../brex_staging.app.mjs";
 
 export default {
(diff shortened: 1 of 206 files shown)


michelle0927 (approved):
Thanks for tackling this! LGTM!

PR #15142, opened 2025-01-02
Biz/dj 2786 fix issue with loading dynamic props

bzwrk:
## WHY

Dynamic props aren't loading (this only happens if we need to reload props more than once, eg. Google sheets, where we fetch dynamicProps for the first time after selecting the sheet. But we also need to fetch the dynamic props after selecting a worksheet, and so on.) The root cause appears to be `useQuery`, which only executes `queryFn` once for a given queryKey. Ideally, useQuery should run once for a given `reloadPropIdx`. So far the only thing that works is invalidating the `queryKey`.

## Summary by CodeRabbit

- **Version Update**
	- Updated `@pipedream/connect-react` package version from `1.0.0-preview.15` to `1.0.0-preview.16`

- **Improvements**
	- Enhanced query key handling in form context hook to improve dynamic property management

### DIFF
--- a/packages/connect-react/src/hooks/form-context.tsx
+++ b/packages/connect-react/src/hooks/form-context.tsx
@@ -129,12 +129,13 @@ export const FormContextProvider = <T extends ConfigurableProps>({
     configuredProps,
     dynamicPropsId: dynamicProps?.id,
   };
+  const queryKey = reloadPropIdx ? `dynamicProps:${reloadPropIdx}` : "dynamicProps"
   const {
     isFetching: dynamicPropsQueryIsFetching,
     // TODO error
   } = useQuery({
     queryKey: [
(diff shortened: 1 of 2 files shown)


packages/connect-react/src/hooks/form-context.tsx
```diff
@@ -129,12 +129,17 @@ export const FormContextProvider = <T extends ConfigurableProps>({
...
+    ...componentReloadPropsInput,
+  }
+
   const {
     isFetching: dynamicPropsQueryIsFetching,
     // TODO error
   } = useQuery({
     queryKey: [
       "dynamicProps",
+      queryKeyInput
```
bzwrk, line 142:
```suggestion
      queryKeyInput,
```

PR #15438, opened 2025-01-28
Biz/dj 2837 show label and keep selected remote option dropdown enabled

bzwrk:
## Release Notes for @pipedream/connect-react v1.0.0-preview.25

- **New Features**
  - Now displays prop labels instead of values after selecting dynamic props

- **Bug Fixes**
  - Resolved issue with remote options not reloading correctly when form component is re-rendered
  - Improved handling of input options in select controls

- **Performance**
  - Streamlined option processing logic in select components
  - Enhanced state management for configured properties on component load
  - Added flexible handling for input values with specific properties

### DIFF
--- a/packages/connect-react/src/utils/component.ts
+++ b/packages/connect-react/src/utils/component.ts
@@ -40,7 +40,7 @@ export function valuesFromOptions<T>(value: unknown | T[] | PropOptions<T>): T[]
     }
     return results
   }
-  if (value && typeof value === "object" &&  "__lv" in value && Array.isArray(value.__lv)) {
+  if (value && typeof value === "object" && Array.isArray(value.__lv)) {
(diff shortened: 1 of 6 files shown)


packages/connect-react/src/utils/component.ts
```diff
@@ -40,6 +40,9 @@ export function valuesFromOptions<T>(value: unknown | T[] | PropOptions<T>): T[]
     }
     return results
   }
+  if (value && typeof value === "object" &&  "__lv" in value && Array.isArray(value.__lv)) {
```
jverce, line 43: **Nitpick**
I think this should be the same
```suggestion
  if (typeof value === "object" && Array.isArray(value.__lv)) {
```

jverce (approved):

PR #15457, opened 2025-01-29
Google drive: adding selection between file URL or path

GTFalcao:
Adjusting the behavior in two Google Drive actions

## Summary by CodeRabbit

- **New Features**
	- Introduced new flexible property handling for file uploads and updates in Google Drive.
	- Added functionality to upload files using either a URL or a local file path with improved visibility control.
	- Implemented a structured way to manage file upload preferences through the `updateType` constant.

- **Chores**
	- Incremented package and component versions:
		- Google Drive package: v0.8.9 → v0.8.10
		- Update file action: v0.1.7 → v1.0.0
		- Upload file action: v0.1.10 → v1.0.0


michelle0927 (dismissed):

GTFalcao:
The issue here was, even when the File URL prop was hidden, its value was still there - and File URL took precedence over File Path when uploading a file. So the file was actually being uploaded, but it was uploading from the hidden URL instead of the provided path.

Fixed so that the component only looks at either the URL or the path, depending on what the user selected (or neither, if they selected metadata only).

michelle0927 (approved):

PR #15506, opened 2025-02-06
Hubspot custom object request updates

GTFalcao:
Adhering to the [breaking change announced by HubSpot](https://developers.hubspot.com/changelog/breaking-change-removed-support-for-referencing-custom-object-types-by-base-name)

Custom objects are now referenced by their `fullyQualifiedName` instead of `name`.

Components affected by this change:
- Create Associations
- Create Custom Object
- Search CRM
- Update Custom Object
- New Note Created
- New Task Created

We also have 'New or Updated Custom Object' and 'New Custom Object Property Change', but those were not modified since they already use the custom object's ID instead of its name.

## Summary by CodeRabbit

- **Chores**
  - Upgraded the HubSpot integration to version 1.0.0 across all actions and sources.
- **Refactor**
  - Refined the display of custom object options to provide clearer labels and more consistent identifiers for an improved user experience.

### DIFF
--- a/components/hubspot/actions/add-contact-to-list/add-contact-to-list.mjs
+++ b/components/hubspot/actions/add-contact-to-list/add-contact-to-list.mjs
@@ -4,7 +4,7 @@ export default {
   key: "hubspot-add-contact-to-list",
   name: "Add Contact to List",
   description: "Adds a contact to a specific static list. [See the documentation](https://legacydocs.hubspot.com/docs/methods/lists/add_contact_to_list)",
-  version: "0.0.13",
+  version: "0.0.14",
   type: "action",
   props: {
     hubspot,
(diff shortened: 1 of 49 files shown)


lcaresia (approved):

GTFalcao:
This issue seems to already exist in the current live component, when selecting a custom object. Should be fixed now nonetheless.

GTFalcao:
@vunguyenhung I looked into the API response, and out of the 17 properties of this object, only 1 is a "searchable property": `pipedream_object` which is labeled as "Pipedream TEXT".

I changed the prop to show the property label instead of only the value (in this case, label is "Pipedream TEXT" and value is `pipedream_object`). This should be more clear to the user.

lcaresia (approved):

PR #15673, opened 2025-02-19
[FIX] SerpApi - issue with no_cache and q props

### LINKED ISSUE #15593: SerpApi
### App
SerpApi

### Summary:
The SerpApi "Scrape Search" pipedream template is not working as expected

### Details:
Hi there,

My name is Sonika, and I am a Customer Success Engineer at SerpApi. 

A user reached out to us about the "Scrape Search" pipedream template not updating the response for different queries in test runs and just repeatedly returning a single saved response. This issue refers to the SerpApi integration here: https://pipedream.com/apps/serpapi

I looked into it and noticed the no_cache parameter is not getting set properly in the built-in "Scrape Search" API call template provided.

no_cache should be set as a String "true" value whereas the template is setting it as a boolean value which could be causing the issue.

There seemed to be an additional issue besides no_cache not being set correctly - the SerpApi call seems to us inputs from the previous run, even when the query is updated in the current run using a variable. This could be somehow related to the no_cache parameter or could be related to the query parameter being a list, instead of a just a single string.

For reference, here is our documentation for this API: https://serpapi.com/search-api

Thank you for looking into this. Please let me know how I can help in fixing this. 

### Screenshots:

Screenshot

jcortes:
## WHY

Resolves #15593

## Summary by CodeRabbit

- **Refactor**
  - Simplified the search query input by transitioning from multiple values to a single string for a more straightforward experience.
  
- **Documentation**
  - Enhanced guidance on formulating search queries, now including details on using advanced search parameters.
  
- **Chores**
  - Updated component versions and dependency support to ensure improved compatibility and performance.

### DIFF
--- a/components/serpapi/serpapi.app.mjs
+++ b/components/serpapi/serpapi.app.mjs
@@ -12,9 +12,9 @@ export default {
       options: constants.ENGINES,
     },
     q: {
-      type: "string[]",
+      type: "string",
       label: "Query",
-      description: "The query or queries you want to search. When passing multiple queries you need to use a comma `,` to separate them",
+      description: "The query you want to search. You can use anything that you would use in a regular Google search. e.g. `inurl:`, `site:`, `intitle:`. SerpApi also supports advanced search query parameters such as `as_dt` and `as_eq`. See the [full list](https://serpapi.com/advanced-google-query-parameters) of supported advanced search query parameters.",
     },
     device: {
       type: "string",
--- a/components/serpapi/actions/scrape-search/scrape-search.mjs
+++ b/components/serpapi/actions/scrape-search/scrape-search.mjs
@@ -4,7 +4,7 @@ export default {
   key: "serpapi-scrape-search",
   name: "Scrape Search",
   description: "Scrape the results from a search engine via SerpApi service. [See the documentation](https://serpapi.com/search-api)",
-  version: "0.0.2",
+  version: "0.0.3",
   type: "action",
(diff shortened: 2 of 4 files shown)


components/serpapi/serpapi.app.mjs
```diff
@@ -12,9 +12,9 @@ export default {
       options: constants.ENGINES,
     },
     q: {
-      type: "string[]",
+      type: "string",
       label: "Query",
-      description: "The query or queries you want to search. When passing multiple queries you need to use a comma `,` to separate them",
+      description: "Parameter defines the query you want to search. You can use anything that you would use in a regular Google search. e.g. `inurl:`, `site:`, `intitle:`. We also support advanced search query parameters such as `as_dt` and `as_eq`. See the [full list](https://serpapi.com/advanced-google-query-parameters) of supported advanced search query parameters.",
```
GTFalcao, line 17:
```suggestion
      description: "The query you want to search. You can use anything that you would use in a regular Google search. e.g. `inurl:`, `site:`, `intitle:`. SerpApi also supports advanced search query parameters such as `as_dt` and `as_eq`. See the [full list](https://serpapi.com/advanced-google-query-parameters) of supported advanced search query parameters",
```
Minor text improvement suggestions

GTFalcao (commented):
I left one text suggestion attached, but this can be moved forward

GTFalcao (approved):

PR #15905, opened 2025-03-12
[FEATURE] Return Record key as well as data in Data Stores: Get Record action

### LINKED ISSUE #15858: [FEATURE] Return key as well as data in Data Stores: Get Record action
In other steps in my workflow I need to know what the key and/or data source name has been used in the [Data Stores: Get Record](https://github.com/PipedreamHQ/pipedream/blob/master/components/data_stores/actions/get-record-or-create/get-record-or-create.mjs) action. This isn't possible currently.

Export these prop values from the **Data Stores: Get Record action** action step (and probably all the other Data Stores steps that are relevant).

**Do you have a workaround?**
The workaround is to add a custom Node JS step with a prop to set these values, then reference those in the Data Stores: Get Record step, and other steps. However, this adds an unnecessary step.

N/A

**Additional context**
N/A

michelle0927:
Resolves #15858 

- Updated "Get record (or create one if not found)" and "Delete a single record" to export the `key`.
- Pipedream doesn't currently support a way to access the Data Store name from within the component.

### DIFF
--- a/components/data_stores/actions/get-record-or-create/get-record-or-create.mjs
+++ b/components/data_stores/actions/get-record-or-create/get-record-or-create.mjs
@@ -4,7 +4,7 @@ export default {
   key: "data_stores-get-record-or-create",
   name: "Get record (or create one if not found)",
   description: "Get a single record in your [Pipedream Data Store](https://pipedream.com/data-stores/) or create one if it doesn't exist.",
-  version: "0.0.11",
+  version: "0.0.12",
   type: "action",
   props: {
     app,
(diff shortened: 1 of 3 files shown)


components/data_stores/actions/get-record-or-create/get-record-or-create.mjs
```diff
@@ -46,37 +46,41 @@ export default {
...
-      await this.dataStore.set(this.key, parsedValue);
-      $.export("$summary", `Successfully added a new record with the key, \`${this.key}\`.`);
-    }
+      // Include TTL information in the return value if it was set
+      if (this.ttl) {
+        return {
+          value: parsedValue,
+          ttl: this.ttl,
+        };
+      }
```
binarymonkey84, line 77: This looks like it returns too early, and therefore doesn't export the summary or key.

PR #15994, opened 2025-03-21
Fixing app main file imports

GTFalcao:
All these packages were pointing to a non-existent file in their main field, meaning they couldn't be imported in a code step.

Can be reproduced with a Node.js code step such as the following:

```
import app from "@pipedream/spotify"
  props: {
    app,
  },
  async run({steps, $}) {
    return this.app.$auth;
  },
})
```

## Summary by CodeRabbit

- **Chores**
  - Minor version bumps applied across multiple integrations.
  - Main module entry points have been updated to use the modern ES module format.
  - Packaging adjustments were made for greater consistency across components.

### DIFF
--- a/components/algolia/package.json
+++ b/components/algolia/package.json
@@ -1,8 +1,8 @@
 {
   "name": "@pipedream/algolia",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "Pipedream Algolia Components",
-  "main": "algolia.app.js",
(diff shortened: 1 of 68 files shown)


michelle0927 (approved):
LGTM! (Reviewing because I noticed this wasn't assigned a reviewer). Moving to Ready for Release because I don't think this needs QA? Feel free to move it back to Ready for QA if I'm mistaken.

PR #15999, opened 2025-03-21
[FEATURE] Slack App Improvements

### LINKED ISSUE #15953: [FEATURE]  Slack App Improvements
Slack API calls are subject to rate limits, and we need to improve our handling of those limits
1. **Pagination**: Slack API calls that support pagination are [rate limited more heavily](https://api.slack.com/apis/rate-limits#pagination) when called without limits.  Unfortunately the default behavior for some of these calls (e.g. [users.list](https://api.slack.com/methods/users.list#arg_limit)) is to make a non-limited request.  We should, by default, **always pass a reasonable limit in our requests to endpoints that support pagination**. [Example fix](https://github.com/PipedreamHQ/pipedream/blob/ed81fb51789a6fcce17fa6ea334a7bf430890db5/components/slack/slack.app.mjs#L606)
2. **SDK Rate Limit Handling**: By default, the Slack Node SDK [handles rate limited requests](https://tools.slack.dev/node-slack-sdk/web-api/#rate-limits) by blocking for N seconds (N is determined by the server's 429 backoff response), and then retrying the request.
    **Anyc Options**: We should have our default behavior be  `rejectRateLimitedCalls` for any async options calls, and display a notification to the user.
    **Workflow Executions**: we may want to block and retry workflow executions automatically, but we should handle that explicitly, and try to notify the user when it happens.
3. **UserNames**: The `userNames()` function recursively calls `users.list` until it retrieves the full list of users in a workspace.  We should not be using this function anymore.
  **Listing Users**: For the `list-users` action (and any other actions/sources that use the `users/list` API endpoint), we should instead be calling with pagination, default to only retrieving one page, and providing an option to retrieve subsequent pages (but only if the user configures it).
  **Resolving User Names from ID's**: In most cases we're currently calling `userNames()`, it's to translate Slack User ID's to human readable User Names.  In these cases, we should call `users.profile.get` if we have a small number of users to look up, or provide a function that pages through `users.list` calls and looks for a given set of UserID's, stopping when they are all resolved.

Task List:

- [x] Find all API calls we're making that support pagination, and make sure we default to using pagination
- [x] Switch our sdk client to `rejectRateLimitedCalls` by default.
- [ ] For async options calls, handle rate limits by notifying the user that they've been rate limited
- [x] For workflow executions, handle rate limits explicitly by catching rate limit errors and sleeping (e.g. using `setTimeout`), but notify user (e.g. via logging to standard output) that execution was delayed for this reason.
- [ ] (Nice to Have) Use `$.flow.rerun` to pause and retry rather than `setTimeout`.
- [ ] (Nice to Have) Provide an option to fail and notify on workflow executions that hit rate limits.
- [x] Remove the `userNames()` function
- [x] Change `list-users` (and any other components that use `users.list`) to use pagination, and provide options for page size and number of pages to retrieve (defaulting to 1 page, and a reasonable page size).
- [x] Where possible/reasonable, use `users.profile.get` (tier 4 rate limit) instead of `users.list` (tier 2 rate limit)
- [x] If still needed, provide a helper function `userNameLookup(list_of_ids)` which pages through `users.list`, and returns a map of ID's to Names, stopping when all ID's are found.  By default, this function should immediately fail if it's rate limited.

michelle0927:
Resolves #15953 

## Summary by CodeRabbit

- **New Features**
  - Introduced configurable pagination for data listings (channels, files, group members, and messages) to give users more control.

- **Refactor**
  - Streamlined Slack integration calls for messaging, file management, and profile updates, enhancing reliability and consistency.
  - Enhanced error handling for rate limits and improved user name retrieval methods.

- **Chore**
  - Updated version identifiers across multiple Slack components to reflect internal improvements.

### DIFF
--- a/components/slack/slack.app.mjs
+++ b/components/slack/slack.app.mjs
@@ -1,5 +1,7 @@
 import { WebClient } from "@slack/web-api";
 import constants from "./common/constants.mjs";
+import get from "lodash/get.js";
+import retry from "async-retry";
 
 export default {
   type: "app",
@@ -15,27 +17,24 @@ export default {
         const types = [
           "im",
         ];
-        let [
-          userNames,
-          conversationsResp,
-        ] = await Promise.all([
-          prevContext.userNames ?? this.userNames(),
-          this.availableConversations(types.join(), prevContext.cursor),
-        ]);
+        let conversationsResp
+          = await this.availableConversations(types.join(), prevContext.cursor, true);
         if (channelId) {
-          const { members } = await this.sdk().conversations.members({
+          const { members } = await this.listChannelMembers({
             channel: channelId,
+            throwRateLimitError: true,
           });
           conversationsResp.conversations = conversationsResp.conversations
             .filter((c) => members.includes(c.user || c.id));
         }
+        const userIds = conversationsResp.conversations.map(({ user }) => user);
+        const userNames = await this.userNameLookup(userIds);
         return {
           options: conversationsResp.conversations.map((c) => ({
             label: `@${userNames[c.user]}`,
             value: c.user || c.id,
           })),
           context: {
-            userNames,
             cursor: conversationsResp.cursor,
           },
         };
@@ -50,7 +49,7 @@ export default {
         const types = [
           "mpim",
         ];
-        const resp = await this.availableConversations(types.join(), cursor);
+        const resp = await this.availableConversations(types.join(), cursor, true);
         return {
           options: resp.conversations.map((c) => {
             return {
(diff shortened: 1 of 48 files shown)


GTFalcao (approved):
LGTM!

components/slack/slack.app.mjs
```diff
@@ -450,29 +472,48 @@ export default {
...
+        minTimeout: 3000,
+      };
+      return retry(async (bail) => {
+        try {
+          return await apiCall();
+        } catch (error) {
+          const statusCode = get(error, "code");
+          if (statusCode === 429) {
+            if (throwRateLimitError) {
+              bail(`Rate limit exceeded. ${error}`);
```
GTFalcao, line 507: Do you know if this message will be shown in the async options UI?

michelle0927: Hi @GTFalcao. I did some testing, made some minor adjustments, and yes the message is displayed in the async options!
Screenshot 2025-03-26 at 3 13 54 PM

GTFalcao (approved):

PR #16024, opened 2025-03-25
Fireflies - error handling

michelle0927:
Add configuration errors to Fireflies actions.

## Summary by CodeRabbit

- **Bug Fixes**
	- Improved error handling in meeting actions to ensure required IDs are provided, delivering clearer feedback when issues occur.
	- Enhanced error reporting for meeting queries to better communicate API errors.

- **Chores**
	- Updated version numbers and upgraded dependencies for enhanced stability and reliability.

### DIFF
--- a/components/fireflies/actions/find-meeting-by-id/find-meeting-by-id.mjs
+++ b/components/fireflies/actions/find-meeting-by-id/find-meeting-by-id.mjs
@@ -1,11 +1,12 @@
 import fireflies from "../../fireflies.app.mjs";
 import queries from "../../common/queries.mjs";
+import { ConfigurationError } from "@pipedream/platform";
 
 export default {
   key: "fireflies-find-meeting-by-id",
   name: "Find Meeting by ID",
   description: "Locates a specific user meeting by its unique ID. [See the documentation](https://docs.fireflies.ai/graphql-api/query/transcript)",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "action",
   props: {
     fireflies,
@@ -17,6 +18,10 @@ export default {
     },
   },
   async run({ $ }) {
+    if (!this.meetingId) {
+      throw new ConfigurationError("Meeting ID is required");
+    }
+
     const meeting = await this.fireflies.query({
       $,
       data: {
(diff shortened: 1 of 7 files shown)


components/fireflies/actions/find-recent-meeting/find-recent-meeting.mjs
```diff
@@ -17,6 +18,10 @@ export default {
     },
   },
   async run({ $ }) {
+    if (!this.userId) {
+      throw new ConfigurationError("User ID is required");
+    }
+
     const { data: { user: { recent_meeting: meetingId } } } = await this.fireflies.query({
```
danhsiung, line 25: @michelle0927 In an error case, what would the `query`` response be?  If the `query` response is { error: ["message"] }, what will the behavior be?

michelle0927: If the `userId` was invalid, it would return the error: “Cannot read properties of null (reading 'recent_meeting’)”.
Updated to throw a configuration error if that query returns an error.

For me, both actions return: "Error: You need to be subscribed to a paid plan to perform this action".

components/fireflies/actions/find-meeting-by-id/find-meeting-by-id.mjs
```diff
@@ -26,6 +31,11 @@ export default {
         },
       },
     });
+
+    if (meeting.errors?.length) {
```
danhsiung, line 35: So, minor coding style nit here, but rather than requiring all callers of [`proxy`](https://github.com/PipedreamHQ/pipedream/blob/ea0ddf2c66cacd0e3cec38dc47dfd8434862ade3/components/fireflies/fireflies.app.mjs#L73) handle these errors, which can be error prone, it might be more consistent to have the base function detect an error response (e.g. by finding an `error` key in the json response), and raising the `ConfigurationError` at that point. That way `upload-audio` any future actions, as well as async options can all benefit from the same error handling. Does that make sense?

components/fireflies/actions/find-recent-meeting/find-recent-meeting.mjs
```diff
@@ -26,6 +31,12 @@ export default {
         },
       },
     });
+
+    if (user.errors?.length) {
```
danhsiung, line 35: Same comment as above.

components/fireflies/actions/find-recent-meeting/find-recent-meeting.mjs
```diff
@@ -39,6 +50,11 @@ export default {
         },
       },
     });
+
+    if (meeting.errors?.length) {
```
danhsiung, line 54: Same comment as above.

PR #16026, opened 2025-03-25
fix(types): add missing `ServiceDB` interface

joscha:
any property of `type: "$.service.db"` becomes an instance of `ServiceDB` at runtime.

## WHY

So the implementation in the implementation methods can benefit from it.

### DIFF
--- a/types/src/index.ts
+++ b/types/src/index.ts
@@ -247,6 +247,11 @@ export interface ServiceDBProp extends BasePropInterface {
   type: "$.service.db";
 }
 
+export interface ServiceDB {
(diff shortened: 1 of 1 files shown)


types/src/index.ts
```diff
@@ -247,6 +247,11 @@ export interface ServiceDBProp extends BasePropInterface {
   type: "$.service.db";
 }
 
+export interface ServiceDB {
+  get<T extends JSONValue>(key: string): T | undefined;
+  set(key: string, value: JSONValue): void;
```
joscha, line 252:
```suggestion
  set(key: string, value: JSONValue | undefined): void;
```

PR #16072, opened 2025-03-27
[TRIGGER] Airtable Polling Trigger with filter

### LINKED ISSUE #15982: [TRIGGER] Airtable Polling Trigger with filter
The actual Airtable-Trigger let you:
- get instant updates (which fire a lot when editing a table)
- polling on new/modified/deleted....in view

But there is very relevant trigger not available.
The API supports filtering on polling via the formula field

It would be great to use this field also in the trigger (it is implemented in the get_records-Action).

A usecase would be:
- You can define an Airtable field with a modified-date from a specific field (or field set)
- Filtering on this field like IS_AFTER({${modifieddatefield}}, DATETIME_PARSE("${lastworkflowrun}", "YYYY-MM-DD HH:mm:ss"))` would give all records where the field changed since last call of this workflow

Filter-Formula-Possibilitities: https://support.airtable.com/docs/formula-field-reference
Airtable API used for the trigger: https://airtable.com/developers/web/api/list-records
n8n has this implemented in their trigger: https://github.com/n8n-io/n8n/blob/c06ce765f11dcde4731d3739e1aa5f27351c3cc2/packages/nodes-base/nodes/Airtable/AirtableTrigger.node.ts

jcortes:
## WHY

Resolves #15982

## Summary by CodeRabbit

- **New Features**
  - Enhanced error handling for smoother record retrieval.
  - Enabled dynamic filtering with customizable criteria for more flexible record searches.
  
- **Chores**
  - Updated version numbers across integration components to ensure consistency and reliability.

### DIFF
--- a/components/airtable_oauth/actions/create-comment/create-comment.mjs
+++ b/components/airtable_oauth/actions/create-comment/create-comment.mjs
@@ -4,7 +4,7 @@ export default {
   key: "airtable_oauth-create-comment",
   name: "Create Comment",
   description: "Create a comment on a selected record. [See the documentation](https://airtable.com/developers/web/api/create-comment)",
-  version: "0.0.8",
+  version: "0.0.9",
   type: "action",
   props: {
     ...common.props,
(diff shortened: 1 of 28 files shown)


michelle0927 (approved):
LGTM!

PR #16214, opened 2025-04-08
Notion markdown followup

GTFalcao:
The `Retrieve Content` prop was renamed back to `Retrieve Children` and changed to a three-option prop instead of a boolean (still optional).

- 'All Children' works the same as before, recursively retrieving all child blocks.
- 'Sub-Pages Only' filters these children to include only sub-pages.
- 'None' (same as default of not choosing anything, will only retrieve the page block itself).

I've special-cased this to check for a boolean value in the code and keep the same behavior (true = 'all children', false = 'none') in case users update the action and don't change the prop's value, so it'll function exactly as it did before.

Finally, the markdown is now returned as well as the blocks, such that enabling the 'Retrieve as Markdown' prop returns this:
```
{
  markdown: 'my content'
  block: {
    id: 'block_id',
    children: [ ... ],
    // ...
  }
}
```

while not enabling it keeps the same behavior as before it existed.
```
{
  id: 'block_id',
  children: [ ... ],
  // ...
}
```

## Summary by CodeRabbit

- **New Features**
	- Enhanced content retrieval now offers flexible filtering options, allowing you to select between retrieving all content, sub-pages only, or none.
	- Added support for optionally returning page content as markdown alongside standard details.
- **Improvements**
	- Streamlined retrieval configurations provide a more intuitive and consistent experience when managing nested content.
- **Version Updates**
	- Updated version numbers reflect the latest enhancements and improvements.

### DIFF
--- a/components/notion/actions/retrieve-block/retrieve-block.mjs
+++ b/components/notion/actions/retrieve-block/retrieve-block.mjs
@@ -4,7 +4,7 @@ export default {
   key: "notion-retrieve-block",
   name: "Retrieve Page Content",
   description: "Get page content as block objects or markdown. Blocks can be text, lists, media, a page, among others. [See the documentation](https://developers.notion.com/reference/retrieve-a-block)",
-  version: "0.1.0",
+  version: "0.2.0",
   type: "action",
   props: {
     notion,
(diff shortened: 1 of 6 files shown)


components/notion/notion.app.mjs
```diff
@@ -325,7 +325,8 @@ export default {
         block_id: blockId,
       });
     },
-    async retrieveBlockChildren(block, params = {}) {
+    async retrieveBlockChildren(block, subpagesOnly = false) {
```
GTFalcao, line 328: For the record, I've checked that all instances of this method being called did not include a `params` argument, so it was always initialized as an empty object - hence I changed it to a variable declaration instead.

GTFalcao:
I'm also skipping the mass-version-bumping that is usually requird for the check to pass, and opting to update only the versions of the components that are affected by the change (double-checked).

dannyroosevelt (approved):

GTFalcao:
We shipped this one due to its urgency, but code review and QA would be helpful nonetheless - we can follow up with any adjustments needed

components/notion/actions/retrieve-block/retrieve-block.mjs
```diff
@@ -15,33 +15,50 @@ export default {
...
+    if ([
+      true,
+      "All Children",
+      "Sub-Pages Only",
+    ].includes(retrieveChildren)) {
+      block.children = await this.notion.retrieveBlockChildren(block, subpagesOnly);
     }
-    $.export("$summary", `Successfully retrieved block${this.retrieveChildren
-      ? ` with ${block.children.length ?? 0} children`
+    $.export("$summary", `Successfully retrieved block${retrieveChildren
```
michelle0927, line 52: This line gets an error when `retrieveChildren` is set to `None`. Should add `&& retrieveChildren !== "None"`:

`$.export("$summary", `Successfully retrieved block${retrieveChildren && retrieveChildren !== "None"`

michelle0927 (commented):
I went ahead and reviewed this because it didn't have a reviewer assigned. Found one small issue.

PR #16384, opened 2025-04-21
Improve Apify actions

### LINKED ISSUE #16305: [ACTION] Improve Apify actions
In the Apify Run Action action, there are a couple issues:
1. The action isn't returning the results you would expect (they probably kick off an async job that requires we wait for the results, or potentially give them a webhook to post results to — we should potentially use $.flow.suspend() here)
2. Some of the additionalProps are flagged as `optional: false`, but it seems like they actually are optional

michelle0927:
Resolves #16305 

## Summary by CodeRabbit

- **New Features**
	- Default values are now displayed in property descriptions and assigned for supported types in the "Run Actor" action, improving clarity when configuring properties.
	- Added support for running actors asynchronously with new options for timeout, memory, max items, total charge limits, and webhooks.
	- Introduced a "Build" selector to specify actor builds when running actors.
	- Added new methods to run actors synchronously or asynchronously via updated APIs.

- **Bug Fixes**
	- The "Run Actor" feature now uses a synchronous API endpoint for improved request handling.

- **Chores**
	- Updated version numbers for several actions, sources, and the package to reflect recent changes.
	- Clarified actor ID format descriptions for better user understanding.

### DIFF
--- a/components/apify/actions/get-dataset-items/get-dataset-items.mjs
+++ b/components/apify/actions/get-dataset-items/get-dataset-items.mjs
@@ -5,7 +5,7 @@ export default {
   key: "apify-get-dataset-items",
   name: "Get Dataset Items",
   description: "Returns data stored in a dataset. [See the documentation](https://docs.apify.com/api/v2/dataset-items-get)",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "action",
   props: {
     apify,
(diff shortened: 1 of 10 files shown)


luancazarine (dismissed):

luancazarine (approved):

PR #16536, opened 2025-05-01
Gmail 'Create Draft' improvement

### LINKED ISSUE #15556: [FEATURE] Gmail - Create Draft with "Sender" and "Signature"
I'm prepraring Gmail mail drafts with Pipedream and have to manually set the sender and the signature before sending. 

The "Create Draft" Action should have the possibility to set the 
- sender email address 
- email signature

**Do you have a workaround?**
do it manual

I used the Zapier action, which was possible to set these two values

GTFalcao:
Closes #15556 

Note: only signatures that are explicitly set as the sender email's signature can be selected (those that are not, are not returned by Gmail's API, but a custom value can still be used)

Regarding PR checks, only updating the version of 1 component since no others use the signature property which was added in the app file's method.

## Summary by CodeRabbit

- **New Features**
  - Added options to specify the sender's email address and include an email signature when creating Gmail drafts.

- **Improvements**
  - Enhanced signature handling so that HTML signatures are automatically appended to the email body when using HTML format.
  - Clarified the description for the signature property to indicate it applies only to HTML email bodies.

- **Chores**
  - Updated package version numbers.

### DIFF
--- a/components/gmail/actions/create-draft/create-draft.mjs
+++ b/components/gmail/actions/create-draft/create-draft.mjs
@@ -6,7 +6,7 @@ export default {
   key: "gmail-create-draft",
   name: "Create Draft",
   description: "Create a draft from your Google Workspace email account. [See the documentation](https://developers.google.com/gmail/api/reference/rest/v1/users.drafts/create)",
-  version: "0.0.5",
+  version: "0.1.0",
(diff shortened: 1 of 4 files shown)


michelle0927 (approved):
LGTM!

PR #16830, opened 2025-05-26
[ACTION] Gmail Find Email actions return too much text #16824

### LINKED ISSUE #16824: [ACTION] Gmail Find Email actions return too much text
When using any of the Gmail actions that return the contents of an email (the triggers probably fall into this category also), the Gmail API / our components return way too much text for an LLM to meaningfully process.

We should potentially add a prop to let users / LLMs return text-only versions, without all the HTML.

luancazarine:
Resolves #16824


lcaresia (approved):

components/gmail/actions/find-email/find-email.mjs
```diff
@@ -14,6 +15,12 @@ export default {
         "q",
       ],
     },
+    withTextPayload: {
+      type: "boolean",
+      label: "With Text Payload",
```
dannyroosevelt, line 20:
```suggestion
      label: "Return payload as plaintext",
```

PR #17255, opened 2025-06-24
Google Analytics - add async options for properties

### LINKED ISSUE #17133: Google Analytics
### App
Google Analytics

### Summary:
Asking the app to run GA or GA4 reports returns errors

### Details:
When requesting reports from the Google Analytics app, most of the times it will return errors from Google.

The main issues for the errors seem to be:

1. The app is unaware of which properties are connected with the authenticated account, and does not ask Google Analytics API for them, os it makes up its own properties

2. If specifically instructed to run a report for a named property ID, it will often submit that property ID in the string format "property/123556". However, Google Analytics expects to receive the property ID in the format "123456".

I assume this could be easily fixed if the app had some actions for requesting account ID and properties associated with the account ID.

### Screenshots:
No screenshots included

### LINKED ISSUE #17147: [FEATURE] update property (and maybe viewId) to use async options in Google Analytics components
Wherever possible for Google Analytics triggers and actions, we should make sure we're using async options. It seems like we can potentially retrieve a list of `property` values via their `/v1beta/properties` API, but we should confirm.

This might be the relevant endpoint: https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1beta/accountSummaries/list#PropertySummary

michelle0927:
Resolves #17147 
Resolves #17133

### DIFF
--- a/components/google_analytics/actions/create-ga4-property/create-ga4-property.mjs
+++ b/components/google_analytics/actions/create-ga4-property/create-ga4-property.mjs
@@ -5,7 +5,7 @@ export default {
   key: "google_analytics-create-ga4-property",
   name: "Create GA4 Property",
   description: "Creates a new GA4 property. [See the documentation](https://developers.google.com/analytics/devguides/config/admin/v1/rest/v1beta/properties/create)",
-  version: "0.1.0",
+  version: "0.1.1",
   type: "action",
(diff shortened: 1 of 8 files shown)


GTFalcao (approved):
LGTM!

PR #17267, opened 2025-06-24
Configure file stash sync for Zoho CRM components

js07:
## WHY

Adding dir props to the Zoho CRM Download Attachment and Upload Attachment actions indicates that a stash ID should be passed when running these actions via the Connect API. This allows files written to the ephemeral /tmp directory in one action to be read by another, separately executed action since both can sync /tmp with the same File Stash directory.


vunguyenhung:

Test report
https://vunguyenhung.notion.site/Configure-file-stash-sync-for-Zoho-CRM-components-21fbf548bb5e81a39e39df0b920d644f

PR #17373, opened 2025-06-30
Configure file stash sync: Happy Scribe-Puppeteer

js07:
Adds dir props to relevant components for apps starting with 'H' to 'P'.

## WHY

The 'dir' prop indicates that a stash ID should be passed when running these actions via the Connect API. This allows files written to the ephemeral /tmp directory in one action to be read by another, separately executed action since both can sync /tmp with the same File Stash directory.


js07:
.

PR #17687, opened 2025-07-17
Calendly - Improve List Events action and New Event Scheduled source

### LINKED ISSUE #17679: [FEATURE] Calendly - Improve List Events action and New Event Scheduled source
Currently, the `List Events` action and `New Event Scheduled` source fetch events for org scope by default. Users on an org that don't have permission won't be able to use those components.

- Add a new prop for user to specify the scope they want to fetch events for: organization/user/group . Default value should be user
- On source creation, use the specified scope to fetch new events. If user don’t have permission, show permission error.

See API Doc: https://developer.calendly.com/api-docs/2d5ed9bbd2952-list-events

**Do you have a workaround?**
No

**Additional context**
See context here: https://pipedream-users.slack.com/archives/G01J50YGJGP/p1752726799179699?thread_ts=1752724766.867609&cid=G01J50YGJGP

michelle0927:
Resolves #17679

### DIFF
--- a/components/calendly_v2/actions/list-events/list-events.mjs
+++ b/components/calendly_v2/actions/list-events/list-events.mjs
@@ -4,15 +4,39 @@ export default {
   key: "calendly_v2-list-events",
   name: "List Events",
   description: "List events for an user. [See the documentation](https://calendly.stoplight.io/docs/api-docs/b3A6NTkxNDEy-list-events)",
-  version: "0.0.5",
+  version: "0.0.6",
   type: "action",
   props: {
     calendly,
+    alert: {
+      type: "alert",
+      alertType: "info",
+      content: `
+      Select "authenticatedUser" scope to return events for the authenticated user
+      Select "organization" scope to return events for that organization (requires admin/owner privilege)
+      Select "user" scope to return events for a specific User in your organization (requires admin/owner privilege)
+      Select "group" scope to return events for a specific Group (requires organization admin/owner or group admin privilege)`,
+    },
+    scope: {
+      type: "string",
+      label: "Scope",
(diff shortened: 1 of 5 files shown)


components/calendly_v2/sources/new-event-scheduled/new-event-scheduled.mjs
```diff
@@ -18,6 +18,72 @@ export default {
...
       },
     },
+    alert: {
+      type: "alert",
+      alertType: "info",
+      content: `
+      Select "authenticatedUser" scope to return events for the authenticated user
+      Select "organization" scope to return events for that organization (requires admin/owner privilege)
+      Select "user" scope to return events for a specific User in your organization (requires admin/owner privilege)
+      Select "group" scope to return events for a specific Group (requires organization admin/owner or group admin privilege)`,
```
GTFalcao: Any chance we'll need the scope prop for other Calendly components in the future? If so I'd move the alert and the additionalProps to a common file (and scope to a propDefinition)

If it's just these two it's probably fine to keep it as is

components/calendly_v2/actions/list-events/list-events.mjs
```diff
@@ -51,14 +88,35 @@ export default {
...
+
+    props.user.hidden = this.scope !== "user";
+    props.user.optional = this.scope !== "user";
+
+    return {};
+  },
   async run({ $ }) {
     const params = {
-      organization: this.organization,
+      inviteeEmail: this.inviteeEmail,
```
vunguyenhung:
```suggestion
      invitee_email: this.inviteeEmail,
```

PR #17689, opened 2025-07-17
Frontapp new components

### LINKED ISSUE #17465: Front
### App
Front
https://dev.frontapp.com/docs/welcome

### Sources
- teammate.created           
- teammate.updated            
- inbox.created               
- inbox.updated               
- message_template.created    
- message_template.updated    
- message_template.deleted 

### Actions
- Create teammate → /teammates
- Update teammate → /teammates/{teammate_id}

- Create inbox → /inboxes
- Update inbox → /inboxes/{inbox_id}

- List templates → /message_templates
- Create template → /message_templates
- Update template → /message_templates/{template_id}
- Delete template → /message_templates/{template_id}

GTFalcao:
Closes #17465 

## Summary by CodeRabbit

* **New Features**
  * Added actions to create inboxes, create and delete message templates, update teammates, and list message templates in FrontApp.
  * Introduced sources to detect when a new conversation or message template is created.
  * Added test event data for new conversation and message template sources.

* **Improvements**
  * Enhanced app with new properties and methods for managing message templates, folders, teammates, and inboxes.
  * Updated action and source descriptions for clarity and consistency.
  * Improved event source filtering to support conditional event emission.

* **Bug Fixes**
  * Minor metadata and version updates to ensure consistency across actions and sources.
  * Fixed safe handling of optional fields in utilities to prevent errors.


components/frontapp/actions/list-message-templates/list-message-templates.mjs
```diff
@@ -0,0 +1,61 @@
...
+  key: "frontapp-list-message-templates",
+  name: "List Message Templates",
+  description: "List the message templates. [See the documentation](https://dev.frontapp.com/reference/list-message-templates).",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    frontApp,
+    sortBy: {
+      type: "string",
+      label: "Sort By",
```
GTFalcao: Note: the docs do not mention a page or per_page parameter for this endpoint at all

luancazarine (approved):

components/frontapp/actions/create-message-template/create-message-template.mjs
```diff
@@ -0,0 +1,115 @@
...
+        "inboxId",
+      ],
+      optional: true,
+    },
+    attachments: {
+      propDefinition: [
+        frontApp,
+        "attachments",
+      ],
+    },
```
michelle0927, line 51: I believe we're supposed to be adding `syncDir` to components that work with the /tmp directory.
```suggestion
    attachments: {
      propDefinition: [
        frontApp,
        "attachments",
      ],
    },
    syncDir: {
      type: "dir",
      accessMode: "write",
      sync: true,
    },
```

components/frontapp/actions/list-message-templates/list-message-templates.mjs
michelle0927: The docs show it uses a `page_token` instead. This action should have a `maxResults` prop and return a paginated list as the result. Note: You can look at how the `messageTemplateId` prop uses pagination.

components/frontapp/sources/common/polling-ids.mjs
```diff
@@ -0,0 +1,55 @@
...
+      },
+    },
+  },
+  methods: {
+    _getSavedIds() {
+      return this.db.get("savedIds") || [];
+    },
+    _setSavedIds(ids) {
+      this.db.set("savedIds", ids);
+    },
```
michelle0927: Would it be simpler to use a timestamp to check the events, rather than storing all the ID's?

If that's not an option, storing the ID's in an object (e.g. { 123: true, 456: true }) would reduce execution time. Checking whether savedIds[id] is `true` is faster than using savedIds.includes().

michelle0927 (changes requested):

components/frontapp/actions/create-message-template/create-message-template.mjs
michelle0927: The `syncDir` prop has to do with File Stash. https://pipedream.com/docs/connect/components/files

components/frontapp/sources/common/polling-ids.mjs
GTFalcao: Thanks for the suggeston, I've changed the two sources to adhere to the timestamp format, with a few modifications to the base file to support them

components/frontapp/actions/list-message-templates/list-message-templates.mjs
```diff
@@ -0,0 +1,34 @@
...
+  type: "action",
+  props: {
+    frontApp,
+    maxResults: {
+      propDefinition: [
+        frontApp,
+        "maxResults",
+      ],
+    },
+  },
```
michelle0927, line 49: Thanks for making that change. I notice you removed the `sortBy` and `sortOrder` props. I think it's nice to give the user those options if they're available. What do you think?

components/frontapp/sources/common/base.mjs
```diff
@@ -30,10 +30,18 @@ export default {
       let responseArray = [];
 
       for await (const item of items) {
-        responseArray.push(item);
+        // If filterFn is provided, use it to filter items, otherwise add all items
+        if (!filterFn || filterFn(item, lastTs)) {
```
michelle0927, line 37: I don't see the `filterFn` being used in any of the sources. Without that, won't it paginate through everything instead of stopping when it gets to items older than `lastTs`?

components/frontapp/actions/list-message-templates/list-message-templates.mjs
GTFalcao: I figured I'd follow the pattern of other actions and use the "latest" default approach, but there's no harm in adding the props, sure. Adding them back in

components/frontapp/sources/common/base.mjs
GTFalcao: Thanks, you're right, I made a mistake here when moving code around to keep it reusable - added the filterFn parameters back to the two sources that need them

michelle0927 (approved):

Issue #17399, opened 2025-07-02
[BUG] Error thrown when configuring `folderId` prop in `microsoft_excel-new-row-added` trigger


dannyroosevelt:
https://pipedream-connect-demo.vercel.app/connect/demo?type=trigger&app=microsoft_excel&component=microsoft_excel-new-row-added

```
Error: Minified React error #31; visit https://react.dev/errors/31?args[]=object%20with%20keys%20%7Bvalue%2C%20label%7D for the full message or use the non-minified dev environment for full errors and additional helpful warnings.
```

![Image](https://github.com/user-attachments/assets/1446377d-b276-490e-b3c1-94f508cdc325)

mariczne:
The issue is that `ControlSelect` doesn't handle options of mixed type (it assumes all elements are the type of 0-th element).

We can somehow work around this by unifying the types in in `RemoteOptionsContainer`. `stringOption` is already a bit of a misnomer here, but I kept it for a cleaner diff: 

```patch
diff --git a/packages/connect-react/src/components/RemoteOptionsContainer.tsx b/packages/connect-react/src/components/RemoteOptionsContainer.tsx
index 5dd526f79..0eaebfcc9 100644
--- a/packages/connect-react/src/components/RemoteOptionsContainer.tsx
+++ b/packages/connect-react/src/components/RemoteOptionsContainer.tsx
@@ -127,10 +127,14 @@ export function RemoteOptionsContainer({ queryEnabled }: RemoteOptionsContainerP
       if (stringOptions?.length) {
         const options = [];
         for (const stringOption of stringOptions) {
-          options.push({
-            label: stringOption,
-            value: stringOption,
-          });
+          if (typeof stringOption === 'object' && stringOption !== null && 'value' in stringOption && 'label' in stringOption) {
+            options.push(stringOption);
+          } else {
+            options.push({
+              label: stringOption,
+              value: stringOption,
+            });
+          }
         }
         _options = options;
       }
```

PR #17747, opened 2025-07-22
Fixes GH-17399

dannyroosevelt:


packages/connect-react/src/utils/type-guards.ts
jverce, line 1: These utility functions should be documented, they seem to be pretty important but some of them are not quite easy intuitive in terms of their purpose.

Issue #17817, opened 2025-07-26
[ACTION] Zep - Memory APIs


vunguyenhung:
**Is there a specific app this action is for?**
https://pipedream.com/apps/zep

**Please provide a link to the relevant API docs for the specific service / operation.**
- Get Threads
- Get Messages for Thread

See API Doc here: https://help.getzep.com/sdk-reference/memory/get

Raised by a customer here: https://pipedream-users.slack.com/archives/CMZG4EBJ9/p1753524554247159

vunguyenhung:
Hi everyone, all test cases are passed! Ready for release! 

Test report
https://vunguyenhung.notion.site/ACTION-Zep-Memory-APIs-254bf548bb5e815e9e08dc3af22f5d38

PR #17897, opened 2025-07-31
[ACTION] Zep - Memory APIs #17817

lcaresia:
## Summary by CodeRabbit

* **New Features**
  * Introduced actions to retrieve session information, session memory, session messages, and a list of all sessions.
* **Chores**
  * Updated version numbers for several actions, sources, and the package.
  * Upgraded a platform dependency to a newer version.


components/zep/actions/get-sessions/get-sessions.mjs
```diff
@@ -0,0 +1,19 @@
...
+  props: {
+    zep,
+  },
+  async run({ $ }) {
+    const response = await this.zep.listSessions({
+      $,
+    });
+    $.export("$summary", `Successfully retrieved ${response.sessions.length} sessions`);
+    return response;
+  },
```
michelle0927: `get-sessions` should return a paginated list of sessions. The prop, `sessionId` shows it accepts `order_by` and `asc` props. I'd include those as optional for the user, and also a `maxResults` prop.

michelle0927 (changes requested):
This is failing checks because it needs version updated for `new-message`, `new-session`, and `session-updated`.

PR #17911, opened 2025-08-01
[BUG] Microsoft Calendar using Get Busy/Free action throwing error

### LINKED ISSUE #17492: [BUG]
**Describe the bug**
When trying to get the schedule of my Microsoft Calendar using "Get Busy/Free" action i am facing the error 

`When trying to read a null collection parameter value in JSON Light, a node of type 'PrimitiveValue' with the value '["hassan@replicaide.com"]' was read from the JSON reader; however, a primitive 'null' value was expected.`

as can be understood from the error that msf no longer accept the param emails list when i tried to send nothing in the emails param i can't use the action because emails is the required column

**To Reproduce**

1. Go to get_schedule under Outlook Calendar
2. try to fill all the common information and hit the action

**Expected behavior**
it should have shown me the free/busy schedule of my account

**Screenshots**

Image

jcortes:
## WHY

Resolves #17492

## Summary by CodeRabbit

* **New Features**
  * Improved input handling for schedules by supporting multiple input formats.
  * Added validation to ensure schedules are provided before proceeding.

* **Enhancements**
  * Updated the label for the schedules input from "Emails" to "Schedules" for clarity.

* **Chores**
  * Updated component and package versions.

### DIFF
--- a/components/microsoft_outlook_calendar/actions/get-schedule/get-schedule.mjs
+++ b/components/microsoft_outlook_calendar/actions/get-schedule/get-schedule.mjs
@@ -1,16 +1,18 @@
+import { ConfigurationError } from "@pipedream/platform";
 import microsoftOutlook from "../../microsoft_outlook_calendar.app.mjs";
+import utils from "../../common/utils.mjs";
 
 export default {
   key: "microsoft_outlook_calendar-get-schedule",
   name: "Get Free/Busy Schedule",
   description: "Get the free/busy availability information for a collection of users, distributions lists, or resources (rooms or equipment) for a specified time period. [See the documentation](https://learn.microsoft.com/en-us/graph/api/calendar-getschedule)",
-  version: "0.0.2",
+  version: "0.0.3",
   type: "action",
   props: {
     microsoftOutlook,
     schedules: {
       type: "string[]",
-      label: "Emails",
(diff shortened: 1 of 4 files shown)


jcortes:
> LGTM! Moving to QA
> 
> One comment/question: I wonder if the `schedules` prop could be improved to use async options by using this endpoint: https://learn.microsoft.com/en-us/graph/api/user-list

Hi @michelle0927 actually I was testing this endpoint and seems like is not working on Microsoft side. Here it is a test I made from the graphql explorer itself

image

This is the grapql explorer tool https://developer.microsoft.com/en-us/graph/graph-explorer

I've also tested with `curl`
```
curl https://graph.microsoft.com/v1.0/users\?$select\=id,displayName,mail \
-H "Authorization: Bearer $API_KEY"
```

The error was:
```
{"@odata.context":"https://graph.microsoft.com/v1.0/$metadata#users","value":[{"error":{"code":"InternalServerError","message":"Invalid response. Expected list response must have value property.","innerError":{"date":"2025-08-02T20:09:00","request-id":"4cf8738b-9645-4f6f-afe7-2d96d5d845e1","client-request-id":"4cf8738b-9645-4f6f-afe7-2d96d5d845e1"}}}
```

PR #17938, opened 2025-08-04
Dependency import fixes

GTFalcao:
- Added the missing dependencies that apps import to their `package.json` files, mostly `@pipedream/platform` and `@pipedream/types`;
- Moved `@pipedream/types` to `devDependencies` instead of `dependencies`.


GTFalcao:
I'll merge this since there are no components to QA. This is just fixing package dependency patterns

PR #17965, opened 2025-08-06
Revise HubSpot prop type

### LINKED ISSUE #17964: [BUG] Revise HubSpot prop types
**Describe the bug**
This is from a business customer.

On HubSpot Create Association action > Association Type, putting the custom experssion as string value resulted in the error below
Image

This is because our `associationType` prop has type `string`, but the value we set on the option is integer (response from HubSpot)
```
    associationType: {
      type: "string", // <- Type is String
      label: "Association Type",
      description: "Type of the association",
      async options({
        fromObjectType, toObjectType,
      }) {
        if (!fromObjectType || !toObjectType) {
          return [];
        }
        const { results: associationTypes } = await this.getAssociationTypes({
          fromObjectType,
          toObjectType,
        });
        return associationTypes.map((associationType) => ({
          label: associationType.label ?? `${fromObjectType}_to_${toObjectType}`,
          value: associationType.typeId, // <- The associationType.typeId is integer, which is from HubSpot response
        }));
      },
    },
```

**Expected behavior**
- Please update the `Create Association action > Association Type` to have the type integer
- Please revise the all existing HubSpot actions for similar issues.

michelle0927:
Resolves #17964

### DIFF
--- a/components/hubspot/hubspot.app.mjs
+++ b/components/hubspot/hubspot.app.mjs
@@ -367,7 +367,7 @@ export default {
       },
     },
     associationType: {
-      type: "string",
+      type: "integer",
       label: "Association Type",
       description: "Type of the association",
       async options({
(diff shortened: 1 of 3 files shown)


luancazarine (approved):

PR #17975, opened 2025-08-07
Linear - pin sdk version

andrewjschuang:
## WHY

The Linear API key components were getting this error:

SyntaxError
The requested module '@linear/sdk' does not provide an export named 'LinearClient'

### DIFF
--- a/components/linear_app/actions/create-issue/create-issue.mjs
+++ b/components/linear_app/actions/create-issue/create-issue.mjs
@@ -5,7 +5,7 @@ export default {
   key: "linear_app-create-issue",
   name: "Create Issue",
   description: "Creates a new issue in Linear. Requires team ID and title. Optional: description, assignee, project, state. Returns response object with success status and issue details. Uses API Key authentication. See Linear docs for additional info [here](https://developers.linear.app/docs/graphql/working-with-the-graphql-api#creating-and-editing-issues).",
-  version: "0.4.10",
+  version: "0.4.11",
   props: {
     linearApp,
     teamId: {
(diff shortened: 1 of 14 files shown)


ttrushin:
@andrewjschuang
This just caused all Node.JS code actions to start failing when trying to deploy or run... 
Can someone please address this right away? 

Heads up that Linear's SDK is WAY ahead of the version you pinned. Try using something solid and stable like the recent version: ^44.1.0

Screenshot 2025-08-07 at 13 44 55

andrewjschuang:
This was supposed to be a hot fix for another issue. Guess it's better to revert first and then address in a separate place. I'll do this right now, apologize for the inconvenience

ttrushin:
> This was supposed to be a hot fix for another issue. Guess it's better to revert first and then address in a separate place. I'll do this right now, apologize for the inconvenience

Thanks @andrewjschuang! We saw that other issue occasionally too. If you consider pinning to `^44.1.0` (which is recent and has been extremely stable for us), I think both issues would be solved.

andrewjschuang:
I wonder if that will work though, I think it doesn't support that syntax when you import the `@pipedream/linear` package. I realized reverting won't work because it issues another deployment, so I'll just have to fix it and use the latest linear version without pinning it

andrewjschuang:
Blocked on https://github.com/linear/linear/issues/830. Having issues publishing the component.

andrewjschuang:
@ttrushin I republished the Linear components, I had to use the latest version because of the pinning issues. Gladly the Linear team fixed it very quickly

PR #18038, opened 2025-08-13
Azure DevOps

michelle0927:


michelle0927:
@vunguyenhung 
I think the problem may be that Microsoft has deprecated Oauth authentication for DevOps (https://learn.microsoft.com/en-us/azure/devops/integrate/get-started/authentication/oauth?view=azure-devops#azure-devops-oauth-deprecated).

The docs here (https://learn.microsoft.com/en-us/rest/api/azure/devops/hooks/publishers/list-event-types?view=azure-devops-rest-5.0&tabs=HTTP) say to use a "personal access token".

I tested and created this PR which is a work in progress, but shows that using a personal access token does work to get event types, create webhooks, and delete webhooks. Strangely, the endpoint to get organizations still seems to need the oauth_access_token.

The first link I pasted above recommends migrating to Microsoft Entra ID OAuth. I think we should either do that or add a personal_access_token field to the $auth for Azure Devops.

vunguyenhung:
Hi @michelle0927, thank you for your information.

Here's the next steps:
1. I've created a new ticket here to add the new app for Azure DevOps using Microsoft Entra ID OAuth: https://github.com/PipedreamHQ/pipedream/issues/18106
2. I've added the personal access token prop to the app (named `personal_access_token`), you can use this property in your action code
image
3. In the future, we can shift the Azure DevOps app to use Personal Access Token only, so that user will have 2 options for authentication

PR #18119, opened 2025-08-19
Cleaning up TS sample files - part 2

GTFalcao:
Changing sample .ts apps to .mjs, removing their .gitignore and tsconfig.json, and updating their package.json to reflect this update.

There are more app files - I'm splitting the result into a few separate PRs to avoid publishing more than a hundred packages at once. This is part 2 of 4.


luancazarine (approved):

PR #18215, opened 2025-08-28
HubSpot - adding retry to search endpoint 

GTFalcao:
The issue is the limit of 5 requests per second for the search endpoint specifically. We theoretically already enforce this per component, but each user can be running multiple components simultaneously (many HubSpot actions and sources need to use the search endpoint).

On top of reducing the max requests per second per component from 4 to 2, I'm implementing retrying up to 5 times on a 429 response (specifically for the search endpoint). I've added a bit of exponential backoff, but total delay between requests only goes up to 17 seconds to avoid component runs taking too long.


vunguyenhung:

Test report
https://vunguyenhung.notion.site/HubSpot-adding-retry-to-search-endpoint-25dbf548bb5e811d81c6e99fb1afee4f

PR #18368, opened 2025-09-15
Notion property building improvements

michelle0927:
Resolves #18349 

## Summary by CodeRabbit

- New Features
  - When creating a page from a database, the new page is now explicitly parented to the selected data source.

- Bug Fixes
  - Clearer configuration errors when property values don’t match expected types during Notion property conversion.

- Chores
  - Version bumps across Notion components and actions (pages, databases, blocks, and file uploads); no behavior changes beyond the updates noted above.

PR #18381, opened 2025-09-16
Notion property building improvements

### LINKED ISSUE #18349: Notion property building improvements
In the Notion "Create Page from Data Source" action (`create-page-from-database` due to the recent migration to Data Sources), the `properties` prop seems like it's not always being properly parsed.

It would be best if we could validate the properties input, and if it doesn't follow the selected Data Source's property schema, throw a ConfigurationError explaining this (as the API errors returned don't seem very descriptive of this).

This is a follow-up issue to the Notion API migration #18212 , which was merged first because several live components were broken (due to the major update of the SDK)

michelle0927:
Resolves #18349 
Follow-up to PR #18368

### DIFF
--- a/components/notion/common/notion-page-properties.mjs
+++ b/components/notion/common/notion-page-properties.mjs
@@ -13,43 +13,68 @@ const NOTION_PAGE_PROPERTIES = {
     type: "string",
     example: "New Beauty Title",
     options: () => undefined,
-    convertToNotion: (property) => ({
-      title: utils.buildTextProperty(property.value),
-    }),
+    convertToNotion: (property) => {
+      if (typeof property.value !== "string") {
+        throw new Error;
+      }
+      return {
+        title: utils.buildTextProperty(property.value),
+      };
+    },
   },
   rich_text: {
     type: "string",
     example: "A beauty text value",
     options: () => undefined,
-    convertToNotion: (property) => ({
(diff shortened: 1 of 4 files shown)


components/notion/common/notion-page-properties.mjs
```diff
@@ -13,43 +13,68 @@ const NOTION_PAGE_PROPERTIES = {
     type: "string",
     example: "New Beauty Title",
     options: () => undefined,
-    convertToNotion: (property) => ({
-      title: utils.buildTextProperty(property.value),
-    }),
+    convertToNotion: (property) => {
+      if (typeof property.value !== "string") {
+        throw new Error;
```
GTFalcao, line 18: Should we add a message to these errors, in case they do get thrown?

GTFalcao (commented):
Thanks for the fix @michelle0927 , I left one comment regarding the errors with no message, but will move it to QA

components/notion/common/notion-page-properties.mjs
michelle0927: They're caught in `/common/base-page-builder.mjs` on line 111:

> throw new ConfigurationError(`Error converting property \`${property.name}\` to Notion format. Must be of type \`${NOTION_CONVERTER[property.type]?.type}\`.`);

GTFalcao (approved):

PR #18437, opened 2025-09-22
Jira - adding pagination to search-issues-with-jql

GTFalcao:
(reported by a customer)

Component did not have pagination, only max results, limiting the amount of responses that could be retrieved.

I've increased the max page size (1000, or less if `maxResults` is less) and added pagination, raising the max `maxResults` to 5000. While technically we can allow larger values, I think at this point the user should improve their query to narrow down the results, as the execution time can skyrocket.

## Summary by CodeRabbit

- New Features
  - JQL search now supports pagination and returns a consolidated issues list.
  - Increased maxResults limit from 100 to 5000 for larger queries.
  - Response shape updated to return { issues } for easier consumption.
  - Pagination helpers exposed to control page size and total results.

- Documentation
  - Clarified descriptions for JQL and maxResults, including updated limits and behavior.

- Chores
  - Version bumps: action API to 0.1.0 and Jira component package to 1.1.0.

### DIFF
--- a/components/jira/actions/search-issues-with-jql/search-issues-with-jql.mjs
+++ b/components/jira/actions/search-issues-with-jql/search-issues-with-jql.mjs
@@ -26,6 +26,7 @@ export default {
       optional: true,
       default: 50,
       min: 1,
+      max: 5000,
     },
     fields: {
       type: "string",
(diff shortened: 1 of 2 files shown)


michelle0927 (approved):
LGTM!

PR #18438, opened 2025-09-22
Adjusting isRelevant calls to always emit if ts is falsy

GTFalcao:
Some HubSpot sources currently ignore the "last timestamp" value if it is undefined (e.g. the first time the source runs).

This PR removes the `ts` check from the individual methods it was on, and instead expands that behavior to all sources, i.e. to the instances of calling the `isRelevant` method.

### DIFF
--- a/components/hubspot/package.json
+++ b/components/hubspot/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/hubspot",
-  "version": "1.7.6",
+  "version": "1.7.7",
   "description": "Pipedream Hubspot Components",
   "main": "hubspot.app.mjs",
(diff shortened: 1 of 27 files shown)


lcaresia (approved):

vunguyenhung:

https://vunguyenhung.notion.site/HubSpot-sources-fixing-ts-comparison-on-initial-run-276bf548bb5e8114b3b2e0f3a67b66de

GTFalcao:
I'm merging this earlier due to the majority of the sources having issues with event emissions currently. The 2 remaining issues reported should've been fixed also, but we can work it out on a subsequent PR if needed.

PR #18469, opened 2025-09-25
Hubspot New Deal in Stage - limit initial events

GTFalcao:
This limits the events emitted when `ts` is falsy, i.e. on initial event emission.


michelle0927 (dismissed):
LGTM!

PR #18476, opened 2025-09-26
17438 afosto

### LINKED ISSUE #17438: Afosto
### App
Afosto
https://developers.afosto.app/

### Sources
- order.shipped
- order.cancelled

### Actions
- Create shipment record → /orders/{order_id}/shipments
- Update shipment status → /orders/{order_id}/shipments/{shipment_id}
- Fetch invoices → /orders/{order_id}/invoices

luancazarine:
Resolves #17438

### DIFF
--- a/components/afosto/actions/add-information-to-cart/add-information-to-cart.mjs
+++ b/components/afosto/actions/add-information-to-cart/add-information-to-cart.mjs
@@ -1,0 +1,452 @@
+import app from "../../afosto.app.mjs";
+
+export default {
+  key: "afosto-add-information-to-cart",
+  name: "Add Information to Cart",
+  description: "Add customer information to a cart. [See the documentation](https://afosto.com/docs/developers/storefront-js-client/custom-checkout/collecting-customer-data/)",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    app,
+    cartId: {
+      propDefinition: [
+        app,
+        "cartId",
+      ],
+    },
+    email: {
+      type: "string",
+      label: "Email",
+      description: "The email of the customer",
+      optional: true,
+    },
+    isGuest: {
+      type: "boolean",
+      label: "Is Guest",
+      description: "Whether the customer is a guest",
+      optional: true,
+    },
+    givenName: {
+      type: "string",
+      label: "Given Name",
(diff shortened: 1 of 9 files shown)


components/afosto/actions/add-information-to-cart/add-information-to-cart.mjs
GTFalcao, line 1: This action is so large that GitHub even omitted it by default. Can we organize this code a little better?

I don't think we need all of these props either, this results in very confusing UX. Please limit the action to use the required props, and most relevant optional ones. Alternatively, since all the billing and shipping properties are strings, they could be two object props (billing info and shipping info). Links to the documentation can be added to an alert prop or to the object props' descriptions.

Also the GraphQL query should probably be kept in a separate file

components/afosto/actions/add-item-to-cart/add-item-to-cart.mjs
```diff
@@ -0,0 +1,78 @@
...
+        "cartId",
+      ],
+    },
+  },
+  async run({ $ }) {
+    try {
+      const response = await this.app.query({
+        $,
+        data: JSON.stringify({
+          query: `mutation AddItemToCart(
```
GTFalcao: Same here, please move the query to a separate file

components/afosto/actions/add-information-to-cart/add-information-to-cart.mjs
```diff
@@ -0,0 +1,450 @@
...
+        countryCode: billingAddressData?.country_code,
+        postalCode: billingAddressData?.postal_code,
+      };
+
+      const response = await this.app.query({
+        $,
+        maxBodyLength: Infinity,
+        headers: {
+          "DisablePreParseMultipartForm": "true",
+        },
```
GTFalcao: Just double checking, are these settings actually needed / mentioned by the API?

GTFalcao (commented):
Hey @luancazarine I think we need to improve the file and code organization a bit on this one.

Also not sure what the point of the try/catch blocks is, the thrown errors have different structures across components, wouldn't it be better to let our native error handling deal with the responses? Or are we intentionally formatting them differently here?

Issue #10912, opened 2024-03-13
[Components] taggun


pipedream-component-development:
# taggun

vunguyenhung:
Hi everyone, all test cases are passed! Ready for release! 

Test report
https://vunguyenhung.notion.site/Components-taggun-28dbf548bb5e81f192ffc09a0daf2751

PR #18508, opened 2025-09-30
[Components] taggun #10912

lcaresia:

### DIFF
--- a/components/taggun/actions/url-data-extraction/url-data-extraction.mjs
+++ b/components/taggun/actions/url-data-extraction/url-data-extraction.mjs
@@ -1,0 +1,64 @@
+import app from "../../taggun.app.mjs";
+
+export default {
+  key: "taggun-url-data-extraction",
+  name: "URL Data Extraction",
+  description: "Provide a URL for a receipt or invoice to extract clear and basic data. [See the documentation](https://developers.taggun.io/reference/url-simple)",
+  version: "0.0.1",
+  type: "action",
+  props: {
+    app,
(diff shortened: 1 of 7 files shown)


components/taggun/actions/url-data-extraction/url-data-extraction.mjs
```diff
@@ -0,0 +1,64 @@
+import app from "../../taggun.app.mjs";
+
+export default {
+  key: "taggun-url-data-extraction",
+  name: "URL Data Extraction",
```
GTFalcao: I would rename this to "Extract Data from URL" to better follow action naming conventions

components/taggun/actions/verbose-url-data-extraction/verbose-url-data-extraction.mjs
```diff
@@ -0,0 +1,64 @@
+import app from "../../taggun.app.mjs";
+
+export default {
+  key: "taggun-verbose-url-data-extraction",
+  name: "Verbose URL Data Extraction",
```
GTFalcao: Can't "verbose" be a boolean option in the "Extract Data from URL" component, therefore eliminating the need for this separate component, since it seems to have the same props?

lcaresia:
> Wasn't `extract-data-from-url` meant to replace `url-data-extraction` and `verbose-url-data-extraction`? They're both still present, but I think they should be removed.

You right, I forgot to remove, thanks for that! Fixed.

PR #18511, opened 2025-09-30
Adding 'annotations' to component typing for actions

GTFalcao:
## Summary by CodeRabbit

- New Features
  - Actions gain optional annotations to indicate destructive, idempotent, open‑world, or read‑only hints—improving clarity in UIs and integrations.
- Chores
  - Package version bumped to 0.3.3.

### DIFF
--- a/types/src/index.ts
+++ b/types/src/index.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/ban-types */
 /* eslint-disable @typescript-eslint/no-explicit-any */
 export type JSONValue =
   | string
@@ -70,7 +69,7 @@ export interface HTTPResponse {
   /**
    * Http Body
    */
-  body: string | Buffer | NodeJS.ReadableStream;
+  body: string | Buffer | ReadableStream;
   /**
    * If true, issue the response when the promise returned is resolved, otherwise issue
    * the response at the end of the workflow execution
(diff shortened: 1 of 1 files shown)


types/src/index.ts
```diff
@@ -384,6 +384,11 @@ export interface Action<
   type: "action";
   methods?: Methods & ThisType<PropThis<ActionPropDefinitions> & Methods>;
   props?: ActionPropDefinitions;
+  annotations?: {
```
dannyroosevelt, line 386: you should add `idempotentHint?` while you're at it, even though we aren't backfilling that value for any components right now

js07 (approved):
I might be doing something wrong, but I get some TypeScript errors when trying to run this package's tests (via `npm run test`).
```
./node_modules/@types/cacheable-request/index.d.ts:26:42 - error TS2709: Cannot use namespace 'ResponseLike' as a type.
        cb?: (response: ServerResponse | ResponseLike) => void
```
Since those don't seem to cause the build to fail, I think they shouldn't be blocking. LGTM!

Edit: Looks like the `pnpm publish` actually failed in [this check](https://github.com/PipedreamHQ/pipedream/actions/runs/18141851023/job/51634397693?pr=18511).

To fix this issue, I think you can tell TypeScript to use only type definitions in the current directory, rather than looking for type definitions up the directory tree (e.g. in `../node_modules/@types`). For example, in `types/tsconfig.json`, add the following to `compilerOptions`:

```json
    "typeRoots": ["./node_modules/@types"]
```

types/src/index.ts
```diff
@@ -70,7 +69,7 @@ export interface HTTPResponse {
   /**
    * Http Body
    */
-  body: string | Buffer | NodeJS.ReadableStream;
+  body: string | Buffer | ReadableStream;
```
js07, line 72: **-Non-blocking-**

I'm not sure if it matters, but I think the instances of `ReadableStream` in this file could instead be `Readable`. And `WritableStream` could be `Writeable`. For example:

```ts
import {
  Readable, Writable,
} from "stream";
// ...
body: string | Buffer | Readable;
// ...
export interface IFile {
  delete(): Promise<void>;
  createReadStream(): Promise<Readable>;
  createWriteStream(contentType?: string, contentLength?: number): Promise<Writable>;
  toEncodedString(encoding?: string, start?: number, end?: number): Promise<string>;
  toUrl(): Promise<string>;
  toFile(localFilePath: string): Promise<void>;
  toBuffer(): Promise<Buffer>;
  fromReadableStream(readableStream: Readable, contentType?: string, contentSize?: number): Promise<IFile>;
```

If you call [`File.fromReadableStream(readStream)`](https://pipedream.com/docs/workflows/data-management/file-stores/reference#file-fromreadablestream-%3Fcontenttype%2C-%3Fcontentlength) with a stream derived from [`fs.createReadStream()`](https://nodejs.org/api/fs.html#fscreatereadstreampath-options), for example, then the `readStream` would have the type [`fs.ReadStream`](https://nodejs.org/api/fs.html#class-fsreadstream), which extends [`stream.Readable`.](https://nodejs.org/api/stream.html#class-streamreadable)

GTFalcao: Good catch - I'll follow up on this. Shipping the current change for now to unblock building components

PR #18520, opened 2025-09-30
Upgrade to the v2 SDK

jverce:
# Description
Use the new v2 version of the Pipedream SDK in the `connect-react` package. This change involves migrating to the new types mostly, but also runtime changes involving the API calls.

The runtime behaviour should not be affected from a user's perspective, except for consumers of the `connect-react` package itself, since some components (e.g. `FrontendClientProvider`) expect their consumers to inject a client instance of the same SDK version. For this reason, this change bumps the **major** version of this package.

### DIFF
--- a/packages/connect-react/src/components/Field.tsx
+++ b/packages/connect-react/src/components/Field.tsx
@@ -43,7 +43,7 @@ export function Field<T extends ConfigurableProp>(props: FieldProps<T>) {
   const app = "app" in field.extra
     ? field.extra.app
     : undefined;
-  if (app && !app.auth_type) {
+  if (app && !app.authType) {
     return null;
   }
 
@@ -59,7 +59,7 @@ export function Field<T extends ConfigurableProp>(props: FieldProps<T>) {
   // XXX use similar pattern as app below for boolean and checkboxing DOM re-ordering?
 
   return (
-    <div {...getProps("field", baseStyles, props as FieldProps<ConfigurableProp>)}>
+    <div {...getProps("field", baseStyles, props)}>
       <Label text={labelText} field={field} form={form} />
       <Control field={field} form={form} />
       <Description markdown={prop.description} field={field} form={form} />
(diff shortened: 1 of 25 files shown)


packages/connect-react/src/components/Field.tsx
```diff
@@ -43,7 +43,7 @@ export function Field<T extends ConfigurableProp>(props: FieldProps<T>) {
   const app = "app" in field.extra
     ? field.extra.app
     : undefined;
-  if (app && !app.auth_type) {
+  if (app && !app.authType) {
     return null;
   }
```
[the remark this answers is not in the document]
jverce: Not using the local package anymore

packages/connect-react/src/hooks/form-context.tsx
```diff
@@ -220,26 +239,27 @@ export const FormContextProvider = <T extends ConfigurableProps>({
 
   // XXX fix types of dynamicProps, props.component so this type decl not needed
   const configurableProps = useMemo(() => {
-    let props = dynamicProps?.configurableProps || formProps.component.configurable_props || [];
+    let props: unknown = dynamicProps?.configurableProps || formProps.component.configurableProps || [];
```
tjk: why can this not get typed correctly?

packages/connect-react/src/hooks/form-context.tsx
```diff
@@ -252,11 +272,11 @@ export const FormContextProvider = <T extends ConfigurableProps>({
     if (prop.type === "app") {
```
tjk: Can you not use the `isConfigurablePropApp` thing here?

https://github.com/PipedreamHQ/pipedream/pull/18520/files#diff-c18032975be0d931201cfcd24805903f57cbe5f01b7d2909614d672b7a776ca0R15-R17

jverce: Fixed here: [`b4fb4f2` (#18520)](https://github.com/PipedreamHQ/pipedream/pull/18520/commits/b4fb4f29ca26e2940965b07f1b3b1b71ca595144)

packages/connect-react/src/hooks/form-context.tsx
jverce: Fixed here: [`0fcde9b` (#18520)](https://github.com/PipedreamHQ/pipedream/pull/18520/commits/0fcde9bd60610ecd2beaa545a8372fd29800aedd)

PR #18609, opened 2025-10-02
MCP annotations follow-up

GTFalcao:
Republishing components that had publish errors

## Summary by CodeRabbit

- Chores
  - Standardized version updates across multiple integrations (CodeScene, Discogs, Freshmarketer, Lokalise, PostHog, Pro Ledger, Transloadit) to improve consistency. No functional changes.
- Dependencies
  - Upgraded Transloadit SDK to v4 for improved compatibility and long-term support.
- Refactor
  - Updated Transloadit import usage to align with the new SDK version.
- Notes
  - No changes to user-facing behavior or workflows; these updates focus on stability, consistency, and compatibility.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1917,8 +1917,7 @@ importers:
 
   components/braintree: {}
 
-  components/brandfetch:
-    specifiers: {}
+  components/brandfetch: {}
 
   components/brandmentions: {}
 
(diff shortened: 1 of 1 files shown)


michelle0927 (commented):
Hi @GTFalcao. I think you need to update some `package.json` versions per the Component Checks.

js07 (approved):
Thanks for fixing these components!

GTFalcao:
Thanks for noting that @michelle0927 , this was intentional: since we only updated the MCP annotations, we chose not to republish the packages, only the components; given that the MCP annotations aren't really needed in a scenario where the package is imported, and this would slow down publishing a bit.

I only updated the "transloadit" app package because that one was actually broken due to a new major version of their SDK.

PR #18666, opened 2025-10-07
Types: replacing ReadableStream and WritableStream

GTFalcao:
Follow-up to #18511 

Note: I pointed all packages that import types to a symlink to build them with these changes applied. The build was successful.

## Summary by CodeRabbit

- Refactor
  - Updated public type definitions to use Node.js stream types for HTTP bodies and file streaming operations, standardizing read/write stream behavior across the API.
- Chores
  - Bumped package version to 0.3.4.

### DIFF
--- a/types/src/index.ts
+++ b/types/src/index.ts
@@ -1,4 +1,6 @@
-import { Readable, Writable } from "stream";
+import {
+  Readable, Writable,
+} from "stream";
 
 /* eslint-disable @typescript-eslint/no-explicit-any */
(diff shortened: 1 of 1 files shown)


types/src/index.ts
```diff
@@ -1,3 +1,5 @@
+import { Readable, Writable } from "stream";
```
js07: To fix the linter error on this line:

```suggestion
import {
  Readable, Writable,
} from "stream";
```

PR #18667, opened 2025-10-07
Applying MCP annotations to TS components - batch 1/2

GTFalcao:


components/clientary/actions/create-invoice/create-invoice.ts
```diff
@@ -44,7 +49,9 @@ export default defineAction({
       try {
         invoiceItemsAttributes = JSON.parse(this.invoiceItemsAttributes);
       } catch (err) {
-        throw new ConfigurationError("`Estimate Items Attributes` must be a valid JSON Array string");
+        throw new ConfigurationError(`\`Estimate Items Attributes\` must be a valid JSON Array string
+          
+${err.message}`);
```
js07: **Nit(optional,non-blocking)**: We could omit `(err)` here, instead, since `err.message` seems like it should be redundant with the static error message.

```ts
      } catch {
        throw new ConfigurationError("`Estimate Items Attributes` must be a valid JSON Array string");
```

(Also applies to `create-estimate`)

js07 (dismissed):
LGTM. Just made one minor optional suggestion.

js07 (approved):
LGTM!

PR #18740, opened 2025-10-13
Gmail: fixing address parsing in replyAll option

GTFalcao:
When using the `replyAll` option in the Gmail "Send Email" action, it would split the "from" and "to" values by comma in an attempt to retrieve multiple comma-separated emails. This caused a problem when the email itself included a comma in the name (e.g. `"Falcao, Guilherme" <guilherme@nowhere.com>`

Using addressparser from nodemailer ensures proper parsing of these fields

Note: I only bumped the version of the `send-email` component, because it is the only component that uses the `replyAll` prop, and the logic was only changed in the block that checks for this prop.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved parsing and normalization of email addresses for To, CC, and BCC fields, reducing errors with complex or combined addresses.
  * More accurate address handling when composing replies and Reply All, ensuring correct recipients are included.

* **Chores**
  * Bumped Gmail integration and send-email action versions for a patch release.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -2044,8 +2044,7 @@ importers:
         specifier: ^3.0.3
         version: 3.0.3
 
-  components/browserflow:
-    specifiers: {}
+  components/browserflow: {}
 
   components/browserhub:
     dependencies:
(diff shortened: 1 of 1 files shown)


vunguyenhung:

Test report
https://vunguyenhung.notion.site/Gmail-fixing-address-parsing-in-replyAll-option-28cbf548bb5e819e960ee39b1f82b279

PR #18822, opened 2025-10-21
Outlook Calendar - use listCalendarView

### LINKED ISSUE #18789: Outlook Calendar: use calendarView instead of listEvents
We currently use [List Events](https://learn.microsoft.com/en-us/graph/api/user-list-events?view=graph-rest-1.0&tabs=http) to obtain events in the Microsoft Outlook Calendar app.

It seems that this does not return individual instances of recurring events, while [List calendarView](https://learn.microsoft.com/en-us/graph/api/user-list-calendarview?view=graph-rest-1.0&tabs=http) would return them.

Please investigate and confirm if this is the case, and if so, we should add the option to list recurring events. This should be added in a backwards-compatible way, such as a boolean prop that defaults to false.

Additional context:
https://stackoverflow.com/questions/75737172/missing-events-when-querying-from-calendar-api

michelle0927:
Resolves #18789 

## Summary by CodeRabbit

* **New Features**
  * Option to include recurring events when listing events, with optional start/end date-time to query a calendar view.

* **Chores**
  * Bumped component and action versions across the calendar integration.
  * Added/updated package dependencies and reorganized shared source wiring to centralize Microsoft Outlook prop and imports.


luancazarine (approved):

PR #18824, opened 2025-10-21
HubSpot: add useQuery to object props

### LINKED ISSUE #18790: HubSpot: add useQuery support for searching objects
HubSpot has many object types. When listing records in `async options`, we currently use the `createOptions` method that in turn calls `listObjectsInPage`. This paginates through, potentially, thousands of records.

We could add the `useQuery` property to all props that currently list records, so that when a query is entered, they *search* for records instead, using the `searchCRM` method. [According to the documentation](https://developers.hubspot.com/docs/api-reference/crm-objects-v3/search/post-crm-v3-objects-objectType-search) this method supports a `query` parameter, which may be appropriate for doing the search - please evaluate if this is the case. The alternative is to use `filters` with a specific property for each object type.

michelle0927:
Resolves #18790 

## Release Notes

* **New Features**
  * Enhanced search capabilities for object lookups in HubSpot actions, enabling query-based filtering for contacts, leads, and object selections.

* **Improvements**
  * Updated multiple HubSpot actions and sources with performance and stability improvements across the integration.

### DIFF
--- a/components/hubspot/actions/add-contact-to-list/add-contact-to-list.mjs
+++ b/components/hubspot/actions/add-contact-to-list/add-contact-to-list.mjs
@@ -5,7 +5,7 @@ export default {
   name: "Add Contact to List",
   description:
     "Adds a contact to a specific static list. [See the documentation](https://legacydocs.hubspot.com/docs/methods/lists/add_contact_to_list)",
-  version: "0.0.26",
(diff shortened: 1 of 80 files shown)


jcortes (approved):
Hi @michelle0927 lgtm! Ready for QA!

PR #18857, opened 2025-10-24
[FIX] Front New Conversation Tag rate limits

### LINKED ISSUE #18834: [BUG] Front "New Conversation Tag" rate limits
May not be an actual bug, but we need to investigate the rate limits on Front. This source is set to run every 1min40sec and is hitting rate limits on every run, though they state to retry after 5-10 seconds in most cases:

Image

We should check if the source is making more requests than it has to, if there's any way to optimize this, and if we can add retrying with exponential backoff on 429 responses.

Additionally, the customer reported the source was not emitting events as intended even when it was not being rate limited (on a 15 min timer) - please investigate.

jcortes:
## WHY

Resolves #18834

## Summary by CodeRabbit

* **New Features**
  * Automatic request retry with configurable rate-limit handling and error recovery.
  * Pagination delay option to control pacing of paginated requests.

* **Bug Fixes**
  * Conversation summary now falls back to an identifier when subject is missing.

* **Chores**
  * Added retry dependency and bumped component, action, and source versions.

### DIFF
--- a/components/frontapp/package.json
+++ b/components/frontapp/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/frontapp",
-  "version": "0.8.1",
+  "version": "0.8.3",
   "description": "Pipedream Frontapp Components",
   "main": "frontapp.app.mjs",
   "keywords": [


components/frontapp/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/frontapp",
-  "version": "0.8.2",
+  "version": "0.8.1",
```
michelle0927:
```suggestion
  "version": "0.8.3",
```

michelle0927 (approved):
LGTM! Ready for QA

PR #18888, opened 2025-10-28
18869 feature   upgrade notion create page in data source action to support templates

### LINKED ISSUE #18869: [FEATURE] - Upgrade Notion "Create Page in Data Source" Action to Support Templates
The Notion API now supports applying page templates when creating pages within data sources: https://developers.notion.com/docs/creating-pages-from-templates

It'd be fantastic if Pipedream users could use this ability and apply templates when creating new pages within the [Create Page from Data Source](https://github.com/PipedreamHQ/pipedream/tree/master/components/notion/actions/create-page-from-database) action.

This action already uses my [notion-helper library](https://github.com/TomFrankly/notion-helper), and I have just updated the library to provide robust methods for handling template application. I've also written a [usage guide for working with templates](https://github.com/TomFrankly/notion-helper/blob/main/guides/Template%20Usage%20Guide.md).

Ideally, the action would be updated to surface a Template property if the chosen data source has any templates. Additionally, if the data source has a **default** template, the property would surface a "Default" option.

(Note how my guide provides an example for checking for the presence of a default template before creating a page. It's always a good idea to check before each page-creation request, as a data source might no longer have a default template set in the future).

Once a user has selected a template, the updated action can use a `verifyTemplateReady()` callback in `createPage()` to verify that a page has fully applied the template before applying any child blocks.

luancazarine:
Resolves #18869

## Release Notes

* **New Features**
  * Template selection options for creating database pages: create empty pages, use default templates, or specify a custom template ID.
  * Added ability to retrieve available templates.

* **Updates**
  * Version increments across multiple Notion actions and sources for stability and consistency.
  * Updated Notion client library dependency to latest compatible version.

### DIFF
--- a/components/notion/actions/create-page-from-database/create-page-from-database.mjs
+++ b/components/notion/actions/create-page-from-database/create-page-from-database.mjs
@@ -115,12 +115,15 @@ export default {
           label: "Template ID",
           description: "The ID of the template to use for the page. [See the documentation](https://developers.notion.com/docs/creating-pages-from-templates) for more information.",
           options: async({ prevContext }) => {
+            const params = {
+              data_source_id: this.parentDataSource,
+            };
+            if (prevContext?.nCursor) {
+              params.start_cursor = prevContext.nCursor;
+            }
             const {
               templates, next_cursor: nCursor,
-            } = await this.notion.listTemplates({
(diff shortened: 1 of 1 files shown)


michelle0927 (approved):
LGTM! Ready for QA.

PR #19726, opened 2026-01-17
Improving TypeScript build

dylburger:
## WHY

Speed up and simplify TypeScript build

## Summary by CodeRabbit

* **Refactor**
  * Migrated components to a new ES module build pipeline and simplified TypeScript configuration across packages.

* **Chores**
  * Streamlined CI and local hooks: unified package manager setup, upgraded Node to v20, enforced frozen lockfile installs, and removed legacy package store caching.
  * Reworked build/publish scripts and removed unused dev tooling.
  * Consolidated .gitignore entries and publish hooks.

* **Maintenance**
  * Bumped one RSS source metadata version.


components/waitwhile/package.json
```diff
@@ -16,9 +16,6 @@
   "publishConfig": {
     "access": "public"
   },
-  "scripts": {
```
dylburger, line 19: Randomly, this was the only `package.json` file that had a `build` script (not necessary in the old or new system), so I removed it.

.gitignore
```diff
@@ -1,24 +1,28 @@
-*.md.orig.*
-*.md.toc.*
-gh-md-toc
+# general
```
dylburger, line 1: the diff looks odd, just look at the latest file — I'm just grouping these rules into sections

dylburger:
Reopening to trigger CI

.github/workflows/components-pr.yaml
```diff
@@ -49,28 +49,20 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
-      - uses: pnpm/action-setup@v4.2.0
+
+      - name: Setup pnpm
+        uses: pnpm/action-setup@v4.2.0
         with:
           version: 9.14.2
-      - name: Get pnpm store directory
```
dylburger, line 55: The **Setup Node.js** step (using `actions/setup-node`) has built-in caching with `cache: 'pnpm'`. Since that action added the `cache` option ([see docs](https://github.com/actions/setup-node)), we no longer needed the manual caching step here, and were duplicating work and delaying jobs. 

I removed that manual caching everywhere and confirmed we should be getting cache hits at relevant steps.

.github/workflows/components-pr.yaml
```diff
@@ -49,28 +49,20 @@ jobs:
...
         with:
-          node-version: 18
-          registry-url: https://registry.npmjs.org/
+          node-version: 20
           cache: 'pnpm'
-      - name: Install Dependencies
-        run: pnpm install -r
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
dylburger, line 65: Two changes here and in the rest of the GitHub actions that run `pnpm install`:

- The `-r` flag is no longer available in `pnpm install` ([see v9 docs](https://pnpm.io/9.x/cli/install#tldr))
- This always should have used `--frozen-lockfile`, since that catches cases where the lockfile wasn't updated in the commit, and makes builds reproducible since the installs must match the lockfile.

.github/workflows/pipedream-sdk-test.yaml
```diff
@@ -18,28 +18,19 @@ jobs:
...
-          key: ${{ runner.os }}-pnpm-store-${{ hashFiles('**/pnpm-lock.yaml') }}
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
 
-      - name: Set up Node.js
+      - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: '22'
+          node-version: 22
```
dylburger, line 29: We were already on 22 here and it's working so I'm not changing this to 20 like in the rest of the workflows

.github/workflows/publish-marketplace-content.yaml
```diff
@@ -12,28 +12,20 @@ jobs:
...
+      - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: 18
-          registry-url: https://registry.npmjs.org/
+          node-version: 20
           cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
```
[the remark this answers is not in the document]
dylburger: The first comment is incorrect — v6 correctly implements this cache with `cache: 'pnpm'`

.github/workflows/pull-request-checks.yaml
```diff
@@ -9,6 +9,7 @@ on:
   pull_request:
     branches:
       - master
+  workflow_dispatch:  # Allows manual triggering from Actions tab
```
dylburger, line 12: Added this for testing

components/rss/sources/new-item-in-feed/new-item-in-feed.ts
```diff
@@ -7,7 +7,7 @@ export default defineSource({
   key: "rss-new-item-in-feed",
   name: "New Item in Feed",
   description: "Emit new items from an RSS feed",
-  version: "1.2.9",
+  version: "1.2.10",
```
dylburger, line 10: I've manually tested components, but I revved this version to test this new build on this RSS source e2e

packages/connect-react/package.json
```diff
@@ -21,7 +21,7 @@
     "build": "vite build",
     "lint": "eslint src",
     "lint:fix": "eslint src --fix",
-    "prepare": "pnpm run build",
```
dylburger, line 24: This would run automatically on the husky `pre-push` hook before, when it wasn't needed. This ensures we only run the build before publishing the package.

.github/workflows/components-pr.yaml
GTFalcao: Does `pnpm install` already handle recursively installing dependencies for all app packages now, or are we handling that in another way?

.github/workflows/components-pr.yaml
```diff
@@ -49,28 +49,20 @@ jobs:
...
-          restore-keys: |
-            ${{ runner.os }}-pnpm-store-
-      - name: Setup Node Env
+
+      - name: Setup Node.js
         uses: actions/setup-node@v6
         with:
-          node-version: 18
-          registry-url: https://registry.npmjs.org/
+          node-version: 20
```
dylburger, line 61: Node 18 was EOL'd on April 30, 2025, and we're on Node 20 locally in the repo, so I updated this. I don't expect any issues (and all my tests ran fine), but we should watch for any Node upgrade issues

.github/workflows/pull-request-checks.yaml
```diff
@@ -50,37 +51,26 @@ jobs:
...
-          # so we can check the git diff against all changed files.
-          # By default, the checkout action only returns the last commit,
-          # There's no native way to do this in the checkout action, so
-          # we have to fetch the entire history. See
-          # https://github.com/actions/checkout/issues/266#issuecomment-638346893
-          fetch-depth: 0
-      - uses: pnpm/action-setup@v4.2.0
+          # Fetch enough history to find merge-base with master
+          # 300 commits handles most long-running feature branches
+          fetch-depth: 300
```
GTFalcao, line 56: This is a good optimization, but 300 commits sounds reachable for a branch that runs for longer than usual and merges in commits from master - which would bump up the commit count here, wouldn't it?

Probably a rare occurrence though, I guess we can raise this later if it happens.

.husky/post-merge
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 . "$(dirname "$0")/_/husky.sh"
 
-npx pnpm install -r
+pnpm install --ignore-scripts
```
GTFalcao, line 4: Is there a drawback to using `npx` here to warrant removing it? I think this would force the user to have pnpm installed globally and their global version might differ from the one we specify in the repo

PR #19861, opened 2026-01-27
Update Slack v2 component version and enhance List Channels action

### LINKED ISSUE #19851: [BUG] Slack "List Channels" unexpected behavior
The Slack "List Channels" action has a `channelTypes` prop, matching the `conversations.list` `types` parameter [documented here](https://docs.slack.dev/reference/methods/conversations.list/)

However, when using the "all" option (`public_channel,private_channel`) it seems that less channels are returned, compared to using only the `public_channel` option.

luancazarine:
Resolves #19851

## Summary by CodeRabbit

* **New Features**
  * Channel type selection now supports choosing multiple channel types.

* **Improvements**
  * When no channel types are specified, all channel types are included by default.
  * Channel type description simplified for clarity.

* **Chores**
  * Version metadata updated across Slack v2 actions, sources, and package.

### DIFF
--- a/components/slack_v2/actions/add-emoji-reaction/add-emoji-reaction.mjs
+++ b/components/slack_v2/actions/add-emoji-reaction/add-emoji-reaction.mjs
@@ -4,7 +4,7 @@ export default {
   key: "slack_v2-add-emoji-reaction",
   name: "Add Emoji Reaction",
   description: "Add an emoji reaction to a message. [See the documentation](https://api.slack.com/methods/reactions.add)",
-  version: "0.0.17",
+  version: "0.0.18",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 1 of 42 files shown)


luancazarine:
> This changes the `channelTypes` prop from string to array, but doesn't seem to change the result of what is being passed to the Slack API. Were you able to reproduce the issue and confirm that this change addresses it?

Hi @GTFalcao, Actually, I added more channel type options, and when the client doesn't select any, it's as if they selected all of them. my tests worked perfectly.
Please, if you can, do some tests there as well, and if the problem persists, we need to investigate the channel settings.

GTFalcao:
I'll close this for these reasons:
- I tried to reproduce the original issue with a more robust setup (145 test channels, mixed public and private), and still couldn't repro it, the results were as expected;
- double checked Slack docs again to ensure we're passing params as expected;
- we had initially discussed adding the 1:1 DM and Group DMs as options, but decided against it as these are "channels" in the Slack API, but this is not intuitive for users at all;
- the client was able to find a workaround: running the action with "public", then running it "private", and combining the results, did the trick for them.

PR #19907, opened 2026-01-30
[Platform] Adding filename parameter support to base64 strings

GTFalcao:
This supports base64 data URLs with a "name" or "filename" parameter, which is sometimes used by users. This sets the filename metadata, defaulting to "file.{ext}" if it is not present.

## Summary by CodeRabbit

* **New Features**
  * Added support for extracting and preserving file names from data URLs with explicit name parameters. File names are now properly propagated to metadata when available.


michelle0927 (approved):
LGTM!

GTFalcao:
Merging this, tested locally with various types of base64 string inputs against this same function.

PR #20058, opened 2026-02-18
[Bug] google_calendar-update-following-instances duplicates the first instance when rescheduling a recurring event

### LINKED ISSUE #20056: [Bug]  google_calendar-update-following-instances duplicates the first instance when rescheduling a recurring event
Google calendar

When using the tool to shift a recurring event's time (e.g., 10am → 8am starting from a specific instance), the original instance on the first affected date is not removed — it remains on the calendar alongside the newly created 8am event. All subsequent instances are correctly updated to 8am.

**Unexpected Result:**
 On the first Monday, the user sees two events (8am and 10am). From the following Monday onward, only the 8am event exists.

**Likely issue:** 
The tool trims the original recurrence with `UNTIL=20260216T085959Z` (i.e., up to but not including the new start time of 08:00), but the original series had an instance at 09:00 UTC (10am Stockholm) on that same date. This falls after the `UNTIL` cutoff and should be excluded, but still appears on the calendar. The `UNTIL` cutoff may be calculated incorrectly, or Google Calendar is not properly suppressing the orphaned original instance. I think we can add some logic to delete the original instance after the split to cover our bases.

jcortes:
## WHY

Resolves #20056

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved update flow for following instances of recurring Google Calendar events: deletion now occurs before recurrence trimming, with added error handling that attempts to restore deleted instances and surfaces clear partial-failure messages when operations partially succeed.

* **Chores**
  * Bumped component and action versions to reflect this update.

### DIFF
--- a/components/google_calendar/actions/update-following-instances/update-following-instances.mjs
+++ b/components/google_calendar/actions/update-following-instances/update-following-instances.mjs
@@ -6,7 +6,7 @@ export default {
   key: "google_calendar-update-following-instances",
   name: "Update Following Event Instances",
   description: "Update all instances of a recurring event following a specific instance. This creates a new recurring event starting from the selected instance. [See the documentation](https://developers.google.com/calendar/api/guides/recurringevents#modifying_all_following_instances)",
-  version: "0.0.3",
+  version: "0.0.4",
   type: "action",
   annotations: {
     destructiveHint: true,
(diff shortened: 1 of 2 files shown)


pipedream-component-development:
- [Bug]  google_calendar-update-following-instances duplicates the first instance when rescheduling a recurring event - Actions Batch 1: https://www.notion.so/Bug-google_calendar-update-following-instances-duplicates-the-first-instance-when-rescheduling-a-r-30beb261c847817d8349c8c08935148f

PR #20128, opened 2026-02-26
[Platform] Cleaning up AxiosError and debug export

### LINKED ISSUE #20120: [Platform] Clean up AxiosError thrown by platform axios
We should look into the AxiosError thrown by the `@pipedream/platform` axios module, as it currently exports a lot of information that isn't really helpful for debugging, and clutters Connect responses more than it helps.

We should definitely be keeping the response body, status and headers, but should not be logging the `request` object, for example.

GTFalcao:
Closes #20120 

This removes `config`, `request` and `stack` from the AxiosError before re-throwing it.

Also ensures `response`, included in the AxiosError and also exported as the debug object, has only the response info relevant to the user: `status`, `statusText`, `headers` and `data`

How this can be tested:
- build an .mjs copy of the axios module, e.g. `esbuild lib/axios.ts --bundle --format=esm --platform=node --external:axios --external:querystring --outfile=dist/axios.mjs`;
- copy this into an app folder and import it instead of `@pipedream/platform`, e.g. in `components\trello\trello.app.mjs`, replace `import { axios } from "@pipedream/platform"` with `import axios from "./common/axios.mjs"`
- publish the component and run it (in the builder and/or via Connect) using input that is sure to fail (e.g. a random word instead of an entity ID; I used `components\trello\actions\get-cards-on-board`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved API error handling to sanitize responses, preventing leakage of sensitive data and ensuring a consistent sanitized error is propagated.
  * Debug output now receives a concise, sanitized response summary instead of raw error payloads.

* **Chores**
  * Bumped package version to 3.3.0.

### DIFF
--- a/platform/lib/axios.ts
+++ b/platform/lib/axios.ts
@@ -189,7 +189,7 @@ function convertAxiosError(err: AxiosError): AxiosResponseSummary | undefined {
     };
     delete err.response.request;
     delete (err.response as Partial<typeof err.response>).config;
-    err.name = `${err.message} - see 'debug' export for the full response`;
+    err.name = `${err.name} - ${err.message}`;
     try {
       err.message = JSON.stringify(err.response.data);
     }
--- a/platform/dist/axios.js
+++ b/platform/dist/axios.js
@@ -152,7 +152,7 @@ function convertAxiosError(err) {
         };
         delete err.response.request;
         delete err.response.config;
-        err.name = `${err.message} - see 'debug' export for the full response`;
+        err.name = `${err.name} - ${err.message}`;
         try {
             err.message = JSON.stringify(err.response.data);
         }


platform/lib/axios.ts
```diff
@@ -164,19 +174,39 @@ function stepExport(step: PipedreamStep | undefined, message: unknown, key: stri
...
       console.error("Error trying to convert `err.response.data` to string");
       console.log(error);
     }
   }
-  return err;
+
+  const clean = new Error(message);
+  clean.name = name;
+  return {
+    error: clean,
```
js07: Some components depend on `error.response` or `error.status` being present, e.g. https://github.com/PipedreamHQ/pipedream/blob/d4763f325e4f44baa25a7e0364b8eee1c13303a3/components/calendly_v2/calendly_v2.app.mjs#L240

What do you think about just deleting `request`, `config`, `response.request`, and maybe `stack` from the error object and returning it? For example:

```ts
  if (err.response) {
    delete err.response.request;
  }
  delete err.config;
  delete err.request;

  return err;
```

js07 (approved):
LGTM. Thanks for making these changes!

PR #20152, opened 2026-03-02
Aircall - List Calls, List Contacts

### LINKED ISSUE #20042: [BUG] Aircall MCP - Get Calls returns stale/cached data, ignores query parameters
Describe the bug
The Aircall MCP integration's "Get Call" action, when used to list calls (GET /v1/calls), returns stale cached data instead of current data. It ignores all query parameters including date filters (from, to), user_id, per_page, and order. The same 20 calls are returned every time regardless of what's requested.

To Reproduce
1. Connect Aircall MCP to Claude.ai via Pipedream
2. Ask Claude to list today's calls using the Aircall get-call action
3. Request specific date filtering (e.g. from/to UNIX timestamps for today)
4. Observe that the same stale calls are returned regardless of parameters

Expected behavior
The GET /v1/calls endpoint should return current call data and respect Aircall's supported query parameters: from, to, user_id, direction, per_page, and order.

Additional context
- Fetching a specific call by numeric ID works correctly and returns accurate, current data
- The dashboard shows current calls from multiple agents, but the MCP only returns old calls from months ago
- The integration is also missing key endpoints: GET /v1/users, GET /v1/calls/search
- Authentication is not the issue — individual call lookups work fine across all numbers

### LINKED ISSUE #20078: Aircall
### App
Aircall

### Summary:
Aircall MCP can only get 1st 20 records and cannot process after that

### Details:
What’s happening
1) Aircall defaults to 20 per page

Aircall’s API pagination docs explicitly say per_page default is 20 (min 1, max 50).

So any request like GET /v1/calls (or /contacts, etc.) without per_page will return 20 items.

2) The Pipedream Aircall component just passes params through (no pagination)

In components/aircall/aircall.app.mjs, listCalls is implemented as a thin wrapper:

listCalls(params = {}, $) { return this._makeRequest({ path: "calls", params, $ }); }

So if the MCP tool is calling listCalls() with no params, you’ll get Aircall’s first page only → 20 items.

Same story for contacts: listContacts() calls /contacts with only order: "desc" and no pagination params.

### Screenshots:
No screenshots included

michelle0927:
Resolves #20078 
Resolves #20042 

## Summary by CodeRabbit

* **New Features**
  * Added "List Calls" action with pagination, date-range filters, and optional related data.
  * Added "List Contacts" action with pagination, date-range filters, sorting, and max-results control.

* **Refactor**
  * Improved pagination and parameter handling across Aircall integrations.
  * Added date parsing for date filters.
  * Expanded contact search/list options (from, to, order, max results).

* **Chores**
  * Bumped component and package versions.

### DIFF
--- a/components/aircall/actions/list-contacts/list-contacts.mjs
+++ b/components/aircall/actions/list-contacts/list-contacts.mjs
@@ -52,7 +52,6 @@ export default {
     const results = this.aircall.paginate({
       fn: this.aircall.listContacts,
       args: {
-        debug: true,
         $,
         params: {
           from: parseDate(this.from),


pipedream-component-development:
### For Integration QA:
- Source Test Report: https://www.notion.so/Aircall-Sources-318eb261c84781848252e5581105421d

PR #20170, opened 2026-03-04
ServiceNow - "List Tables" action

### LINKED ISSUE #20164: [ACTION] ServiceNow- Add "List Tables" action to enable table discovery
### **Is there a specific app this action is for?**
ServiceNow

### **Please provide a link to the relevant API docs for the specific service / operation:**
  * [Table API](https://www.servicenow.com/docs/bundle/zurich-api-reference/page/integrate/inbound-rest/concept/c_TableAPI.html) - used to query sys_db_object for all table definitions
* [Aggregate API](https://www.servicenow.com/docs/bundle/zurich-api-reference/page/integrate/inbound-rest/concept/c_AggregateAPI.html) - used to probe table access via /api/now/stats/{table}

### **Proposed Action**
A new `list-accessible-tables` action that returns only the tables the authenticated user can query. It
  works in two steps: first, fetch all table names from `sys_db_object` via the existing `getTableRecords` method; then
  probe each table using the existing `getRecordCountsByField` method (Stats API — `/api/now/stats/{table}`), which
  returns only a count with no record data. Tables returning 200 are included in the result; tables returning `403` are
  filtered out. Probes run in parallel via `Promise.allSettled` to avoid sequential blocking.

  This approach is specifically designed for non-admin users — it tests access empirically using whatever roles the
  authenticated user already has, rather than reading privileged system tables like `sys_security_acl` which requires
  `security_admin`.

### **Why:**
The ServiceNow integration currently has no way for users or AI agents to discover what tables are available in their instance. Every existing action requires the user to already know the table name. 

This is a significant gap because:
  * Table names vary between instances depending on which ServiceNow products are licensed and what custom tables have been configured
  * AI agents using the ServiceNow integration have no discovery mechanism and cannot determine which tables to
  interact with from the available components alone
  * The existing `table` prop uses a search-based dropdown (`useQuery: true`) that requires the user to type a search term of length > 1, so it cannot be used to browse all available tables

Key things to note:
  * No new endpoint required - uses the existing `getTableRecords` method already implemented in [servicenow.app.mjs](https://github.com/PipedreamHQ/pipedream/blob/master/components/servicenow/servicenow.app.mjs),
  targeting `sys_db_object`
  * No new OAuth scopes required - `sys_db_object `is accessible with the same credentials used by all other ServiceNow
  actions
  * Useful fields to return: `name`, `label`, `super_class`, `is_extendable`, `access`, `create_access`, `delete_access`, and `sys_id `
  * Should support a limit param  (recommended cap ~200) - `sys_db_object` can contain thousands of rows on mature instances

michelle0927:
Resolves #20164 

## Summary by CodeRabbit

* **New Features**
  * Added a ServiceNow "List Tables" action that retrieves accessible tables, supports an explicit query or filters for creation date, update date and active status, validates accessibility before returning results, includes a configurable result limit, and returns a summary count of accessible tables.
* **Chores**
  * Component package version bumped to 0.9.0.


michelle0927:
> Hi @michelle0927 lgtm! This is ready for release!

@jcortes Sorry, had to resolve a merge conflict. :)

PR #20176, opened 2026-03-05
20166 add format file ref to file accepting props

### LINKED ISSUE #20166: Add `format: "file-ref"` to file-accepting props

Many component actions across `components/` have props that accept file paths or URLs, but there is no standardized way for the platform or Connect developers to know which props represent file inputs. This means developers cannot render file-upload UI for these props, and users must manually provide `/tmp` paths or URLs with no guidance from the interface. Additionally, some file props only accept a path _or_ a URL but not both, creating an inconsistent experience.

Audit all actions and for every prop that accepts a file path or URL:

1.  **Add `format: "file-ref"`** to the prop definition. This new attribute on [string props](https://pipedream.com/docs/components/contributing/api#user-input-props:~:text=format,-string) signals to the platform and Connect developers that the prop accepts both a URL or `/tmp` path as a reference to a file.
    
2.  **If a prop currently accepts only a file path OR only a URL (but not both)**, update the action to accept both, following the [file-related actions guidelines](https://pipedream.com/docs/components/contributing/guidelines#file-related-actions). Then add `format: "file-ref"`.
    

**How to identify file props:** Look for props where the name, label, or description suggests a file path or URL input — e.g. prop names like `filePath`, `file`, `path`, `fileUrl`, `attachment`, `image`, `document`; labels like `"File Path"`, `"File URL"`, `"File Path or URL"`; descriptions mentioning `/tmp` or file URLs. These are typically `type: "string"` props.

After the change, a file-accepting prop should look like:

```js
filePath: {
  type: "string",
  label: "File Path or URL",
  description: "Provide either a file URL or a path to a file in the `/tmp` directory. [See the documentation on working with files](https://pipedream.com/docs/code/nodejs/working-with-files/).",
  format: "file-ref",
},

```

**Do you have a workaround?**

As a heuristic, developers could search for props whose label contains "File Path or URL", and this works in many cases, but it is not reliable — some file props use different labels (e.g. "File Path", "File URL", "Attachment", "Image") or indicate file input only in the description. The format: "file-ref" attribute provides a definitive, machine-readable signal that eliminates the need for guesswork.

N/A

**Additional context**
- Prior art: PR #20167 adds `format: "file-ref"` to a first batch of apps (Gmail, Google Drive, Jira, Microsoft Outlook, Salesforce, Sharepoint, Slack).
-   Given the number of apps in  `components/`, this work can be split across multiple PRs grouped by app or alphabetical range.
-   Reference docs:
    -   [Component prop definition docs](https://pipedream.com/docs/components/contributing/api#user-input-props)
    -   [File-related actions guidelines](https://pipedream.com/docs/components/contributing/guidelines#file-related-actions)

luancazarine:
Resolves #20166
This PR adds the 'format: "file-ref"' to props that already accept both file path and URL.

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -643,8 +643,8 @@ importers:
   components/airparser:
     dependencies:
       '@pipedream/platform':
-        specifier: 3.2.5
-        version: 3.2.5
+        specifier: ^3.2.5
+        version: 3.3.0
       form-data:
         specifier: 4.0.4
         version: 4.0.4
@@ -719,8 +719,8 @@ importers:
   components/akeneo:
     dependencies:
       '@pipedream/platform':
-        specifier: 3.2.5
-        version: 3.2.5
+        specifier: ^3.2.5
+        version: 3.3.0
       form-data:
         specifier: 4.0.4
         version: 4.0.4
@@ -842,14 +842,14 @@ importers:
   components/alttext_ai:
     dependencies:
       '@pipedream/platform':
-        specifier: 3.2.5
-        version: 3.2.5
+        specifier: ^3.2.5
+        version: 3.3.0
 
   components/alttextify:
     dependencies:
       '@pipedream/platform':
-        specifier: 3.2.5
-        version: 3.2.5
+        specifier: ^3.2.5
+        version: 3.3.0
 
   components/alttextlab:
     dependencies:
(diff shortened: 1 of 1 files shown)


components/airparser/package.json
```diff
@@ -13,7 +13,7 @@
     "access": "public"
   },
   "dependencies": {
-    "@pipedream/platform": "^3.1.1",
+    "@pipedream/platform": "3.2.5",
```
GTFalcao: Why is the version being locked here? Please revert or change to `^3.2.5`

PR #20197, opened 2026-03-06
Box - bug fix for sources

### LINKED ISSUE #20184: [Bug] Deleting a Box trigger via Connect API does not remove the underlying Box webhook, causing 409 conflicts on re-deploy
**Describe the bug**

  When using the Pipedream Connect API to delete a deployed `box-new-file` trigger, the trigger is successfully removed from Pipedream's side, but the underlying Box webhook is **not** cleaned up on Box's
  infrastructure. This leaves an orphaned webhook that causes a 409 Conflict error when attempting to deploy a new trigger for the same Box account/folder target.

  This makes any delete → redeploy cycle for Box triggers unrecoverable through the Connect API alone.

  **To Reproduce**

  1. Deploy a `box-new-file` trigger via the Connect API:
     ```
     POST /v1/connect/{project_id}/triggers/deploy
     {
       "external_user_id": "org_xxx",
       "id": "box-new-file",
       "configured_props": {
         "app": { "authProvisionId": "apn_xxx" },
         "webhookTarget": "{\"id\":\"0\",\"type\":\"folder\"}"
       },
       "webhook_url": "https://example.com/webhook"
     }
     ```
     Trigger deploys successfully (e.g., `dc_abc123`).

  2. Delete the trigger via the Connect API:
     ```
     DELETE /v1/connect/{project_id}/deployed-triggers/dc_abc123?external_user_id=org_xxx
     ```
     Returns success (200/204). Trigger no longer appears in `GET /deployed-triggers`.

  3. Deploy a **new** trigger for the same Box account (same payload as step 1).

  4. Deployment fails with **409 or 422** — Box rejects it because the webhook from step 1 still exists on their side.

  5. Attempting recovery by listing existing triggers via `GET /deployed-triggers` returns **no matching trigger** (it was deleted in step 2), so there is no way to resolve the conflict through the API.

  **Expected behavior**

  When a `box-new-file` trigger is deleted via the Connect API, Pipedream should also call Box's webhook deletion endpoint (`DELETE https://api.box.com/2.0/webhooks/{webhook_id}`) to remove the underlying
  webhook, ensuring a clean state for future deployments. This is how we'd expect teardown to work — the full lifecycle of the external resource should be managed by the component.

  **Screenshots**

  N/A — this is an API-only issue. Sanitized logs from our staging environment:

  ```
  [PIPEDREAM-TRIGGERS] Deploying new files trigger (provider: box)
  [PIPEDREAM-TRIGGERS] Failed to deploy trigger (status: 409/422)
  ```

  **Additional context**

  - **Component:** `box-new-file`
  - **API:** Pipedream Connect API with OAuth client credentials flow
  - **Box webhook target:** Root folder (`id: "0"`, `type: "folder"`)
  - Reproduced in both our staging and production environments
  - Google Drive and Dropbox triggers do **not** have this issue — only Box
  - We attempted to manually clean up orphaned webhooks via the Box API (`DELETE /2.0/webhooks/{id}`), but the Connect API does not expose the underlying Box webhook ID, making this impractical
  - **Workaround:** We currently catch the conflict error and try to find an existing trigger to reuse, but since the trigger was already deleted on Pipedream's side, recovery fails every time
  - This is a blocking issue for our users who frequently update their monitored folder configurations, as any folder change triggers a delete → redeploy cycle

michelle0927:
Resolves #20184 

I was able to confirm that an error is received when creating a webhook for a target that already has an existing webhook in Box, though I wasn't able to reproduce the exact issue the user is seeing. 

I added a call in the `activate` hook that checks for an existing webhook for the selected target, and deletes it if it exists before creating the new one.

## Summary by CodeRabbit

* **New Features**
  * Added webhook listing so users can view existing webhooks.
  * Automatically removes duplicate webhooks for the same target to ensure only one active webhook per destination.
  * Added an informational alert about the single-webhook-per-target limitation.
  * Added a webhook target selection option when configuring new-folder event sources.

* **Chores**
  * Minor package/component version updates.

### DIFF
--- a/components/box/sources/common/common.mjs
+++ b/components/box/sources/common/common.mjs
@@ -49,9 +49,13 @@ export default {
   hooks: {
     async activate() {
       // Delete existing hook with same target if it exists
-      const targetId = this.getTarget().id;
+      const {
+        id: targetId, type: targetType,
+      } = this.getTarget();
       const { entries } = await this.app.listWebhooks();
-      const existingHook = entries?.find((entry) => entry.target.id === targetId);
+      const existingHook = entries?.find(
+        (entry) => entry.target.id === targetId && entry.target.type === targetType,
+      );
       if (existingHook) {
         await this.app.deleteHook({
           hookId: existingHook.id,
--- a/components/box/sources/new-folder/new-folder.mjs
+++ b/components/box/sources/new-folder/new-folder.mjs
@@ -8,6 +8,18 @@ export default {
   type: "source",
   dedupe: "unique",
   ...common,
+  props: {
+    ...common.props,
+    webhookTarget: {
+      propDefinition: [
+        common.props.app,
+        "webhookTarget",
+        () => ({
+          type: "folder",
+        }),
+      ],
+    },
+  },
   methods: {
     ...common.methods,
     getTriggers() {


components/box/sources/common/common.mjs
```diff
@@ -43,6 +48,15 @@ export default {
   },
   hooks: {
     async activate() {
+      // Delete existing hook with same target if it exists
+      const targetId = this.getTarget().id;
+      const { entries } = await this.app.listWebhooks();
+      const existingHook = entries?.find((entry) => entry.target.id === targetId);
```
jcortes: Seems like you also need to filter by type
```suggestion
      const existingHook = entries?.find(
    (entry) => entry.target.id === target.id && entry.target.type === target.type
  );
```

pipedream-component-development:
### For Integration QA:
- Source Test Report: https://www.notion.so/Bug-Deleting-a-Box-trigger-via-Connect-API-does-not-remove-the-underlying-Box-webhook-causing-409--31beb261c847814e8973f6131c434b8e

jcortes:
Hi @michelle0927 please make sure the target is a folder, because if I choose a file I get this error:

image

pipedream-component-development:
- [Bug] Deleting a Box trigger via Connect API does not remove the underlying Box webhook, causing 409 conflicts on re-deploy - Sources: https://www.notion.so/Bug-Deleting-a-Box-trigger-via-Connect-API-does-not-remove-the-underlying-Box-webhook-causing-409--31beb261c847814e8973f6131c434b8e

PR #20250, opened 2026-03-11
Freshchat - bug fix to list-messages

### LINKED ISSUE #20189: [BUG] freshchat-list-messages: from_time parameter appears to work as "until time" due to default descending sort order
## Describe the bug

The `freshchat-list-messages` action has a `From Time` parameter that is documented as:
> "UTC datetime to retrieve messages **after** this timestamp"

However, in practice it behaves as an **"until time"** — returning messages **before** the given timestamp instead of after it.

## Root cause

The Freshchat API returns messages in **descending order** (newest first) by default. 
This means that when `from_time` is set (e.g. `2025-12-01T22:22:07Z`), the API 
correctly filters to messages at or after that time — but because results are sorted 
newest-first, `getPaginatedResults` collects pages starting from the most recent 
messages *down to* the `from_time` boundary, making it appear as if `from_time` 
is an end time rather than a start time.

## Steps to reproduce

1. Add the **List Messages** action for Freshchat in a workflow
2. Set `From Time` to a timestamp in the past (e.g. `2025-12-01T22:22:07Z`)
3. Run the action
4. Observe that the returned messages are **older** than the timestamp, not newer

## Expected behavior

Messages returned should have a `created_time` **equal to or later than** the `from_time` value.

## Actual behavior

Messages returned appear to be from **before** the `from_time` value, as if it is functioning as an "until time".

## Suggested fix

Add `sort_order: "asc"` to the API params so messages are returned oldest-first, 
making `from_time` behave correctly as a start boundary:

**File:** `components/freshchat/actions/list-messages/list-messages.mjs`
```diff
  args: {
    $,
    conversationId,
    params: {
      from_time: fromTime,
+     sort_order: "asc",
    },
  },
```

## Component info

- **Key:** `freshchat-list-messages`  
- **Version:** `0.0.2`  
- **File:** `components/freshchat/actions/list-messages/list-messages.mjs`

## References

- [Freshchat API docs – List messages](https://developers.freshchat.com/api/#list_messages)
- The API docs confirm `from_time` filters to messages where `created_time >= from_time`, but do not specify a default sort order.

michelle0927:
Resolves #20189 

## Summary by CodeRabbit

* **New Features**
  * Message lists are returned in reverse chronological order.
  * Optional creation-time filtering and optional maximum-result limiting are applied after fetching.
  * The reported count reflects the processed (filtered/trimmed) message set.

* **Chores**
  * Component and action version numbers bumped.

### DIFF
--- a/components/freshchat/actions/list-messages/list-messages.mjs
+++ b/components/freshchat/actions/list-messages/list-messages.mjs
@@ -60,7 +60,7 @@ export default {
 
     let messages = response.reverse();
     if (fromTime) {
-      messages = response.filter(( { created_time: ts }) => Date.parse(ts) > Date.parse(fromTime));
+      messages = messages.filter(( { created_time: ts }) => Date.parse(ts) > Date.parse(fromTime));
     }
     if (maxResults) {
       messages = messages.slice(0, maxResults);


pipedream-component-development:
- [BUG] freshchat-list-messages: from_time parameter appears to work as "until time" due to default descending sort order - Actions Batch 1: https://www.notion.so/BUG-freshchat-list-messages-from_time-parameter-appears-to-work-as-until-time-due-to-default-des-320eb261c847819fa16ce2170f4a3900

PR #20303, opened 2026-03-17
Google Drive - bug fix to deduping in update sources

michelle0927:
When concatenating an id and timestamp, e.g. `${fileId}-${ts}`, the resulting string can be greater than 64 characters and therefore fail to emit some events.

This PR updates the sources that use a timestamp in the emitted `id` to hash the id using `md5`.

## Summary by CodeRabbit

* **Chores**
  * Google Drive component and its sources bumped to new patch versions (component 1.6.1; multiple sources incremented).
  * Added md5 dependency.
  * Emitted event/file/comment IDs are now represented as md5 hashes instead of plain concatenated strings, affecting how IDs appear in integrations and logs.


pipedream-component-development:
### For Integration QA:
- Source Test Report: https://www.notion.so/Google-Drive-bug-fix-to-deduping-in-update-sources-Sources-326eb261c84781feb570cb90e72bc41f

PR #20368, opened 2026-03-23
Pipedream Utils - Download multiple files to tmp

michelle0927:
Allows for downloading multiple files to the `/tmp` directory.

## Summary by CodeRabbit

* **New Features**
  * Added an action that saves multiple files to the workflow /tmp directory with per-file error handling and a summary of results.

* **Chores**
  * Bumped package version to 0.1.0.
  * Removed the action that downloaded a file to /tmp.

### DIFF
--- a/components/pipedream_utils/actions/add-files-to-tmp/add-files-to-tmp.mjs
+++ b/components/pipedream_utils/actions/add-files-to-tmp/add-files-to-tmp.mjs
@@ -56,7 +56,15 @@ export default {
         console.log(`Failed to add file: ${file}`);
       }
     }
-    $.export("$summary", `Successfully added ${files.length} files to /tmp`);
+    if (files.length === this.files.length) {
+      $.export("$summary", `Successfully added ${files.length} file${files.length === 1
+        ? ""
+        : "s"} to /tmp`);
+    } else if (files.length > 0) {
+      $.export("$summary", `Successfully added ${files.length} of ${this.files.length} files to /tmp`);
+    } else {
+      throw new Error("Failed to add files to /tmp");
+    }
     return files;
   },
 };


components/pipedream_utils/actions/download-files-to-tmp/download-files-to-tmp.mjs
```diff
@@ -0,0 +1,58 @@
+import pipedream_utils from "../../pipedream_utils.app.mjs";
+import { getFileStreamAndMetadata } from "@pipedream/platform";
+import fs from "fs";
+
+export default {
+  key: "pipedream_utils-download-files-to-tmp",
+  name: "Download Files To /tmp",
+  description: "Downloads a list of files to [your workflow's /tmp directory](https://pipedream.com/docs/code/nodejs/working-with-files/#the-tmp-directory)",
```
GTFalcao: To make this more intuitive, I think we should rename the new component to "Add Files" instead of "Download Files", and remove the old component that only supported a single file URL.

components/pipedream_utils/actions/add-files-to-tmp/add-files-to-tmp.mjs
```diff
@@ -0,0 +1,62 @@
...
+        fs.writeFileSync(filepath, buffer);
+        files.push({
+          filename,
+          filepath,
+        });
+      } catch {
+        console.log(`Failed to add file: ${file}`);
+      }
+    }
+    $.export("$summary", `Successfully added ${files.length} files to /tmp`);
```
GTFalcao: Maybe we should have a separate variable to track successful file uploads, so we can adjust the summary accordingly - currently it would export a success summary even if all files fail

PR #20394, opened 2026-03-26
Fixed Add Labels actions not showing the correct error message

### LINKED ISSUE #20375: [ACTION] Pipedrive - Cannot add multiple labels using custom expressions

Pipedrive

Add Labels - cannot add multiple custom expressions or single

Image

Please see samples

Image

Image

Image

Even this it doesn't work

Image

Image

https://developers.pipedrive.com/docs/api/v1

jcortes:
## WHY

Resolves #20375

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error handling when adding labels to Pipedrive records to provide clearer, user-facing error messages when updates fail.

* **Chores**
  * Updated Pipedrive package and action versions (component v0.12.2, action v0.0.3).

No other user-facing functionality changes.

### DIFF
--- a/components/pipedrive/actions/add-labels/add-labels.mjs
+++ b/components/pipedrive/actions/add-labels/add-labels.mjs
@@ -1,10 +1,11 @@
+import { ConfigurationError } from "@pipedream/platform";
 import pipedriveApp from "../../pipedrive.app.mjs";
 
 export default {
   key: "pipedrive-add-labels",
   name: "Add Labels",
   description: "Adds labels to a lead, person, deal, or organization in Pipedrive. [See the documentation](https://developers.pipedrive.com/docs/api/v1/Leads#updateLead)",
-  version: "0.0.2",
+  version: "0.0.3",
   annotations: {
     destructiveHint: true,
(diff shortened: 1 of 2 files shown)


jcortes:
> I don't think this will fix the issue, but at least it will display a better error. Ready for QA!

Yeah indeed not an issue but we were not showing the error the right way so the user can have a better idea of what's happening, seems like the ids the user were not set at that moment

michelle0927 (approved):
Passed QA, Ready for Release!

tonyskye101:
@michelle0927 @jcortes Would it be possible to add a function that removes a specific label? Currently, the only options available are replacing the existing label or adding a new one without removing the current label.

what I'm aiming is. . .

current label assigned: **label1, label2, label3, label4**
remove specific label: **label4**
present label: **label1, label2, label3**

Actually, the labels are fine as they are, since they are correctly read by the system.

image

image

image

Let me know if I should create a new request for this action.

tonyskye101:
here's a ticket: https://github.com/PipedreamHQ/pipedream/issues/20431

PR #20436, opened 2026-03-30
Egnyte - Search & List Documents

### LINKED ISSUE #20410: Egnyte — add Find/List Documents action for agent discoverability
Egnyte

https://developers.egnyte.com/docs/read/File_System_Management_API_Documentation

## Problem

The Egnyte app has no `propDefinitions` and no mechanism for an agent to discover available documents. All actions rely on raw file paths passed as free-form strings. An agent cannot retrieve a list of documents to know what exists — it must have the exact path in advance, which requires a human in the loop.

## What's needed

Add a **Find/List Documents** action that searches for documents by name or lists documents within a folder path. The action should return document names and their full paths so an agent can identify the correct file before passing it to other actions.

## Notes

- The Egnyte API supports searching files: `GET /pubapi/v1/fs-content-search/` and listing folder contents: `GET /pubapi/v1/fs/{folderPath}`
- This is a prerequisite for PIPE-79 — an agent needs to find a document before it can read or download it
- PIPE-78 covers folder discovery, which is the related blocker for navigating to the right location

***[PIPE-77]***

michelle0927:
Resolves #20410 

## Summary by CodeRabbit

* **New Features**
  * Added "List Documents" action to retrieve folder contents
  * Added "Search" action to search files and folders with advanced filtering (file type, location, sorting options)

* **Chores**
  * Updated component versions across Egnyte integration

### DIFF
--- a/components/egnyte/actions/create-folder/create-folder.mjs
+++ b/components/egnyte/actions/create-folder/create-folder.mjs
@@ -4,7 +4,7 @@ export default {
   key: "egnyte-create-folder",
   name: "Create Folder",
   description: "Creates a new folder in your Egnyte workspace. [See the documentation](https://developers.egnyte.com/docs/File_System_Management_API_Documentation#Create-a-Folder)",
-  version: "0.0.3",
+  version: "0.0.4",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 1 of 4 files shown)


luancazarine:
Hi @michelle0927, I got this error for both actions 
image

PR #20439, opened 2026-03-30
Egnyte - Download File

### LINKED ISSUE #20412: Egnyte — add Read/Download Document action
Egnyte

https://developers.egnyte.com/docs/read/File_System_Management_API_Documentation

## Problem

The Egnyte app has no action for reading or downloading document content. Even if an agent can locate a file (via PIPE-77), there is no way to retrieve its contents. This makes the integration read-only in name only — an agent can upload files but cannot consume the content of existing documents.

## What's needed

Add a **Read/Download Document** action that accepts a file path and returns the document content. For text-based files, content should be returned directly in the response. For binary formats (e.g. Word, PDF), the file should be downloaded to `/tmp` so it can be passed to subsequent actions.

## Notes

- The Egnyte API supports downloading file content: `GET /pubapi/v1/fs-content/{filePath}`
- Ideally the action supports a format conversion option (e.g. to PDF) similar to the SharePoint download action, so agents can process documents regardless of their original format
- This depends on PIPE-77 and PIPE-78 — an agent needs to be able to find the file path before it can download it

***[PIPE-79]***

michelle0927:
Resolves #20412 

## Summary by CodeRabbit

* **New Features**
  * New "Download File" action for Egnyte — save files to the runtime and optionally return text content.
  * Dynamic "Filename" selector: shows files for the chosen folder and a searchable "Folder Path" selector.

* **Chores**
  * Bumped package and component versions across Egnyte integrations and sources.

### DIFF
--- a/components/egnyte/egnyte.app.mjs
+++ b/components/egnyte/egnyte.app.mjs
@@ -18,7 +18,7 @@ export default {
         const { files } = await this.getFolder({
           folderPath,
         });
-        return files.map((file) => file.name);
+        return files?.map?.((file) => file.name) || [];
       },
     },
     folderPath: {
@@ -34,12 +34,12 @@ export default {
               type: "FOLDER",
             },
           });
-          return results.map((result) => result.path);
+          return results?.map?.((result) => result.path) || [];
         }
         const { folders } = await this.getFolder({
           folderPath: "",
         });
-        return folders.map((folder) => folder.path);
+        return folders?.map?.((folder) => folder.path) || [];
       },
     },
   },


components/egnyte/actions/download-file/download-file.mjs
```diff
@@ -0,0 +1,119 @@
...
+
+    const ext = path.extname(filename);
+    const contentType = res.headers?.["content-type"];
+    const result = {
+      filepath,
+      filename,
+    };
+    if (this._isTextFile(contentType, ext)) {
+      result.content = buffer.toString("utf8");
+    }
```
GTFalcao: I think this behavior should be controlled by an optional boolean prop. If the content is too large, it can cause the request to error out (we have certain limits on the data size returned/exported from executions).

GTFalcao (commented):
Just one thing that needs to be addressed before we can move forward

PR #20496, opened 2026-04-02
Salesforce - remove dynamic props from create-opportunity and update-opportunity

### LINKED ISSUE #20467: Salesforce — support Amount field in opportunity create/update
Salesforce

https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/dome_sobject_create.htm

## Problem

The create and update opportunity actions do not expose the `Amount` field as a configurable prop. This is one of the most commonly used fields on an Opportunity record. An agent creating or updating a deal cannot set the dollar value, which makes the action incomplete for typical sales workflows.

## What's needed

Add `Amount` as an optional prop (type `string` or `number`) to both the **Create Opportunity** and **Update Opportunity** actions. The field should be passed through in the request body to the Salesforce REST API as a standard field on the Opportunity object.

## Notes

- Check whether the `fieldsToUpdate` prop or similar mechanism already filters which fields are sent — the fix may involve adjusting that filter rather than adding a new prop
- `Amount` is a standard Salesforce field, not a custom field, so it should be available on all orgs

***[PIPE-82]***

michelle0927:
Resolves #20467 

## Summary by CodeRabbit

* **Chores**
  * Updated Salesforce REST API component to version 1.12.1
  * Updated versions for multiple Salesforce actions to improve stability

### DIFF
--- a/components/salesforce_rest_api/actions/create-account/create-account.mjs
+++ b/components/salesforce_rest_api/actions/create-account/create-account.mjs
@@ -8,7 +8,7 @@ export default {
   key: "salesforce_rest_api-create-account",
   name: "Create Account",
   description: `Creates a Salesforce account. [See the documentation](${docsLink})`,
-  version: "0.3.5",
+  version: "0.3.6",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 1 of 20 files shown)


luancazarine (approved):

PR #20533, opened 2026-04-07
Initial CodeRabbit guidelines

GTFalcao:
## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive Pipedream authoring guidelines for components, actions, and sources covering props, annotations, emits/deduplication, polling/webhook/hybrid lifecycles, dynamic options/pagination, HTTP/request patterns, error-handling, and documentation-quality rules.

* **Chores**
  * Added an automated review configuration enabling en-US review profiles, path-scoped checks, high-level summaries, collapsed walkthroughs, knowledge-base linkage, and auto-reply for non-draft PRs.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -40,6 +40,13 @@ reviews:
         - Component `description` must end with a documentation link `[See the documentation](https://...)`
           and should follow the agent-friendly structure in the guidelines:
           primary purpose → when to use → cross-tool references → parameter notes → gotchas → docs link
+        - When a component accepts a file input, verify:
+          - The file input prop includes `format: "file-ref"`
+          - A `syncDir` prop with `accessMode: "read"` and `sync: true` is declared alongside it
+          - `getFileStreamAndMetadata` or `getFileStream` from `@pipedream/platform` is used to
+            access the file in `run()`, not raw `fs.readFileSync` or direct URL fetching
+        - When a component writes a file to `/tmp`, verify a `syncDir` prop with
+          `accessMode: "write"` and `sync: true` is present
 
     # ── Action components ──────────────────────────────────────────────────────
     - path: "components/*/actions/**/*.mjs"
--- a/.github/pipedream-component-guidelines.md
+++ b/.github/pipedream-component-guidelines.md
@@ -361,6 +361,89 @@ includes `async options()` that calls a paginated API endpoint, it must support 
 
 ---
 
+## File Handling
+
+Components that accept files as input or write files to `/tmp` must follow a consistent
+pattern using helpers from `@pipedream/platform`.
+
+### Reading / uploading files
+
+When a component accepts a file from the user (a URL or a `/tmp` path), the input prop
+must include `format: "file-ref"`:
+
+```javascript
+filePath: {
+  type: "string",
+  label: "File Path or URL",
+  description: "Provide either a file URL or a path to a file in the `/tmp` directory (for example, `/tmp/myFile.pdf`).",
+  format: "file-ref",
+},
+```
+
+The file content is accessed at runtime via `getFileStreamAndMetadata` (or `getFileStream`
+for stream-only access) imported from `@pipedream/platform`:
+
+```javascript
+import { getFileStreamAndMetadata } from "@pipedream/platform";
+
+async run({ $ }) {
+  const { stream, metadata } = await getFileStreamAndMetadata(this.filePath);
+  // stream             — a readable stream of the file contents
+  // metadata.name        — filename
+  // metadata.contentType — MIME type
+  // metadata.size        — byte size
+},
+```
+
+The component must also declare a `syncDir` prop with `accessMode: "read"` to allow the
+platform to sync the referenced file into the execution environment:
+
+```javascript
+syncDir: {
+  type: "dir",
+  accessMode: "read",
+  sync: true,
+  optional: true,
+},
+```
+
+### Writing / downloading files to `/tmp`
+
+When a component writes a file to `/tmp`, declare a `syncDir` prop with
+`accessMode: "write"` so the platform can sync the result out:
+
+```javascript
+syncDir: {
+  type: "dir",
+  accessMode: "write",
+  sync: true,
+},
+```
+
+Write the file using standard Node.js `fs` or a stream pipeline:
+
+```javascript
+import { pipeline } from "stream/promises";
+import fs from "fs";
+
+async run({ $ }) {
+  const { Body: data } = await this.app.downloadFile(...);
+  await pipeline(data, fs.createWriteStream(this.filePath));
+},
+```
+
+### File handling checklist
+
+- `format: "file-ref"` must be present on every prop that accepts a file path or URL as
+  input — without it the platform cannot resolve the reference at runtime
+- A `syncDir` prop with the appropriate `accessMode` must always accompany a file input or
+  output: `"read"` when the component consumes a file, `"write"` when it produces one
+- Use `getFileStreamAndMetadata` when both the stream and metadata (name, content type,
+  size) are needed; use `getFileStream` when only the stream is needed
+- File handling applies equally to actions and sources — it is not restricted to actions
+
+---
+
 ## Runtime Primitives
 
 These values are injected by the Pipedream platform at runtime. They are not importable —


.github/pipedream-action-guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+### `async options` (standard pattern)
+
+When a prop's valid values are dynamic but enumerable, use `async options()`. This is the
+standard pattern and works in both the workflow UI and agent contexts:
+
+```javascript
+pipeline: {
+  type: "string",
+  label: "Pipeline",
+  description: "The pipeline to assign this deal to.",
```
michelle0927, line 119: Not very important, but are we still enforcing that descriptions should only have a period if they're a complete sentence? If so, the examples should reflect that.

.github/pipedream-action-guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+### JSON object props for open-ended inputs
+
+When an action accepts a dynamic set of field name-value pairs (e.g., record properties
+with user-defined custom fields), either `type: "object"` or `type: "string"` is
+acceptable. Both work in agent and workflow contexts:
+
+- **`type: "object"`** — renders a key-value editor in the UI; the platform deserializes
+  the value automatically so no `JSON.parse()` is needed in `run()`.
+- **`type: "string"` with `JSON.parse()` in `run()`** — accepts a raw JSON string; more
+  explicit and portable, but requires manual parsing.
```
michelle0927, line 152: We should parse the input whether the prop type is "object" or "string", either by using a try/catch block or checking the input type. A user can enter an object in a string prop or vice versa.

.github/pipedream-component-guidelines.md
```diff
@@ -0,0 +1,424 @@
...
+```
+
+### Prop types
+
+| Type | Description |
+|------|-------------|
+| `"string"` | Text input |
+| `"string[]"` | Array of text inputs |
+| `"integer"` | Integer, supports `min` / `max` |
+| `"number"` | Floating-point |
```
michelle0927: I don't think we support `number`, do we?
```suggestion
```

.github/pipedream-component-guidelines.md
```diff
@@ -0,0 +1,424 @@
...
+    return stages.map((s) => ({
+      label: s.name,   // displayed to user
+      value: s.id,     // stored and used at runtime
+    }));
+  },
+},
+```
+
+Requirements:
+- Each option must have both `label` (shown in UI) and `value` (used at runtime).
```
michelle0927: Some props just return the value. Is there a reason to enforce having both `label` and `value` for all async options?

.github/pipedream-component-guidelines.md
```diff
@@ -0,0 +1,424 @@
...
+- If the API paginates results, `async options()` must support `prevContext` for "load more":
```
michelle0927:
```suggestion
- If the API paginates results, `async options()` must support `prevContext` or `page` for "load more":
```

.github/pipedream-component-guidelines.md
```diff
@@ -0,0 +1,424 @@
...
+```javascript
+async options({ prevContext }) {
+  const cursor = prevContext?.cursor ?? null;
+  const { items, nextCursor } = await this.app.getItems({ cursor });
+  return {
+    options: items.map((i) => ({ label: i.name, value: i.id })),
+    context: { cursor: nextCursor },
+  };
+},
+```
```
michelle0927, line 191:
```suggestion

```javascript
async options({ page }) {
  const { items } = await this.app.getItems({ params: { page: page + 1 }});
  return items.map((i) => ({ label: i.name, value: i.id })),
},
```
```

File: .github/pipedream-component-guidelines.md [hunk 7]
```diff
@@ -0,0 +1,424 @@
...
+
+### Decision guide
+
+| Operation type | `readOnlyHint` | `destructiveHint` | `openWorldHint` |
+|---|---|---|---|
+| Fetch / list / search / get | `true` | `false` | `true` |
+| Create / send / post / publish | `false` | `false` | `true` |
+| Update / patch / upsert | `false` | `false` | `true` |
```
michelle0927: I've never been totally sure about this, but shouldn't Updates have `destructiveHint` as `true` since they can irreversibly modify data?

.github/pipedream-component-guidelines.md
```diff
@@ -0,0 +1,424 @@
...
+  pass per-request headers that merge with the standard auth headers, without clobbering them.
+- **Spreading `args`**: All other request options (`method`, `url`, `params`, `data`, etc.)
+  are passed through via `...args`, keeping public methods minimal — they set the
+  method and URL, then forward everything else.
+
+### propDefinitions
+
+Prop definitions shared across more than one component belong in the app file. New
+`propDefinitions` entries must include both `label` and `description`. If the definition
+includes `async options()` that calls a paginated API endpoint, it must support `prevContext`.
```
michelle0927:
```suggestion
includes `async options()` that calls a paginated API endpoint, it must support `prevContext` or `page`.
```

.github/pipedream-source-guidelines.md
```diff
@@ -0,0 +1,363 @@
...
+};
+```
+
+`dedupe: "unique"` is the standard strategy. Pipedream will not emit an event whose `id`
+it has already seen for this source instance.
+
+For deduplication to work correctly:
+- The `id` must be unique per real-world event (not per poll execution)
+- The `id` must be stable — if the source is paused and restarted, the same real-world
+  event must produce the same `id`
```
michelle0927, line 83:
```suggestion
- The `id` must be unique per real-world event (not per poll execution)
- The `id` must be stable — if the source is paused and restarted, the same real-world
  event must produce the same `id`
-  The `id` will only be deduplicated up to 64 characters.
```

.coderabbit.yaml
```diff
@@ -0,0 +1,113 @@
...
+        - New methods follow naming conventions: `_camelCase()` for private helpers,
+          verb-noun for public methods (`getRecords`, `createContact`, `deleteItem`)
+        - Existing methods must not be renamed or refactored — only additions are acceptable,
+          since renames break deployed components that reference those method names
+        - HTTP requests must use `axios` from `@pipedream/platform`, never the bare `axios` package
+        - Each public method should represent one logical API operation and delegate to a shared
+          private `_makeRequest()` (or equivalent) — raw axios calls should not appear in public methods
+        - New `propDefinitions` entries must include both `label` and `description`
+        - If `async options()` is added to a propDefinition and the underlying API paginates,
+          the implementation must handle `prevContext` to support "load more"
```
michelle0927:
```suggestion
        - If `async options()` is added to a propDefinition and the underlying API paginates,
          the implementation must handle `prevContext` or `page` to support "load more"
```

.github/pipedream-action-guidelines.md
GTFalcao: I know we've done some of this in the past, usually when requested by QA, but technically a user can enter any value for any prop - number in string, object in array, etc. and I don't think our individual component code should be handling this - if anything, global execution code should.

.github/pipedream-action-guidelines.md
GTFalcao: Not sure, we can discuss

.github/pipedream-component-guidelines.md
GTFalcao: Yep good catch, I think this was the result of an incorrect grep (there are non-prop constants that use "type": "number" elsewhere)

.github/pipedream-component-guidelines.md
GTFalcao: Good catch, adjusting this to mention the primitive values

.github/pipedream-component-guidelines.md
GTFalcao: Yeah I think update should be flagged as destructive, agreed and changed

.github/pipedream-action-guidelines.md
michelle0927: I agree that it would be better if this were handled globaly.

.github/pipedream-action-guidelines.md
```diff
@@ -0,0 +1,310 @@
...
+## Annotations for Actions
+
+`openWorldHint` must be `true` for all actions. The `readOnlyHint` and `destructiveHint`
+values depend on the API operations performed:
+
+| Action type | `readOnlyHint` | `destructiveHint` |
+|---|---|---|
+| Fetch / get / list / search / describe | `true` | `false` |
+| Create / add / send / publish | `false` | `false` |
+| Update / edit / patch / upsert | `false` | `false` |
```
michelle0927:
```suggestion
| Update / edit / patch / upsert | `false` | `true` |
```

PR #20551, opened 2026-04-09
WICS - bug fix to update-order

### LINKED ISSUE #20376: [BUG] WICS Update Order: TypeError when `lines` prop is null/undefined
## Bug Description

The `update-order` action throws a `TypeError` when the `lines` property is null or undefined.

**Component:** [`components/wics/actions/update-order/update-order.mjs`](https://github.com/PipedreamHQ/pipedream/blob/master/components/wics/actions/update-order/update-order.mjs)

## Steps to Reproduce

1. Use the WICS "Update Order" action
2. Leave the `lines` property empty / don't provide it
3. Run the action

## Expected Behavior

The action should gracefully handle a null/undefined `lines` input — either skipping the line-merge logic or returning the original order lines unchanged.

## Actual Behavior

`parseObject(this.lines)` returns null when `this.lines` is null/undefined, causing:

1. `newLines.find(...)` throws `TypeError: Cannot read properties of null (reading 'find')`
2. `for (const line of newLines)` throws `TypeError: null is not iterable`

## Suggested Fix

Default to an empty array after parsing:

```js
const newLines = parseObject(this.lines) ?? [];
```

Or guard early:

```js
if (!this.lines) return order.lines;
```

michelle0927:
Resolves #20376 

## Summary by CodeRabbit

* **New Features**
  * Added dynamic order number selection with intelligent filtering for order pickers.

* **Bug Fixes**
  * Improved line parsing with safe fallback to prevent failures when line data is missing.

* **Updates**
  * Standardized parameter naming from "order reference" to "order number" across actions and integrations.
  * Bumped component/action versions to deliver these updates.

### DIFF
--- a/components/wics/actions/delete-order-line/delete-order-line.mjs
+++ b/components/wics/actions/delete-order-line/delete-order-line.mjs
@@ -4,7 +4,7 @@ export default {
   key: "wics-delete-order-line",
   name: "Delete Order Line",
   description: "Delete a line from an order. [See the documentation](https://docs.wics.nl/test-environment.html#orders-delete-order-line)",
-  version: "0.0.2",
+  version: "0.0.3",
   type: "action",
   annotations: {
     destructiveHint: true,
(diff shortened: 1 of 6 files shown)


GTFalcao (approved):
LGTM!

PR #20597, opened 2026-04-14
Microsoft Entra ID — add search and filter support to Get Organization Users action

### LINKED ISSUE #20590: Microsoft Entra ID — add search and filter support to Get Organization Users action
Microsoft Entra ID

https://learn.microsoft.com/en-us/graph/api/user-list

## Problem

The `get-organization-users` action lists all enabled users but does not support searching or filtering by name or email. An agent that needs to find a specific user must retrieve the full user list and scan it locally — impractical in large tenants.

## What's needed

Add optional `$search` and `$filter` props to the existing `microsoft_entra_id-get-organization-users` action. When provided, pass them as query parameters to `listUsers()` via the `params` argument. This is consistent with the existing `search-groups` action pattern.

## Notes

- `listUsers()` in `microsoft_entra_id.app.mjs` accepts a `params` object — `$search` and `$filter` can be passed directly
- `$search="displayName:John"` requires `ConsistencyLevel: eventual` header (already set in `listUsers()`)
- Cross-reference the `search-groups` action for the pattern to follow

***[PIPE-35]***

michelle0927:
Resolves #20590 

## Summary by CodeRabbit

* **New Features**
  * Added optional filtering and search parameters to the Microsoft Entra ID organization user retrieval feature. The filter parameter enables customization of query criteria, defaulting to retrieval of enabled accounts only. The new search parameter allows targeted user lookups within your organization, providing improved flexibility and precision for user queries.

### DIFF
--- a/components/microsoft_entra_id/actions/get-organization-users/get-organization-users.mjs
+++ b/components/microsoft_entra_id/actions/get-organization-users/get-organization-users.mjs
@@ -4,7 +4,7 @@ export default {
   key: "microsoft_entra_id-get-organization-users",
   name: "Get Organization Users",
   description: "List all users in the organization. By default returns only enabled accounts. [See the documentation](https://learn.microsoft.com/en-us/graph/api/user-list?view=graph-rest-1.0&tabs=http)",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "action",
   annotations: {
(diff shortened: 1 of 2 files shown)


jcortes (approved):
Hi @michelle0927 lgtm! This is ready for release!

PR #20601, opened 2026-04-14
Monday — add queryParams to Get Board Items Page action

### LINKED ISSUE #20591: Monday — add search and filter support to Get Board Items Page action
Monday

https://developer.monday.com/api-reference/reference/items-page

## Problem

The `get-board-items-page` action retrieves all items from a board but has no search or filter capability. An agent that needs to find a specific item by name or column value must retrieve the full item list and scan locally — impractical for large boards.

## What's needed

Add optional search/filter props to the existing `monday-get-board-items-page` action. The Monday GraphQL `items_page` query supports a `query_params` argument for column-value filtering — expose this so agents can narrow results without fetching all items.

## Notes

- `listBoardItemsPage` in `monday.app.mjs` currently only accepts `boardId` and `cursor` — extend it to pass `query_params` when provided
- Requires a board ID as input — see PIPE-63 (list boards) for that prerequisite

***[PIPE-64]***

michelle0927:
Resolves #20591 

## Summary by CodeRabbit

* **New Features**
  * Added optional query-params support for retrieving board items, enabling filtered/sorted pages and returning expanded item metadata (more fields per item).

* **Chores**
  * Bumped package and component versions across the Monday.com integration.

### DIFF
--- a/components/monday/actions/create-board/create-board.mjs
+++ b/components/monday/actions/create-board/create-board.mjs
@@ -6,7 +6,7 @@ export default {
   name: "Create Board",
   description: "Creates a new board. [See the documentation](https://developer.monday.com/api-reference/reference/boards#create-a-board)",
   type: "action",
-  version: "0.0.11",
+  version: "0.0.12",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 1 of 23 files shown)


GTFalcao (approved):
LGTM!

PR #20607, opened 2026-04-14
Asana - Search Tasks Premium action

### LINKED ISSUE #20593: Asana — update Search Tasks action to use server-side name filtering
Asana

https://developers.asana.com/reference/gettasksforproject

## Problem

The `asana-search-tasks` action accepts a name filter but applies it client-side — the full task list is fetched from the API before filtering. For large projects this is inefficient and can return incomplete results if the task list is paginated.

## What's needed

Update `asana-search-tasks` to pass name/text filters as server-side query parameters to `GET /tasks` where the Asana API supports it, rather than filtering client-side.

## Notes

- `search-tasks` already exists with: name (currently client-side), assignee, section, project, completedSince, modifiedSince
- The `tasks` prop in `asana.app.mjs` calls `this.getTasks({ params: { project } })` → `GET /tasks`
- Requires one of: project, section, or assignee — cannot list tasks without a filter context

***[PIPE-67]***

michelle0927:
Resolves #20593 

The [Get multilple tasks endpoint](https://developers.asana.com/reference/gettasks) being used by action "Search Tasks" doesn't support searching by name, but is available to all users.

The [Search tasks in a workspace endpoint](https://developers.asana.com/reference/searchtasksforworkspace) supports a `text` prop, but is only available to premium users.

I chose to leave the "Search Tasks" action in place and added a "Search Tasks Premium" action that filters by name client-side.

## Summary by CodeRabbit

* **New Features**
  * Added "Search Tasks Premium" with advanced filtering (project, assignee, section, text, completed/modified dates).

* **Bug Fixes**
  * Adjusted several Asana actions to reflect updated metadata and non-destructive annotations.

* **Chores**
  * Bumped package to 0.9.0 and incremented versions across multiple Asana actions and sources.

### DIFF
--- a/components/asana/actions/add-task-to-section/add-task-to-section.mjs
+++ b/components/asana/actions/add-task-to-section/add-task-to-section.mjs
@@ -5,7 +5,7 @@ export default {
   name: "Add Task To Section",
   description: "Add a task to a specific, existing section. This will remove the task from other sections of the project. [See the documentation](https://developers.asana.com/docs/add-task-to-section)",
   key: "asana-add-task-to-section",
-  version: "0.2.11",
+  version: "0.2.12",
   annotations: {
     destructiveHint: true,
     openWorldHint: true,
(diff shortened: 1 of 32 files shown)


GTFalcao (approved):
LGTM - nice catches on the other actions' annotations too

PR #20657, opened 2026-04-21
CodeRabbit improvements + PR Template file

GTFalcao:
Improved some CodeRabbit guidelines and revamped our `pull_request_template` file

## Summary by CodeRabbit

* **Documentation**
  * Added formal component versioning rules: start new components at 0.0.1 and require semver bumps for changes; app-level version must be bumped by the same-or-greater segment
  * Clarified annotations guidance: annotations disallowed for source components and required/validated for action components (including readOnly/destructive/openWorld hints)
  * Updated PR and issue templates with a new summary/checklist and minor wording tweak
* **Chores**
  * Strengthened automated review validation to enforce the above rules

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,3 +1,25 @@
-## WHY
+## Summary
 
-<!-- author to complete -->
+<!-- Add your PR description here -->
+
+
+
+## Checklist
+Please check the following items:
+
+### PR Type
+
+Check all that apply to this PR:
+- [ ] New component(s)
+- [ ] Enhancement to existing component(s)
+- [ ] Bug fixes
+- [ ] Other
+
+### Versioning
+- [ ] All components updated in this PR had their version updated (`0.0.1` for new ones)
+- [ ] The app updated in this PR had its `package.json`'s version updated
+
+### New app
+If this a new app, please submit [an app integration request](https://github.com/PipedreamHQ/pipedream/issues/new?template=app---service-integration.md) - the PR will only be reviewed after the app is integrated.
+- [ ] The app updated in this PR is already integrated
+
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,15 +5,7 @@
 
 
 ## Checklist
-Please check the following items:
-
-### PR Type
-
-Check all that apply to this PR:
-- [ ] New component(s)
-- [ ] Enhancement to existing component(s)
-- [ ] Bug fixes
-- [ ] Other
+Please check the following items before your PR can be reviewed:
 
 ### Versioning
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -1,3 +1,25 @@
...
+
+Check all that apply to this PR:
+- [ ] New component(s)
+- [ ] Enhancement to existing component(s)
+- [ ] Bug fixes
+- [ ] Other
+
+### Versioning
+- [ ] All components updated in this PR had their version updated (`0.0.1` for new ones)
+- [ ] The app updated in this PR had its `package.json`'s version updated
```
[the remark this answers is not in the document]
GTFalcao: Can you explain where you're obtaining this requirement for a CHANGELOG.md entry?

.github/pull_request_template.md
```diff
@@ -1,3 +1,25 @@
...
+<!-- Add your PR description here -->
+
+
+
+## Checklist
+Please check the following items:
+
+### PR Type
+
+Check all that apply to this PR:
```
dylburger: Is this helpful for y'all in review? I feel like this should be implicit in the PR itself?

.github/pull_request_template.md
```diff
@@ -1,3 +1,25 @@
...
+- [ ] New component(s)
+- [ ] Enhancement to existing component(s)
+- [ ] Bug fixes
+- [ ] Other
+
+### Versioning
+- [ ] All components updated in this PR had their version updated (`0.0.1` for new ones)
```
dylburger, line 11: Don't the GitHub actions already check this? Is this just to make this obvious to users when submitting the PR (since people don't check action runs?)

.github/pull_request_template.md
dylburger, line 12: Same as above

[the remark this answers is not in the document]
GTFalcao: Thanks, you can remove these learnings entirely, as well as any other learnings explicitly associated with the connect-react package within this repository (because it is no longer used here)

.github/pull_request_template.md
GTFalcao: They do, I added this partially because people don't check them, and also because they don't run by default for PRs from outside contributors (a maintainer has to approve running them). The goal is to be more immediately transparent to users and also require them to acknowledge these (a possible next step would be only adding the PR to the review queue once the user has interacted on at least some of these checkboxes)

.github/pull_request_template.md
GTFalcao: Actually I had this initially because of the 'submit new app' requirement, but ended up splitting that into its own section, so this doesn't make much sense anymore - removing it, thanks for flagging

dylburger (approved):

PR #20744, opened 2026-04-29
stripe: add opt-in pagination metadata to list actions

### LINKED ISSUE #20702: Stripe 'List Customers' needs more pagination info
Initially reported in "List Customers" but may apply to other Stripe components - all should be checked

We need to return `has_more` in the action response so the user can know whether they need to retrieve another page or not, if they're looking to retrieve a large amount of customers.

Important: this can be a breaking change to the existing return value.

s0s0physm:
## Summary

Fixes #20702.

All six Stripe list actions were calling `.autoPagingToArray()` on the Stripe SDK, which returns a plain array and discards the Stripe list envelope — including `has_more`. Users iterating through large datasets had no way to know whether additional pages existed.

This PR adds an opt-in `Return Pagination Info` boolean prop to each list action. The default behavior is unchanged (returns `Array<T>`), so existing workflows continue to work. When the flag is set to `true`, the action instead returns `{ data, has_more, next_starting_after }` — putting pagination metadata directly in the return value where it's reliably visible to AI agents calling through MCP.

### Changes

- New optional prop `returnPaginationInfo` (boolean, default `false`) on all six list actions
- When `false` (default): identical to original behavior — returns `Array<T>` via `.autoPagingToArray({ limit })`
- When `true`: returns `{ data: T[], has_more: boolean, next_starting_after: string | null }`. Internally fetches `limit + 1` items, detects `has_more`, and trims back to `limit` — agents see at most `limit` items either way
- Added `startingAfter` / `endingBefore` cursor props to `list-refunds` and `list-payment-intents` (the other four already had them)
- Action descriptions updated to explain when to set the flag and how `next_starting_after` chains into the next call's `Starting After` prop
- All six actions: `0.1.4` → `0.2.0` (minor — new optional prop, no breaking change)
- `@pipedream/stripe`: `0.8.1` → `0.9.0` (minor)

### Why opt-in flag instead of always returning the new shape

The default array return must stay intact for backwards compatibility — countless workflows reference `steps.list_X.$return_value` as an iterable. Putting pagination info behind a flag means existing workflows are untouched, while agents and pagination-aware callers can opt in.

### Actions changed

- `stripe-list-customers`
- `stripe-list-invoices`
- `stripe-list-payouts`
- `stripe-list-balance-history`
- `stripe-list-refunds`
- `stripe-list-payment-intents`

## Test plan

- [ ] Existing workflows referencing `steps.list_customers.$return_value` continue to receive an array — no regression
- [ ] Run `stripe-list-customers` with `Return Pagination Info: true` against a Stripe account with more than `limit` customers — confirm response is `{ data, has_more: true, next_starting_after: "cus_xxx" }`
- [ ] Run with `Return Pagination Info: true` and fewer records than `limit` — confirm `has_more: false` and `next_starting_after: null`
- [ ] Pass the returned `next_starting_after` value as `Starting After` on a second call — confirm it returns the next page
- [ ] Confirm `startingAfter` cursor works on `list-refunds` and `list-payment-intents`

## Summary by CodeRabbit

* **New Features**
  * Optional "return pagination info" mode for list actions that returns { data, has_more, next_starting_after } to support next-page navigation.

* **Improvements**
  * Added optional startingAfter/endingBefore inputs for finer pagination control.

* **Bug Fixes**
  * List operations now trim to the requested limit, report accurate counts, and indicate when more results are available.

* **Chores**
  * Component and action versions bumped for release.

### DIFF
--- a/components/stripe/actions/cancel-or-reverse-payout/cancel-or-reverse-payout.mjs
+++ b/components/stripe/actions/cancel-or-reverse-payout/cancel-or-reverse-payout.mjs
@@ -4,7 +4,7 @@ export default {
   key: "stripe-cancel-or-reverse-payout",
   name: "Cancel Or Reverse A Payout",
   type: "action",
-  version: "0.1.4",
+  version: "0.1.5",
   annotations: {
     destructiveHint: true,
     openWorldHint: true,
--- a/components/stripe/actions/cancel-payment-intent/cancel-payment-intent.mjs
+++ b/components/stripe/actions/cancel-payment-intent/cancel-payment-intent.mjs
@@ -4,7 +4,7 @@ export default {
   key: "stripe-cancel-payment-intent",
   name: "Cancel A Payment Intent",
   type: "action",
-  version: "0.1.4",
+  version: "0.1.5",
   annotations: {
     destructiveHint: true,
     openWorldHint: true,
--- a/components/stripe/actions/cancel-subscription/cancel-subscription.mjs
+++ b/components/stripe/actions/cancel-subscription/cancel-subscription.mjs
@@ -4,7 +4,7 @@ export default {
   key: "stripe-cancel-subscription",
   name: "Cancel Subscription",
   description: "Cancel a subscription. [See the documentation](https://docs.stripe.com/api/subscriptions/cancel?lang=node)",
-  version: "0.0.3",
+  version: "0.0.4",
   annotations: {
     destructiveHint: true,
     openWorldHint: true,
--- a/components/stripe/actions/capture-payment-intent/capture-payment-intent.mjs
+++ b/components/stripe/actions/capture-payment-intent/capture-payment-intent.mjs
@@ -4,7 +4,7 @@ export default {
   key: "stripe-capture-payment-intent",
   name: "Capture a Payment Intent",
   type: "action",
-  version: "0.1.4",
+  version: "0.1.5",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/stripe/actions/confirm-payment-intent/confirm-payment-intent.mjs
+++ b/components/stripe/actions/confirm-payment-intent/confirm-payment-intent.mjs
@@ -5,7 +5,7 @@ export default {
   key: "stripe-confirm-payment-intent",
   name: "Confirm A Payment Intent",
   type: "action",
-  version: "0.1.4",
+  version: "0.1.5",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/stripe/actions/create-billing-meter/create-billing-meter.mjs
+++ b/components/stripe/actions/create-billing-meter/create-billing-meter.mjs
@@ -4,7 +4,7 @@ export default {
   key: "stripe-create-billing-meter",
   name: "Create Billing Meter",
   type: "action",
-  version: "0.0.3",
+  version: "0.0.4",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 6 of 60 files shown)


pipedream-component-development:

pipedream-component-development:

s0s0physm:
Thanks for the review @coderabbitai. Two responses:

**Lint errors (multiline-ternary):** Fixed in 117216184a. Extracted `count`, `noun`, and `moreSuffix` into locals before building each `$summary` string. Also cleaned up the existing `eslint-disable-next-line` comments on `list-payouts` and `list-refunds` with the same pattern.

**Version bump (suggested major / 1.0.0):** These comments were against earlier commits where the action's return shape changed unconditionally. The PR has since been refactored to use an opt-in `Return Pagination Info` boolean prop:
- Default (flag = `false`): identical behavior to before, returns `Array<T>` via `.autoPagingToArray({ limit })`. No breaking change for existing workflows.
- Opt-in (flag = `true`): returns `{ data, has_more, next_starting_after }`.

Since adding a new optional prop with no default-behavior change is a minor bump per the [component versioning guidelines](https://github.com/PipedreamHQ/pipedream/blob/master/.github/pipedream-component-guidelines.md#versioning), `0.1.4` → `0.2.0` is correct.

GTFalcao (commented):
LGTM - we should definitely test some of these calls before merging

mariano-pd:
QA Testing results:

The functionality for the new `Return Pagination Info` boolean prop works perfectly ✅

We just require a minor change. This is what Stripe official docs mention about the `limit` prop  _A limit on the number of objects to be returned. Limit can range between 1 and 100, and the default is 10._

We should change the limit (in all actions) to be 10 by default not 100, and add a validation that doesn't allow the user to input a number greater than 100. We should also add an additional validation that shows an input warning when trying to add a number less or equal to 0.

image

s0s0physm:
@mariano-pd Thanks for testing! 

Main concern is that capping the limit to 10 might silently break existing workflows that return more records. Lets keep it at 100, but I can add the validations you suggested.

s0s0physm:
@mariano-pd - Pushed updates that address most of the validations, though with one tweak on the limit cap. 
                                                                                                                                                                                                                                        
  #### On the limit validations:                                                                                                                                                                                                            
   
  Capping `Limit` at 100 universally would silently break workflows that don't set the prop and rely on auto-pagination. The default mode uses Stripe's `autoPagingToArray` to fetch up to 10000 records across pages. So the cap fires   
  only when `Return Pagination Info` is enabled, which is the only mode where Stripe's 100-per-call limit actually applies.
                                                                                                                                                                                                                                          
  #### What changed:  

  - `ConfigurationError` when `Return Pagination Info` is true and `Limit > 100`, with a message explaining how to fix it                                                                                                                 
  - `ConfigurationError` when both `Starting After` and `Ending Before` are set (Stripe rejects this anyway)
  - Per-action override of the `Limit` description so users see the dual-mode behavior at the prop level                                                                                                                                  
  - Moved `Return Pagination Info` to the app's `propDefinitions`, closing CodeRabbit's earlier refactor suggestion
  - Updated the `Return Pagination Info` description to explain the cursor-chaining pattern                                                                                                                                               
                                                                                                                                                                                                                                          
  #### What didn't change:                                                                                                                                                                                                                  
                                                                                                                                                                                                                                          
  - Default `Limit` stays at 100. Changing it to 10 Could silently cause breaking changes in the results for any workflow using the default.
  - The `≤ 0` warning is already handled by `min: 1` on the central propDefinition, which blocks input at the UI level.

mariano-pd (approved):
QA testing successful! ✅

Image

PR #20756, opened 2026-04-30
HubSpot - Get Engagements action

### LINKED ISSUE #20727: HubSpot — add Get Engagements action
HubSpot

https://developers.hubspot.com/docs/api/crm/objects

## Problem

There is no way to read the content of engagement objects (notes, tasks, calls, emails, meetings) by ID using the CRM v3 API. `list-crm-associations` returns only record IDs — agents can see a contact has 3 notes but cannot retrieve their content. `get-crm-objects` is limited to standard CRM object types and does not include engagement types.

## What's needed

A standalone **Get Engagements** action that accepts an engagement type (note, task, call, email, or meeting) and one or more object IDs, and returns full content properties. Calls `POST /crm/v3/objects/{objectType}/batch/read`.

## Notes

- `batchGetObjects` in `hubspot.app.mjs` uses `POST /crm/v3/objects/{objectType}/batch/read` and can be reused — add `ENGAGEMENT_TYPES` as the object type options
- `ENGAGEMENT_TYPES` constant exists in `common/object-types.mjs` with values: `NOTE`, `TASK`, `MEETING`, `EMAIL`, `CALL`
- Key content properties: notes → `hs_note_body`; tasks → `hs_task_subject`, `hs_task_body`; calls → `hs_call_body`; emails → `hs_email_subject`, `hs_email_text`; meetings → `hs_meeting_title`, `hs_meeting_notes`
- Required OAuth scopes (verify against app registry): `crm.objects.notes.read`, `crm.objects.tasks.read`, `crm.objects.calls.read`, `crm.objects.emails.read`, `crm.objects.meetings.read`

***[PIPE-89]***

michelle0927:
Resolves #20727 

## Summary by CodeRabbit

* **New Features**
  * Added a HubSpot action to retrieve multiple engagements by ID with selectable engagement type, optional property selection (auto-populates properties if omitted), and a concise result summary showing the retrieved count with correct singular/plural wording.

* **Chores**
  * HubSpot component package version bumped to 1.13.0.

### DIFF
--- a/components/hubspot/actions/get-engagements/get-engagements.mjs
+++ b/components/hubspot/actions/get-engagements/get-engagements.mjs
@@ -58,6 +58,7 @@ export default {
     const properties = this.properties?.length
       ? this.properties
       : (await this.hubspot.getProperties({
+        $,
         objectType: this.engagementType,
       }))?.results.map((property) => property.name);
 
--- a/components/hubspot/actions/get-engagements/get-engagements.mjs
+++ b/components/hubspot/actions/get-engagements/get-engagements.mjs
@@ -1,0 +1,97 @@
+import hubspot from "../../hubspot.app.mjs";
+import { ENGAGEMENT_TYPE_OPTIONS } from "../../common/constants.mjs";
+
+export default {
+  key: "hubspot-get-engagements",
+  name: "Get Engagements",
+  description: "Retrieves one or more engagements by their IDs. Use this action to fetch details about multiple engagements, such as their types, timestamps, and associated CRM objects. Requires engagement IDs, which you can obtain from the **List CRM Objects** action. [See the documentation](https://developers.hubspot.com/docs/api-reference/legacy/crm/objects/objects/batch/get-objects)",
+  version: "0.0.1",
+  annotations: {
+    destructiveHint: false,
+    openWorldHint: true,
+    readOnlyHint: true,
+  },
+  type: "action",
+  props: {
+    hubspot,
+    engagementType: {
+      type: "string",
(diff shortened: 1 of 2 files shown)


components/hubspot/actions/get-engagements/get-engagements.mjs
```diff
@@ -0,0 +1,84 @@
...
+    hubspot,
+    engagementType: {
+      type: "string",
+      label: "Engagement Type",
+      description:
+        "One of: `notes`, `tasks`, `meetings`, `emails`, or `calls`. "
+        + "Defines the CRM object type listed and which properties belong in **Object Properties**.",
+      async options() {
+        return ENGAGEMENT_TYPE_OPTIONS;
+      },
```
GTFalcao, line 25: Can this be a simple options property instead?

```suggestion
      options: ENGAGEMENT_TYPE_OPTIONS,
```

GTFalcao (commented):
Left one minor comment

components/hubspot/actions/get-engagements/get-engagements.mjs
michelle0927: It could, but I thought it would be better if the other props (`engagementIds` and `properties`) are updated when `engagementType` is changed. It has to use async options to accomplish that.

GTFalcao: Ah I see, not very intuitive - I'd think regular options (or any prop without options even) would also trigger a refresh on subsequent props that depend upon them.

GTFalcao (approved):

PR #20759, opened 2026-04-30
Tableau - List Sites and List Workbooks actions

### LINKED ISSUE #20730: Tableau — add List Sites action
Tableau

https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_site.htm

## Problem

The `siteId` prop returns only the single site from the authenticated user's current session — it cannot return more than one site. There is no standalone action that exposes site listing, making multi-site Tableau environments unsupported.

## What's needed

A standalone **List Sites** action that returns available site IDs and names via the current session endpoint. Note: the Tableau REST API's `GET /api/{version}/sites` endpoint requires server admin privileges for full site enumeration — a standalone action would be limited to returning the current user's site.

## Notes

- `siteId` prop calls `this.getCurrentSession()` → `GET /sessions/current`, returning only the session's site
- The `GET /sites` endpoint lists all sites but requires system administrator access
- This is a prerequisite for the List Workbooks action, which requires a site ID as input

***[PIPE-69]***

michelle0927:
Resolves #20730 

## Summary by CodeRabbit

* **New Features**
  * Added "List Sites" action with pagination and permission-aware behavior
  * Added "List Workbooks" action with filtering, sorting, fields, and pagination
  * Added a new app method to support site listing for Tableau actions

* **Chores**
  * Bumped component and package versions (Tableau components and package updated)

### DIFF
--- a/components/tableau/actions/create-project/create-project.mjs
+++ b/components/tableau/actions/create-project/create-project.mjs
@@ -4,7 +4,7 @@ export default {
   key: "tableau-create-project",
   name: "Create Project",
   description: "Creates a project on the specified site. You can also create project hierarchies by creating a project under the specified parent project on the site. [See the documentation](https://help.tableau.com/current/api/rest_api/en-us/REST/rest_api_ref_projects.htm#create_project)",
-  version: "0.0.3",
+  version: "0.0.4",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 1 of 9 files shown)


GTFalcao (dismissed):
LGTM!

mariano-pd:
For **list-sites** action we should add a validation that if the page number doesn't exist then show a message to the user, and we should add in the page size label, that the maximum is 1000.

image
<br/>
For **list-workbooks** we should add the missing props for filter-expression, sort-expression and field-expression, and we should add in the page size label, that the maximum is 1000. Perhaps we can also display a prettier error message to the user when a page number doesn't exists. 

image

image

mariano-pd:
I ran into an error testing **list-sites** about auth expired. It is weird because it has only been 3 hours since I first connected the account.

image

michelle0927:
> I ran into an error testing **list-sites** about auth expired. It is weird because it has only been 3 hours since I first connected the account.
> 

@mariano-pd I ran into that as well and had to reconnect the account. Maybe it's because we're using the same account so it's not always looking at the latest oauth token? It's not something that can be addressed in the component though, so I think this is ready for release?

mariano-pd (approved):
QA testing successful! ✅

PR #20766, opened 2026-05-01
Jotform - List Forms

### LINKED ISSUE #20751: Jotform — add List Forms action
Jotform

https://api.jotform.com/docs/#user-forms

## Problem

There is no standalone action to list Jotform forms. The `formId` prop is accessible via `retrieve_options` within existing actions, but agents have no dedicated tool to list forms directly.

## What's needed

A standalone **List Forms** action that calls `GET /user/forms` and returns form IDs and titles. Optionally accepts a `teamId` to filter by Jotform Enterprise team.

## Notes

- `formId` prop calls `this.getForms()` → `GET /user/forms`
- The `getForms` method already exists in `jotform.app.mjs` and can be reused
- Optional `teamId` prop enables filtering by Jotform Enterprise team

***[PIPE-80]***

michelle0927:
Resolves #20751 

## Summary by CodeRabbit

* **New Features**
  * Added "List Forms" action to retrieve all Jotform forms with pagination and summary.

* **Documentation**
  * Clarified action documentation link wording for easier reference.

* **Chores**
  * Bumped Jotform package and source/component versions.
  * Incremented several action versions to reflect updates.

### DIFF
--- a/components/jotform/actions/list-forms/list-forms.mjs
+++ b/components/jotform/actions/list-forms/list-forms.mjs
@@ -1,4 +1,5 @@
 import common from "../common/common.mjs";
+import { ConfigurationError } from "@pipedream/platform";
 
 export default {
   ...common,
@@ -25,10 +26,16 @@ export default {
         common.props.jotform,
         "max",
       ],
+      description: "Maximum number of forms to return. Default is 20. Maximum is 1000.",
       optional: true,
+      min: 1,
+      max: 1000,
     },
   },
   async run({ $ }) {
+    if (this.max > 1000 || this.max < 1) {
+      throw new ConfigurationError(`\`Max\` is set to "${this.max}" but must be a positive integer between 1 and 1000.`);
+    }
     const params = {
       $,
       max: this.max,
--- a/components/jotform/actions/list-forms/list-forms.mjs
+++ b/components/jotform/actions/list-forms/list-forms.mjs
@@ -1,0 +1,47 @@
+import common from "../common/common.mjs";
+
+export default {
+  ...common,
+  key: "jotform-list-forms",
+  name: "List Forms",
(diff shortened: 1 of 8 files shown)


components/jotform/actions/list-forms/list-forms.mjs
```diff
@@ -0,0 +1,47 @@
...
+        common.props.jotform,
+        "max",
+      ],
+      optional: true,
+    },
+  },
+  async run({ $ }) {
+    const params = {
+      $,
+      max: this.max,
```
GTFalcao, line 41: I think `max` needs to be inside a params object here, since the getForms method destructures its parameter as ` $, params, teamId` - this is what CodeRabbit flagged in its comment as well

```suggestion
      params: {
        max: this.max,
      }
```

GTFalcao (changes requested):
Left one comment that is important before we move forward

components/jotform/actions/list-forms/list-forms.mjs
michelle0927: Thanks. But actually, the `max` prop is used in the `paginate` method and then removed from `params`.

```
    async *paginate(resourceFn, params = {}) {
      const { max } = params;
      delete params.max;
      let count = 0;
      while (true) {
        const { content: items } = await resourceFn(params);
        for (const item of items) {
          yield item;
          count ++;
          if (count >= max) {
            break;
          }
        }
        const {
          limit = 20,
          offset = 0,
        } = params;
        if (items.length < limit || count >= max) {
          return;
        }
        params.offset = offset + limit;
      }
    },
```

GTFalcao (dismissed):

mariano-pd:
QA testing results:

Minor bug: It allows to input negative values for max items, and the label doesn't establish the limit.

image

mariano-pd (approved):
QA testing successful ✅

Image

PR #20793, opened 2026-05-05
fix(google_drive): escape search term special chars and document tokenized matching

dannyroosevelt:
## Summary

- **Bug fix**: `nameSearchTerm` was not escaped before being interpolated into the Drive query string in `commonSearchQuery.mjs`. Single quotes in file names (e.g. `O'Brien's Doc`) would generate a malformed query. Now escapes `\` and `'` per the Drive query syntax spec.
- **Bug fix**: `find-folder` had the same issue — it was escaping `'` but not `\`. Added backslash escaping to match.
- **Agent guidance**: Updated action descriptions for `find-file`, `find-folder`, `find-forms`, and `find-spreadsheets` to document that `Search Name` uses Google Drive's tokenized full-text matching. Special characters like `&` act as token boundaries, not literal characters, so agents/users should pass a distinctive word or phrase rather than the full title. Also updated the `fileNameSearchTerm` prop description in the app for the same reason.

## Context

Reported by a Connect developer: searching for files with `&` in the name (e.g. "Learn Pricing Tool & ROI calc") consistently failed on the first attempt. The `&` issue is a Google Drive API tokenization behavior (not a syntax bug), but the lack of single-quote escaping is a real correctness bug. Both are addressed here.

## Test plan

- [ ] Search for a file whose name contains a single quote — confirm it returns results instead of a Drive API error
- [ ] Search for a file whose name contains `&` using a word on one side of the `&` — confirm it finds the file
- [ ] Verify action descriptions surface the tokenization note in the UI

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced special character handling in Google Drive search operations for improved search reliability.

* **Documentation**
  * Updated search action descriptions with guidance on tokenized full-text matching and special character escaping in search queries.

### DIFF
--- a/components/google_drive/actions/find-file/find-file.mjs
+++ b/components/google_drive/actions/find-file/find-file.mjs
@@ -5,8 +5,8 @@ import googleDrive from "../../google_drive.app.mjs";
 export default {
   key: "google_drive-find-file",
   name: "Find File",
-  description: "Search for a specific file by name. [See the documentation](https://developers.google.com/drive/api/v3/search-files) for more information",
-  version: "0.1.20",
+  description: "Search for a specific file by name. The `Search Name` field uses Google Drive's tokenized full-text matching — pass a distinctive word or short phrase rather than the full title when the name contains special characters like `&` or `'`. [See the documentation](https://developers.google.com/drive/api/v3/search-files) for more information",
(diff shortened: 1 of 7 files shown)


michelle0927 (approved):
LGTM!

PR #20837, opened 2026-05-11
Coderabbit updates

GTFalcao:
Updating guidelines for naming conventions and component version bumping

## Summary by CodeRabbit

* **Documentation**
  * Clarified when component version bumps are required (only for behavior/interface/implementation changes).
  * Added guidance on semantic versioning for new components and a CI-versions subsection allowing patch bumps for non-impactful shared-file changes.
  * Extended naming conventions to require UPPER_SNAKE_CASE for module-level constants.

* **Chores**
  * Refined internal review configuration to reduce false-positive version-bump alerts.

[Review Change Stack](https://app.coderabbit.ai/change-stack/PipedreamHQ/pipedream/pull/20837)

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -19,16 +19,15 @@ reviews:
         Review this file against the Pipedream component model described in
         .github/pipedream-component-guidelines.md.
 
-        - Verify the `version` field is incremented from its previous value and the bump
-          follows semantic versioning (see the Versioning section of the guidelines):
-          new components start at `0.0.1`; bug fixes and non-breaking refactors → patch;
-          new optional props or backwards-compatible additions → minor; removed/renamed props
-          or breaking behavior changes → major; flag missing version bumps only when the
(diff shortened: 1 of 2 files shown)


michelle0927:
@GTFalcao, Historically, we've often gotten errors publishing components when we don't enforce this guideline, but I assume that's a risk we're ok with. If we're going to change this guideline, I think we should change the PR check as well so that user's aren't receiving contradictory information.

michelle0927 (approved):

michelle0927:
@GTFalcao Since this isn't merged yet, I wonder if we could add something to it regarding doc links. The tool/eval monster is regularly producing made-up doc links that return 404. Can coderabbit check that the links are valid?

GTFalcao:
I lost track of this PR. Since we have a next pass of updates incoming and the docs links is one thing that came up, I'll merge this so we have the base up-to-date

PR #21048, opened 2026-06-02
Asana - bug fixes

### LINKED ISSUE #21033: Asana — search-projects times out on workspace-scale data
Asana

https://developers.asana.com/reference/getprojects
https://developers.asana.com/docs/pagination

## Problem

Read actions in the Asana component handle pagination inconsistently, and most don't paginate at all:

- 6 actions make a single API call and return whatever comes back, with no `limit` / `offset` / `next_page` handling: `search-projects`, `search-tasks`, `search-tasks-premium`, `search-user-projects`, `search-sections`, `get-tasks-from-task-list`. Large workspaces return a "query too large" error or a silently truncated result.
- 3 actions expose `offset` / `limit` props but return one page only, with no signal to the agent that more pages exist: `list-teams`, `list-users`, `list-workspaces`.
- 1 action paginates correctly: `list-task-stories` (loops on `next_page.offset` with a `maxResults` guard, default 100).

A separate bug was found. Both `search-user-projects` (line ~47) and `getOrganizations` in `asana.app.mjs` (line ~386) pass an `async` callback to `Array.prototype.filter`. `Array.filter` doesn't wait for async results, so every item passes through regardless of the actual condition.

- `search-user-projects` returns all projects in the workspace, ignoring user membership.
- `getOrganizations` returns all workspaces, ignoring `is_organization`.

## What's needed

1. Apply the `list-task-stories` auto-paginate pattern to the 6 actions with no pagination. Loop on `next_page.offset`, respect a `maxResults` prop (default 100, configurable).
2. For `list-teams`, `list-users`, and `list-workspaces`, return a `truncated: true` flag and the next-page offset in the response when more pages exist, so the agent knows to ask for more.
3. Fix the async-filter bug in `search-user-projects` and `getOrganizations`. Use `Promise.all` + a sync filter, or await lookups serially.

## Notes

- Asana's pagination uses `offset` / `limit` with a max of 100 per page.
- `list-task-stories` is the working template: `do/while` loop, `next_page.offset` cursor, `maxResults` guard.
- Pair this fix with the analytical-reads feature work in #21032.

***[AI-14823]***

michelle0927:
Resolves #21033 

## Summary by CodeRabbit

* **New Features**
  * Added maxResults controls, improved pagination across searches and list actions, and clearer summaries; task-list retrieval now targets the user's personal My Tasks inbox.

* **Bug Fixes**
  * More reliable workspace/organization detection and streamlined workspace/team/user lookup guidance.

* **Chores**
  * Version bumps across Asana actions, sources, and the Asana component; updated prop descriptions and help text.


components/asana/actions/search-projects/search-projects.mjs
```diff
@@ -21,7 +21,7 @@ export default {
     },
     workspace: {
       label: "Workspace",
-      description: "The workspace or organization to filter projects on.",
+      description: "The workspace or organization to filter projects on. Use the **List Workspaces** action to find available workspace GIDs.",
       type: "string",
       optional: true,
```
ashwins01: looks like Asana expects a workspace when making a paginated call, since we now default to a paginated call, getting an error when a request for search-projects is made without workspace.
```
    {
      "ts": 1780656069258,
      "k": "error",
      "err": {
        "message": "{\"errors\":[{\"message\":\"Need to specify a workspace to paginate!\",\"help\":\"For more information on API status codes and how to handle them, read the docs on errors: https://developers.asana.com/docs/errors\"}]}",
        "name": "AxiosError - Request failed with status code 400",
        "isAxiosError": true,
        "code": "ERR_BAD_REQUEST",
```

components/asana/asana.app.mjs
ashwins01, line 390: Adding a query-param `?opt_fields=is_organization` will remove the need to fetch each of the workspaces returned by this endpoint below to see whether the `is_organization` value is `true` for the workspace or not. Reduces our N+1 workspace fetch to N
doc ref: https://developers.asana.com/reference/getworkspaces

components/asana/asana.app.mjs
```diff
@@ -383,13 +390,17 @@ export default {
...
+        workspaces.map(async (workspace) => {
+          const { data } = await this.getWorkspace({
+            workspaceId: workspace.gid,
+          });
+          return data?.is_organization
+            ? workspace
+            : null;
+        }),
+      );
+      return results.filter(Boolean);
```
ashwins01: Adding a query-param `?opt_fields=is_organization` in the above `getWorkspaces` call will remove the need to fetch each of the workspaces to check whether the `is_organization` value is `true` for the workspace or not. Reduces our N+1 workspace fetch to N
doc ref: https://developers.asana.com/reference/getworkspaces

ashwins01 (changes requested):
A couple of changes which we can address to optimise the overall implementation.
Can we add a `min` & `max` to `maxResults` across actions since Asana states they support limits of only `1` to `100` ?

components/asana/actions/search-user-projects/search-user-projects.mjs
```diff
@@ -35,28 +34,58 @@ export default {
...
-    });
+      hasMore = next;
+      params.offset = next?.offset;
 
-    $.export("$summary", "Successfully retrieved projects of user");
+      if (data.length === 0) break;
+
+      const membership = await Promise.all(
+        data.map(async (project) => {
+          const { data: detail } = await this.asana.getProject({
```
ashwins01: This might hit Asana's concurrency rate-limit if user has set `maxResults` to more than 50 and the `getProject` calls are fired simultaneously. Asana seems to have a max cap of 50 concurrent requests for `GET` api's, ref: https://developers.asana.com/docs/rate-limits#concurrent-request-limits

Does the pipedream-axios package internally trigger retries for 429's ? or should we chunk these so as to not hit the concurrency limits ?

components/asana/asana.app.mjs
```diff
@@ -272,6 +272,15 @@ export default {
         };
       },
     },
+    maxResults: {
+      type: "integer",
+      label: "Max Results",
+      description: "The maximum number of results to return",
+      default: 100,
+      min: 1,
+      max: 100,
```
ashwins01: Contrary to my previous comment, should we decouple `maxResults` and `limit` of each paginate call ? By keeping `max` for `maxResults` somewhere in the thousands say 9999 so it'll be the total number of results returned post auto-pagination and `limit` for each paginated api call will always be 100 (the upper-range in Asana) ?

ashwins01 (approved):
Looks great, moving to QA!

mariano-pd (approved):
QA testing successful!

PR #21049, opened 2026-06-02
Google file search — default to all drives when no drive is selected

### LINKED ISSUE #21024: Google file search — default to all drives when no drive is selected
Google Drive (helper) / Google Sheets / Google Slides / Google Docs

https://developers.google.com/workspace/drive/api/reference/rest/v3/files/list

## Problem

When no `driveId` is selected, the Google file-search helpers in `google_drive`, `google_sheets`, `google_slides`, and `google_docs` silently exclude Shared Drive content. Agents can't find Shared Drive files without pre-selecting the drive.

## What's needed

Three changes:

1. **Update `getListFilesOpts` in `google_drive/common/utils.mjs`** — when no `drive` is passed, default the request to `corpora: "allDrives"` with both `supportsAllDrives: true` and `includeItemsFromAllDrives: true`, so `files.list` returns items from My Drive and every Shared Drive the user can access. Used by `find-file`, `find-folder`, `find-forms`, `find-spreadsheets`, `list-files`, and `google_docs-find-document` (which imports this helper).

2. **Update `listSheetsOptions` in `google_sheets.app.mjs` and `listPresentationsOptions` in `google_slides.app.mjs`** — apply the same default when no `driveId` is passed.

3. **Add a `limitToMyDrive` opt-out prop** mirroring the one introduced in #20680 for `list-files`. Wire it onto the affected actions so callers (especially agents) can narrow back to a My-Drive-only scan when the broad default isn't appropriate. Reuse the same prop shape and inverted mapping (`limitToMyDrive: true` → passes `includeItemsFromAllDrives: false`).

Verified via the Connect proxy: with only `supportsAllDrives=true`, `files.list` returned My-Drive items only. Adding `includeItemsFromAllDrives=true` and `corpora=allDrives` also returned Shared-Drive items.

## Notes

- This is a behavior change. Workflows that silently exclude Shared Drives today will start matching them after this ships. Bump the minor version on each affected component and call it out in release notes.
- This pairs with #20680: that PR fixed the driveId-set branch of `getListFilesOpts` (was 403'ing on Shared Drives). This issue fixes the no-driveId branch. Together they cover both paths.
- Audit confirmed: Drive, Sheets, Slides, and Docs are the only Google apps in this repo that call `files.list`. Forms, Gmail, Calendar, Tasks, Contacts, and Photos don't surface a Drive file picker that hits this endpoint.
- Sources (`new-or-modified-files`, `new-files-instant`, etc.) also call `getListFilesOpts`. They're not in the agent path, so need to change.

***[PIPE-114]***

michelle0927:
Resolves #21024 

## Summary by CodeRabbit

* **New Features**
  * Added a "Limit to My Drive" option for file/search flows in Drive, Sheets, and Slides.

* **Bug Fixes**
  * Improved drive-search behavior to better distinguish personal vs. shared-drive queries.
  * Default drive selection now defaults to "All Drives" when left unset.

* **Chores**
  * Numerous component, action, source, and package version bumps across Google Drive, Sheets, Slides, and Docs.


components/google_drive/sources/changes-to-specific-files/changes-to-specific-files.mjs
```diff
@@ -16,7 +16,7 @@ export default {
   key: "google_drive-changes-to-specific-files",
   name: "Changes to Specific Files",
   description: "Watches for changes to specific files, emitting an event when a change is made to one of those files. To watch for changes to [shared drive](https://support.google.com/a/users/answer/9310351) files, use the **Changes to Specific Files (Shared Drive)** source instead.",
```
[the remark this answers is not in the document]
michelle0927: For new sources being added, a doc link should be included. However, if the source already exists, it doesn't need a doc link added. Applies to all sources.

components/google_drive/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/google_drive",
-  "version": "1.10.0",
+  "version": "1.11.0",
```
ashwins01, line 3: should we update this version in `google_docs` aswell ? even though the latest version of `google_drive` will get downloaded in the execution-environment ?

ashwins01 (changes requested):
The `watchedDrive`/`drive` prop is being defaulted to `My Drive` even though we want it to default to `All Drives` when no drive is selected which would cause the `corpora` to always return `user` when no drive is selected. I tried the `find-file` action and I was able to repro this issue. Seems like its worth taking another look at, would you mind taking another look at it ?

components/google_drive/sources/common-webhook.mjs
```diff
@@ -19,7 +19,6 @@ export default {
         googleDrive,
         "watchedDrive",
       ],
-      description: "Defaults to My Drive. To select a [Shared Drive](https://support.google.com/a/users/answer/9310351) instead, select it from this list.",
```
ashwins01, line 22: should we keep the description, but let it say that this defaults to all-drives including shared-drives ?

```
description: "Defaults to **All Drives** (My Drive and all Shared Drives you have access to) when no selection is made. To limit to your personal drive or a [Shared Drive](https://support.google.com/a/users/answer/9310351), select it from this list.",
```

components/google_sheets/actions/update-row/update-row.mjs
```diff
@@ -10,9 +10,9 @@ export default {
   key: "google_sheets-update-row",
   name: "Update Row",
   description: "Update a row in a spreadsheet. [See the documentation](https://developers.google.com/sheets/api/reference/rest/v4/spreadsheets.values/update)",
-  version: "0.1.19",
+  version: "0.1.21",
```
ashwins01:
```suggestion
  version: "0.1.20",
```

ashwins01 (dismissed):
Looks good to me, moving to QA!

mariano-pd (approved):
QA testing successful ✅

PR #21055, opened 2026-06-03
fix(monday): fix pagination crash when board has >25 items

### LINKED ISSUE #20978: Monday: monday-get-board-items-page crashes with TypeError for boards >25 items (cursor-loop destructure asymmetry)
## Summary

`monday-get-board-items-page` throws `TypeError: Cannot read properties of undefined (reading 'cursor')` on any Monday board that has more than one `items_page` (Monday's default page size is 25). The crash happens inside the cursor pagination loop on the very first iteration, before any rows are returned.

## Repro

1. Connect a Monday account in Pipedream.
2. Pick or create a board with **26 or more items**.
3. Trigger `monday-get-board-items-page` for that board.
4. Observe the action throw `TypeError: Cannot read properties of undefined (reading 'cursor')` instead of returning items.

Verified empirically on a 28-item board.

## Stack trace (from production logs)

```
TypeError: Cannot read properties of undefined (reading 'cursor')
    at Object.run (file:///tmp/__pdg__/dist/code/<hash>/code/actions/get-board-items-page/get-board-items-page.mjs:74:23)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async file:///var/task/component_maker.mjs:196:13
    at async captureObservations (/var/task/node_modules/@lambda-v2/component-runtime/src/captureObservations...
```

The crash site (line 74) is the destructure on the second pagination call. (Modern Node would emit "Cannot destructure property 'cursor' of '…' as it is undefined" — the older V8 form above is what Pipedream's runtime surfaces, but the underlying error is the same.)

## Files

- [`components/monday/actions/get-board-items-page/get-board-items-page.mjs`](https://github.com/PipedreamHQ/pipedream/blob/master/components/monday/actions/get-board-items-page/get-board-items-page.mjs) (action, version 0.0.3 at the time of report)
- [`components/monday/common/queries.mjs`](https://github.com/PipedreamHQ/pipedream/blob/master/components/monday/common/queries.mjs) (the GraphQL query in use)

## Why it affects every multi-page board

The action's `async run()` makes two destructures of the same `listBoardItemsPage` GraphQL response, but in *different* shapes:

**Initial call (lines 60–66, works):**
```js
const {
  data: {
    boards: [
      { items_page: pageItems },
    ],
  },
} = response;
```

**Cursor-loop call (lines 70–77, throws):**
```js
while (cursor) {
  const {
    data: {
      boards: {
        items_page: {
          cursor: nextCursor, items: nextItems,
        },
      },
    },
  } = await this.monday.listBoardItemsPage({ boardId: +this.boardId, cursor });
```

The `listBoardItemsPage` GraphQL query selects `boards (ids: [\$boardId]) { items_page { … } }`, which always returns `data.boards` as an array per Monday's schema (`boards: [Board]`). The initial destructure treats it as an array (`boards: [{…}]`) — correct. The cursor-loop destructure treats it as an object (`boards: {…}`) — so `data.boards.items_page` reads as `undefined` on an array, and the next nested step destructures `cursor` off `undefined`, which throws.

## Impact

Affects any board with >25 items, which covers most production Monday boards in real-world use (project trackers, task lists, CRMs, etc.). Sub-25-item boards work fine because the cursor loop never enters.

I'm not aware of any open issue tracking this — the existing Monday issues / PRs I found (#18610, #18714, #20591, #20601) cover the initial implementation and a queryParams addition, but none touched the cursor loop.

Happy to provide additional logs or test against a fix.

michelle0927:
The cursor loop in get-board-items-page destructured `data.boards` as an object but the Monday API always returns it as an array. This caused a TypeError ("Cannot read properties of undefined (reading 'cursor')") on any board with more than 25 items (the default page size).

Resolves PipedreamHQ/pipedream#20978

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed pagination response handling for board items retrieval.

* **Chores**
  * Updated component versions.


michelle0927:
@coderabbit review

ashwins01 (approved):
LGTM, moving to QA.

mariano-pd (approved):
QA testing approved ✅

PR #21056, opened 2026-06-03
Microsoft Outlook Calendar - add deleted calendar event trigger

### LINKED ISSUE #21054: microsoft_outlook_calendar: add deleted calendar event trigger
## Problem
The `microsoft_outlook_calendar` app has instant triggers for new and updated calendar events, but no option for deleted events.

## Scope
Add a new source: `microsoft_outlook_calendar-deleted-calendar-event` ("Calendar Event Deleted (Instant)").
- API doc: https://learn.microsoft.com/en-us/graph/api/subscription-post-subscriptions
    - Add`userId` (optional) prop: target another user's events for delegated/admin scenarios. When provided, the subscription resource is `/users/{userId}/events`. Defaults to `/me/events`.
    - [Outlook event subscriptions](https://learn.microsoft.com/en-us/graph/change-notifications-overview) only support `/me/events` and `/users/{id}/events`. 

## Notes
- Follows the same pattern as `new-calendar-event` and `updated-calendar-event`: uses the shared Microsoft Graph subscription lifecycle in `components/microsoft_outlook/sources/common/common.mjs`.
- The deleted event cannot be re-fetched after deletion, so the emit should use `resourceData.id` from the notification instead of calling `getCalendarEvent`. Since the event is gone, the emit shape needs explicit design: at minimum include `eventId` (from `resourceData.id`), notification timestamp, and `changeType`.
- Trigger description should state plainly what gets emitted: a Microsoft Graph change notification containing the deleted event's `id` and metadata. The full event body is not retrievable post-deletion, and there is no way to scope the trigger to a specific calendar.
- Prop labels and descriptions should be clear enough for an agent to configure without external context.

[WDINT-432]

michelle0927:
Resolves #21054 

Adds a new instant source that emits when a Calendar event is deleted. Emits notification metadata (eventId, changeType, subscriptionId, tenantId, notificationTimestamp) since deleted events cannot be fetched after deletion.

## Summary by CodeRabbit

* **New Features**
  * Detects deleted events in Microsoft Outlook calendar integration, enabling workflows to trigger on event deletions.
* **Bug Fixes**
  * Prevents errors during calendar hook deactivation when a stored hook identifier is missing.
* **Chores**
  * Bumped Microsoft Outlook Calendar component version (v8.4.0 → v8.5.0) and Microsoft Outlook package (v1.13.0 → v1.13.1); minor version bumps for several Outlook sources.

### DIFF
--- a/components/microsoft_outlook/package.json
+++ b/components/microsoft_outlook/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/microsoft_outlook",
-  "version": "1.13.0",
+  "version": "1.13.1",
   "description": "Pipedream Microsoft Outlook Components",
   "main": "microsoft_outlook.app.mjs",
   "keywords": [
--- a/components/microsoft_outlook/sources/common/common.mjs
+++ b/components/microsoft_outlook/sources/common/common.mjs
@@ -67,6 +67,7 @@ export default {
     },
     async deactivate() {
       const hookId = this.db.get("hookId");
+      if (!hookId) return;
       await this.microsoftOutlook.deleteHook({
         hookId,
       });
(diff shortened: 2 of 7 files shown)


GTFalcao (approved):

mariano-pd (approved):
QA testing successful! ✅

PR #21153, opened 2026-06-12
Bug fix to Notion create-database action

### LINKED ISSUE #21102: [BUG] Notion "Create Database" action: bare-string property values produce an opaque 400 validation_error (parseObject silently passes invalid input through)
**Describe the bug**
The Notion **Create Database** action (`notion-create-database`, version `0.1.5`) passes the user-supplied `properties` object straight to Notion's `databases.create` endpoint with no validation or schema transformation. When a property's value is a bare string (the property _type_ name, e.g. `"title"`, `"select"`) instead of a property-schema object (e.g. `{"title": {}}`), the action does not catch or correct it. Instead, `common/utils.mjs` → `parseObject()` attempts `JSON.parse("title")`, the parse throws, and the catch block silently returns the original string unchanged. That malformed value reaches Notion and triggers a hard `400 validation_error` on `body.initial_data_source.properties.*` that cannot be worked around from the action's configuration UI.

Relevant code:

- `components/notion/actions/create-database/create-database.mjs` — builds `initial_data_source.properties:
utils.parseObject(this.properties)` and forwards it directly to `notion.createDatabase()`.
- `components/notion/common/utils.mjs` — `parseObject()` string branch:
```javascript
if (typeof obj === "string") {
  try {
    return JSON.parse(obj);
  } catch {
    return obj;   // <-- silently returns the invalid bare string
  }
}
```
- `components/notion/notion.app.mjs` — `createDatabase()` forwards the payload to `databases.create()` with no further shaping.

**To Reproduce**
1. Add the Notion **Create Database** action to a workflow and connect a Notion account.
2. Set **Parent Page** to any page you can edit.
3. In the **Properties** field, enter an object whose values are property type _names_ as strings (an intuitive mistake given the prop's free-form `object` type), e.g.:
```json
{
  "Name": "title",
  "Status": "select",
  "Notes": "rich_text"
}
```
4. Run the action.
5. See the action fail with a Notion `400 validation_error` referencing `body.initial_data_source.properties.Name` (and the other keys), instead of any guidance that the value should be an object such as `{"title": {}}`.

**Expected behavior**
Either:
- The action validates the properties input and returns a clear, actionable error explaining that each value must be a property-schema object (e.g. `{"title": {}}`), and/or
- `parseObject()` does not silently swallow a failed `JSON.parse` of a property value — a value that cannot be parsed into the required object shape should surface as a descriptive configuration error rather than being passed through to Notion as an invalid payload, and/or
- The action accepts a friendlier shorthand (property name → type name) and builds the correct Notion property-schema object itself.

At minimum, a malformed `properties` value should not result in an opaque Notion 400 with no hint about the required shape.

**Screenshots**
N/A (configuration + error output can be added if helpful).

**Additional context**
The core issue is the combination of (a) a free-form `properties` object prop with no validation, and (b) `parseObject()`'s `catch` block returning the original string on JSON-parse failure. Together these turn a common input mistake into an unhelpful 400. Notion's create-database endpoint requires each property value to be a [property-schema object](https://developers.notion.com/reference/property-schema-object) keyed by type (e.g. `{"title": {}}`, `{"select": {"options": [...]}}`), never a bare type-name string.

michelle0927:
Resolves #21102 

**Summary**
- parseObject() in common/utils.mjs silently returns bare strings when JSON.parse fails, so user-supplied property values like "title" or "select" passed through unchanged to Notion's API
- Notion's databases.create endpoint requires property values to be schema objects (e.g. {"title": {}}) — bare strings cause an opaque 400 validation_error with no helpful message
- After parsing this.properties, the action now normalizes any remaining bare-string values into proper property schema objects ("title" → {"title": {}}) before sending to Notion

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved property handling for Notion database creation to ensure correct formatting of string and object-type properties.

### DIFF
--- a/components/notion/actions/create-database/create-database.mjs
+++ b/components/notion/actions/create-database/create-database.mjs
@@ -7,7 +7,7 @@ export default {
   key: "notion-create-database",
   name: "Create Database",
   description: "Create a database and its initial data source. [See the documentation](https://developers.notion.com/reference/database-create)",
-  version: "0.1.7",
+  version: "0.1.8",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
@@ -37,6 +37,23 @@ export default {
     },
   },
   async run({ $ }) {
+    const parsedProperties = utils.parseObject(this.properties);
+    const properties = parsedProperties && typeof parsedProperties === "object"
+      ? Object.fromEntries(
+        Object.entries(parsedProperties).map(([
+          key,
+          value,
+        ]) => [
+          key,
+          typeof value === "string"
+            ? {
+              [value]: {},
+            }
+            : value,
+        ]),
+      )
+      : parsedProperties;
+
     const response = await this.notion.createDatabase({
       parent: {
         type: "page_id",
@@ -51,7 +68,7 @@ export default {
         },
       ],
       initial_data_source: {
-        properties: utils.parseObject(this.properties),
+        properties,
       },
     });
 
--- a/components/notion/actions/create-database/create-database.mjs
+++ b/components/notion/actions/create-database/create-database.mjs
@@ -43,14 +43,33 @@ export default {
         Object.entries(parsedProperties).map(([
           key,
           value,
-        ]) => [
-          key,
-          typeof value === "string"
-            ? {
-              [value]: {},
+        ]) => {
+          if (typeof value === "string") {
+            return [
+              key,
+              {
+                [value]: {},
+              },
+            ];
+          }
+          // Normalize { type: "X" } objects missing their type-key: { type: "checkbox" } → { type: "checkbox", checkbox: {} }
+          if (value && typeof value === "object" && "type" in value) {
+            const typeKey = value.type;
+            if (typeKey && typeof typeKey === "string" && !(typeKey in value)) {
+              return [
+                key,
+                {
+                  ...value,
+                  [typeKey]: {},
+                },
+              ];
             }
-            : value,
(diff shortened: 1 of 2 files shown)


michelle0927:
## Eval Results

Ran 4 evals against the private-registry build of this PR using `claude-haiku-4-5-20251001`. All pass.

| # | Eval | Status | Tool Calls |
|---|---|---|---|
| 11 | Create Database With String-Shorthand Property Types | ✅ PASS | 2 |
| 12 | Create Database With Select and Number Properties | ✅ PASS | 2 |
| 13 | Create Database With Date and URL Properties | ✅ PASS | 2 |
| 14 | Create Database With Multi-Select and People Properties | ✅ PASS | 2 |

**4/4 passing** · avg 2 tool calls · 30s total

### What was tested

Each eval prompts the model to create a Notion database with a different mix of property types, using the natural shorthand the model reaches for (`{ type: "title" }`, `"checkbox"`, etc.). The evals verify both that the API call succeeds and that the database appears in a follow-up Notion search.

**Evals 11–13** were already covered by the original PR fix (bare-string normalization). **Eval 14** (multi_select + people) exposed a second gap: the model also sends `{ type: "X" }` objects (with a `type` key but no `X: {}` key), which Notion rejects. A follow-on commit extends the normalization to cover that pattern.

ashwins01 (approved):
Looks good to me, moving to QA!

PR #21179, opened 2026-06-16
Fixing Outlook trigger datetime sort

### LINKED ISSUE #21159: [BUG] Shared Microsoft Inbox Trigger timestamp bug causing dropped emails
**Describe the bug**
The trigger uses "sendDateTime" instead of "receivedDateTime" for advancing the cursor. This causes emails that have a delayed receive time to be skipped if another email arrives earlier (this can be the case in emails with large attachments). 

**To Reproduce**
Tough to reproduce as it is a race condition in Microsoft world but this is the architecture of the bug:
1. Send email with very large attachments that will take some time to process
2. Send email with no attachments immediately after 
3. The email with no attachments will be processed, sentTime will advance
4. The email sent in 1 will arrive, but will be skipped due to not having an earlier sentDateTime

**Expected behavior**
Email 1 should be processed when it arrives.

GTFalcao:
Closes #21159 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved new email detection for shared folders by using the message received date for filtering and ordering, improving polling accuracy and checkpointing behavior.
* **Chores**
  * Bumped the Microsoft Outlook package version for the shared-folder email source.

### DIFF
--- a/components/microsoft_outlook/package.json
+++ b/components/microsoft_outlook/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/microsoft_outlook",
-  "version": "1.14.0",
+  "version": "1.14.1",
   "description": "Pipedream Microsoft Outlook Components",
   "main": "microsoft_outlook.app.mjs",
   "keywords": [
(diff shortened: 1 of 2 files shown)


vetrivigneshwaran (approved):
LGTM

PR #21177, opened 2026-06-16
microsoft_outlook: find-email typed triage filters and folder support

michelle0927:
Resolves #21175 

## Summary by CodeRabbit

* **New Features**
  * Added actions to retrieve a specific mail folder and a shared folder.
  * Added “List Shared Folders” with optional exact name filtering, recursion, hidden-folder control, and pagination.
  * Enhanced “Find Email” with additional filters (sender, received date range, importance, flagged, and attachments) and expanded folder scoping.
* **Improvements**
  * Updated “List Folders” and shared folder listing to support exact Display Name matching.
  * Added a Folder ID input for targeting a specific mail folder when listing/counting messages.
* **Chores**
  * Version bumps across multiple Outlook actions, sources, and the Microsoft Outlook package.

PR #21191, opened 2026-06-17
microsoft_outlook: include total count in find-email list responses

### LINKED ISSUE #21176: microsoft_outlook: include total count in find-email list responses
When \`find-email\` returns a paginated list, \`\$summary\` reflects the page size, not the total. An agent seeing "Found 25 messages" has no way to know there are hundreds more.

\`countOnly\` doesn't cover this. It returns \`{ count: N }\` instead of a list, so an agent that needs both has to make two separate calls.

The Graph API returns \`@odata.count\` alongside the first page when \`\$count=true\` is set. Adding that to the initial list request lets the response include a \`totalCount\` field next to the messages array.

**Ask**

Add \`\$count=true\` and \`ConsistencyLevel: eventual\` to list requests. Return \`totalCount\` alongside the messages array so agents know upfront whether there are more results than the page returned.

---

Linear: WDINT-437/WDINT-444

michelle0927:
Resolves #21176 

To be merged after https://github.com/PipedreamHQ/pipedream/pull/21177.

## Release Notes

* **New Features**
  * Enhanced **Find Email** to better support property-scoped search and return total counts when available during listing.
  * Updated Microsoft Outlook message pagination to include total-count metadata when the API provides it.

* **Documentation**
  * Expanded **Get Message** action descriptions with clearer `includeAttachments` guidance and an example workflow.

* **Chores**
  * Bumped versions across multiple Microsoft Outlook actions/components and the Microsoft Outlook package (metadata-only updates).


components/microsoft_outlook/actions/find-email/find-email.mjs
```diff
@@ -22,7 +22,7 @@ export default {
     + " Example: `find-email(isRead=false, folderScope=\"inbox\", countOnly=true)` → `{ count: 47 }` for unread inbox count."
     + " Example: `find-email(search=\"Eval-Festivus\", folderScope=\"inbox\", maxResults=5)` → array of messages with `id`, `subject`, `from`, `receivedDateTime`, `isRead` (no body — call Get Message next for body)."
     + " [See the documentation](https://learn.microsoft.com/en-us/graph/api/user-list-messages)",
-  version: "0.4.0",
+  version: "0.5.0",
```
GTFalcao:
```suggestion
  version: "1.0.0",
```
Response schema is a breaking change, we have to do a major bump - CodeRabbit was right here

GTFalcao (approved):

mariano-pd (approved):
QA testing successful!

PR #21232, opened 2026-06-23
LinkedIn API version update

### LINKED ISSUE #20457: Updated LinkedIn and LinkedIn Ads API version to latest.
LinkedIn sunsets their API versions a year after they are released. To remain compliant with their changes, we should assess versions quarterly. 

See here for more details:
https://learn.microsoft.com/en-us/linkedin/marketing/integrations/migrations?view=li-lms-2024-11

GTFalcao:
Closes #20457 

I'm updating both at once because the constant is not shared among them, so it's not necessary to split the PR into two in this case

Also verified the changes between these API versions do not impact the components we have, so safe to ship the update

## Summary by CodeRabbit

* **Chores**
  * Updated LinkedIn and LinkedIn Ads integration component versions across multiple actions and sources.
  * Refreshed the integration header/version value used for API requests.
  * Bumped the LinkedIn Ads package version to keep the component set aligned for continued compatibility and stability.


michelle0927 (approved):
Looks good, as long as you're sure the version update doesn't require any changes to the components. Ready for QA!

PR #21254, opened 2026-06-26
Spotify rate limit improvements

GTFalcao:
Reworked the `new-track-by-artist` trigger so that it no longer fetches tracks for albums that had no change in the number of tracks, on every single run. It still does this once a day as a fallback, in case an album had changes without changing the amount of tracks (rare).

Albums that had the amount of tracks change since the last run always immediately retrieve the tracks.

Also adjusted the app-wide retry logic to use delays of 5-10-15 seconds for the 1st/2nd/3rd retry, respectively (previously it was 0-5-10 seconds, likely due to an oversight)

## Summary by CodeRabbit

* **New Features**
  * Enhanced artist track updates to more accurately detect changes and refresh listings with less frequent full rescans.
  * Improved album/track fetching by deduplicating album IDs and using more reliable pagination to expand coverage.
* **Bug Fixes**
  * Improved handling of Spotify HTTP 429 responses by adjusting retry timing and when requests stop retrying.
* **Chores**
  * Bumped Spotify action/source versions (and the Spotify package version) for compatibility and consistent updates.


michelle0927 (approved):
LGTM! Ready for QA.

PR #21266, opened 2026-06-26
Fixing hubspot new-deal-property-change trigger

### LINKED ISSUE #21267: [BUG] HubSpot "New Deal Property Change" initial deployment emits
This HubSpot trigger currently seems to not emit the initial events on the `deploy()` hook as is the correct path (to honor the `emit_on_deploy` parameter when deploying the source).

We need to adjust this so the trigger emits initial events (up to a max amount) and advances the cursor for the next event emission, all in the `deploy()` hook rather than the first `run()`.

GTFalcao:
Now correctly emits retroactive events on deploy rather than on first run (so that the opt-out parameter of emitting on deploy is honored when set)

Closes #21267 

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved initial sync to cap how many recent deal updates are emitted during deploy/backfill, reducing large backfill bursts.
  * Refined cursor behavior so first-time runs advance correctly even when no updated deals are found, preventing repeated reprocessing.
  * Improved event emission logic to reliably sort and limit initial events, and to update the cursor based on emitted timestamps.

* **Chores**
  * Bumped package version(s).

### DIFF
--- a/components/hubspot/package.json
+++ b/components/hubspot/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/hubspot",
-  "version": "1.18.0",
+  "version": "1.18.1",
   "description": "Pipedream Hubspot Components",
   "main": "hubspot.app.mjs",
   "keywords": [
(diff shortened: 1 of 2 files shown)


vetrivigneshwaran (approved):
LGTM

PR #21284, opened 2026-06-29
fix(sharepoint): carry per-item driveId through file picker

### LINKED ISSUE #21280: SharePoint — download-files / retrieve-file-metadata fail for programmatic callers
SharePoint

https://learn.microsoft.com/en-us/graph/api/driveitem-get

## Problem

SharePoint Download Files and Retrieve File Metadata fail for programmatic callers with "ObjectHandle is Invalid" when the file IDs passed don't belong to the drive the action is querying. Both actions require a driveId and fetch every file via GET /drives/{driveId}/items/{fileId} (getDriveItem in sharepoint.app.mjs), with no check that the file IDs belong to that drive. A valid item ID paired with a wrong or mismatched driveId returns a 400 and the action throws "Failed to fetch all selected files." This is easy to hit for an agent because a Graph item ID carries no drive context, and the driveId and file IDs are independent inputs.

## What's needed

Make file resolution tolerant of plain item IDs that come from search or list results. Two workable options within the current code: make driveId optional so resolution falls through to the existing /sites/{siteId}/drive/items/{id} default-drive path, and/or carry each selected item's own parentReference.driveId through the picker so selections resolve against their own drive regardless of the top-level driveId. Graph cannot resolve a bare item ID with no drive context, so a drive or site reference is still required.

## Notes

- Failing path: getDriveItem GET /drives/{driveId}/items/{fileId} (sharepoint.app.mjs), called per file from the shared file-picker base.
- The shared file-picker base is SharePoint-only; OneDrive does not use it.
- Both download-files and retrieve-file-metadata go through the same getDriveItem call, so a fix in the shared base covers both.

***[PIPE-122]***

michelle0927:
## Summary

- Carries each selected item's `parentReference.driveId` through the SharePoint file picker so `Download Files` and `Retrieve File Metadata` resolve each file against its own drive instead of the top-level `driveId` prop.
- Falls back to the top-level `driveId` when the per-item value is absent; when both are absent, `getDriveItem` falls through to the existing `/sites/{siteId}/drive/items/{id}` default-drive path.
- Versions bumped: `sharepoint-download-files` 0.0.4 → 0.0.5, `sharepoint-retrieve-file-metadata` 0.0.4 → 0.0.5, `@pipedream/sharepoint` 0.10.0 → 0.10.1.

Both actions go through the same shared `file-picker-base.mjs`, so the fix lands in one place and covers both.

Closes #21280

## Why

Both actions previously called `GET /drives/{driveId}/items/{fileId}` with the top-level `driveId` for every selection. A valid Graph item ID paired with a mismatched `driveId` returns "ObjectHandle is Invalid" (404), which is easy for an agent or programmatic caller to hit since Graph item IDs carry no drive context.

## Test plan

- [x] Happy-path evals via the MCP eval harness — `sharepoint-download-files` and `sharepoint-retrieve-file-metadata` exercised end-to-end (site → drive → folder → file → action), 3/3 pass.
- [x] Direct MCP probe with `driveId` divergence (top-level `driveId` deliberately set to a different site's drive while the per-item JSON carries the correct drive). Result table:

  | Scenario | top-level `driveId` | per-item `driveId` | result |
  |---|---|---|---|
  | Correct top-level, no per-item | correct | absent | pass |
  | Wrong top-level, correct per-item (the fix) | wrong | correct | **pass** |
  | Wrong top-level, no per-item (old behavior) | wrong | absent | **fail**: `The resource could not be found` |

  The third row is exactly the bug from the issue; the second row shows the fix recovers from it.

## Summary by CodeRabbit

* **New Features**
  * SharePoint file and folder selections now include drive context, improving downstream handling and metadata accuracy.
  * File metadata retrieval now uses the correct drive per selected item when available, with a fallback to the configured drive.

* **Chores**
  * Updated versions across SharePoint actions, sources, and the SharePoint package to the latest release levels.

### DIFF
--- a/components/sharepoint/actions/create-folder/create-folder.mjs
+++ b/components/sharepoint/actions/create-folder/create-folder.mjs
@@ -4,7 +4,7 @@ export default {
   key: "sharepoint-create-folder",
   name: "Create Folder",
   description: "Create a new folder in SharePoint. [See the documentation](https://learn.microsoft.com/en-us/onedrive/developer/rest-api/api/driveitem_post_children?view=odsp-graph-online)",
-  version: "0.0.8",
+  version: "0.0.9",
   type: "action",
   annotations: {
     destructiveHint: false,
--- a/components/sharepoint/actions/create-item/create-item.mjs
+++ b/components/sharepoint/actions/create-item/create-item.mjs
@@ -4,7 +4,7 @@ export default {
   key: "sharepoint-create-item",
   name: "Create Item",
   description: "Create a new item in Microsoft Sharepoint. [See the documentation](https://learn.microsoft.com/en-us/graph/api/listitem-create?view=graph-rest-1.0&tabs=http)",
-  version: "0.0.14",
+  version: "0.0.15",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/sharepoint/actions/create-link/create-link.mjs
+++ b/components/sharepoint/actions/create-link/create-link.mjs
@@ -5,7 +5,7 @@ export default {
   key: "sharepoint-create-link",
   name: "Create Link",
   description: "Create a sharing link for a DriveItem. [See the documentation](https://docs.microsoft.com/en-us/graph/api/driveitem-createlink?view=graph-rest-1.0&tabs=http)",
-  version: "0.0.10",
+  version: "0.0.11",
   type: "action",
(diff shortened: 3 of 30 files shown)


GTFalcao (approved):

mariano-pd (approved):
QA testing successful!

PR #21286, opened 2026-06-30
Google Slides - 'Get Presentation' action

### LINKED ISSUE #21279: Google Slides — add Get Presentation action
Google Slides

https://developers.google.com/slides/api/reference/rest/v1/presentations/get

## Problem

Google Slides has no action to get a presentation by ID. The closest action, Find Presentation, only resolves a presentation through the file picker and returns the entire presentation. An agent that already has a presentation ID and only needs the title or basic metadata has no way to fetch it without pulling the whole presentation.

## What's needed

A new Get Presentation action that accepts a presentation ID and returns the presentation, with an option to limit the response to specific fields (for example title and metadata only) so it stays lightweight. The app already has a getPresentation method and a presentationId prop, and presentations.get supports a fields mask. The presentationId prop currently uses a drive picker, so the action should also accept a plain ID or URL.

## Notes

- Endpoint: presentations.get; supports a `fields` partial-response mask.
- Find Presentation does not search by name; it already gets by ID via the picker and returns the full payload, so this is a lighter, ID-first variant.
- Mirror the conventions in google_docs Get Document.
- Related: WDINT-438 (Google Docs metadata/fields option), same request family.

***[PIPE-118]***

GTFalcao:
Clsoes #21279

## Summary by CodeRabbit

* **New Features**
  * Added a **Get Presentation** action for Google Slides.
  * Retrieve presentations using either an ID or a full editable Slides URL.
  * Supports optional **field selection** to limit returned presentation details.
* **Bug Fixes**
  * Published (non-editable) Slides URLs are now rejected with a clear configuration error.
* **Chores**
  * Updated version numbers across the Google Slides component, related actions, and the new presentation source (including component version **0.5.0**).

### DIFF
--- a/components/google_slides/actions/create-image/create-image.mjs
+++ b/components/google_slides/actions/create-image/create-image.mjs
@@ -4,7 +4,7 @@ export default {
   key: "google_slides-create-image",
   name: "Create Image",
   description: "Creates an image in a slide. [See the documentation](https://developers.google.com/workspace/slides/api/reference/rest/v1/presentations/request#CreateImageRequest)",
-  version: "0.0.4",
+  version: "0.0.5",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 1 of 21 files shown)


michelle0927 (approved):
LGTM!

mariano-pd (approved):
QA testing successful!

PR #21287, opened 2026-06-30
Fixing readOnlyHint on several components

### LINKED ISSUE #21281: Mark read-only MCP actions as readOnlyHint: true across multiple apps
Multiple — AWS (DynamoDB Query, DynamoDB Scan, Redshift Query Database), Supabase (Select Row), Notion (Query Database), Bitbucket (Get Issue), Coda (List Docs), Google Chat (Get Message), Sentry (List Issue Events), Typeform (Get Form)

N/A — this is an MCP annotation correction, not an API change.

## Problem

These actions only read data but are annotated readOnlyHint: false, so MCP clients treat them as state-changing and surface unnecessary approval steps. That adds friction for agents calling tools that don't change anything.

## What's needed

Flip readOnlyHint from false to true on each of the components listed above. readOnlyHint is set per component in each action's annotations, so each is a one-line change plus a version bump. The existing destructiveHint: false and openWorldHint: true values are correct and stay as-is.

## Notes

- Redshift Query Database is read-only by design but builds SQL from string fragments, so the read-only guarantee rests on intended use; still appropriate to mark readOnlyHint: true.

***[PIPE-121]***

GTFalcao:
Closes #21281 

## Summary by CodeRabbit

* **New Features**
  * Updated several app integrations so more data retrieval actions are now marked as read-only, improving how they’re surfaced in the product.
* **Bug Fixes**
  * Refreshed action and package versions across AWS, Bitbucket, Coda, Google Chat, Sentry, Supabase, and Typeform integrations.
  * No runtime behavior was changed for the updated actions.

### DIFF
--- a/components/aws/actions/dynamodb-query/dynamodb-query.mjs
+++ b/components/aws/actions/dynamodb-query/dynamodb-query.mjs
@@ -12,11 +12,11 @@ export default {
     The query operation finds items based on primary key values.
     [See docs](https://docs.aws.amazon.com/AWSJavaScriptSDK/v3/latest/clients/client-dynamodb/classes/querycommand.html)
   `),
-  version: "0.5.3",
+  version: "0.5.4",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
-    readOnlyHint: false,
+    readOnlyHint: true,
   },
   type: "action",
   props: {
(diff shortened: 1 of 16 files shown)


vetrivigneshwaran (approved):
LGTM

PR #21295, opened 2026-07-01
[21272] fix(trello): make "board" optional prop

### LINKED ISSUE #21272: [BUG] Trello MCP "Get Card" requires unused "board" prop
**Describe the bug**
The Trello **Get Card** action (`components/trello/actions/get-card/get-card.mjs`, v0.2.5) requires a `board` prop that the action never uses. The `run()` method calls `this.app.getCard({ cardId: this.cardId, ... })` and only ever sends `cardId` — `this.board` is collected from the user but discarded. Because neither the action nor the `board` propDefinition in `trello.app.mjs` sets `optional: true`, `board` is mandatory.

The `board` prop exists only to scope the `cardId` async-options dropdown in the visual builder (the `cardId` propDefinition passes `(c) => ({ board: c.board })` to filter the card list). When the action is invoked programmatically — e.g. via MCP/API — the caller already has the card ID or short link, but is still forced to supply a board ID that has no effect on the result. Since a Trello card URL exposes only the short link (e.g. `https://trello.com/c/wmpPenNY`), there is no way to derive the board from it, so callers are forced to guess or iterate across every board to satisfy a required field the API call ignores.

The underlying `getCard` method issues `GET /cards/{cardId}`, and Trello's REST endpoint accepts a card ID **or short link** directly with no board context — so the board requirement is unnecessary.

**To Reproduce**

1. Invoke the Trello **Get Card** action programmatically (via MCP or the API), supplying only a valid `cardId` / card short link (e.g. `wmpPenNY`).
2. The action refuses to run, demanding a `board` value.
3. Supply any board ID to get past the required prop.
4. Observe that the resulting `GET /cards/{cardId}` call succeeds using `cardId` alone — confirming the `board` value was never needed.

**Expected behavior**
Get Card should retrieve a card from just a `cardId` (or short link), with no board required. The `board` prop should be `optional: true` — kept available to filter the card dropdown in the visual builder when present, but not blocking programmatic callers who already have the card identifier.

**Screenshots**
N/A

**Additional context**
- Affected file: `components/trello/actions/get-card/get-card.mjs` (v0.2.5).
- The `board` prop is declared via `propDefinition: [app, "board"]`; neither it nor the `board` propDefinition in `trello.app.mjs` sets `optional: true`, so it defaults to required.
- `getCard` in `trello.app.mjs`: `getCard({ cardId, ...args }) => this._makeRequest({ path: `/cards/${cardId}`, ...args })`.
- Suggested fix: mark the `board` prop as `optional: true` in `get-card.mjs`. The same pattern likely affects other Trello actions whose `cardId`/`list` pickers depend on a `board` prop that `run()` never sends to the API.

ashwins01:
## Summary

Closes https://github.com/PipedreamHQ/pipedream/issues/21272
Make `board` an optional prop
Add example `id` for `board`

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved card selection behavior so options load safely when no board is selected (no unnecessary card lookups).
  * Updated the Get Card action so the board input can be omitted without issues.
* **Chores**
  * Updated Trello component/action/source versions to the latest release metadata.


pipedream-component-development:

pipedream-component-development:

components/trello/actions/archive-card/archive-card.mjs
```diff
@@ -18,6 +18,7 @@ export default {
         app,
         "board",
       ],
+      optional: true,
```
vetrivigneshwaran: cardId still depends on board via `app, "cards", (c) => ({ board: c.board }), and getCards({ boardId: undefined })` becomes  /boards/undefined/cards. Make board required again or guard the cardId options callback.

components/trello/actions/remove-label-from-card/remove-label-from-card.mjs
```diff
@@ -18,6 +18,7 @@ export default {
         app,
         "board",
       ],
+      optional: true,
```
vetrivigneshwaran: Keep board required or guard the dependent option loaders. cardId and labelId still call getCards()/findLabel() with boardId: c.board, so making board optional lets those loaders run with undefined and hit invalid Trello requests when the field isn’t set.

vetrivigneshwaran (changes requested):

components/trello/actions/archive-card/archive-card.mjs
ashwins01: removed changes outside `get-card` action to keep the PR scoped to just the issue

components/trello/actions/remove-label-from-card/remove-label-from-card.mjs
ashwins01: removed changes outside `get-card` action to keep the PR scoped to just the issue

components/trello/actions/get-card/get-card.mjs
```diff
@@ -19,6 +19,7 @@ export default {
         common.props.app,
         "board",
       ],
+      optional: true,
```
vetrivigneshwaran, line 22: cardId still depends on board via app, "cards", (c) => ({ board: c.board }), and getCards({ boardId: undefined }) becomes /boards/undefined/cards. Make board required again or guard the cardId options callback.

vetrivigneshwaran (changes requested):

components/trello/actions/get-card/get-card.mjs
ashwins01: There's a null-check present for `board` in `components/trello/trello.app.mjs` which returns `[]` instead of triggering an api call to trello.

vetrivigneshwaran (approved):
LGTM

PR #21302, opened 2026-07-01
fix(google_drive): preserve list-files pagination when a custom fields mask is supplied

### LINKED ISSUE #21178: Google Drive - 401 on google_drive-list-files MCP calls
### App
Google Drive

### Summary:
401 on google_drive-list-files MCP calls

### Details:
I am experiencing a persistent issue with the Google Drive integration that started today. This workflow used to work perfectly, but now it is consistently throwing a 401 Unauthorized error.

What we are calling: We are using the Google Drive "List Files" action, requesting specific fields: files(id,mimeType,name,createdTime,modifiedTime).

Error Details: PipedreamError: Status code: 401

Here are a few recent Request IDs for your logs:

8c88652a-e100-40ed-8391-5457cff8957b
266e7a2a-3799-4f40-b6bb-9de4fbdf3a59

### Screenshots:
No screenshots included

michelle0927:
## Summary

Fixes a pagination regression in `google_drive-list-files` that surfaces whenever a caller supplies a partial-response `fields` mask scoped to `files(...)` — the exact scenario reported in #21178.

### The bug

Google Drive's `files.list` honors partial-response masks literally. A request with `fields=files(id,name)` returns a response body containing only that shape — the top-level `nextPageToken` field is dropped. `list-files`'s do/while loop reads `pageToken = res.data.nextPageToken`, gets `undefined`, and exits after the first page — silently returning a truncated list.

Reproduced live against a connected account with the reporter's exact mask (`files(id,mimeType,name,createdTime,modifiedTime)`) — `nextPageToken` was absent from the response.

### The fix

When the user's `fields` mask omits `nextPageToken`, auto-prepend it before the API call so the pagination cursor comes back:

```diff
  if (this.fields) {
-   opts.fields = this.fields;
+   let fieldsValue = this.fields;
+   if (!fieldsValue.includes(PAGINATION_TOKEN_FIELD)) {
+     fieldsValue = `${PAGINATION_TOKEN_FIELD},${fieldsValue}`;
+   }
+   opts.fields = fieldsValue;
  }
```

The `fields` prop description gains a note explaining the constraint so callers passing their own mask directly know why it matters. The `nextPageToken` string is centralized in a new `PAGINATION_TOKEN_FIELD` constant in `common/constants.mjs`.

### Versions

- `list-files`: `0.3.0 → 0.3.1`
- `@pipedream/google_drive`: `1.12.0 → 1.12.1`

### On the 401 side of #21178

The reporter's issue title mentions HTTP 401, which turns out to be **separate from the pagination bug and not fixable at the component level**. Every Google integration in this repo (`google_calendar`, `google_docs`, `google_sheets`, `gmail`) constructs its client with `new OAuth2()` (no args) and sets only `access_token` — the platform layer owns token refresh. `$auth.oauth_client_secret` isn't exposed for any Google app, so no component-side refresh is possible. That side of the report likely needs either the reporter to reconnect their Google Drive account (refresh_token expired/revoked) or platform-team investigation into token refresh for MCP invocations. Referencing the issue but not closing it, since the 401 root cause is upstream of this component.

Refs #21178

## Test plan

- [x] ESLint clean across the changed files.
- [x] Reproduced the bug against a live Google Drive account via Pipedream Connect: request with the reporter's exact `fields` mask returned no `nextPageToken`, confirming the loop-exit condition.
- [x] Confirmed the fix's behavior: with `nextPageToken` auto-prepended to the mask, the response includes the pagination cursor and the loop continues to subsequent pages.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file listing pagination so results continue across pages when using custom field masks.
  * Added guidance so users know pagination will keep working even if `nextPageToken` isn’t included manually.

* **Chores**
  * Updated Google Drive component and package versions across multiple actions and sources for the latest release.


ashwins01 (approved):
The pagination fix LGTM. Should someone from the platform team be notified about the 401 or is this a known issue and the expectation is for the user to periodically reconnect their google-drive account ?

PR #21372, opened 2026-07-13
fix(ci): fail component version check on version downgrades

ashwins01:
## Summary

The `Components Checks "check_version"` job only verified that a component's version line was touched in the diff, not that the new value was greater than the one on the base branch. A downgrade such as `0.0.2` -> `0.0.1` passed, even though the lower version has already been published to the registry and cannot be republished.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved component version validation against the base branch.
  * Added detection for version downgrades and versions that do not increase.
  * Updated automated fix suggestions to recommend the next valid publishable version.
  * Enhanced failure messages to clearly identify versioning issues and affected files.


vetrivigneshwaran (approved):
LGTM. 

.github/actions/git-diff-on-components/src/index.js:171

git show <base>:<path> does a literal tree lookup with no rename detection. If a component's directory/file is renamed in the same PR that also lowers its version, filePath never existed under that name at baseCommit, so this returns null and getComponentsWithVersionDowngrade silently skips the check for that file. A PR that renames an app folder while dropping 0.0.2 → 0.0.1 on the new path would pass CI. Worth at least a follow-up issue, since catching downgrades is the whole point of this PR.

michelle0927:
> LGTM.
> 
> .github/actions/git-diff-on-components/src/index.js:171
> 
> git show : does a literal tree lookup with no rename detection. If a component's directory/file is renamed in the same PR that also lowers its version, filePath never existed under that name at baseCommit, so this returns null and getComponentsWithVersionDowngrade silently skips the check for that file. A PR that renames an app folder while dropping 0.0.2 → 0.0.1 on the new path would pass CI. Worth at least a follow-up issue, since catching downgrades is the whole point of this PR.

If a component is renamed, its key is changed, so it's essentially a new component and should have version "0.0.1".

PR #21400, opened 2026-07-15
Hubspot trigger adjustments for retroactive events

GTFalcao:
The expected behavior is for triggers to emit retroactive events only on the `deploy()` hook, with the `run()` method only serving subsequent events, never retroactive ones. This allows users to optionally bypass retroactive events with `emit_on_deploy: true`

This was previously fixed in one Hubspot source (`new-deal-property-change` at #21266 ), but many others still behaved incorrectly, which is what this PR is fixing.

## Summary by CodeRabbit

* **New Features**
  * HubSpot sources now emit a limited catch-up sample during initial deployment.
  * Sync begins from the deployment moment to reduce reprocessing of older events.
  * Enhanced timestamp-based syncing for engagements, emails, messages, threads, forms, and other HubSpot content.
  * More consistent initial load pagination and cursor handling.

* **Bug Fixes**
  * Reduced the initial deployment sample cap from 25 to 10.
  * Prevented retroactive emissions when deploy-time cursor information is missing.
  * Improved email pagination/cursor advancement.

* **Chores**
  * Bumped versions across affected HubSpot sources and the HubSpot package.

### DIFF
--- a/components/hubspot/sources/new-engagement/new-engagement.mjs
+++ b/components/hubspot/sources/new-engagement/new-engagement.mjs
@@ -38,7 +38,7 @@ export default {
   + " [Postal Mail](https://developers.hubspot.com/docs/api-reference/latest/crm/activities/postal-mail/search/search-postal-mail)"
   + " [Tasks](https://developers.hubspot.com/docs/api-reference/latest/crm/activities/tasks/search/search-tasks)"
   + " [See the documentation](https://developers.hubspot.com/docs/api-reference/latest/crm/using-object-apis)",
-  version: "0.0.50",
+  version: "0.0.51",
   dedupe: "unique",
   type: "source",
   props: {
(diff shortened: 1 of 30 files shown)


components/hubspot/sources/new-message-in-thread/new-message-in-thread.mjs
```diff
@@ -19,42 +22,74 @@ export default {
...
+        const {
+          results = [], paging,
+        } = await this.hubspot.listMessages({
+          threadId: this.threadId,
+          params,
+        });
+        messages.push(...results);
+        params.after = paging?.next?.after;
+      } while (params.after);
+      return messages;
```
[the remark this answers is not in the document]
GTFalcao: This is true, but we don't seem to have a viable alternative:
https://developers.hubspot.com/docs/reference/api/conversations/inbox-and-messages/v3
https://community.hubspot.com/t5/APIs-Integrations/Sort-parameter-not-working/m-p/913402

How would you approach this?

components/hubspot/sources/new-engagement/new-engagement.mjs
```diff
@@ -38,7 +38,7 @@ export default {
   + " [Postal Mail](https://developers.hubspot.com/docs/api-reference/latest/crm/activities/postal-mail/search/search-postal-mail)"
   + " [Tasks](https://developers.hubspot.com/docs/api-reference/latest/crm/activities/tasks/search/search-tasks)"
   + " [See the documentation](https://developers.hubspot.com/docs/api-reference/latest/crm/using-object-apis)",
-  version: "0.0.49",
+  version: "0.0.51",
```
michelle0927:
```suggestion
  version: "0.0.50",
```

components/hubspot/sources/new-event/new-event.mjs
```diff
@@ -8,7 +10,7 @@ export default {
   key: "hubspot-new-event",
   name: "New Events",
   description: "Emit new event for each new Hubspot event. Note: Only available for Marketing Hub Enterprise, Sales Hub Enterprise, Service Hub Enterprise, or CMS Hub Enterprise accounts",
-  version: "0.0.47",
+  version: "0.0.49",
```
michelle0927:
```suggestion
  version: "0.0.48",
```

components/hubspot/sources/new-message-in-thread/new-message-in-thread.mjs
```diff
@@ -19,42 +22,75 @@ export default {
...
+    let newest = after;
+    for (const message of messages) {
+      const ts = this.getTs(message);
+      if (ts > after) {
+        this.emitEvent(message);
+      }
+      if (ts > newest) {
+        newest = ts;
+      }
+    }
```
michelle0927, line 88: Nitpick: If messages are sorted oldest to newest, the newest ts will always be the last one.
```suggestion
    if (!messages?.length) {
      return;
    }
    for (const message of messages) {
      const ts = this.getTs(message);
      if (ts > after) {
        this.emitEvent(message);
      }
    }
    const newestTs = this.getTs(messages[messages.length - 1]);
    if (newestTs > newest) {
      newest = newestTs;
    }
```

components/hubspot/sources/new-thread-created/new-thread-created.mjs
```diff
@@ -1,50 +1,100 @@
...
+    let newest = after;
+    for (const thread of threads) {
+      const latest = Date.parse(thread.latestMessageTimestamp) || this.getTs(thread);
+      if (this.getTs(thread) > after) {
+        this.emitEvent(thread);
+      }
+      if (latest > newest) {
+        newest = latest;
+      }
+    }
```
michelle0927, line 92: Same nitpick as above.

michelle0927 (changes requested):

components/hubspot/sources/new-message-in-thread/new-message-in-thread.mjs
GTFalcao: This is true but I think I'd rather keep the more future-proof version (especially since "oldest first" is not the most common pattern) given the performance cost is very minimal

michelle0927 (approved):

PR #21429, opened 2026-07-20
Add batch/pagination support to Google Drive list-files and download-file

### LINKED ISSUE #21206: Google Drive
### App
Google Drive

### Summary:
No limit option

### Details:
I have a workflow where I need to download 5 files per run but since there is no limit on there and I have 100+ files in a folder there is no way for me to make this work efficiently.

### Screenshots:
No screenshots included

michelle0927:
## Summary

Adds batch / pagination support to the Google Drive **list-files** and **download-file** actions so large folders can be processed in set batches, per the request in #21206.

## `list-files` (0.3.1 → 0.4.0)
- **`Max Results`** — caps how many files are returned per run. `pageSize` is aligned to the number still needed, so the returned `nextPageToken` lands exactly after the last kept file — **no files are skipped** when resuming (a naive "fetch a full page then trim" approach would drop the remainder).
- **`Page Token`** input + returns **`{ files, count, nextPageToken, isComplete }`** — pass the previous run's `nextPageToken` back in to continue where it stopped (stateless cursor).
- Replaced the `additionalProps()`-based `filterType` with a static optional prop.

> **Design note:** the issue suggested persisting the cursor via a Data Store. Pipedream doesn't allow a `data_store` prop to be optional (it would become mandatory on every run, including simple "list everything" calls), so this uses a stateless `pageToken` in/out instead — fully optional and equally able to resume across runs.

> ⚠️ **Output shape change:** `list-files` now returns an object (`{ files, ... }`) instead of a bare array. Bumped as a minor version.

## `download-file` (0.1.27 → 0.2.0)
- **`Files`** (`fileIds`, string[]) — download multiple files in a single run. The legacy single **`File`** (`fileId`) is retained for backwards compatibility; both are merged and de-duped.
- A single file returns the original object shape; a batch returns `{ files, count }`. Errors if `Destination File Path` is set with multiple files.

## Testing
Published both to a private registry and ran targeted MCP evals — **5/5 passed**:
- `list-files`: folder listing; `Max Results` cap (returned exactly 2, flagged more remain); page-token resume (fetched next page with no overlap).
- `download-file`: single-file (backwards compat); batch (one call with 3 file IDs returned all three).

Closes #21206

## Summary by CodeRabbit

* **New Features**
  * Download multiple Google Drive files in one run (deduplicated), while preserving the existing single-file flow.
  * List files with selectable match type, per-run `Max Results` limiting, and `Page Token` pagination with resume support.
* **Bug Fixes**
  * Improved export-format option handling for downloads when selecting batches.
  * Added validation for download/list input rules and limits.
* **Chores**
  * Updated action versions and user-facing descriptions; listing results now return `{ files, count, nextPageToken, isComplete }` to reflect pagination state.

### DIFF
--- a/components/google_drive/package.json
+++ b/components/google_drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/google_drive",
-  "version": "1.12.3",
+  "version": "1.13.0",
   "description": "Pipedream Google_drive Components",
   "main": "google_drive.app.mjs",
   "keywords": [
--- a/components/google_drive/package.json
+++ b/components/google_drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/google_drive",
-  "version": "1.13.0",
+  "version": "2.0.0",
   "description": "Pipedream Google_drive Components",
   "main": "google_drive.app.mjs",
   "keywords": [
--- a/components/google_drive/package.json
+++ b/components/google_drive/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/google_drive",
-  "version": "1.12.2",
+  "version": "1.12.3",
   "description": "Pipedream Google_drive Components",
   "main": "google_drive.app.mjs",
   "keywords": [
--- a/components/google_drive/actions/download-file/download-file.mjs
+++ b/components/google_drive/actions/download-file/download-file.mjs
@@ -1,3 +1,4 @@
+import { ConfigurationError } from "@pipedream/platform";
 import fs from "fs";
 import stream from "stream";
 import { promisify } from "util";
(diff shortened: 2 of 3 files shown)


components/google_drive/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/google_drive",
-  "version": "1.12.2",
+  "version": "1.13.0",
```
ashwins01:
```suggestion
  "version": "2.0.0",
```

ashwins01 (dismissed):
LGTM, except for just this one version change.

michelle0927:
## MCP Eval Results — google_drive (PR-scoped)

**5/5 passed** (100%) | 2.6 avg tool calls | 63.9s
Model: `claude-haiku-4-5-20251001` | Registry: `private` | Run against the republished `list-files` (v1.0.0) and `download-file` (v0.2.0)

| # | Eval | Status | Tools | What it verifies |
|---|------|--------|-------|------------------|
| 23 | `[list-files]` List files in a folder | ✅ pass | 2 | Folder-scoped listing returns all 4 files |
| 24 | `[list-files]` Cap results with Max Results | ✅ pass | 2 | `maxResults` returns exactly 2, response flags more remain |
| 25 | `[list-files]` Resume with page token | ✅ pass | 3 | Passing `nextPageToken` back fetches the next page with no overlap |
| 26 | `[download-file]` Download a single file | ✅ pass | 2 | Backwards-compatible single-file download |
| 27 | `[download-file]` Download multiple files in one run | ✅ pass | 4 | One `download-file` call with 3 file IDs downloads all three (batch path) |

Each eval also called a `find-folder`/`find-file` helper to resolve names to IDs before the target tool — expected for these prompts.

> Evals run via MCP against the two changed tools published to a private registry. Covers the new `maxResults` cap, stateless `pageToken` resume, and multi-file batch download.

ashwins01 (approved):

PR #21436, opened 2026-07-21
Fal.ai - allowing custom appId input

GTFalcao:
This allows custom app ids to be entered, but is unfortunately backwards-incompatible with previous appId inputs that did not include the `fal-ai/` prefix *and* had another slash in it (we are automatically prefixing the ones with no slash since these are always assumed to be `fal-ai/` prefixed, to reduce backwards incompatibility).

## Summary by CodeRabbit

- **New Features**
  - Added support for canonical model identifiers, including namespaced and bare model names.
  - Requests now consistently target the correct model-specific endpoints.

- **Improvements**
  - Updated queue, cancellation, status, and response actions to use normalized model identifiers.
  - Released the integration and its actions as version 1.0.0.


vetrivigneshwaran (approved):
LGTM

PR #21476, opened 2026-07-23
fix(google_calendar): harden event tools for AI-agent (MCP) use

dylburger:
## Summary

Hardens the Google Calendar event tools for use as MCP tools driven by LLM agents. Each change addresses a failure mode observed while exercising these components through an AI agent. Every changed file includes an inline comment explaining the rationale.

**`add-attendees-to-event`**
- Rewrote `run()` from a full-event update (`events.update`/PUT, which re-sent read-only fields like `iCalUID`/`etag`/`sequence`) to `events.patch`/PATCH of only `attendees`.
- Merges new invitees into the event's existing attendee list — deduped by email (case-insensitive) and preserving each existing attendee's `responseStatus`.
- Clarified the description so agents route "add people to an event that already exists" here. Previously the tool went unused because agents folded attendees straight into `create-event`.

**`create-event`**
- Agents invoked it for requests it cannot fulfill (setting working hours, changing calendar settings, creating a new calendar), silently producing a stray event. The description now scopes it to events only and instructs the agent to decline those, since no tool exposes those capabilities.

**`delete-event`**
- Agents bulk-deleted events for "clear my whole day" requests without confirming. The description now asks for confirmation before destructive or ambiguous multi-event deletes and clarifies it deletes exactly one event per call.

**`common/create-event-common` (`parseEmails`)**
- Accepts a JSON-stringified array of emails (e.g. `'["a@x.com","b@y.com"]'`) in addition to a comma-separated list — agents commonly emit the former for the string-typed `attendees` prop, and it was previously comma-split into malformed tokens the API rejected. Affects all actions that accept attendees, so their versions are bumped as well.

### Backwards compatibility
All changes are backwards-compatible: no props renamed or removed, return shapes unchanged. The `add-attendees-to-event` rewrite keeps its interface and is behavior-preserving or safer (existing attendees and their RSVP status are preserved; the added dedup only prevents duplicate-attendee entries the API rejects anyway).

### Versioning
- [x] All components updated in this PR had their version updated

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **New Features**
  - Added support for adding attendees to existing calendar events while preserving event details.
  - Added selectable response fields for calendar event listings, including a compact format.
  - Enhanced attendee input handling to accept JSON-formatted attendee arrays.

- **Bug Fixes**
  - Prevented duplicate attendees through case-insensitive matching.
  - Improved success feedback to report newly added attendees.

- **Documentation**
  - Clarified event creation and single-event deletion behavior, including confirmation requirements.

- **Chores**
  - Updated calendar action and package versions.


components/google_calendar/actions/create-event/create-event.mjs
```diff
@@ -6,8 +6,12 @@ import constants from "../../common/constants.mjs";
...
   name: "Create Event",
-  description: "Create an event in a Google Calendar. [See the documentation](https://developers.google.com/calendar/api/v3/reference/events/insert)",
-  version: "1.0.4",
+  // Scope guidance for AI agents: create-event was being invoked for requests it cannot
+  // fulfill — setting working hours, changing calendar settings, or creating a whole new
+  // calendar — silently producing a stray event. The description now scopes it to events
+  // only and tells the agent to decline those, since no tool exposes those capabilities.
+  description: "Create a new event on a Google Calendar — a single or recurring appointment at a specific date/time (optionally with attendees, location, and description). Use this whenever the user wants to add something to their calendar. This creates calendar EVENTS only: do NOT use it to configure calendar settings, working hours, availability, or default preferences, and do NOT use it to create a new calendar — no tool exposes those capabilities, so decline such requests rather than representing them as an event. [See the documentation](https://developers.google.com/calendar/api/v3/reference/events/insert)",
```
dylburger: Scope guidance added here: AI agents were calling create-event for requests it cannot fulfill — setting working hours, changing calendar settings, or creating a new secondary calendar — and silently producing a stray event. The added "do NOT use it to…" language tells the agent to decline those, since no tool exposes those capabilities.

components/google_calendar/actions/delete-event/delete-event.mjs
```diff
@@ -3,8 +3,12 @@ import googleCalendar from "../../google_calendar.app.mjs";
 export default {
   key: "google_calendar-delete-event",
   name: "Delete an Event",
-  description: "Delete an event from a Google Calendar. [See the documentation](https://googleapis.dev/nodejs/googleapis/latest/calendar/classes/Resource$Events.html#delete)",
-  version: "0.1.12",
+  // Confirmation guidance for AI agents: agents were bulk-deleting events for "clear my
+  // whole day" style requests without pausing to confirm. The description now instructs a
+  // confirmation step before destructive or ambiguous multi-event deletes, and makes clear
+  // this deletes exactly one event per call (it does not clear a day on its own).
+  description: "Delete a single event from a Google Calendar. This permanently cancels the event and cannot be undone. Deletes only the one event identified by `eventId` — it does NOT clear a whole day or delete multiple events. When the user asks to delete several events, clear a day, or the target is ambiguous, first confirm with the user (ideally listing what will be deleted) before calling this, and call it once per event. [See the documentation](https://googleapis.dev/nodejs/googleapis/latest/calendar/classes/Resource$Events.html#delete)",
```
dylburger: Confirmation guidance added: agents were bulk-deleting events for "clear my whole day" requests without pausing to confirm. The new wording asks for confirmation before destructive or ambiguous multi-event deletes and makes explicit that this deletes exactly one event per call (it can't clear a day on its own).

components/google_calendar/actions/add-attendees-to-event/add-attendees-to-event.mjs
```diff
@@ -4,8 +4,12 @@ import createEventCommon from "../common/create-event-common.mjs";
...
-  description: "Add attendees to an existing event. [See the documentation](https://googleapis.dev/nodejs/googleapis/latest/calendar/classes/Resource$Events.html#update)",
-  version: "0.0.9",
+  // Routing guidance for AI agents: agents defaulted to create-event (which can also set
+  // attendees), so this tool went unused. The description now steers "add people to an
+  // event that already exists" here. run() (below) PATCHes only the `attendees` field,
+  // merging new invitees into the current list rather than replacing it.
+  description: "Add one or more attendees (invitees) to an event that ALREADY EXISTS on a Google Calendar, without recreating it. New attendees are merged into the event's current attendee list — existing attendees are preserved, not replaced. Use this when the user wants to invite additional people to an event they already have; identify the target event with `eventId` first (e.g. via list-events or get-event). When creating a brand-new event, set its attendees directly in create-event instead of calling this afterward. [See the documentation](https://googleapis.dev/nodejs/googleapis/latest/calendar/classes/Resource$Events.html#patch)",
```
dylburger, line 7: Routing guidance rewritten: agents defaulted to create-event (which can also set attendees), so this tool went unused. The description now steers "add people to an event that already exists" here and states that existing attendees are preserved (merged, not replaced).

components/google_calendar/actions/add-attendees-to-event/add-attendees-to-event.mjs
```diff
@@ -42,25 +46,43 @@ export default {
...
+      (a) => !existingEmails.has(a.email.toLowerCase()),
+    );
+    const attendees = [
+      ...existingAttendees,
+      ...attendeesToAdd,
+    ];
+
+    // PATCH (not UPDATE/PUT) so we only touch `attendees` — avoids echoing back
+    // read-only fields (iCalUID, etag, sequence, …) that a full PUT can reject.
+    const response = await this.googleCalendar.patchEvent({
```
dylburger, line 83: Changed from events.update (PUT) to events.patch (PATCH). The old code spread the entire fetched event back into a PUT body, re-sending read-only fields (iCalUID, etag, sequence, …). PATCH touches only the attendees field; the merge just above preserves existing attendees and their responseStatus and dedupes new invitees by email.

components/google_calendar/actions/common/create-event-common.mjs
```diff
@@ -90,7 +90,25 @@ export default {
     },
     parseEmails(input) {
       if (typeof input === "string") {
-        return input.split(",")
+        const trimmed = input.trim();
+        // Accept a JSON-stringified array too (e.g. '["a@x.com","b@y.com"]'),
+        // which LLM callers commonly emit for a string-typed "array or CSV" prop.
+        // Without this, JSON.stringify output is comma-split into malformed tokens
+        // like `["a@x.com"` that the API rejects as "Invalid attendee email".
+        if (trimmed.startsWith("[")) {
```
dylburger, line 98: parseEmails now accepts a JSON-stringified array (e.g. '["a@x.com","b@y.com"]'), which AI agents commonly emit for this string-typed "array or CSV" prop. Previously that input was comma-split into malformed tokens the API rejected as invalid emails. Plain single emails and comma-separated lists still take the original split path unchanged.

components/google_calendar/actions/create-event/create-event.mjs
ashwins01: Should we add this overtly generalised guidance statement here - "no tool exposes" ?
This might be misleading if not updated proactively when a new tool gets added. Is it not better to not add this direction at all ?

components/google_calendar/actions/delete-event/delete-event.mjs
ashwins01: Would it be better to do `cannot be undone` guidance as part of the action description or append it as part of tool description based on the `destructiveHint: true` annotation ?
I'd lean towards using the annotation.

ashwins01 (dismissed):
LGTM, other than a couple of questions.

components/google_calendar/actions/create-event/create-event.mjs
dylburger: Good catch — the cross-tool claim is the fragile part, agreed. Fixed in aeebcb4: the sentence is now scoped to this action ("it cannot configure calendar settings, working hours, availability, or default preferences, and it cannot create a new calendar"), so it stays true no matter which tools are added later.

I kept scope guidance rather than dropping it, because it was added for a measured failure: agents were calling create-event for "set my working hours to 9–5" and reporting success after creating an event named after the request. The new wording gives the same grounds to decline without asserting anything about the rest of the toolset. Evals 25–27 (abstain on: create a calendar / share a calendar / set working hours) are the regression check on the rewording.

components/google_calendar/actions/delete-event/delete-event.mjs
dylburger: Agreed — and `destructiveHint: true` is already set on this action, so the prose was redundant. "This permanently cancels the event and cannot be undone" is gone in aeebcb4.

What I kept is the part the annotation cannot express: that this deletes exactly one event per call (agents were treating it as "clear my day" and issuing bulk deletes), and that multi-event or ambiguous deletes should be confirmed first. `destructiveHint` is a per-tool boolean — "confirm when the target is ambiguous" is not derivable from it, and in practice clients use annotations for approval UI rather than putting them in the model context, so the behavioral guidance still has to live in the description.

components/google_calendar/actions/add-attendees-to-event/add-attendees-to-event.mjs
```diff
@@ -42,25 +46,43 @@ export default {
...
+        .map((a) => a.email?.toLowerCase())
+        .filter(Boolean),
+    );
+    const attendeesToAdd = newAttendees.filter(
+      (a) => !existingEmails.has(a.email.toLowerCase()),
+    );
+    const attendees = [
+      ...existingAttendees,
+      ...attendeesToAdd,
+    ];
```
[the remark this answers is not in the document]
dylburger: Already fixed in 5d2b18c — `attendeesToAdd` now adds each accepted address to `seenEmails` as it goes, so `["A@example.com", "a@example.com"]` yields one entry. The set is seeded from the existing attendees, so an address already on the event is not re-added and the existing object survives with its `responseStatus` intact.

ashwins01 (approved):
QA passed.

PR #14, opened 2020-04-24
Add Twitter search component

psavkar:

PR #21491, opened 2026-07-24
feat(gmail): eval-driven fixes to the AI-optimized Gmail actions

dylburger:
## What & why

Fixes to the AI-optimized Gmail action set, driven by the MCP eval suite in `pd-connect-eval-monster/evals/gmail` (31 real-user tasks, 5 multi-tool workflows, 6 restraint tests). Confirming run: **29/31 pass² on Sonnet 5**, multi-step 12/12, tool coverage 9/9, precision 1.00 / recall 1.00. A Haiku 4.5 + Sonnet 5 sweep flags **no failure as a genuine tool problem** — every remaining failure passes on at least one model.

## Changes

- **send-email** (`0.3.1` → `0.3.2`, patch) — every reply that passed `inReplyToMessageId` without an explicit `to` returned **HTTP 400 "Recipient address required"**. `getOptionsToSendEmail` only populated `opts.to` on the `replyAll` branch, so the documented "only the original sender is addressed" behavior was never actually implemented. Now defaults to the replied-to message's `Reply-To`/`From`; also covers a `replyAll` self-thread case where filtering out the user's own address left zero recipients, and stops a missing `To`/`Cc` header throwing.
- **create-draft** (`0.2.1` → `0.2.2`, patch) — same defect via the shared app file.
- **find-email** (`0.2.1` → `0.3.0`, minor) — a "catch me up on this week's mail" task produced a confidently wrong digest, describing an offline fence sensor as healthy. The agent had correctly asked for `format:"full"` + `payload`; the 42k-char response exceeded the MCP client's 25k-token result ceiling and was spilled to a file, and the in-tool budget then stripped the requested field, leaving 197-character snippets. The needed sentence sits at character 670 of a 934-character body, so no snippet could contain it and nothing signalled it had been cut. Adds `bodyText` (decoded plain text, ~half the size of raw `payload`), a `bodyChars` cap with `bodyTruncated`/`bodyTotalChars`, `snippetTruncated` on metadata results, lowers the response cap to 30k chars, and **inverts the budget rule: a field the caller explicitly named is never stripped** — whole messages are dropped instead, and the note says how many of how many are shown. All new props are optional; omitting them reproduces prior output.
- **list-thread-messages** (`0.1.1` → `0.2.0`, minor) — same result-ceiling risk; response cap lowered to 30k chars. It already degraded `full` → `metadata` → tail, so it needed the cap rather than a fields projection.
- **modify-labels** (`0.0.2` → `0.0.3`, patch) — three description defects the evals surfaced: the agent skipped the tool in 20% of tasks that required it, trashed an entire label's worth of mail with no confirmation, and believed it could create filter rules for future mail. Adds "use this whenever the user says star / archive / file / label / trash / mark read", a trash-is-destructive confirmation rule, and a "not for filters or future mail" exclusion. Tool recall across the suite went **0.88 → 1.00**.

App `package.json` bumped `2.1.0` → `2.2.0` (largest segment applied: minor).

## Verification

- Suite: 31 evals — 19 single-tool, 12 multi-step, 6 restraint (abstain/clarify/confirm).
- Sonnet 5 pass²: **29/31**. Two flaky, neither actionable at the component level: one trial hit a Claude Code API-level refusal *after* the tool call succeeded, and one restraint test proceeded after the harness deliberately denied its `AskUserQuestion` confirmation attempt.
- Largest single tool result is now 45% of the client's token ceiling (previously over it, which meant the model received a file path instead of the data).

## Summary by CodeRabbit

* **New Features**
  * Find Emails now supports returning only selected fields and limiting message body length.
  * Find Emails improved budget-aware trimming, with clearer per-message body/snippet truncation indicators and more informative summary output.
  * Thread message listings use a stricter response-size cap and mark oversized results as truncated.
* **Bug Fixes**
  * Improved reply/reply-all recipient handling to prevent “missing recipient” failures.
* **Documentation**
  * Expanded action guidance for modifying labels (including stronger TRASH warnings) and list-labels usage notes.


components/gmail/gmail.app.mjs
```diff
@@ -386,11 +388,23 @@ export default {
...
-              opts.bcc = this.parseEmailAddresses(bcc.value);
+              opts.bcc = this.parseEmailAddresses(bcc);
+            }
+            // Replying to a thread the user is the only participant in (a note to
+            // self) filters the recipient list down to nothing. Fall back to the
+            // original sender rather than sending with no recipient.
+            if (!opts.to.length && originalSender) {
+              opts.to = this.parseEmailAddresses(originalSender);
             }
+          } else if (!opts.to?.length && originalSender) {
```
dylburger, line 415: **Why:** eval #16 ("send myself an email, then reply so both sit in one conversation") failed with **HTTP 400 `Recipient address required`** on every reply that passed `inReplyToMessageId` without an explicit `to`. `opts.to` was only ever populated inside the `replyAll` branch, so the behavior both tool descriptions promise — *"otherwise only the original sender is addressed"* — was never actually implemented.

**Change:** default `to` to the replied-to message's `Reply-To`/`From`. The `replyAll` branch above also now falls back to the original sender when filtering out the user's own address empties the list (replying to a note-to-self), and header lookups no longer throw on a missing `To`/`Cc`.

**Result:** #16 and #17 both pass. Patch bump on send-email and create-draft, which share this method.

components/gmail/actions/find-email/find-email.mjs
```diff
@@ -77,6 +105,30 @@ export default {
       default: "metadata",
       optional: true,
     },
+    fields: {
```
dylburger, line 102: **Why:** eval #23 ("catch me up on this week's mail and email me the summary") produced a confidently wrong digest — it described an **offline** fence sensor as healthy. The agent was not being lazy; it asked for exactly the right thing:

```json
{"format":"full","maxResults":50,"fields":["subject","sender","date","snippet","payload"]}
```

The 42k-char response exceeded the MCP client's 25k-token result ceiling (so the client spilled it to a file and handed the model a path), and the in-tool budget then stripped `payload` — the field that had been explicitly requested — leaving 197-character snippets. Measured on the fixture: the sentence the summary needed sits at **character 670 of a 934-character body**. No snippet could contain it, and nothing in a snippet signals that it was cut.

**Change:** `bodyText`, a derived field — decoded plain text, HTML converted, MIME scaffolding and attachments stripped (934 vs 1962 chars for the same email). Requesting it upgrades the underlying fetch to `full` automatically. Paired with `bodyChars` (below) and `snippetTruncated` on metadata results so the default path's lossiness is at least visible.

**Result:** #23 passes. Minor bump — every new prop is optional and omitting them reproduces the previous output.

components/gmail/actions/find-email/find-email.mjs
```diff
@@ -2,7 +2,31 @@ import utils from "../../common/utils.mjs";
 import gmail from "../../gmail.app.mjs";
 
 const DEFAULT_MAX_RESULTS = 25;
-const MAX_RESPONSE_CHARS = 100_000;
+// Sized to stay under a typical MCP client's per-result token ceiling (Claude Code
+// truncates at 25k tokens; message JSON runs ~1.7 chars/token, so 100k chars was ~2.4x
+// over). An oversized result is not merely verbose — the client may spill it to a file
+// and hand the model a path, so the model sees NOTHING. Measured: a 56-message search
+// returned 42k chars and was spilled.
+const MAX_RESPONSE_CHARS = 30_000;
```
dylburger: **Why:** 100k chars was ~2.4x over a typical MCP client's per-result token ceiling. Going over is not merely verbose — the client writes the result to a file and hands the model a path, so **the model sees none of the data**.

**Change:** cap lowered to 30k, and the over-budget strategy now depends on who chose the shape: *if the caller named `fields`, whole messages are dropped and the note reports "48 of 50 shown"; if it named none, messages are compacted so counts stay accurate.* The old code did the opposite for caller-chosen fields, which is what silently removed `payload` above. Visible incompleteness is recoverable (narrow and retry); invisible fidelity loss produces a wrong answer nothing downstream can detect. `bodyText` shrinks toward a 400-char floor before anything is dropped, flagging each cut message.

**Result:** largest tool result is now 45% of the ceiling; eval #18 ("how many field reports?", 56 messages) still counts correctly because that query names no fields.

components/gmail/actions/list-thread-messages/list-thread-messages.mjs
```diff
@@ -1,7 +1,11 @@
 import utils from "../../common/utils.mjs";
 import gmail from "../../gmail.app.mjs";
 
-const MAX_RESPONSE_CHARS = 100_000;
+// Same ceiling as Find Emails: an MCP client may spill an oversized result to a file
+// and hand the model a path, so a result over the limit is worse than a trimmed one —
+// the model sees none of it. This action already degrades before dropping (full →
+// metadata → tail), so it needs the lower cap, not a `fields` prop.
+const MAX_RESPONSE_CHARS = 30_000;
```
dylburger: **Why:** same result-ceiling exposure as Find Emails — an oversized thread gets spilled to a file and the model sees nothing.

**Change:** cap lowered to 30k chars. This action already degraded `full` → `metadata` → tail-drop, so it needed the lower ceiling rather than a fields projection.

**Result:** no thread read hit the ceiling in the confirming run. Minor bump: interface unchanged, but large threads now degrade earlier than before.

components/gmail/actions/modify-labels/modify-labels.mjs
```diff
@@ -7,6 +7,10 @@ export default {
   description:
     "Add and/or remove labels on one or more Gmail messages in a single call."
     + " In Gmail, most inbox-state operations are label mutations under the hood, so this one tool covers archive / trash / untrash / star / unstar / mark-read / mark-unread / apply-label / remove-label."
+    + "\n\n**Use this whenever the user asks you to star, unstar, flag, archive, file, sort, label, tag, categorise, move, trash, delete, restore, or mark mail as read or unread** — there is no separate tool for any of those. Pair it with **Find Emails** to turn a description of the mail (\"the invoice from billing\", \"everything from last week\") into the `messageIds` this tool needs."
```
dylburger, line 10: **Why:** three distinct defects, all description-level, all surfaced by evals. (1) The agent **skipped this tool in 20% of the tasks that required it** — evals #14/#20/#22/#25 reached for it only sometimes; under tool search the description IS the retrieval key. (2) Eval #31 ("trash every email in my Park Operations label") trashed 16 messages with no confirmation. (3) Eval #27 ("set up a filter that labels incoming mail") had the agent gathering filter criteria it could never act on, because nothing said filters are out of scope.

**Change:** an explicit "use this whenever the user says star / unstar / archive / file / label / trash / mark read…" list; a trash-is-destructive rule requiring confirmation for bulk deletes the user did not enumerate; and a "do NOT use for filters, rules, or anything applying to mail that has not arrived" exclusion.

**Result:** tool recall across the suite went **0.88 → 1.00**; #27 passes; #31 now asks before trashing. Patch bump — description only, no code change.

dylburger (commented):
Per-change rationale inline — each comment cites the eval that motivated it. Full suite, seed fixtures and run reports live in pd-connect-eval-monster/evals/gmail.

michelle0927 (approved):
LGTM!

michelle0927:
## MCP Eval Results — Gmail (PR #21491)

**6/6 pass · 0 fail · 0 warn (100%)**

| # | Eval | Status |
|---|---|---|
| 10 | Find email and read the body ("catch me up") | ✅ PASS |
| 11 | Find email and get its full thread | ✅ PASS |
| 12 | Star an email found by subject | ✅ PASS |
| 13 | Find email by body keyword | ✅ PASS |
| 14 | Summarize an email (routes to bodyText projection) | ✅ PASS |
| 15 | Mark email as read | ✅ PASS |

PR #21496, opened 2026-07-26
feat(google_tasks): AI-optimize task actions

### LINKED ISSUE #21472: google_tasks: AI-optimize MCP action set
Google Tasks today has solid basics — create, update, delete, and list tasks and task lists. What's missing is the quick, natural things an agent tends to reach for. We'd like these improved and new tools added so an agent can work the way people actually manage to-dos, and so each tool is easy for an agent to reach for and get right.

- Check off a task without doing a full update — e.g. "mark 'send invoice' done"
- Move a task to another list or reorder it
- Find tasks across lists by due date or keyword
- Work with subtasks

**API reference:** https://developers.google.com/workspace/tasks/reference/rest

**Add marker:**
```
// x-pd-ai: optimized
```

jjnawaaz:
## Summary

This PR AI-optimizes the Google Tasks component by adding new task management actions and improving the app for agent workflows.

### Added
- Complete Task action
- Move Task action
- Create Subtask action
- Search Tasks action

### Enhanced
- Added support for query parameters in `insertTask()` to enable subtask creation.
- Added `moveTask()` helper method.
- Added reusable `keyword` prop definition.

### AI Optimization
- Added `// x-pd-ai: optimized` marker to the app and action files.

Closes #21472

## Summary by CodeRabbit

* **New Features**
  * Added actions to complete tasks, create subtasks, move tasks, and search tasks.
  * Added keyword and due-date filtering across task titles, notes, and schedules.
  * Enabled moving tasks between lists and changing their position or parent.

* **Improvements**
  * Improved task creation, updates, pagination, and task-list retrieval.
  * Enhanced support for optional task-list and task filters.
  * Updated Google Tasks package, action, and source versions.


components/google_tasks/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,116 @@
...
+      );
+
+      tasks.push(
+        ...listTasks.map((task) => ({
+          ...task,
+          taskListId: taskList.id,
+          taskListTitle: taskList.title,
+        })),
+      );
+    }
```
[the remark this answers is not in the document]
jjnawaaz: The current sequential implementation is intentional to avoid issuing concurrent requests across many task lists, which could increase the likelihood of hitting API rate limits. Given that this is a low-value optimization and the existing implementation is simpler, I'd prefer to keep the current approach unless profiling shows it to be a bottleneck.

jjnawaaz:
The current implementation follows the existing request pattern used throughout this component. Refactoring to a shared request helper would require touching all request methods and is outside the scope of this feature PR.

components/google_tasks/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,118 @@
...
+
+    const keyword = this.keyword?.toLowerCase();
+    const dueDate = this.due?.split("T")[0];
```
vetrivigneshwaran: Due-date filtering is done entirely client-side (`task.due?.split("T")[0] === dueDate`) even though the tasks.list API natively supports `dueMin`/`dueMax` query params. Using the API params would reduce payload and interact correctly with pagination limits rather than filtering a truncated in-memory set.

components/google_tasks/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,118 @@
+// x-pd-ai: optimized
+import { ConfigurationError } from "@pipedream/platform";
+import app from "../../google_tasks.app.mjs";
+
+export default {
+  key: "google_tasks-search-tasks",
+  name: "Search Tasks",
+  description:
+    "Searches tasks across all task lists using a keyword, due date, or both. Use this action to find tasks that match specific criteria without modifying them. To create, update, complete, or move tasks, use the corresponding task actions instead. [See the docs here](https://developers.google.com/workspace/tasks/reference/rest)",
```
vetrivigneshwaran: Update the doc link to `https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/list`

Line 9 uses [See the docs here] instead of the required `[See the
  documentation]` link text. Replace the generic “corresponding task
  actions” reference with bold names of the existing actions.

components/google_tasks/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,118 @@
...
+        "due",
+      ],
+    },
+    maxResults: {
+      propDefinition: [
+        app,
+        "maxResults",
+      ],
+    },
+    showCompleted: {
```
vetrivigneshwaran, line 37: The action exposes a "Show Completed" prop and passes `showCompleted`, but per Google's tasks.list docs, `showHidden` must ALSO be true to return tasks completed in first-party clients (web/mobile). Since the action never sends `showHidden`, enabling Show Completed will still not surface most completed tasks - so searching for a completed to-do by keyword/due will return nothing, contradicting the issue's "mark 'send invoice' done"/find-tasks intent

components/google_tasks/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,118 @@
...
+      );
+    }
+
+    const params = {
+      maxResults: this.maxResults,
+      showCompleted: this.showCompleted,
+      showDeleted: this.showDeleted,
+    };
+
+    const taskLists = await this.app.paginate(
```
vetrivigneshwaran, line 82: HTTP calls do not thread the step `$` context. The new action reuses `app.paginate`, which invokes `fn(this, ...)` with `this` bound to the app object, so the platform axios receives the app (not `$`) as its context - losing the request-logging/retry benefits the guidelines require. Pre-existing helper pattern, but this new action inherits it

components/google_tasks/actions/create-subtask/create-subtask.mjs
```diff
@@ -0,0 +1,92 @@
...
+    parent: {
+      propDefinition: [
+        app,
+        "taskId",
+        ({ taskListId }) => ({
+          taskListId,
+        }),
+      ],
+      label: "Parent Task",
+      description: "The parent task under which the subtask will be created.",
```
vetrivigneshwaran: parent and previous accept opaque Google Tasks IDs. Their descriptions
  need an inline example and instructions for obtaining each ID from a
  task-list response.

  As per coding guidelines, descriptions for non-obvious IDs must explain
  the expected format and where to obtain the value.

components/google_tasks/actions/move-task/move-task.mjs
```diff
@@ -0,0 +1,98 @@
...
+        }),
+      ],
+    },
+    destinationTasklist: {
+      propDefinition: [
+        app,
+        "taskListId",
+      ],
+      label: "Destination Task List",
+      description:
```
vetrivigneshwaran, line 39: destinationTasklist, parent, and previous accept opaque Google Tasks
  IDs. Their descriptions need an inline example and instructions for
  obtaining each ID from a task-list response.

  As per coding guidelines, descriptions for non-obvious IDs must explain
  the expected format and where to obtain the value.

components/google_tasks/actions/move-task/move-task.mjs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    const res = await this.app.moveTask(
+      $,
+      this.taskListId,
+      this.taskId,
+      params,
+    );
+
+    $.export("$summary", "Task successfully moved");
```
vetrivigneshwaran:
```suggestion
    $.export("$summary", `Moved task ${this.taskId}`);
```

components/google_tasks/actions/complete-task/complete-task.mjs
```diff
@@ -0,0 +1,43 @@
...
+        }),
+      ],
+    },
+  },
+  async run({ $ }) {
+    const res = await this.app.patchTask($, this.taskListId, this.taskId, {
+      status: "completed",
+    });
+
+    $.export("$summary", "Task successfully completed");
```
vetrivigneshwaran:
```suggestion
    $.export("$summary", `Completed task ${this.taskId}`);
```

components/google_tasks/actions/move-task/move-task.mjs
```diff
@@ -0,0 +1,98 @@
+// x-pd-ai: optimized
+import app from "../../google_tasks.app.mjs";
+
+export default {
+  key: "google_tasks-move-task",
+  name: "Move Task",
+  description:
+    "Moves an existing task to a different position, parent task, or task list. Use this action to reorganize your task hierarchy or reorder tasks without modifying the task's title, notes, due date, or completion status. To update those properties, use the Update Task action instead. [See the docs here](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/move)",
```
vetrivigneshwaran:
```suggestion
"Moves an existing task to a different position, parent task, or task list. Use this action to reorganize your task hierarchy or reorder tasks without modifying the task's title, notes, due date, or completion status. To update those properties, use the **Update Task** action instead. [See the documentation](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/move)",
```

components/google_tasks/actions/create-subtask/create-subtask.mjs
```diff
@@ -0,0 +1,92 @@
...
+      parent: this.parent,
+    };
+
+    if (this.previous) {
+      params.previous = this.previous;
+    }
+
+    const res = await this.app.insertTask($, this.taskListId, data, params);
+
+    $.export("$summary", "Subtask successfully created");
```
vetrivigneshwaran:
```suggestion
    $.export("$summary", `Created subtask: ${this.title}`);
```

components/google_tasks/actions/complete-task/complete-task.mjs
```diff
@@ -0,0 +1,43 @@
+// x-pd-ai: optimized
+import app from "../../google_tasks.app.mjs";
+
+export default {
+  key: "google_tasks-complete-task",
+  name: "Complete Task",
+  description:
+    "Marks an existing task as completed by setting its status to `completed`. Use this action when you want to complete a task without changing its title, notes, due date, or other properties. To modify those fields, use the Update Task action instead. [See the docs here](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/patch)",
```
vetrivigneshwaran:
```suggestion
"Marks an existing task as completed by setting its status to `completed`. Use this action when you want to complete a task without changing its title, notes, due date, or other properties. To modify those fields, use the **Update Task** action instead. [See the documentation](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/patch)",
```

components/google_tasks/actions/create-subtask/create-subtask.mjs
```diff
@@ -0,0 +1,92 @@
+// x-pd-ai: optimized
+import app from "../../google_tasks.app.mjs";
+
+export default {
+  key: "google_tasks-create-subtask",
+  name: "Create Subtask",
+  description:
+    "Creates a new subtask under an existing parent task. Use this action when you want to organize related work into a task hierarchy while keeping the parent task unchanged. To create a top-level task instead, use the Create Task action. [See the docs here](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/insert)",
```
vetrivigneshwaran:
```suggestion
"Creates a new subtask under an existing parent task. Use this action when you want to organize related work into a task hierarchy while keeping the parent task unchanged. To create a top-level task instead, use the **Create Task** action. [See the documentation](https://developers.google.com/workspace/tasks/reference/rest/v1/tasks/insert)",
```

components/google_tasks/google_tasks.app.mjs
```diff
@@ -120,58 +135,103 @@ export default {
       return data;
     },
     async getTaskLists(ctx = this, params = {}) {
-      return axios(ctx, this._getRequestParams({
-        path: "/users/@me/lists",
-        method: "GET",
-        params,
-      }));
+      return axios(
```
vetrivigneshwaran: Centralize the HTTP request boundary.

  Add _makeRequest(ctx, options) and route all ten HTTP methods through
  it. _getRequestParams() should only build request options; axios is
  already imported from @pipedream/platform.

vetrivigneshwaran (changes requested):

components/google_tasks/actions/search-tasks/search-tasks.mjs
```diff
@@ -0,0 +1,119 @@
...
+      );
+
+      tasks.push(
+        ...listTasks.map((task) => ({
+          ...task,
+          taskListId: taskList.id,
+          taskListTitle: taskList.title,
+        })),
+      );
+    }
```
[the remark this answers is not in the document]
jjnawaaz: I agree that applying maxResults after the cross-list search would provide a more accurate global result limit. I considered that approach, but it would require changing the search algorithm to page through every task list until the global match limit is reached, rather than using the existing pagination helper. Since this is a more substantial behavioral change affecting pagination semantics, I kept the current implementation consistent with the existing component patterns for this feature PR. I'd be happy to revisit this as a follow-up improvement if desired.

vetrivigneshwaran (approved):
LGTM

PR #21543, opened 2026-07-30
fix(microsoft_outlook): find-email defaults to whole-mailbox search, add subject filter

### LINKED ISSUE #21541: microsoft_outlook: Find Email search should cover the whole mailbox by default
## Problem

The **Find Email** action only returns inbox messages when a user searches without choosing a folder. A recent change (#21177) set the **Folder Scope** option to default to the inbox, so any search that doesn't explicitly set it is limited to the inbox.

The result: a user asking for something like "my emails from the last seven days" gets far fewer results than they see in the Outlook app, because Sent, Archive, and other folders are left out. From the user's point of view the search looks broken.

## What's needed

Searches should return results from across the whole mailbox by default, matching what a user sees when they search in Outlook.

The one case that should stay inbox-scoped is a count query (e.g. "how many unread emails do I have"), so the number matches the unread badge in Outlook rather than being inflated by Sent, Drafts, and Junk.

| Query | Today | Intended |
| --- | --- | --- |
| Find Email — search / list | Inbox only | Whole mailbox |
| Find Email — count only | Inbox only | Inbox (unchanged) |

Choosing a folder explicitly should keep working exactly as it does now.

## Notes

- The inbox default was introduced in #21177 to keep unread counts accurate; the goal here is to keep that count behavior while restoring whole-mailbox search.
- This can be done without a breaking change — no options removed or renamed, and explicit Folder Scope values behave as before.
- The action and Folder Scope descriptions should make the default behavior clear so the agent knows when it's searching the whole mailbox versus the inbox.
- Excluding specific folders (Drafts/Junk/Sent) by default was considered but is fragile and harder to maintain; scoping by the type of request is the simpler path.

## Suggested description copy

Example wording to make the default behavior clear to the agent

**Folder Scope line in the action description:**

> Folder scope defaults by intent: search and list requests scan the whole mailbox (all folders), while count-only requests default to the inbox so unread counts match what the user sees in Outlook. Set Folder Scope explicitly to override — `all` for a mailbox-wide count, or `inbox` (or another folder) to narrow a search.

**Folder Scope prop description:**

> Scope the search to a mail folder. If left unset, the scope depends on the request: count-only queries default to `inbox` (so unread counts match Outlook's inbox badge), and all other searches default to `all` (the whole mailbox). Set it explicitly to override — `all` searches every folder; `inbox`, `sentitems`, `drafts`, `deleteditems`, `junkemail`, `archive`, or a folder ID restrict to that one folder. Use the List Folders action to get folder IDs. Not for use with a shared folder.

[WDINT-569]

michelle0927:
## Summary
- `find-email` search/list requests (`Count Only = false`) now scan the whole mailbox by default instead of being implicitly Inbox-scoped, matching what a user sees when searching in Outlook. Count-only requests stay Inbox-scoped by default so the count matches Outlook's unread-inbox badge. `Folder Scope` can still be set explicitly to override either behavior.
- Adds a `subject` filter prop (`contains(subject, ...)`), separate from the free-text `search` prop.

Closes #21541

## Test plan
- [x] `node --check` and `eslint` pass on the changed file
- [x] Ran the eval suite for `microsoft_outlook` (5 existing evals) — 5/5 passed
- [x] Added a new coverage eval targeting `find-email` specifically (subject filter + whole-mailbox default) — passed, found 52 matching messages with no folder restriction

## Summary by CodeRabbit

* **New Features**
  * Added optional subject-text filtering to the Outlook “Find Email” action.
  * Improved folder-scope behavior for mailbox-wide and inbox-only searches.
  * Clarified descriptions for filtering, ordering, and count options.

* **Bug Fixes**
  * Added validation and ordering support for compatible search filters.
  * Improved count handling for list and search results.
  * Enhanced handling of explicit folder scopes and mailbox-wide searches.


components/microsoft_outlook/actions/find-email/find-email.mjs
ashwins01:
```suggestion
    if (this.folderScope && this.sharedFolderId) {
```
since `inbox` is not the default value now.

components/microsoft_outlook/actions/find-email/find-email.mjs
```diff
@@ -172,14 +170,18 @@ export default {
...
       || this.importance
       || hasFlagged
       || hasHasAttachments;
 
     if (hasSearch && (hasFilterProps || this.orderBy)) {
-      throw new ConfigurationError("`search` cannot be combined with `filter`, `from`, `receivedAfter`, `receivedBefore`, `importance`, `flagged`, `hasAttachments`, or `orderBy` — Graph does not support `$search` with `$filter` or `$orderby`. Use filter props alone, or remove them when using `search`.");
+      throw new ConfigurationError("`search` cannot be combined with `filter`, `subject`, `from`, `receivedAfter`, `receivedBefore`, `importance`, `flagged`, `hasAttachments`, or `orderBy` — Graph does not support `$search` with `$filter` or `$orderby`. Use filter props alone, or remove them when using `search`.");
+    }
+    if (this.subject && this.orderBy) {
+      throw new ConfigurationError("`subject` cannot be combined with `orderBy` — Microsoft Graph does not support sorting results when a `contains()` filter (used by `subject`) is applied. Remove `orderBy`, or filter without `subject` (e.g. use `filter` on an indexed property such as `receivedDateTime` instead).");
```
ashwins01: I went through [the pinned endpoint docs](https://learn.microsoft.com/en-us/graph/api/user-list-messages?view=graph-rest-1.0) ("Using filter and orderby in the same query") and I don't think this is the rule ?
```
Properties that appear in `$orderby` must also appear in `$filter`. Properties that appear in `$orderby` are in the same order as in `$filter`. Properties that are present in `$orderby` appear in `$filter` before any properties that aren't. Failing to do this results in error code `InefficientFilter`.
```
I'd suggest to drop the guard and let Graph's `InefficientFilter` surface. Practically, `subject` + `orderBy: "receivedDateTime desc"` (newest matching email first) seems to be one of the most natural things to ask for here, and this hard-blocks it.

components/microsoft_outlook/actions/find-email/find-email.mjs
```diff
@@ -9,20 +9,13 @@ export default {
...
-    + " [See the documentation](https://learn.microsoft.com/en-us/graph/api/user-list-messages)",
-  version: "1.0.2",
+    "Find (search, list, or count) email messages in a Microsoft Outlook mailbox via Microsoft Graph."
+    + " By default a search or list request (Count Only = false) scans the WHOLE mailbox (`/me/messages`, all folders including Sent, Archive, etc.), matching what you see when searching in Outlook;"
+    + " a count-only request (Count Only = true) with no explicit Folder Scope stays inbox-scoped (`/me/mailFolders/inbox/messages`) and counts ALL inbox messages by default, not just unread ones — also set `Is Read` to `false` to count only unread messages (matching Outlook's unread inbox badge)."
+    + " Set Folder Scope explicitly to override this behavior for either mode."
+    + " To search a shared mailbox folder instead, use **Find Shared Folder Email**."
+    + " [See the documentation](https://learn.microsoft.com/en-us/graph/api/user-list-messages?view=graph-rest-1.0)",
```
ashwins01, line 17: 1. Is mentioning the actual endpoints unwanted information ? should we just keep it descriptive without mentioning the actual microsoft-graph endpoints ? also `isRead` instead of `Is Read` in this description, basis the prop-key.
2. We've also removed the mention of the `userId` & `sharedFolderId` props, while still having them, and redirecting the user/agent to `Find Shared Folder Email` action. Should we remove the `sharedFolderId` prop ? Would you know if theres any advantage of keeping this here while also having another action to specifically read emails from a shared-folder ?

ashwins01 (changes requested):

components/microsoft_outlook/actions/find-email/find-email.mjs
michelle0927: Updated descriptions to reflect documented rules. Updated code to order the filter parameter appropriately.

components/microsoft_outlook/actions/find-email/find-email.mjs
```diff
@@ -172,46 +181,138 @@ export default {
...
+      const leading = [];
+      for (const { key } of orderKeys) {
+        for (const part of filterParts) {
+          if (part.key === key && !leading.includes(part)) leading.push(part);
+        }
+      }
+      const rest = filterParts.filter((part) => !leading.includes(part));
+      filterParts.length = 0;
+      filterParts.push(...leading, ...rest);
+    }
```
ashwins01, line 282: 1. `$orderby` with no filter `throws`. The guard never checks `filterParts.length`, so with zero filter conditions every sort property is "missing".
2. `isRead=false` + `orderBy="receivedDateTime desc"` also `throws`. `isRead` is in the filter, `receivedDateTime` isn't, so the guard fires.

Suggesting the below code block and letting MS Graph decide on whether it is an invalid req.
```
if (this.orderBy && !this.countOnly && !rawFilter) {
  const orderKeys = this.orderBy
    .split(",")
    .map((clause) => this.canonicalOrderKey(clause.trim().split(/\s+/)[0]))
    .filter(Boolean);

  // Reorder so the sorted properties lead the filter, in $orderby order.
  const leading = [];
  for (const key of orderKeys) {
    for (const part of filterParts) {
      if (part.key === key && !leading.includes(part)) {
        leading.push(part);
      }
    }
  }
  const rest = filterParts.filter((part) => !leading.includes(part));
  filterParts.length = 0;
  filterParts.push(...leading, ...rest);
}
```

ashwins01 (changes requested):

ashwins01 (approved):
LGTM. Ready for QA!

mariano-pd (approved):
QA testing successful!

PR #21569, opened 2026-08-05
fix(linear): bound AI-agent response sizes and unblock team-scoped reads

dannyroosevelt:
## Problem

Linear's read actions return every field of every record, and two of them required a `teamId` the user rarely supplies. That's workable for a workflow step reading one value out of a result; for an AI agent, where the whole response enters the model's context, it's a correctness problem.

Measured against a real workspace through the MCP eval harness:

| Action | Avg output | Worst |
|---|---|---|
| Get Teams | 45 KB | 46 KB |
| Get Current User | 43 KB | 43 KB |
| List Views | 26 KB | **44 KB** |
| Search Issues | — | **34 KB** |
| List Projects | 12 KB | 27 KB |
| List Workflow States | — | **22 KB** |

Three evals failed outright by crossing the 25k-token MCP output ceiling. Past that limit Claude Code writes the result to a file and hands the model a **path** — so the model never sees the data, while the tool call still reports success and the agent answers from nothing. There is no error to notice.

`Get Current User` was the clearest case: 43 KB of which the user object was 546 bytes, the rest a nested team list nobody asked for — and it's typically the *first* call an agent makes.

Separately, `teamId` being required on Search Issues and List Projects left an agent with no strategy but to enumerate teams: one task brute-forced four teams and still missed the issue, another called List Projects 20 times.

## Change

**Payload control — additive.** An optional `fields` prop on eight actions (`sections` on Get Current User):

- **Blank returns exactly today's output.** The helper returns the original array by reference when nothing is requested, so existing workflows reading `steps.x.$return_value[0].someField` are untouched. No prop renames, no removals, no default-shape change.
- **`compact`** expands to a curated set per resource, always keeping `id` so the record stays actionable.
- **Explicit subsets** are validated against the documented field list *unioned with the keys the API actually returned* — a wrong guess fails with the valid names instead of silently returning records containing only an id, and a field Linear adds later still works.
- Descriptions name the expensive fields, so the choice is informed rather than guessed.

Get Current User selects whole sections and **skips the fetch** for ones not requested — each is its own round trip, so `compact` saves the request as well as the bytes.

**`teamId` is now optional** on Search Issues and List Projects, matching List Workflow States. Relaxing a requirement can't break an existing caller; omitting it searches everything the credential can see.

**Search Issues is bounded.** It accumulated up to 200 issues across pages on a keyword search. At ~3.0 KB per full-width Issue that's ~600 KB — and even 50 came to 150 KB (~37k tokens), still over the ceiling. The default is now a **local** 25 (~67 KB, ~16.8k tokens, measured), deliberately *not* a change to the shared `DEFAULT_MAX_RECORDS`, which also backs the app's generic paginate helper that the event sources use. `pageInfo` still reports further pages, so this bounds without hiding.

**List Comments** now directs callers to pass `issueId`; without it, it searched every team's comments and was observed returning unrelated threads that merely shared a word.

## Results

| | Before | After |
|---|---|---|
| Get Current User | 43 KB | **1.1 KB** |
| Get Teams | 45 KB | **3.0 KB** |
| List Projects | 12 KB | **2.6 KB** |
| List Views | 26 KB | **7.5 KB** |
| Eval pass rate (2 trials) | 42% | **88%** |
| Cost per suite run | $7.45 | $2.80 |
| Tool calls, worst eval | 37 | 3 |

## Notes for review

- **Why post-fetch projection, not the GraphQL selection set:** Get Teams, List Views and List Initiatives go through the `@linear/sdk` client, which owns its own selection set and gives the action no say — that's *why* Get Teams shipped `inviteHash` and `cycleCooldownTime`. Post-fetch projection is the one approach covering both those and the raw-GraphQL actions. Narrowing the fragments server-side is a reasonable follow-up for the GraphQL-backed ones.
- **Why the code is in `components/linear_app`:** the `components/linear` OAuth actions are thin re-exports, so the API-key app benefits too. Those wrappers override `description`, so they carry the same guidance and get matching version bumps. `components/linear` now requires `@pipedream/linear_app ^0.12.0` — the old `^0.11.1` would not have resolved the new code, since caret doesn't cross a 0.x minor.
- **`projectRecords` skips lazy relation getters.** On an SDK model `comment.user` returns a LinearFetch promise; naming it in a compact set emitted a pending promise. Caught by the contract test below.
- Follows the pattern from #21476 (`google_calendar`), including the `compact` shorthand and validate-against-known-fields behavior.
- Some version numbers skip ahead — iterating against a private registry consumes a version per publish, and it rejects republishing at the same semver.

## Testing

A contract test calls each of the nine changed tools twice over MCP, bare and compact, asserting: the bare response is unchanged, compact is materially smaller, `id` survives projection, and an invalid field name is rejected rather than silently returning stripped records. **36/36 checks pass**, reductions 38–98%. It caught three bugs the eval suite could not see, including two actions that shipped without the prop taking effect at all.

Eval suite: 24 evals at 2 trials, **21 pass / 1 fail / 2 warn**. Known residual: an agent that explicitly raises `limit` for a bulk "export everything" request can still exceed the ceiling — the description warns, but a description is weaker than a bound. Making `compact` the default would close it and is a deliberate, separate compatibility decision.

## Summary by CodeRabbit

* **New Features**
  * Added compact and custom field selection for teams, initiatives, projects, views, issues, comments, workflow states, and view issues.
  * Added configurable response sections for current-user details.
  * Enabled workspace-wide project and issue searches without requiring a team.
  * Added compact response defaults, pagination guidance, and improved result limits.
* **Bug Fixes**
  * Added validation for unsupported fields and response sections.
* **Documentation**
  * Expanded guidance on payload sizes, filtering, available fields, and compact options.


dannyroosevelt:
### ⚠️ Merge ordering — needs a follow-up PR

`components/linear`'s actions are thin re-exports of `@pipedream/linear_app`, and this monorepo sets `link-workspace-packages=false`, so that dependency resolves from **npm**, not the workspace.

That means the two halves of this PR land at different times:

| | Status on merge |
|---|---|
| `components/linear_app/*` (API key) | ✅ correct — `fields` implemented and shipped together |
| `components/linear/*` (OAuth) | ⚠️ descriptions document `fields`, but resolve `linear_app@0.11.x`, which doesn't have it |

I originally bumped the dependency to `^0.12.0`, but `linear_app@0.12.0` doesn't exist on npm until this merges and CI publishes it — that broke `pnpm install --frozen-lockfile` with `ERR_PNPM_OUTDATED_LOCKFILE` across four jobs, so it's reverted to `^0.11.1` here.

**Follow-up required:** once `@pipedream/linear_app@0.12.0` is published, bump `components/linear/package.json` to `^0.12.0` and regenerate `pnpm-lock.yaml`. Caret won't cross a 0.x minor, so without that step the OAuth actions keep resolving the old package.

Happy to split this into two PRs (linear_app first, linear second) if that's the preferred sequencing here.

PR #21575, opened 2026-08-05
fix(telegram_bot_api): clear stale webhook before registering new one

VibhanshuPandey:
## Summary

Telegram instant sources (`message-updates`, `new-updates`, `channel-updates`, `new-bot-command-received`) stop receiving events after re-deployment due to a stale webhook/secret mismatch.

### Root cause

`webhooks.mjs` (the shared base) has two problems:

**1. `deploy()` throws on any existing webhook**
```js
if (result.url?.length > 0) {
  throw new ConfigurationError("...");
}
```
When a source is deleted with `ignoreHookErrors: true` (common in programmatic/Connect usage), the `deactivate()` hook can fail silently, leaving Telegram's webhook registered. The next deployment then always throws `ConfigurationError`, and users have to manually call `deleteWebhook` to recover.

**2. `activate()` doesn't clear state before registering**
```js
async activate() {
  const secret = uuid();
  this.setSecret(secret);
  await this.telegramBotApi.createHook(this.http.endpoint, ...);
}
```
If `deactivate()` failed silently, `createHook()` registers a **new** Pipedream endpoint with a **new** UUID secret. The old source endpoint can still be alive with the **old** secret. Telegram now delivers to the new URL, but `run()` checks `x-telegram-bot-api-secret-token` against `this.getSecret()`. If the old source is alive with the old secret, the check passes on the first event (before the secret is rotated) but fails or behaves inconsistently on subsequent events — causing the source to appear to "fire only once."

### Fix

**`deploy()`** — replace the `ConfigurationError` throw with a `deleteHook()` call. A stale webhook from a prior source should not block redeployment; clearing it and proceeding is always the correct behavior.

**`activate()`** — add `deleteHook()` before `createHook()`. This makes activation idempotent: regardless of what was previously registered, the new source always starts from a clean slate with the correct secret.

### Versions bumped

| Source | Before | After |
|--------|--------|-------|
| `telegram_bot_api-message-updates` | 0.1.7 | 0.1.8 |
| `telegram_bot_api-new-updates` | 0.1.6 | 0.1.7 |
| `telegram_bot_api-channel-updates` | 0.1.6 | 0.1.7 |
| `telegram_bot_api-new-bot-command-received` | 0.0.6 | 0.0.7 |

### Test plan

- [ ] Deploy a Telegram `message-updates` source with a bot that already has a webhook registered → should deploy successfully (previously threw `ConfigurationError`)
- [ ] Send multiple messages to the bot → all should appear as events (previously only the first message fired)
- [ ] Delete the source and redeploy → should register a fresh webhook cleanly
- [ ] Verify `deactivate()` still removes the webhook on a clean stop

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Telegram webhook activation and deployment by removing matching stale webhooks before registering new ones.
  * Preserved configuration errors for incompatible or unrelated webhooks.
  * Improved handling of invalid or unrecognized webhook URLs during configuration.

* **Maintenance**
  * Updated Telegram integration and event source version metadata.
  * Improved webhook management reliability while leaving unrelated webhook configurations untouched.

### DIFF
--- a/components/telegram_bot_api/sources/common/webhooks.mjs
+++ b/components/telegram_bot_api/sources/common/webhooks.mjs
@@ -5,6 +5,22 @@ import { ConfigurationError } from "@pipedream/platform";
 
 const PIPEDREAM_WEBHOOK_DOMAIN = "pipedream.net";
 
+/**
+ * Returns true only when `url` is hosted on a pipedream.net subdomain or
+ * exactly on pipedream.net itself. Parsing the hostname prevents path-based
+ * spoofing (e.g. https://evil.com/pipedream.net would pass a naive .includes()
+ * check but fails here because the hostname is evil.com).
+ */
+function isPipedreamWebhook(url) {
+  try {
+    const { hostname } = new URL(url);
+    return hostname === PIPEDREAM_WEBHOOK_DOMAIN
+      || hostname.endsWith(`.${PIPEDREAM_WEBHOOK_DOMAIN}`);
+  } catch {
+    return false;
+  }
+}
+
 export default {
   props: {
     telegramBotApi,
@@ -31,7 +47,7 @@ export default {
        */
       const { result } = await this.telegramBotApi.getWebhookInfo();
       if (result.url?.length > 0) {
-        if (result.url.includes(PIPEDREAM_WEBHOOK_DOMAIN)) {
+        if (isPipedreamWebhook(result.url)) {
           console.log(`Removing stale Pipedream webhook (${result.url}) before deploying...`);
           await this.telegramBotApi.deleteHook();
         } else {
@@ -61,13 +77,17 @@ export default {
       this.setSecret(secret);
       /**
        * Clear any stale Pipedream-owned webhook before registering the new one.
-       * This ensures the new secret is always authoritative. Non-Pipedream webhooks
-       * are not explicitly deleted here — deploy() guards that path — and Telegram's
-       * setWebhook call below replaces whatever is registered regardless.
+       * If the existing webhook belongs to an external (non-Pipedream) service,
+       * throw ConfigurationError — Telegram's setWebhook would overwrite it
+       * otherwise, silently breaking the external integration.
        */
       const { result } = await this.telegramBotApi.getWebhookInfo();
-      if (!result.url?.length || result.url.includes(PIPEDREAM_WEBHOOK_DOMAIN)) {
-        await this.telegramBotApi.deleteHook();
+      if (result.url?.length > 0) {
+        if (isPipedreamWebhook(result.url)) {
+          await this.telegramBotApi.deleteHook();
+        } else {
+          throw new ConfigurationError("[Telegram only supports](https://core.telegram.org/bots/api#setwebhook) a single webhook at a time, for a given Bot. To get around this, you can reuse an existing Telegram Bot source or disable the active source and try again. [View all your sources here](https://pipedream.com/sources).");
+        }
       }
       await this.telegramBotApi.createHook(this.http.endpoint, this.getEventTypes(), secret);
     },
--- a/components/telegram_bot_api/sources/common/webhooks.mjs
+++ b/components/telegram_bot_api/sources/common/webhooks.mjs
@@ -1,6 +1,9 @@
 import telegramBotApi from "../../telegram_bot_api.app.mjs";
 import constants from "./constants.mjs";
 import { v4 as uuid } from "uuid";
+import { ConfigurationError } from "@pipedream/platform";
+
+const PIPEDREAM_WEBHOOK_DOMAIN = "pipedream.net";
 
 export default {
   props: {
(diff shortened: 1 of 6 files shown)


pipedream-component-development:

pipedream-component-development:

VibhanshuPandey:
Addressed in 6b19d22: added ownership verification before deletion in both hooks.

**deploy()**: checks if the existing URL contains `pipedream.net`. Stale Pipedream-owned webhooks are deleted before registering the new one. External/non-Pipedream URLs fall through to the original `ConfigurationError` unchanged.

**activate()**: calls `getWebhookInfo()` first and only issues `deleteHook()` when the registered URL is Pipedream-owned or empty. A non-Pipedream URL is not explicitly deleted (deploy() guards that path, and Telegram's `setWebhook` call replaces it anyway per API semantics).

Added `PIPEDREAM_WEBHOOK_DOMAIN = 'pipedream.net'` as a named constant.

VibhanshuPandey:
**Re: ownership check on latest commit (bf5c246)**

CodeRabbit is suggesting a source-specific check — e.g. `result.url === this.db.get('webhookEndpoint')` — to ensure only this source's own previously-registered endpoint is deleted. That approach is the right design for same-source redeployments, but it cannot fix the root bug this PR addresses.

**Why db-based ownership can't solve the orphan case:**

The bug occurs when a prior Pipedream source is deleted with `ignoreHookErrors: true` (Pipedream Connect's default). `deactivate()` runs but silently fails, so Telegram still has the old endpoint registered. When a *new* source is then deployed for the same bot:

- The new source has a fresh `$.service.db` — `db.get('webhookEndpoint')` is `null`
- The orphaned webhook URL belongs to a *different* source's db (now gone)
- `result.url !== null` → would always throw ConfigurationError

The user is still blocked and must manually call `deleteWebhook`. The bug is not fixed.

**Why `isPipedreamWebhook()` is the right scope for this fix:**

`isPipedreamWebhook()` targets the specific failure mode: stale Pipedream-owned endpoints left by `ignoreHookErrors` cleanup. It correctly:
- Cleans up any orphaned `*.pipedream.net` endpoint (fixes the bug)
- Preserves the `ConfigurationError` for external URLs (protects external integrations)

The known trade-off: two *concurrent* Pipedream sources for the same bot could conflict — one's `deploy()`/`activate()` would delete the other's webhook. However this scenario was already broken before this PR (the original code would throw ConfigurationError for *any* existing webhook, including other Pipedream sources), and the real fix for concurrent multi-source use is a platform-level guard in Pipedream Connect's `triggers.deploy()`, not in the component lifecycle hooks.

Happy to add a comment in the code documenting this trade-off if that helps reviewers.

VibhanshuPandey:
@michelle0927 gentle bump on this one — it's been open a week with no maintainer review yet.

Status: all CI green and CodeRabbit SUCCESS at `bf5c246`; `mergeStateStatus` is only BLOCKED on a required approving review. The last outstanding CodeRabbit suggestion (db-based ownership check) was answered in the comment above — a deleted source's `$.service.db` is gone, so `db.get("webhookEndpoint")` returns null and the orphaned-webhook case this PR exists to fix would still throw. Ownership is decided by parsing `new URL(result.url).hostname` rather than a substring match, so `https://evil.com/pipedream.net` is not treated as Pipedream-owned.

Impact while this sits: Telegram instant sources fire exactly once after deploy, then silently drop every subsequent update, because `activate()` overwrites the webhook without clearing the old registration and the running source ends up holding a different secret than the one Telegram signs with. The only symptom is `Could not identify sender identity, exiting...` in the source logs, which looks identical to "no events at all" from the outside. Current workaround is a manual `deleteWebhook` before every redeploy.

Happy to rebase or split it up if that makes review easier. If you're not the right person for this area, could you point me at whoever owns `telegram_bot_api` sources? Thanks!

VibhanshuPandey:
Widening the ask so this isn't waiting on one person — @ashwins01 @vetrivigneshwaran @GTFalcao, would any of you be able to take a look?

Short version: CI is fully green and CodeRabbit is SUCCESS at `bf5c246`, so the only thing holding this is an approving review. Details and the answer to the one open CodeRabbit thread are in the comment directly above.

It's a small, self-contained fix to `components/telegram_bot_api/sources/common/webhooks.mjs` — `deploy()`/`activate()` become idempotent by clearing a stale Pipedream-owned webhook before registering the new one, while still throwing `ConfigurationError` for externally-owned ones. Without it, Telegram instant sources fire once after deploy and then silently discard every later update.

components/telegram_bot_api/sources/common/webhooks.mjs
```diff
@@ -3,6 +3,24 @@ import constants from "./constants.mjs";
...
 
+const PIPEDREAM_WEBHOOK_DOMAIN = "pipedream.net";
+
+/**
+ * Returns true only when `url` is hosted on a pipedream.net subdomain or
+ * exactly on pipedream.net itself. Parsing the hostname prevents path-based
+ * spoofing (e.g. https://evil.com/pipedream.net would pass a naive .includes()
+ * check but fails here because the hostname is evil.com).
+ */
+function isPipedreamWebhook(url) {
```
vetrivigneshwaran: isPipedreamWebhook() matches every *.pipedream.net endpoint. Another Telegram source using the same bot token can therefore be deleted in deploy() or activate(), which stops that source from receiving
updates. Delete the webhook only after matching this.http.endpoint or another source-specific identity. Otherwise, keep the ConfigurationError path.

vetrivigneshwaran (dismissed):
Added review comment and also kindly address the coderabbit review related to security and privacy.

components/telegram_bot_api/sources/common/webhooks.mjs
VibhanshuPandey: Thanks — updated in `bdd19fba`. Ownership is now decided by exact endpoint identity: `isOwnWebhook()` compares the currently registered URL against `this.http.endpoint` (normalized to origin + path), so only this source's own registration is cleared. Any other webhook, whether external or another Pipedream source sharing the bot token, keeps the `ConfigurationError` path — and in `activate()` that error is thrown before `createHook()`, so `setWebhook` can no longer overwrite it either. That also resolves the CodeRabbit authorization finding.

One consequence worth noting: a webhook orphaned by a *deleted* source (deactivate skipped via `ignoreHookErrors`) has a different endpoint, so it is no longer cleaned up automatically and the user gets the `ConfigurationError` telling them to reuse or disable the existing source. That is the safe behaviour given ownership can't be proven, and it is still an improvement over the previous state where the mismatched secret caused every update to be dropped silently.

VibhanshuPandey:
@michelle0927 could you take a look when you have a moment? The review feedback from @vetrivigneshwaran is addressed in `bdd19fba` — webhook cleanup is now scoped to this source's own `this.http.endpoint` (exact origin+path match), so a webhook belonging to another Telegram source on the same bot token, or to an external service, is never deleted and keeps the existing `ConfigurationError` path. All checks are green on that commit.

Happy to adjust further if the approach needs a different shape.

michelle0927 (approved):
Looks good, tests passed! Ready for release!

PR #21311, opened 2026-07-02
[Components] speak_ai - add instant-hook sources and AI Chat / Find Media actions

speakai:
## Summary

This upgrades the existing `@pipedream/speak_ai` app with a set of richer instant-hook sources and two new actions. The change is **fully additive**: every existing component and app method is kept byte-for-byte, and the new components layer on top of the same app file and the same registry conventions (`common/constants.mjs`, `sources/common` base pattern, `test-event.mjs` + `sampleEmit`, action `annotations`).

- Sources: **2 → 7** (5 new)
- Actions: **4 → 6** (2 new)
- App name, author, and homepage are unchanged.

### New sources (instant, webhook-backed)
| Key | Fires when |
| --- | --- |
| `speak_ai-new-transcription` | A media transcript is ready |
| `speak_ai-new-captions` | Caption files (SRT/VTT) are ready |
| `speak_ai-new-sentiment` | Sentiment analysis is ready |
| `speak_ai-new-ai-chat` | An AI Chat response is ready |
| `speak_ai-new-recording` | A recorder submission is received |

### New actions
| Key | What it does | Notes |
| --- | --- | --- |
| `speak_ai-run-ai-chat` | Runs an AI Chat against a folder and/or media. | `readOnlyHint: false`. |
| `speak_ai-find-media` | Looks up a single media item by ID. | `readOnlyHint: true`. |

Each new source registers a webhook with Speak AI when enabled and removes it when disabled, hydrates each delivery with the full insight, stores the webhook id in `db`, uses `dedupe: "unique"`, and emits via `generateMeta`. The existing sources (`new-media-created-instant`, `new-text-analyzed-instant`), actions (`upload-media`, `analyze-text`, `get-transcription`, `list-folder-id-options`), and the shared `common/` base files are byte-identical to upstream. `speak_ai.app.mjs` is a strict superset — only added prop definitions and methods, nothing existing changed.

## Checklist
Please check the following items before your PR can be reviewed:

### Versioning
- [x] All components updated in this PR had their version updated (`0.0.1` for new ones)
- [x] The app updated in this PR had its `package.json`'s version updated (→ `0.3.0`)

### New app
If this is a new app, please submit [an app integration request](https://github.com/PipedreamHQ/pipedream/issues/new?template=app---service-integration.md) - the PR will only be reviewed after the app is integrated.
- [x] The app updated in this PR is already integrated (this is an additive upgrade to the existing `speak_ai` app, not a new app)

### CodeRabbit review
[...]

PR #21579, opened 2026-08-06
feat(speak_ai): documented endpoints, instant sources and AI Chat / Find Media actions

ashwins01:
## Summary

Fixes review comments from #21311 since we are unable to make the changes directly in their fork as `maintainerContribution` is `disabled` by the submitted user.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **New Features**
  - Added actions for finding media insights and running AI chats with prompts, media, and assistant options.
  - Added event triggers for AI chat responses, captions, recordings, sentiment analysis, and completed transcriptions.
  - Added SRT and VTT caption export options.
  - Added tools for retrieving transcripts, exporting media, and managing AI prompts.

- **Improvements**
  - Improved webhook handling, event identification, and duplicate-event prevention.
  - Updated documentation, descriptions, validation, and package version.


vatsal2210:
Thanks @ashwins01 for moving this over and doing the endpoint rewrite, and @vetrivigneshwaran for the detailed review. Agreed on the `/apps/*` paths: those were wrong and never existed. `GET /v1/media/insight/:mediaId`, `POST /v1/media/export/:mediaId/:fileType` and `GET /v1/prompt` are the right calls, and `mediaId` in the path with `userId` as an optional query param is correct for Find Media.

I went back through this branch against our server source rather than the docs, and opened #21582 into `feat/speak_ai` with the result. High level:

**One bug that would ship.** `sources/common/webhook.mjs` stores `response.data._id`, but `POST /v1/webhook` returns the id as `data.webhookId`. It comes back undefined, so `deactivate()` never calls the delete and disabling a source leaves a live subscription delivering forever. It's in the shared base, so it also affects the two already released sources. That one is on us: the `_id` in our Create Webhook 200 example is stale and I'm correcting it.

**Three endpoint details where our API reference was thinner than the endpoint.**

- `POST /v1/prompt` does accept `folderId`. Its validation ends in `.or('folderId', 'mediaIds')`, so either one satisfies the request and neither is required alone. Requiring `mediaIds` drops the folder-wide question, which is a common use. Both are optional again in #21582, with a guard requiring at least one.
- That endpoint does return `promptId`, alongside `messageId`, `state`, `answer`, `totalMedia` and `references`. The 200 example just omits them.
- Every delivery already carries `deliveryId` and `eventType`; the body is built as `{ ...metaData, eventType, deliveryId }` with a fresh id per delivery, reused across retries. So `getEventId() => resource.deliveryId` under `dedupe: "unique"` is correct as written: analyzed and a later reanalyzed emit separately, retries collapse. Flagging it so it doesn't get changed later.

**Also in #21582:** `source: "pipedream"` on webhook create, which the API accepts and which otherwise defaults to `speak`, so these subscriptions are attributable in the user's account. Plus a docs link on List Folder ID Options, the one component without one.

I verified all 11 endpoints the component calls against their routes and validation, confirmed all 16 event names in `events.mjs` are valid enum members, and confirmed every documentation link and anchor resolves. The table is in #21582.

On our side I'm fixing the two stale response examples that caused most of this, in the AI Chat and Webhooks references. Everything now lives under docs.speakai.co, which covers the API, the MCP server and the help content in one place, so the per-endpoint anchors in this branch are the durable links. The MCP tool pages document the same endpoints from the agent side if they're ever useful for prop wording, for example https://docs.speakai.co/mcp/tools/magic-prompt/ask_ai_chat/

Happy to fold #21582 into this branch however is easiest, or to adjust anything. cc @vatsal2210

vatsal2210:
Quick follow-up: the two stale response examples I mentioned are corrected and live.

- [Create Webhook](https://docs.speakai.co/api/webhooks/#post-webhook) now shows what the endpoint actually returns, `{ "status": "success", "data": { "webhookId": "..." } }`, instead of a full webhook document with `_id`. That example is what led to the `data._id` read in this branch, so it was our error to begin with.
- [Ask AI Chat](https://docs.speakai.co/api/ai-chat/#post-prompt) now includes `promptId` and `messageId` in the 200 response, which the endpoint has always returned.

While checking those I found a third one and fixed it too: [List Webhooks](https://docs.speakai.co/api/webhooks/#get-webhook) documented `data` as a bare array when it is `{ totalCount, webhooks: [...] }`. Nothing in this component reads it, so it has no bearing here, but it is right now if anyone goes looking.

#21582 is ready whenever you want it. Happy to rebase or reshape it if that is easier to take. cc @vatsal2210

components/speak_ai/actions/run-ai-chat/run-ai-chat.mjs
```diff
@@ -0,0 +1,80 @@
...
+      ],
+    },
+    folderId: {
+      propDefinition: [
+        app,
+        "folderId",
+      ],
+      description: "Answer the prompt from every media file in this folder. Set this, **Media IDs**, or both",
+      optional: true,
+    },
```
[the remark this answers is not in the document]
vatsal2210: Fixed in #21583, with one correction to the premise.

There was no guidance in the shared definition to preserve. The app file read `description: "The ID of the folder to upload or retrieve files from"` with no format, example or source, so the override was not dropping anything, and four actions inherit that same thin description: Analyze Text, Get Transcription, Run AI Chat and Upload Media. Fixing only the override would have left the other three as they were, so both are updated.

Folder IDs are generated by the same helper as media IDs, a sha256 digest truncated to 12 characters, so the example is `905c208f1c07`.

The `mediaIds` formatting point is correct and is fixed, with bold kept for cross-component references.

vatsal2210:
Thanks for merging #21582 so quickly, @ashwins01.

CodeRabbit then flagged the `folderId` description on Run AI Chat, which came from that PR. #21583 addresses it, into `feat/speak_ai` again.

Two things worth surfacing from it rather than leaving in the diff:

**The gap was in the app file, not just the one call site.** The shared `folderId` definition read "The ID of the folder to upload or retrieve files from" with no format, example or source, and four actions inherit it. Both it and the Run AI Chat override now follow the pattern `mediaId` already uses.

**Speak AI IDs are not all one format,** which is easy to get wrong and is not written down anywhere obvious. `mediaId`, `folderId`, `promptId` and `messageId` are 12-character ids from a truncated sha256. `userId`, `webhookId` and `companyId` are 24-character Mongo ObjectIds. I checked both examples already in this branch and they are correct as they stand: `f8eb3c22bec3` for `mediaId`, and `6a708504253783a639e51914` for `userId` on Find Media.

That check found a mistake on our side, not yours: our AI Chat reference was showing `promptId` as a 24-character ObjectId. Corrected, and the format is now stated rather than left to be inferred from an example: https://docs.speakai.co/api/ai-chat/#post-prompt

The three remaining CodeRabbit comments from earlier are on parts of the branch outside our changes, so I have left those to you. Happy to pick any of them up if it helps. cc @vatsal2210

components/speak_ai/actions/find-media/find-media.mjs
```diff
@@ -0,0 +1,48 @@
...
+  },
+  type: "action",
+  props: {
+    app,
+    mediaId: {
+      propDefinition: [
+        app,
+        "mediaId",
+      ],
+      description: "The media file to look up",
```
vetrivigneshwaran:
```suggestion
      description: "A Speak AI media ID, for example `b4994aa1267c`. Get it from the `mediaId` field returned by Speak AI.",
```

components/speak_ai/actions/run-ai-chat/run-ai-chat.mjs
```diff
@@ -0,0 +1,80 @@
...
+        app,
+        "prompt",
+      ],
+    },
+    folderId: {
+      propDefinition: [
+        app,
+        "folderId",
+      ],
+      description: "Answer the prompt from every media file in this folder. Set this, **Media IDs**, or both",
```
vetrivigneshwaran:
```suggestion
description: "A Speak AI folder ID, for example `905c208f1c07`. Get it from the `folderId` field returned by Speak AI. Answer the prompt from every media file in this folder. Set this, **Media IDs**, or both.",
```

components/speak_ai/actions/get-transcription/get-transcription.mjs
```diff
@@ -27,6 +27,7 @@ export default {
           folderId,
         }),
       ],
+      description: "The media file to retrieve the full transcription for",
```
vetrivigneshwaran:
```suggestion
      description: "A Speak AI media ID, for example `b4994aa1267c`. Get it from the `mediaId` field returned by Speak AI.",
```

components/speak_ai/sources/new-sentiment/new-sentiment.mjs
```diff
@@ -0,0 +1,36 @@
...
+      const { data } = await this.app.getInsight({
+        mediaId: resource.mediaId,
+      });
+      return data;
+    },
+    generateMeta(resource) {
+      return {
+        id: this.getEventId(resource),
+        summary: `New Sentiment Analysis: ${resource.mediaId}`,
+        ts: Date.now(),
```
vetrivigneshwaran:
```suggestion
       ts: new Date(resource.data.updatedAt ?? resource.data.createdAt).getTime(),
```

Use the API event timestamp.

  Line 31 assigns the execution time. A delayed delivery or retry then changes the emitted event timestamp. The fixture provides resource.data.updatedAt in  components/speak_ai/sources/new-sentiment/test-event.mjs, Line 22.

components/speak_ai/sources/new-recording/new-recording.mjs
```diff
@@ -0,0 +1,35 @@
...
+      const { data } = await this.app.getInsight({
+        mediaId: resource.mediaId,
+      });
+      return data;
+    },
+    generateMeta(resource) {
+      return {
+        id: this.getEventId(resource),
+        summary: `New Recording Captured: ${resource.mediaId}`,
+        ts: Date.now(),
```
vetrivigneshwaran:
```suggestion
       ts: new Date(resource.data.updatedAt ?? resource.data.createdAt).getTime(),
```

Line 30 assigns the execution time. A delayed delivery or retry the changes the emitted event timestamp. The fixture provides  resource.data.updatedAt in  components/speak_ai/sources/new-recording/test-event.mjs, Line 19.

components/speak_ai/sources/new-transcription/new-transcription.mjs
```diff
@@ -0,0 +1,36 @@
...
+      const { data } = await this.app.getTranscript({
+        mediaId: resource.mediaId,
+      });
+      return data;
+    },
+    generateMeta(resource) {
+      return {
+        id: this.getEventId(resource),
+        summary: `New Transcription: ${resource.mediaId}`,
+        ts: Date.now(),
```
vetrivigneshwaran:
```suggestion
        ts: new Date(resource.data.updatedAt ?? resource.data.createdAt).getTime(),
```
Line 31 assigns the execution time. A delayed delivery or retry then  changes the emitted event timestamp. The fixture provides  resource.data.createdAt in  components/speak_ai/sources/new-transcription/test-event.mjs, Line 60.

components/speak_ai/sources/common/webhook.mjs
```diff
@@ -62,26 +67,34 @@ export default {
...
         ...args,
       });
     },
   },
   async run({ body }) {
+    if (!body || typeof body !== "object" || Array.isArray(body)) {
+      console.log("Skipping delivery: the webhook body is missing or is not an event object.");
+      return;
+    }
     await this.processResource(body);
```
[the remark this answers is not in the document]
vatsal2210: I looked into this one and I do not think it applies to sources, so I have not made the change. Flagging rather than silently skipping.

Sources are not passed a `$`. The step-context argument exists for actions, whose signature is `run({ $ })`. A source's `run` receives the triggering event instead, and the established webhook sources in this repo are consistent on that:

```
components/calendly_v2/sources/common/common.mjs:135:  async run(event)
components/shopify/sources/common/webhook.mjs:74:     async run(event)
components/shopify/sources/common/polling.mjs:36:     async run()
```

So there is no `$` in `run({ body })` to thread through `processResource()` and into `getData()`. Adding one would mean passing `undefined`, and `axios(undefined, ...)` is worse than what is there now.

The current behaviour is already correct for this reason. `_makeRequest({ $ = this, ... })` defaults to `this`, and for a source `this` **is** the platform context that `@pipedream/platform`'s `axios` expects, so `listPrompts()`, `getInsight()`, `getTranscript()` and `exportMedia()` do get context logging and error handling when called from a source. The default exists precisely so the same app methods work from both actions, which pass `$`, and sources, which cannot.

Happy to be corrected if there is a source pattern that does receive a step context, but I could not find one in the repo.

The other two comments in this batch are handled: the action descriptions were expanded and the `Speak Ai` casing fixed, and the remaining ID-description gap is in #21583, which also picks up Get Transcription, which had the same problem and was not flagged.

vatsal2210:
@ashwins01 I have gone through every open review comment on this branch, ours and yours alike, so the path to `master` is clear. Status:

**1. Document non-obvious ID inputs** — mostly already done by you: `mediaId`, `mediaIds` and the `userId` on Find Media all carry a format, example and source. What remained was overrides replacing that guidance. #21583 now covers all three: `folderId` on Run AI Chat, `mediaId` on Find Media, and `mediaId` on Get Transcription. The last one had the identical problem and was not flagged, so it is fixed there too rather than coming back in another round. That is every propDefinition override in the app that sets its own description; none are left carrying an ID without a format.

**2. Pass `$` through webhook source API calls** — I do not think this one applies, and I have replied on the thread rather than making the change. Sources are not passed a `$`; `run({ $ })` is the action signature, while a source's `run` receives the event. `calendly_v2` and `shopify` webhook sources are both `async run(event)`. There is no `$` to thread through, and `_makeRequest({ $ = this, ... })` already resolves to the right context for a source. Happy to be shown otherwise.

**3. Expand the action descriptions** — done on your side. All five read well now, and the `Speak Ai` casing is fixed.

So the only thing outstanding is merging #21583. CI here is green: lint, spellcheck and publish all pass, and `mergeable_state` is `blocked` only on review approval rather than on anything failing.

Once #21583 is in, I believe this branch is ready for `master`. If anything else surfaces in review, send it my way and I will turn it round quickly, whether it is in our components or elsewhere in the branch. cc @vetrivigneshwaran @vatsal2210

components/speak_ai/sources/new-ai-chat/new-ai-chat.mjs
```diff
@@ -0,0 +1,37 @@
...
+  methods: {
+    ...common.methods,
+    getEvents() {
+      return [
+        events.CHAT_STATUS,
+      ];
+    },
+    async getData(resource) {
+      const { data: { history } } = await this.app.listPrompts();
+      const match = history.find(({
```
vetrivigneshwaran: Match prompt history records by id.

  GET /prompt returns history records with id, while chat.status
  provides promptId and messageId. The current predicate misses matching
  records and causes the wrapper to emit the raw webhook resource as data.
  Compare history.id with resource.promptId; keep the fallback.

vetrivigneshwaran (changes requested):

components/speak_ai/actions/find-media/find-media.mjs
vatsal2210: Applied in #21583, using your wording exactly. Live on the branch now:

```js
description: "A Speak AI media ID, for example `b4994aa1267c`. Get it from the `mediaId` field returned by Speak AI.",
```

The format checks out: media IDs come from a sha256 digest truncated to 12 characters, so `b4994aa1267c` is the right shape.

components/speak_ai/actions/get-transcription/get-transcription.mjs
vatsal2210: Applied in #21583, same wording. Live on the branch now:

```js
description: "A Speak AI media ID, for example `b4994aa1267c`. Get it from the `mediaId` field returned by Speak AI.",
```

Worth noting this one was not in the original batch of comments. It had the identical gap, so I picked it up rather than leave it for another round.

components/speak_ai/actions/run-ai-chat/run-ai-chat.mjs
vatsal2210: Applied in #21583, and you were right on the formatting where CodeRabbit was not. Live on the branch now:

```js
description: "A Speak AI folder ID, for example `905c208f1c07`. Get it from the `folderId` field returned by Speak AI. Answer the prompt from every media file in this folder. Set this, **Media IDs**, or both.",
```

I briefly kept `mediaIds` in code formatting, because CodeRabbit had asked for that specifically earlier in this PR and marked it resolved. Checking the standard settles it your way:

> Enclose sample input values in backticks (`), refer to other props using **bold**
> — https://pipedream.com/docs/components/guidelines/

So the sample ID keeps backticks and the sibling prop is **Media IDs**, which also matches the `ConfigurationError` this action already throws. Reverted in `929f75b`.

components/speak_ai/sources/new-sentiment/new-sentiment.mjs
vatsal2210: Fixed in #21583, and you are right that `Date.now()` was wrong here. One correction to the suggested line, because it would throw.

`processResource` passes the hydrated payload as generateMeta's **second** argument:

```js
this.$emit({ ...resource, data }, this.generateMeta(resource, data));
```

`resource` is the raw webhook body, which carries `eventType`, `deliveryId`, `state` and `mediaId` but no `data` key. So `resource.data.updatedAt` dereferences `undefined` and every delivery would fail. The `data` visible in `test-event.mjs` is the emitted event, which is the merged `{ ...resource, data }`, not what generateMeta receives.

So the sources now declare `generateMeta(resource, data)` and read the second argument, with the parse shared on the source base rather than repeated per file:

```js
getEventTs(data) {
  const ts = Date.parse(data?.updatedAt ?? data?.createdAt ?? "");
  return Number.isNaN(ts) ? Date.now() : ts;
},
```

Applied to all seven sources, not just the three flagged. New AI Chat had the same gap and carries `createdAt`, and the two pre-existing instant sources inherit it from the shared base. New Captions falls back to now by design, since its payload is the caption text itself and has no date.

components/speak_ai/sources/new-recording/new-recording.mjs
vatsal2210: Fixed in #21583, and you are right that `Date.now()` was wrong here. One correction to the suggested line, because it would throw.

`processResource` passes the hydrated payload as generateMeta's **second** argument:

```js
this.$emit({ ...resource, data }, this.generateMeta(resource, data));
```

`resource` is the raw webhook body, which carries `eventType`, `deliveryId`, `state` and `mediaId` but no `data` key. So `resource.data.updatedAt` dereferences `undefined` and every delivery would fail. The `data` visible in `test-event.mjs` is the emitted event, which is the merged `{ ...resource, data }`, not what generateMeta receives.

So the sources now declare `generateMeta(resource, data)` and read the second argument, with the parse shared on the source base rather than repeated per file:

```js
getEventTs(data) {
  const ts = Date.parse(data?.updatedAt ?? data?.createdAt ?? "");
  return Number.isNaN(ts) ? Date.now() : ts;
},
```

Applied to all seven sources, not just the three flagged. New AI Chat had the same gap and carries `createdAt`, and the two pre-existing instant sources inherit it from the shared base. New Captions falls back to now by design, since its payload is the caption text itself and has no date.

components/speak_ai/sources/new-transcription/new-transcription.mjs
vatsal2210: Fixed in #21583, and you are right that `Date.now()` was wrong here. One correction to the suggested line, because it would throw.

`processResource` passes the hydrated payload as generateMeta's **second** argument:

```js
this.$emit({ ...resource, data }, this.generateMeta(resource, data));
```

`resource` is the raw webhook body, which carries `eventType`, `deliveryId`, `state` and `mediaId` but no `data` key. So `resource.data.updatedAt` dereferences `undefined` and every delivery would fail. The `data` visible in `test-event.mjs` is the emitted event, which is the merged `{ ...resource, data }`, not what generateMeta receives.

So the sources now declare `generateMeta(resource, data)` and read the second argument, with the parse shared on the source base rather than repeated per file:

```js
getEventTs(data) {
  const ts = Date.parse(data?.updatedAt ?? data?.createdAt ?? "");
  return Number.isNaN(ts) ? Date.now() : ts;
},
```

Applied to all seven sources, not just the three flagged. New AI Chat had the same gap and carries `createdAt`, and the two pre-existing instant sources inherit it from the shared base. New Captions falls back to now by design, since its payload is the caption text itself and has no date.

components/speak_ai/sources/new-ai-chat/new-ai-chat.mjs
vatsal2210: I checked this one against our server and I do not think the change is right, so I have not made it. Flagging rather than skipping quietly.

`GET /v1/prompt` history records carry `promptId` and `messageId`, alongside `id`. The mapping in our controller builds each record as:

```js
const mappedMessages = messages.map((prompt) => ({
  id: prompt._id.toString(),   // the message document Mongo _id
  promptId: data.promptId,     // the chat id
  messageId: prompt.messageId, // the message id
  ...
}));
```

So `id` is the message document Mongo `_id`, which is a 24 character ObjectId. `promptId` is a different value in a different format, a 12 character id from a truncated sha256, the same family as `mediaId` and `folderId`.

Comparing `history.id` with `resource.promptId` would compare a 24 character ObjectId against a 12 character id, so it would never match. That would make the fallback fire on every delivery and emit the raw webhook resource as `data`, which is the failure the comment is trying to prevent.

The predicate as written already matches on the two fields the `chat.status` webhook carries, and both exist on the history records, so it resolves correctly today:

```js
const match = history.find(({ promptId, messageId }) =>
  (resource.promptId && promptId === resource.promptId)
  || (resource.messageId && messageId === resource.messageId));
```

Happy to be corrected if you are seeing it miss in practice, in which case I would want to know what the delivery body looked like, since that would point at something on our side rather than in the component.

vatsal2210:
@vetrivigneshwaran all seven of your comments are answered on the threads. Five are applied and live on #21583; two I have pushed back on with evidence rather than applied, because as written they would break at runtime. Short version, detail is on each thread:

- The three ID descriptions use your wording exactly.
- The timestamp point is right and is fixed, but `resource.data.updatedAt` would throw. `processResource` passes the hydrated payload as generateMeta's **second** argument, and `resource` is the raw webhook body with no `data` key. The `data` you see in `test-event.mjs` is the emitted event, which is the merged `{ ...resource, data }`. The sources now take the second argument, via one shared `getEventTs()` helper, and it is applied to all seven sources rather than the three flagged.
- On `**Media IDs**` you were right and CodeRabbit was wrong. The guidelines say to refer to other props in bold, so I reverted my earlier change.
- On matching prompt history by `id`: history records already carry `promptId` and `messageId`. `id` is the message document's Mongo `_id`, a 24 character ObjectId, while `promptId` is a 12 character id, so comparing them would never match and would make the fallback fire on every delivery. Left as is, happy to revisit if you are seeing it miss in practice.

I then went through the component against the [contribution guidelines](https://pipedream.com/docs/components/guidelines/) to get ahead of the next round rather than trickle these out. Three more, all in #21583:

**A regression in the prompt history lookup.** New AI Chat Response scans `GET /prompt`, which defaults to 25 records per page. This was raised on the earlier PR and fixed by widening it, and the rewrite dropped it. Restored as a named constant with the reason attached. Worth knowing: the endpoint validates a `promptId` query param but the controller ignores it, so a targeted lookup is not available and a wider page really is the fix. I will raise that separately on our side.

**Unguarded response reads.** Four actions interpolated `response.data.mediaId` directly into `$summary`, and Get Transcription returned `response.data.insight.transcript`, which throws for a media file that has not finished analyzing. This is the same shape CodeRabbit flagged for `promptId` on Run AI Chat earlier, so I applied that same fix to all five.

**JSDoc coverage.** Nine app methods were undocumented. All are now, including a note on `_makeRequest` explaining why `$` defaults to `this`, since that detail has already produced one incorrect review suggestion.

One I deliberately did not change, since it is a behaviour change rather than a description fix: `userId` on Find Media is free text, and the guidelines prefer async options over typed IDs. `GET /v1/media/team` could back it as a dropdown. Happy to add it here or leave it for a follow-up, whichever you prefer.

Everything is in #21583, which targets `feat/speak_ai`. cc @ashwins01 @vatsal2210

vatsal2210:
Thank you @ashwins01! Looking forward for @vetrivigneshwaran review and getting merged to master soon! Our clients are awaiting the Pipedream workflow integration with @speakai!.

vetrivigneshwaran (approved):
LGTM, Moving to Ready for QA

PR #21583, opened 2026-08-06
fix(speak_ai): document the Folder ID format, example and source

vatsal2210:
Targets `feat/speak_ai`, following #21582. Addresses the CodeRabbit comment on `run-ai-chat.mjs` about the `folderId` description.

## What the comment asked for

> Restore the non-obvious `folderId` guidance. This inline description overrides the shared `folderId` definition. It does not state the ID format, include an example, or explain where to obtain the value. Preserve that guidance from the app definition. Refer to the sibling input as `mediaIds` in code formatting.

The formatting point is right and is fixed here. The first part had a slightly wrong premise, and the fix is bigger than it suggested.

## The shared definition never had that guidance either

There was nothing to preserve. The app file read:

```js
folderId: {
  type: "string",
  label: "Folder ID",
  description: "The ID of the folder to upload or retrieve files from",
```

No format, no example, no source. So the gap was in the app file, and four actions inherit it: **Analyze Text**, **Get Transcription**, **Run AI Chat** and **Upload Media**. Fixing only the Run AI Chat override would have left the other three thin. Both are updated here, following the pattern `mediaId` already uses.

## The example is a real format

Folder IDs come from the same generator as media IDs, a sha256 digest truncated to 12 characters:

```js
const hashKey = crypto.createHash('sha256').update(randomString).digest('hex').substr(0, 12);
```

`mediaId`, `folderId`, `promptId` and `messageId` all default to it, so `905c208f1c07` is the correct shape.

Worth recording, because it is easy to get wrong: **these IDs are not all one format.** The four above are 12-character ids, while `userId`, `webhookId` and `companyId` are 24-character Mongo ObjectIds. I checked the two examples already in this branch and both are correct as they stand: `f8eb3c22bec3` for `mediaId` and `6a708504253783a639e51914` for `userId` on Find Media. Nothing to change there.

I also corrected our own API reference off the back of this. The AI Chat response example had been showing `promptId` as a 24-character ObjectId, which is the wrong family. It now shows a 12-character id and states the format: https://docs.speakai.co/api/ai-chat/#post-prompt

## Notes

- No version bumps; both files are unreleased on this branch.
- `node --check` passes on both.

### DIFF
--- a/components/speak_ai/actions/analyze-text/analyze-text.mjs
+++ b/components/speak_ai/actions/analyze-text/analyze-text.mjs
@@ -41,7 +41,10 @@ export default {
       mediaId,
     });
 
-    $.export("$summary", `Successfully analyzed text with ID \`${response.data.mediaId}\`.`);
+    const analyzedId = response?.data?.mediaId;
+    $.export("$summary", analyzedId
+      ? `Successfully analyzed text with ID \`${analyzedId}\`.`
+      : "Successfully analyzed text.");
     return response;
   },
 };
--- a/components/speak_ai/actions/find-media/find-media.mjs
+++ b/components/speak_ai/actions/find-media/find-media.mjs
@@ -18,7 +18,7 @@ export default {
         app,
         "mediaId",
       ],
-      description: "The media file to look up",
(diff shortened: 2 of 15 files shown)


pipedream-component-development:

pipedream-component-development:

PR #21584, opened 2026-08-06
feat(amplitude): AI-optimized amplitude components for MCP

### LINKED ISSUE #21465: amplitude: AI-optimize MCP action set
Amplitude today can only send event data in. An agent can write data but can't read any of the analytics back out. We'd like this improved and new tools added so an agent can actually answer product questions, and so each tool is easy for an agent to reach for and get right.

- Pull a chart or query results — e.g. "active users in the last 30 days"
- Look up a user and their event history
- Get funnel, retention, or segmentation numbers
- Create or manage cohorts

**API reference:** https://amplitude.com/docs/apis

**Add marker:**
```
// x-pd-ai: optimized
```

michelle0927:
Resolves #21465 

## What & why

AI-optimized Amplitude action set for the MCP tool surface: 8 new actions covering
Dashboard REST (event segmentation, retention, funnels, user search/activity) and
Behavioral Cohorts (list/get/create), plus a fix to the pre-existing send-event-data
action's overly-strict required props. Hardened against a 12-eval suite run directly
against a live connected Amplitude account.

## Changes (per component)

- **amplitude.app.mjs** — added one method per endpoint (matches this repo's other
  AI-optimized apps' pattern). Also fixed `analyticsRequest()` sending
  `password: this.$auth.secret_key` instead of `this.$auth.api_secret` — every
  Basic-Auth call (7 of 9 actions) was silently sending an undefined password,
  403'ing every time despite passing all static checks. Confirmed via eval run:
  fixing this alone took the suite from 17% to 75%.
- **get-retention-analysis** (new → `0.0.1`) — `retentionMode` defaults to `nday`
  (Amplitude's own docs say `n-day`; that value 400s on the live API — confirmed
  directly). Added guidance to prefer `interval: 7`/`30` over long date ranges.
- **get-event-segmentation** (new → `0.0.1`) — fixed a worked example that used the
  literal string `"Any Active Event"` (Amplitude UI label text, not a valid
  `event_type`) — replaced with the confirmed-working built-in `_active`/`_new`.
- **send-event-data** (`0.0.2` → `0.1.0`, minor) — `userId`/`deviceId` were both
  required; Amplitude only requires one or the other. Relaxed to optional with a
  runtime guard, rewrote the description, improved `$summary`.
- **get-cohort, list-cohorts, create-cohort, get-funnel-analysis,
  get-user-activity, search-users** (new → `0.0.1` each) — worked examples,
  flat MCP-compatible schemas, response shaping (`fields` param on
  list-cohorts/get-user-activity), auto-pagination for get-user-activity.

## Verification

- Suite: `evals/amplitude` — 12 evals, 12/12 on Sonnet 5 (`claude_code_default`
  profile, default 1 trial), all 9 tools exercised.
- Report: `evals/amplitude/results/2026-08-06T17-23-07-561Z.report.md`
- Every fix above was confirmed against the live Amplitude API directly, not
  inferred from docs — several contradict Amplitude's own published docs.

## Summary by CodeRabbit

* **New Features**
  * Added Amplitude actions for cohort creation, listing, downloading, status tracking, and searching.
  * Added event segmentation, funnel analysis, retention analysis, and user activity retrieval.
  * Added configurable date ranges, pagination, filtering, segmentation, grouping, metrics, and result limits.
  * Added regional analytics support and cohort export processing with member limits and truncation reporting.
* **Bug Fixes**
  * Improved event submission validation by requiring a user or device identifier.
  * Added validation for analysis formulas and retention bracket settings.
* **Improvements**
  * Enhanced response summaries and filtering for clearer results.
  * Updated the Amplitude component version to 0.2.0.


michelle0927 (commented):
Per-change rationale inline — each comment says what was found and how it was confirmed. Suite + report in pd-connect-eval-monster/evals/amplitude (12/12 on Sonnet 5).

components/amplitude/amplitude.app.mjs
```diff
@@ -24,6 +36,18 @@ export default {
...
       return axios($, config);
     },
+    analyticsRequest({
+      $, path, ...opts
+    }) {
+      return axios($, {
+        url: `${this.analyticsBaseUrl()}${path}`,
+        auth: {
+          username: this.$auth.api_key,
+          password: this.$auth.api_secret,
```
michelle0927, line 81: **Why:** `analyticsRequest()` sent `password: this.$auth.secret_key`, but Amplitude's actual credential field (confirmed against the connected account's real credentials) is `api_secret`. `secret_key` doesn't exist on `$auth`, so every Basic-Auth call sent an undefined password. **Change:** fixed the field name. **Result:** this alone took the eval suite from 17% to 75% pass rate — every one of the 7 actions using this method was 403'ing ("Invalid API/Secret Key combination") despite passing every static check, because a static audit can't catch a wrong credential field name.

components/amplitude/amplitude.app.mjs
```diff
@@ -36,5 +60,112 @@ export default {
...
+    },
+    /**
+     * Amplitude's User Activity endpoint has no cursor, only `offset`/`limit`
+     * (per-request cap 1000), and may return slightly more events than
+     * requested to avoid splitting a session. This walks `offset` forward in
+     * `LIMIT_MAX`-sized pages until `params.limit` events are collected or a
+     * page returns fewer than requested (the stream is exhausted), so a
+     * caller asking for more than 1000 events no longer silently gets page 1.
+     */
+    async getUserActivity({
```
michelle0927, line 142: **Why:** Amplitude's User Activity endpoint has no cursor, only `offset`/`limit` (per-request cap 1000) — a caller asking for more than 1000 events silently got page 1 with no signal more existed. **Change:** `getUserActivity` now pages `offset` forward internally up to a documented ceiling (`USER_ACTIVITY_MAX_RESULTS` in constants.mjs) and returns a `truncated` flag when it stops short of the stream's actual end. **Result:** the action's own interface (`{$, params}`) is unchanged; callers just stop getting silently truncated data.

components/amplitude/actions/get-retention-analysis/get-retention-analysis.mjs
```diff
@@ -0,0 +1,115 @@
...
+    },
+    endDate: {
+      type: "string",
+      label: "End Date",
+      description: "End date, inclusive, in `YYYYMMDD` format (the `end` param). Example: `20240805`.",
+    },
+    retentionMode: {
+      type: "string",
+      label: "Retention Mode",
+      description: "Retention calculation mode (the `rm` param). One of `bracket`, `rolling`, `nday`. Defaults to `nday`. (Amplitude's own docs say `n-day` with a hyphen — that value is rejected by the live API with a 400; the working value is `nday`, no hyphen, confirmed directly against the API.) `rolling` returns an unbounded, much larger response — prefer `nday` unless you specifically need rolling retention, and if using `rolling` over a long date range, raise `interval` (e.g. to `7` or `30`) to keep the response size manageable.",
```
michelle0927, line 44: **Why:** `retentionMode` defaulted to `n-day` — exactly as documented on amplitude.com/docs/apis/analytics/dashboard-rest — but the live API rejects that value with 400 "Invalid retention method." Confirmed directly against the API: the working value is `nday` (no hyphen). This is also why an earlier example on this same prop used the literal string "Any Active Event" (Amplitude UI label text, not a valid `event_type`) — replaced with the confirmed-working `_new`/`_active`. **Result:** an eval targeting this action went from two straight 400s (the model correctly gave up on the documented default and fell back to `rolling`, which then produced a response large enough to exceed the MCP output ceiling) to a clean pass. Also added guidance to prefer `interval: 7`/`30` over date ranges wider than ~2 weeks, since daily granularity alone can still produce an oversized response even with the mode fixed.

components/amplitude/actions/get-event-segmentation/get-event-segmentation.mjs
```diff
@@ -0,0 +1,104 @@
...
+    readOnlyHint: true,
+    destructiveHint: false,
+    openWorldHint: true,
+  },
+  props: {
+    app,
+    event: {
+      type: "string",
+      label: "Event",
+      description: "A single JSON-encoded event definition (the `e` param). `event_type` is required; `filters` and `group_by` are optional. Use a real event name from your project (e.g. `Purchase`, `Sign Up`), or Amplitude's built-in `_active`/`_new` events to query overall activity. Example: `{\"event_type\":\"_active\"}` (Amplitude's built-in \"any active event\" — do NOT use the literal string \"Any Active Event\", which is Amplitude UI label text, not a valid `event_type` value, and returns a 400).",
```
michelle0927, line 25: **Why:** the worked example used `{"event_type":"Any Active Event"}` — that's Amplitude UI label text for a virtual/aggregate event, not a value the REST API accepts (confirmed: 400 "Invalid chart definition" every time, tested 3x). The model sometimes copied this literally from the example into real calls. **Change:** replaced with the confirmed-working built-in `_active` event and an explicit warning against the literal UI-label string. **Result:** removes a real, reproducible 400 that a static schema/annotation check can't catch — it's a bad example VALUE, not a bad schema.

components/amplitude/actions/send-event-data/send-event-data.mjs
```diff
@@ -245,6 +249,9 @@ export default {
     },
   },
   async run({ $ }) {
+    if (!this.userId && !this.deviceId) {
```
michelle0927, line 252: **Why:** `userId` and `deviceId` were both required, but Amplitude's HTTP V2 API only requires one or the other — an agent couldn't send a device-only anonymous event (a normal use case: pre-login/anonymous tracking) even though the API supports it. **Change:** relaxed both to `optional: true` and added this runtime guard so omitting both still fails fast with a clear message instead of a confusing API-level rejection. **Result:** strictly more permissive (any existing caller providing both still works identically) — minor bump, not major.

components/amplitude/actions/list-cohorts/list-cohorts.mjs
```diff
@@ -0,0 +1,53 @@
...
+  },
+  props: {
+    app,
+    includeSyncInfo: {
+      type: "boolean",
+      label: "Include Sync Info",
+      description: "When `true`, include cohort sync metadata in the response (the `includeSyncInfo` param).",
+      optional: true,
+    },
+    fields: {
```
michelle0927, line 25: **Why:** the raw `cohorts` array can include large fields per record (`definition` — the cohort's full filter logic, `owners`/`viewers` email arrays, `syncMetadata`) that burn context on every call for a list action. **Change:** added an additive `fields` param defaulting to a curated subset (`id, name, size, lastMod, appId`); omitting it reproduces the prior full-default-field output, the large fields are opt-in. **Result:** matches the same pattern this repo's other AI-optimized apps use (e.g. writer's list-applications).

components/amplitude/actions/create-cohort/create-cohort.mjs
```diff
@@ -0,0 +1,89 @@
...
+    openWorldHint: true,
+  },
+  props: {
+    app,
+    cohortName: {
+      type: "string",
+      label: "Name",
+      description: "Name of the cohort (the `name` field). Example: `Power Users Q3`.",
+    },
+    appId: {
```
michelle0927, line 23: **Why:** `appId` (the numeric Amplitude project ID) isn't retrievable through any Amplitude REST API this app exposes — there's no `list-projects`-style endpoint. **Change:** the description says so explicitly and points to where a human finds it (Amplitude Settings > Projects) instead of implying a lookup tool exists. **Note for reviewers:** this is a genuine Amplitude API gap, not an oversight — confirmed there's no cohort-creation path that omits it, and no endpoint that returns it.

components/amplitude/actions/get-retention-analysis/get-retention-analysis.mjs
```diff
@@ -0,0 +1,115 @@
...
+      options: RETENTION_MODES,
+      optional: true,
+    },
+    brackets: {
+      type: "string",
+      label: "Brackets",
+      description: "Bracket ranges as a JSON-encoded array, required only when Retention Mode is `bracket` (the `rb` param). Example: `[[0,5]]`.",
+      optional: true,
```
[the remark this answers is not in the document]
michelle0927: This has been resolved.

components/amplitude/actions/get-event-segmentation/get-event-segmentation.mjs
```diff
@@ -0,0 +1,104 @@
...
+      label: "End Date",
+      description: "End date, inclusive, in `YYYYMMDD` format (the `end` param). Example: `20240805`.",
+    },
+    metric: {
+      type: "string",
+      label: "Metric",
+      description: "Metric to compute (the `m` param). One of: `uniques`, `totals`, `pct_dau`, `average`, `histogram`, `sums`, `value_avg`, `formula`. Defaults to `uniques`.",
+      options: SEGMENTATION_METRICS,
+      optional: true,
+    },
```
ashwins01, line 45: can the `formula` query param be supported here as well ? `formula` is required when `metric` is set to `formula`.

components/amplitude/actions/get-retention-analysis/get-retention-analysis.mjs
```diff
@@ -0,0 +1,140 @@
...
+    }
+    const response = await this.app.getRetentionAnalysis({
+      $,
+      params: {
+        se: this.startEvent,
+        re: this.returnEvent,
+        start: this.startDate,
+        end: this.endDate,
+        rm: this.retentionMode,
+        rb: this.brackets,
```
ashwins01, line 126: There's a bunch of validations happening for `this.brackets` when `this.retentionMode === "brackets"`, but `this.brackets` is still passed here even when the `retentionMode` is something else.
Should we add `this.retentionMode === "bracket"` check here as well before passing `rb` ?

components/amplitude/amplitude.app.mjs
ashwins01: There's a bunch of inline props which can be moved as shared-props - `startDate`, `endDate`, `interval`, `segmentDefinitions`, `groupBy` & `limit`.

components/amplitude/amplitude.app.mjs
```diff
@@ -36,5 +60,112 @@ export default {
...
+        path: "/api/3/cohorts",
+        params,
+      });
+    },
+    getCohort({
+      $, cohortId,
+    }) {
+      return this._analyticsRequest({
+        $,
+        path: `/api/3/cohorts/${cohortId}`,
```
ashwins01: Is this endpoint correct ? The linked docs seem to indicate `/api/5/cohorts/request/:id`

components/amplitude/actions/list-cohorts/list-cohorts.mjs
ashwins01, line 25: I believe combining `includeSyncInfo` & `fields` wouldn't work to display the sync metadata, since the `COHORT_DEFAULT_FIELDS` does not include it ?

ashwins01 (changes requested):
Looks good overall, added a couple of comments, nothing major.

components/amplitude/amplitude.app.mjs
ashwins01: @michelle0927, would you mind sharing the resolution for this ? I wasn't able to find any changes wrt this.

components/amplitude/actions/get-user-activity/get-user-activity.mjs
```diff
@@ -0,0 +1,79 @@
...
+      optional: true,
+    },
+    direction: {
+      type: "string",
+      label: "Direction",
+      description: "Chronological direction (the `direction` param). One of `earliest`, `latest`. Defaults to `latest`.",
+      options: USER_ACTIVITY_DIRECTIONS,
+      optional: true,
+    },
+    limit: {
```
ashwins01, line 44: Can the shared-prop `limit` from `amplitude.app.mjs` be used here ?

components/amplitude/amplitude.app.mjs
michelle0927: Sorry, I meant to comment on this and forgot to. The `/api/3/cohorts/:id` endpoint works, but is not documented. The documented endpoint(s) requires using a 3 step process to get a cohort, so I figured it was simpler to leave it as a single endpoint.
However, upon further reflection, I decided it's better to keep in line with the most recent documentation. So I've split `get-cohort` into 3 separate actions: `request-cohort-download`, `get-cohort-download-status`, and `download-cohort-file`.

michelle0927:
## MCP Eval Results — amplitude

**12/12 passed** (100%) | 1.2 avg tool calls | 62.0s

| # | Eval | Status | Tools | Notes |
|---|------|--------|-------|-------|
| 1 | How many people used the product yesterday | PASS | 1/5 | — |
| 2 | What cohorts exist in Amplitude | PASS | 1/5 | — |
| 3 | Log a product event | PASS | 1/5 | — |
| 4 | [get-event-segmentation] Unique active users yesterday | PASS | 1/1 | — |
| 5 | [search-users] Find the canary test user | PASS | 1/1 | — |
| 6 | [get-user-activity] Canary user's recent events | PASS | 1/1 | — |
| 7 | [list-cohorts] List all cohorts | PASS | 1/1 | — |
| 8 | [download-cohort-file] Download a specific cohort's members | PASS | 3/3 | — |
| 9 | [get-retention-analysis] New-to-active retention | PASS | 1/1 | — |
| 10 | [get-funnel-analysis] Signup to active funnel | PASS | 1/1 | — |
| 11 | [create-cohort] Create a static cohort from a known user | PASS | 1/1 | — |
| 12 | [send-event-data] Send a purchase event | PASS | 1/1 | — |

> Model: `claude-haiku-4-5-20251001` | Registry: `private` | All 11 tools exercised (100% coverage)

components/amplitude/common/utils.mjs
```diff
@@ -0,0 +1,153 @@
...
+    cleanup();
+    const combined = new PassThrough();
+    combined.write(peeked);
+    // If the source already ended, its "end" event has already fired, so
+    // piping it now would never forward that (already-past) end to
+    // `combined`, hanging any downstream reader — end it directly instead.
+    if (ended) {
+      combined.end();
+    } else {
+      stream.pipe(combined);
```
ashwins01, line 46:
```suggestion
    } else {
      stream.once("error", err => combined.destroy(err))
      stream.pipe(combined);
```

components/amplitude/actions/download-cohort-file/download-cohort-file.mjs
```diff
@@ -0,0 +1,55 @@
...
+      propDefinition: [
+        app,
+        "requestId",
+      ],
+      description: "The `request_id` returned by **Request Cohort Download**, after **Get Cohort Download Status** reports `JOB COMPLETED`. Example: `req_456`.",
+    },
+    maxMembers: {
+      type: "integer",
+      label: "Max Members",
+      description: `Maximum number of member records to return (the full file is still downloaded and decompressed in memory, so this only caps the returned/reported result — useful to avoid an oversized response for very large cohorts). Defaults to ${COHORT_DOWNLOAD_MAX_MEMBERS}. The true total is always reported in \`memberCount\`, with \`truncated: true\` if it exceeds this cap.`,
```
ashwins01:
```suggestion
      description: `Maximum number of member records to return. Defaults to ${COHORT_DOWNLOAD_MAX_MEMBERS}. The true total is always reported in \`memberCount\`, with \`truncated: true\` if it exceeds this cap.`,
```

components/amplitude/common/constants.mjs
```diff
@@ -0,0 +1,110 @@
...
+// is. If still in progress when the budget runs out, the caller is expected
+// to call the status action again.
+export const COHORT_DOWNLOAD_POLL_INTERVAL_MS = 3000;
+export const COHORT_DOWNLOAD_POLL_BUDGET_MS = 35000;
+
+// Default cap on member records returned by Download Cohort File. Very large
+// cohorts are fully buffered and decompressed in memory before parsing (no
+// streaming), so an unbounded result risks excessive memory use and an
+// oversized step export; this keeps the default response bounded while still
+// reporting the true total via `memberCount` and a `truncated` flag.
```
ashwins01:
```suggestion
// Default cap on member records returned by Download Cohort File.
```

components/amplitude/actions/list-cohorts/list-cohorts.mjs
```diff
@@ -0,0 +1,59 @@
+// x-pd-ai: optimized
+import app from "../../amplitude.app.mjs";
+import { COHORT_DEFAULT_FIELDS } from "../../common/constants.mjs";
+import { pluck } from "../../common/utils.mjs";
+
+export default {
+  key: "amplitude-list-cohorts",
+  name: "List Cohorts",
+  description: `List all behavioral cohorts in the Amplitude project. Returns a \`cohorts\` array; each entry defaults to ${COHORT_DEFAULT_FIELDS.join(", ")}. Pass \`fields\` to get more, e.g. \`description\`, \`published\`, \`archived\`, or the large \`definition\` (the cohort's filter logic) / \`owners\`/\`viewers\` (email arrays). Use this to discover valid cohort IDs for **Get Cohort**. Example: call with no parameters -> returns \`{cohorts: [{id: "abc123", name: "Power Users Q3", size: 4200, lastMod: 1722873600000, appId: 849238}, ...]}\`. [See the documentation](https://amplitude.com/docs/apis/analytics/behavioral-cohorts#get-all-cohorts).`,
```
ashwins01: Can we modify this description, since `Get Cohort` is no longer there ?

ashwins01 (dismissed):
LGTM, just a couple of stale descriptions to be modified. Ready for QA.

ashwins01 (approved):

PR #21602, opened 2026-08-10
feat(salesforce_rest_api): add New Chatter Feed Item and Feed Comment sources

### LINKED ISSUE #21588: [TRIGGER]
**Name of app / service**
Salesforce (salesforce_rest_api)

**Link to developer documentation**

FeedItem: https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_feeditem.htm
FeedComment: https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_feedcomment.htm

**Is lack of support preventing you from moving forward, or do you have a workaround?**

Blocking. The existing Salesforce "New/Updated Record" sources poll on Case.LastModifiedDate. Chatter activity on a case does not update Case.LastModifiedDate, so new Chatter posts and comments on a Case are never emitted.

For our customer this means internal notes and @mentions on a case — the channel their agents use to interact with our AI agent — never reach us. The only workaround today is custom Apex maintained in the customer's own Salesforce org, which we would like to move away from.

**Are there specific actions or triggers you'd like to see for this app?**

A trigger emitting new FeedItem records, ideally with:
- a filter on ParentId object type (e.g. Case only)
- Type, Body, ParentId, CreatedById and CreatedDate in the payload
- an option to exclude items created by the connected/authenticating user, to avoid self-trigger loops

A companion trigger on FeedComment is also needed — comments are a separate object and are currently equally invisible.

For Neople

michelle0927:
## Closes #21588

Adds two **instant-or-polling** Salesforce sources that emit new Chatter activity, which the existing New/Updated Record sources can't see: Chatter posts/comments don't update the parent record's `LastModifiedDate`, so those sources never fire for them.

### New sources
- **New Chatter Feed Item (Instant or Polling)** — `salesforce_rest_api-new-feed-item` — emits new `FeedItem` (posts)
- **New Chatter Feed Comment (Instant or Polling)** — `salesforce_rest_api-new-feed-comment` — emits new `FeedComment` (replies)

Both poll their object on `CreatedDate` (with a webhook/instant path that falls back to timer polling automatically) and support:
- **Parent Object Type** filter — appends `AND Parent.Type = '<value>'` (e.g. `Case`) to the SOQL `WHERE`, traversing the polymorphic `ParentId`.
- **Exclude Self** — appends `AND CreatedById != '<authenticatedUserId>'` to drop events authored by the connected user, avoiding self-trigger loops. The user id is resolved via the userinfo endpoint and cached in `db`.

Payloads include `Type`, `Body`/`CommentBody`, `ParentId`, `FeedItemId` (comments), `CreatedById`, and `CreatedDate` — matching the fields requested in the issue.

### Shared code
- New `sources/common/common-feed.mjs` holds the logic shared by both sources (extends `common-new-record.mjs`): props, historical `deploy()` via direct SOQL, `processTimerEvent` with the extra-conditions query, and a `getMinFields()` hook each source supplies.
- `common-new-record.mjs`: `generateTimerMeta` now resolves the object type via `this.getObjectType()` instead of the `objectType` prop, so it works for sources that fix the type via a method (behavior-identical for the existing `new-record-instant` / `new-case-instant`).

### Testing
Verified against a Developer Edition org via the REST API that every call each source makes succeeds: `FeedItem`/`FeedComment` describe cleanly, the `deploy()` SOQL query returns records, and `GET /sobjects/{FeedItem,FeedComment}/{id}` retrieves them. Created a test post + reply and confirmed both are queryable and retrievable by id.

## Summary by CodeRabbit

- **New Features**
  - Added Salesforce Chatter FeedItem and FeedComment event sources.
  - Supports real-time delivery with polling fallback for reliable event capture.
  - Added filtering by parent object and an option to exclude posts or comments authored by the connected account.
  - Added replay of recent records during deployment and chronological event delivery.

- **Improvements**
  - Updated Salesforce actions and event sources to their latest versions.
  - Improved selectable-field handling when Salesforce metadata is unavailable.
  - Improved event retrieval with pagination and more reliable record filtering.


vetrivigneshwaran (approved):
LGTM, moving to ready for QA

PR #21610, opened 2026-08-11
Fix location name formatting in Google My Business components

### LINKED ISSUE #21517: [BUG] Google Business Profile batch reviews double-prefixes canonical resource names
### Component

`google_my_business-get-reviews-multiple-locations` version `0.0.4`

### Problem

The action constructs each request body value as:

```ts
`accounts/${account}/locations/${locationName}`
```

When callers pass canonical Google resource names, such as `accounts/123` and `locations/456`, the upstream request becomes:

```text
accounts/accounts/123/locations/locations/456
```

Google rejects this with `INVALID_ARGUMENT` because the path does not match a known template. Other Google Business Profile actions and API responses naturally expose canonical resource names, so this is easy to trigger when composing actions.

### Expected behavior

Normalize both inputs with the app helper before constructing `locationNames`, for example:

```ts
const accountId = this.app.getCleanName(account)
locationNames: locationNames.map((name) =>
  `accounts/${accountId}/locations/${this.app.getCleanName(name)}`
)
```

This keeps existing bare-ID inputs working while also accepting canonical resource names.

### LINKED ISSUE #21592: [BUG] Google My Business – New Review Created trigger throws TypeError: items?.slice is not a function on deploy
**Describe the bug**
the component crashes on deploy when the reviews list response isn't an array (e.g. location with zero reviews / undefined reviews field).

**To Reproduce**
1. add the trigger "Google Business Profile > New Review Created
2. connect a GBP account, select account + location, deploy + time
3. click save and continue -> error

**Expected behavior**
This trigger saves and started checking new review ever X time. 

**Screenshots**
See attached

Image

**Additional context**
This worked on an old gmail address we are using. That old email is no longer active. The new email works on other GMB steps just having issue with the trigger.

Priyadharshan-Pdm:
Normalize Google resource names once, in the app layer, so every action and source accepts both bare IDs and canonical resource names.

Closes #21517 and  #21592 

## Summary

**The bug.** Google Business Profile IDs circulate in two forms: bare (`123`, `456`) and canonical (`accounts/123`, `accounts/123/locations/456`). The components assumed the bare form and concatenated directly into request paths. Prop dropdowns already emit bare IDs, so the UI path worked — but Google's API returns the *canonical* form in every `name` field, so any value wired in from an upstream step (including this connector's own **List Accounts** / **List Locations** actions) produced `accounts/accounts/123/locations/locations/456` and failed with `INVALID_ARGUMENT`.

**The reported failure.** `get-reviews-multiple-locations` and the `new-review-created-multiple-locations` source built body entries as `` `accounts/${account}/locations/${locationName}` `` from raw props. Both now normalize first.

**The root-cause fix — `app/google_my_business.app.ts`.** Rather than patch each call site, all 7 request methods now pass IDs through the existing `getCleanName` helper at the point of interpolation: `listLocations`, `listReviews`, `listPosts`, `createPost`, `getReview`, `updateReviewReply`, `batchGetReviews`. `getCleanName` returns the last path segment, so it is idempotent — bare IDs are unaffected and every canonical shape collapses correctly.

- `batchGetReviews` additionally rebuilds `data.locationNames`. It is the only endpoint that carries location names in the request **body** rather than the URL — which is precisely why the reported bug slipped past the existing URL-level guard.
- Removed the `cleanUrl` regex from `_httpRequest` (`/\/(accounts|locations)\/\1\//g`). It repaired the finished URL string instead of the input, only collapsed *adjacent identical* segments, and could not see request bodies. It fixed `accounts/accounts/123`, but silently passed through `locations/accounts/123/locations/456` and any canonical `review` name. With normalization at the source it is dead code, and leaving it would mask this same class of bug again.
- `listAccounts` is deliberately left un-normalized — its `parentAccount` is a query parameter that Google requires in canonical `accounts/{id}` form, so normalizing it would break a currently-working action. A regression check pins this.

**Version-only bumps.** No behavior change, but required because they depend on the modified app file: `create-post`, `create-update-reply-to-review`, `get-specific-review`, `list-accounts`, `list-all-reviews`, `list-locations`, `list-posts`, plus the `new-post-created` and `new-review-created` sources.

**Testing.** All 13 changed files pass `pnpm exec eslint` (exit 0). `pnpm run build` compiles every component to `dist/*.mjs`, and `findBadKeys`, `checkComponentAppProp` and `findDuplicateKeys` pass.

Behavior was verified with a harness that loads the **built** artifacts and stubs `_httpRequest`, capturing each URL exactly as the app methods construct it — this proves the removed `cleanUrl` regex has nothing left to repair, rather than hiding behind it. Every ID was supplied in bare and in each canonical form: **63/63 combinations across all 7 endpoints** produce the correct URL and body, including the `listAccounts` guard. The originally reported action and source were checked against 4 input shapes each — **6/8 failing before the fix, 8/8 after**, with bare-ID input unchanged to confirm no regression.

**Not verified against a live Google Business Profile account** — no request was sent to Google; all checks are static and the harness stubs the HTTP layer. Removing `cleanUrl` is a genuine behavior change, so an input shape outside those 63 combinations would now fail loudly rather than being half-repaired.

## Checklist

### Versioning

### New app

### CodeRabbit review
`

## Bug Fixes

* Improved handling of account and location identifiers across Google Business Profile actions and notifications.
* Improved reliability when retrieving reviews across multiple locations, including when no reviews are returned.
* Improved consistency for account, location, review, post, and reply requests.
* Review-count reporting now safely handles missing review data and updates tracking only when reviews are emitted.

## Updates

* Updated Google Business Profile actions, event sources, and integration components to the latest versions.

### DIFF
--- a/components/google_my_business/actions/create-post/create-post.ts
+++ b/components/google_my_business/actions/create-post/create-post.ts
@@ -12,7 +12,7 @@ export default defineAction({
   key: "google_my_business-create-post",
   name: "Create Post",
   description: `Create a new local post associated with a location. [See the documentation](${DOCS_LINK})`,
-  version: "0.0.6",
+  version: "0.0.7",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/google_my_business/actions/create-update-reply-to-review/create-update-reply-to-review.ts
+++ b/components/google_my_business/actions/create-update-reply-to-review/create-update-reply-to-review.ts
@@ -8,7 +8,7 @@ export default defineAction({
   key: "google_my_business-create-update-reply-to-review",
   name: "Create or Update Reply to Review",
   description: `Create or update a reply to the specified review. [See the documentation](${DOCS_LINK})`,
-  version: "0.0.5",
+  version: "0.0.6",
   annotations: {
     destructiveHint: true,
     openWorldHint: true,
--- a/components/google_my_business/actions/get-reviews-multiple-locations/get-reviews-multiple-locations.ts
+++ b/components/google_my_business/actions/get-reviews-multiple-locations/get-reviews-multiple-locations.ts
@@ -8,7 +8,7 @@ export default defineAction({
   key: "google_my_business-get-reviews-multiple-locations",
   name: "Get Reviews from Multiple Locations",
   description: `Get reviews from multiple locations at once. [See the documentation](${DOCS_LINK})`,
-  version: "0.0.4",
+  version: "0.0.5",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
@@ -69,11 +69,14 @@ export default defineAction({
       account, locationNames, pageSize, orderBy, ignoreRatingOnlyReviews,
     } = this;
 
+    const accountId = this.app.getCleanName(account);
+
     const params: BatchGetReviewsParams = {
       $,
-      account,
+      account: accountId,
       data: {
-        locationNames: locationNames?.map((locationName: string) => `accounts/${account}/locations/${locationName}`),
+        locationNames: locationNames?.map((locationName: string) =>
+          `accounts/${accountId}/locations/${this.app.getCleanName(locationName)}`),
         pageSize,
         orderBy,
         ignoreRatingOnlyReviews,
--- a/components/google_my_business/actions/get-specific-review/get-specific-review.ts
+++ b/components/google_my_business/actions/get-specific-review/get-specific-review.ts
@@ -8,7 +8,7 @@ export default defineAction({
   key: "google_my_business-get-specific-review",
   name: "Get a Specific Review",
   description: `Return a specific review by name. [See the documentation](${DOCS_LINK})`,
-  version: "0.0.4",
+  version: "0.0.5",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/google_my_business/actions/list-accounts/list-accounts.ts
+++ b/components/google_my_business/actions/list-accounts/list-accounts.ts
@@ -9,7 +9,7 @@ export default defineAction({
   key: "google_my_business-list-accounts",
   name: "List Accounts",
   description: `Lists all accounts accessible to the authenticated user. [See the documentation](${DOCS_LINK})`,
-  version: "0.0.1",
+  version: "0.0.2",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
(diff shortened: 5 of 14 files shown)


Priyadharshan-Pdm:
## Root cause analysis

**Symptom.** `Get Reviews from Multiple Locations` (`v0.0.4`) failed with `INVALID_ARGUMENT` whenever **Account Name** or **Location Names** received a canonical resource name (`accounts/123`, `locations/456`) instead of a bare ID. Easy to hit when composing steps — Google returns the canonical form in every `name` field, including from our own **List Accounts** / **List Locations** actions.

**Cause.** The action concatenated raw props into a resource name:

```ts
`accounts/${account}/locations/${locationName}`   // -> accounts/accounts/123/locations/locations/456
```

The app's `getCleanName` helper (`accounts/123` → `123`, idempotent on bare IDs) was applied to prop dropdown *values*, but never to incoming prop values at request time. Dropdown selection therefore worked, which is why this went unnoticed.

**Why the existing `cleanUrl` guard missed it.** It worked correctly — and the request still failed:

| Stage | Value |
|---|---|
| Action builds body | `["accounts/accounts/123/locations/locations/456"]` ❌ |
| `batchGetReviews` builds URL | `…/accounts/accounts/123/locations:batchGetReviews` ❌ |
| `_httpRequest` runs `cleanUrl` | `…/accounts/123/locations:batchGetReviews` ✅ repaired |
| Body sent to Google | `["accounts/accounts/123/locations/locations/456"]` ❌ untouched |

`cleanUrl` is a string op on `url`; the body lives in `data`, spread through `...args` to axios verbatim. `batchGetReviews` is the **only** endpoint carrying resource names in the body rather than the path — so no version of that regex could have caught this. It was also incomplete for URLs: the `\1` backreference only matched adjacent identical segments, so `locations/accounts/123/locations/456` passed straight through.

**Fix.** Normalize at the point an ID enters a request, in the app layer:
- All 7 request methods pass IDs through `getCleanName` before interpolation.
- `batchGetReviews` also rebuilds `data.locationNames` — the body case the URL guard couldn't reach.
- `cleanUrl` removed: now dead code, and keeping it would mask this class of bug again.
- `listAccounts` deliberately excluded — its `parentAccount` is a query param Google requires in canonical form. Pinned by a regression check.
- Same defect existed verbatim in the `new-review-created-multiple-locations` source; fixed alongside.

**Verification.** A harness runs the built `dist/*.mjs` with `_httpRequest` stubbed, capturing each URL as the app methods construct it (proving the removed regex has nothing left to repair). **63/63** combinations across all 7 endpoints pass, with every ID in bare and canonical form. The reported action + source: **6/8 failing before, 8/8 after**, bare-ID input unchanged. eslint, `pnpm build`, `findBadKeys`, `checkComponentAppProp`, `findDuplicateKeys` all pass. Not exercised against a live Google account — checks are static.
`

Priyadharshan-Pdm:
QA done using Connect

michelle0927 (approved):
LGTM!

PR #21617, opened 2026-08-11
[21606] Google docs action changes

### LINKED ISSUE #21606: google_docs: add delete-table and full-table write actions
Google Docs

**Please provide a link to the relevant API docs.**
https://developers.google.com/docs/api/reference/rest/v1/documents/request

## Problem

1. There is no way to delete a table. When an agent clears a table's contents, the empty grid stays in the document as blank cells. No tool removes the table itself, so the user has to delete it by hand in Google Docs.
2. There is no way to write a full table of data in one step. Insert Table creates an empty grid only. Filling it means writing one cell at a time, which is slow and error-prone.

## What's needed

Two new actions on the Google Docs app (OAuth). Both run through the `documents.batchUpdate` endpoint (`POST /v1/documents/{documentId}:batchUpdate`). No new scopes are required.

### Tables

| Action | Endpoint | Notes |
|---|---|---|
| Delete Table | `POST /v1/documents/{documentId}:batchUpdate`<br>(`DeleteContentRange`) | Removes an entire table, not just its text. The API has no `DeleteTable` request and no table ID, so the tool needs a deterministic way to pick which table (for example, order in the document). Deletes over the table's content range. |
| Write Table | `POST /v1/documents/{documentId}:batchUpdate`<br>(`InsertText`, batched) | Accepts the full table as structured data (header row plus body rows) and writes every cell in a single `batchUpdate`. This is the primary ask: it removes the box-by-box filling that is slow and error-prone. Cell indices shift as content is inserted, so writes must be ordered to stay valid. Could extend Insert Table to accept data, or ship as a separate action. |

### Tool Descriptions

Update tool descriptions to support agent drive table creation and updates.

**Write Table description:** 
"Create a table and fill it with data in a single step. Provide the entire table (all rows and columns, including a header row) at once. Use this instead of inserting cells or text one at a time. The action places every value in the correct cell for you."

**Insert Table:** add one sentence to the existing description, do not rewrite it. 
Current text: "_Insert an empty table with the given number of rows and columns into a Google Doc. Use **Find Document** to resolve a document's name to its ID. [See the documentation](https://developers.google.com/docs/api/reference/rest/v1/documents/request#InsertTableRequest)_". 
* After the first sentence, add: "If you already have the data, use **Write Table** instead so you don't have to fill cells one by one."
* The Find Document hint and docs link stay as-is.

**Delete Table description:**
"Remove an entire table, structure and contents, from the document. Use this to clear a table completely, including an empty table left behind after its text was removed."

**Prop descriptions (Write Table):** define the data shape, for example a 2D array of cell values, one inner array per row (`[["Name","Role"],["Ada","Engineer"]]`), header row first. Optional `hasHeaderRow` boolean to style the first row.

## Notes
- Identifying the target table is the main open design question. The Docs API exposes no table ID or selector, so the tool needs a reliable way to choose which table to act on.
- Sheets-linked tables have a hard API limit: the Docs API does not expose or preserve a table's link to a Google Sheet. A table written or rebuilt through the API is static and unlinked, and linked tables cannot be refreshed via the API. Deleting a linked table works, since it is treated as normal content.
- Keep these tools' props static (no async dropdowns) so they stay usable by the agent.

vetrivigneshwaran:
Closes: https://github.com/PipedreamHQ/pipedream/issues/21606

Added 2 new actions Delete Table and  Write table, Also updated the description of Insert table.

components/google_docs/actions/delete-table/delete-table.mjs — removes an entire table via deleteContentRange over the table's structural range. Since the Docs API has no table ID, tables are identified by Table Number (1-indexed, document order, defaults to 1 for the common "clean up the one table I just emptied" case). destructiveHint: true (irreversible removal).

components/google_docs/actions/write-table/write-table.mjs — inserts a table sized to the data and fills every cell in one step. Takes rows as a JSON string (2D array, following the exact convention already established by google_sheets-add-rows), optional hasHeaderRow to bold row 1, and the existing shared position prop. destructiveHint: false.

**Supporting changes:**

- google_docs.app.mjs: refactored _buildRequestForPosition to extract _resolvePositionIndex (no behavior change — verified all 4 existing consumers still get identical output), and added flattenTables, deleteTable, and writeTable app methods. writeTable re-fetches the doc after insertion (the API's insertTable reply carries no location info), locates the new table by matching its startIndex to the requested position (or takes the last table for "end"), then fills cells in reverse document order in a single batchUpdate — verified this ordering logic in isolation with a simulated document structure (fill order came out strictly descending, and short/jagged rows correctly leave trailing cells blank).

- common/utils.mjs: added a pure flattenTables() helper (only top-level tables — nested tables in a cell aren't counted, and this limitation is documented in both actions' prop descriptions).

- insert-table.mjs: added the exact cross-reference sentence from the issue pointing to Write Table, kept the Find 
Document hint and docs link as-is; bumped 1.0.1 → 1.0.2.

- package.json: 1.1.0 → 1.2.0 (minor, new actions).

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **New Features**
  * Added actions to create and populate Google Docs tables, with optional header-row formatting.
  * Added the ability to delete a selected top-level table by table number.
  * Added support for placing tables at the beginning, end, or a specified document position.
  * Added validation for table data, positions, and selected tables.
  * Added summaries of table updates after creation or deletion.
* **Documentation**
  * Updated table insertion guidance to recommend Write Table when data is already available.

### DIFF
--- a/components/google_docs/actions/append-text/append-text.mjs
+++ b/components/google_docs/actions/append-text/append-text.mjs
@@ -4,7 +4,7 @@ export default {
   key: "google_docs-append-text",
   name: "Append Text",
   description: "Append text to an existing document. [See the documentation](https://developers.google.com/docs/api/reference/rest/v1/documents/request#InsertTextRequest)",
-  version: "0.1.12",
+  version: "0.1.13",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/google_docs/actions/create-document-from-template/create-document-from-template.mjs
+++ b/components/google_docs/actions/create-document-from-template/create-document-from-template.mjs
@@ -5,7 +5,7 @@ export default {
   key: "google_docs-create-document-from-template",
   name: "Create Document From Template",
   description: "Create a new Google Doc by copying a template document and substituting `{{placeholder}}` tokens with values. Use **Find Document** to resolve the template's name to its ID. Returns `{documentId, title, url}`. [See the documentation](https://developers.google.com/docs/api/reference/rest/v1/documents/request#ReplaceAllTextRequest)",
-  version: "1.0.1",
+  version: "1.0.2",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/google_docs/actions/create-document/create-document.mjs
+++ b/components/google_docs/actions/create-document/create-document.mjs
@@ -4,7 +4,7 @@ export default {
   key: "google_docs-create-document",
   name: "Create Document",
   description: "Create a new Google Doc with an optional body. The body is rendered as Markdown, so you can include headings (`# Title`), bold/italic, bullet/numbered lists, links, and code. Optionally place the doc in a specific Drive folder. Returns `{documentId, title, url}`. [See the documentation](https://developers.google.com/docs/api/reference/rest/v1/documents/create)",
-  version: "1.0.1",
+  version: "1.0.2",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/google_docs/actions/delete-table/delete-table.mjs
+++ b/components/google_docs/actions/delete-table/delete-table.mjs
@@ -1,0 +1,54 @@
+import { ConfigurationError } from "@pipedream/platform";
+import googleDocs from "../../google_docs.app.mjs";
+
+export default {
+  key: "google_docs-delete-table",
+  name: "Delete Table",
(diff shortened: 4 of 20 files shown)


michelle0927 (approved):
LGTM! Ready for QA.

mariano-pd (approved):
QA testing successful! ✅

PR #21502, opened 2026-07-28
[21458] feat (Brex) : add card, expense, and user management actions

Priyadharshan-Pdm:
## Summary

Closes #21458  and will do a followup PR for shiping these actions to brex_staging 

Moved to blocked actions into separate [PR](https://github.com/PipedreamHQ/pipedream/pull/21624)

## Checklist
Please check the following items before your PR can be reviewed:

### Versioning
- [ ] All components updated in this PR had their version updated (`0.0.1` for new ones)
- [ ] The app updated in this PR had its `package.json`'s version updated

### New app
If this is a new app, please submit [an app integration request](https://github.com/PipedreamHQ/pipedream/issues/new?template=app---service-integration.md) - the PR will only be reviewed after the app is integrated.
- [ ] The app updated in this PR is already integrated

### CodeRabbit review
After the PR is opened, and if new changes are pushed, CodeRabbit will automatically review it. Do not 'mark as resolved' CodeRabbit's comments, but reply to them instead, whether you agree (and update the PR accordingly) or disagree.
- [ ] I have addressed or acknowledged all of CodeRabbit's review comments

## Summary by CodeRabbit

- **New Features**
  - Added actions to retrieve individual users and their spending limits.
  - Added actions to list users and cash accounts.
  - Added card transaction search with merchant, date, amount, and cardholder filters.
- **Enhancements**
  - Added email filtering, expanded limit details, configurable result caps, and pagination.
  - Added clear result counts, formatted amounts, and truncation summaries.
  - Standardized spend-duration and card configuration options across Brex actions.

PR #21624, opened 2026-08-12
Brex Actions for Cards, Transactions and Expenses

### LINKED ISSUE #21458: brex: AI-optimize MCP action set
Brex today can create a card, invite a user, list transactions for a couple of account types, and set a user's limit. That's a narrow slice of what teams do with Brex. We'd like these improved and new tools added so an agent can handle more of the spend and card work, and so each tool is easy for an agent to reach for and get right.

- Look up a specific transaction and its receipt
- Search transactions across accounts by merchant, amount, or date
- Get a card's details, balance, or limit
- Freeze or cancel a card
- List the cards or people in the account

**API reference:** https://developer.brex.com/

**Add marker:**
```
// x-pd-ai: optimized
```

Priyadharshan-Pdm:
## Summary
Closes #21458  , This PR consists of only the Actions that are blocked due to test cards and scopes

Previous [Merged PR ](https://github.com/PipedreamHQ/pipedream/pull/21502)

Completes the Brex integration with the card and expense surface, so an agent can issue and control cards and reconcile spend end to end — look up a card, freeze or cancel it, adjust its limit, then search expenses across card, bill-pay, and reimbursement spend and pull the receipts behind them. This is the deliberate follow-up to the earlier Brex PR, which shipped only the actions that had a live test run behind them; these are the ones that could not be verified against the sandbox account because they need API scopes and a test card it does not have (see TEST).

## WHAT

**New card actions (7, all keyed off a card ID from `list-cards`):**
- `list-cards`, `list-card-accounts` — discovery. `list-cards` filters by cardholder or status, but Brex has no server-side status filter, so `ACTIVE`/`SHIPPED`/`LOCKED`/`TERMINATED` is applied after fetching. `list-card-accounts` returns account-level balance, limit, and statement period — not per-card limits
- `get-card` — one card with status, last four, and vendor-card spend limit. Corporate cards return `spend_controls: null` because they draw on the cardholder's limit, so the description routes those to **Get User Limit**
- `freeze-card` / `unfreeze-card` — reversible lock pair
- `cancel-card` — permanent termination; description steers to **Freeze Card** for anything temporary, since this cannot be undone
- `update-card-limit` — vendor cards only (`limit_type: CARD`); corporate cards route to **Set Limit for User**

**New expense actions (2):**
- `search-expenses` — spans card, bill-pay, and reimbursement spend with receipt and approval state, filterable by merchant, amount, date, person, type, or status
- `get-expense` — one expense plus receipts. Receipt download links expire 15 minutes after the response, so the description says to download rather than store them

**New discovery actions (2):**
- `list-departments`, `list-locations` — turn a department or location name into the ID that **Invite User** takes. They exist because the dropdowns that used to supply those IDs were removed (see below)

**Existing actions (10 updated):** descriptions rewritten to behavioral form with cross-references to the tools that supply their IDs. `search-card-transactions` now expands `expense_id`, which is the bridge from a settled card posting to the expense holding its receipt.

**App + shared:**
- `common-app.mjs` — 11 new methods (`getCard`, `listCardsPaginated`, `lockCard`, `unlockCard`, `terminateCard`, `updateCard`, `getExpense`, `listExpensesPaginated`, `listCardAccountsPaginated`, `listLocationsPaginated`, `listDepartmentsPaginated`)
- `common/options.mjs` — new shared option lists: card type/status, limit type, card action reason, expense type/status/payment-status/expand, spend duration
- `common/utils.mjs` — **bug fix:** `formatMoney` divided every amount by 100, which is wrong for zero-decimal currencies. It now resolves minor-unit digits per ISO 4217 via `Intl`, so `700` renders as `7.00 USD` but stays `700 JPY`, falling back to 2 digits for codes the runtime rejects
- **bug fix:** the user-limit path now wraps `userId` in `encodeURIComponent`
- `maxResults` is capped at 2000 (`MAX_LIMIT_PER_PAGE` 100 × `MAX_PAGES` 20) rather than an arbitrary number — past that the paginator stops, so a larger value could never be satisfied
- `new-transfer-event` source — description now names what each event type means (`TRANSFER_PROCESSED` vs `TRANSFER_FAILED`) instead of "please specify the events you want to watch"
- Removed the `async options()` dropdowns from the shared `user`, `location`, and `department` propDefinitions, per the MCP cleanup bar. Each now takes a plain ID, with the example shape in its description and the discovery tool that supplies it named there — **List Users**, **List Locations**, **List Departments**. The latter two are added in this PR so those IDs stay reachable to an agent without the dropdown; `getUsers`, `getLocations`, and `getDepartments` are gone
- Migrated the remaining 14 brex actions from the legacy `// x-pd-ai: optimized` marker comment to the top-level `ai: "optimized"` field, matching the convention master's backfill (#21876) established — brex held 14 of the repo's last 15 stragglers

**Versioning:** package bumped `0.2.1` → `0.3.0`. The 11 new actions ship at `0.0.1`; `0.0.2` → `0.0.3` for `get-user`, `get-user-limit`, `list-cash-accounts`, `list-users`, `search-card-transactions`; `0.1.2` → `0.1.3` for `create-card`, `invite-user`, `set-limit-for-user`, and both `list-transactions-*`; `new-transfer-event` source `0.1.1` → `0.1.2`.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **New Features**
  - Added Brex card management actions to retrieve, list, freeze, unfreeze, cancel, and update card spending limits.
  - Added card account listing, expense retrieval, and advanced expense search with filtering and pagination.
  - Added standardized options for card statuses, action reasons, expense types, statuses, and expandable details.
  - Increased supported result limits to 2,000 records.
- **Bug Fixes**
  - Improved currency formatting for currencies with different decimal conventions.
  - Clarified card transaction search results and receipt details.


components/brex/actions/create-card/common.mjs
```diff
@@ -1,18 +1,20 @@
...
       type: "string",
       label: "Card Name",
-      description: "Card Name",
+      description: "A label for the card, shown in the Brex dashboard and printed on physical cards, e.g. `AWS Vendor Card`.",
     },
     cardType: {
       type: "string",
       label: "Card Type",
-      description: "Card Type",
+      description: "Must be `VIRTUAL`, for a card usable immediately. `PHYSICAL` is not supported yet — Brex requires a mailing address to ship a card and this action does not collect one, so selecting it raises a configuration error.",
```
vetrivigneshwaran: Restrict cardType to VIRTUAL. options.cardType exposes PHYSICAL, but run() rejects that value with ConfigurationError before the Brex  request. Expose only VIRTUAL in the static options and keep the runtime  guard as a safety net.

components/brex/actions/invite-user/common.mjs
```diff
@@ -5,17 +5,17 @@ export default {
...
       type: "string",
       label: "First Name",
-      description: "User first name",
+      description: "The user's first name, e.g. `Jane`.",
     },
     lastName: {
       type: "string",
       label: "Last Name",
-      description: "User last name",
+      description: "The user's last name, e.g. `Doe`.",
```
vetrivigneshwaran, line 13: Remove UI-centric wording.

  Select both assumes a graphical UI. State the input operation directly,
  such as “Set both event types to emit an event for either outcome.”

  As per path instructions: prop descriptions must “Avoid UI-centric
  language (“select from the dropdown”, “choose from the list”).”

components/brex/common-app.mjs
```diff
@@ -12,19 +12,19 @@ export default {
...
       label: "Posted At Start",
-      description: "Shows only transactions with a posted_at_date on or after this date-time. This parameter is the date-time notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). Example: `2022-12-12T23:59:59.999Z`",
+      description: "Return only transactions posted on or after this date-time, in [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) notation, e.g. `2022-12-12T23:59:59.999Z`. Brex can post transactions with a lag, so when polling repeatedly, overlap this with your previous window by a day or more and de-duplicate results by transaction `id` rather than advancing it to the exact end of the last poll.",
       optional: true,
     },
     max: {
       type: "integer",
       label: "Max",
-      description: "The maximum amount of registered that will be fetched. Defaults to `500`.",
+      description: "The maximum number of records to fetch. Defaults to `500`.",
```
vetrivigneshwaran:
```suggestion
      description: "The maximum number of records to fetch. Defaults to `100`.",
```

  Correct the advertised pagination default.

  max is optional. When a caller omits it, _paginateItems() uses
  DEFAULT_MAX_RESULTS, which is 100. The description states 500.

components/brex/common-app.mjs
```diff
@@ -12,19 +12,19 @@ export default {
     postedAtStart: {
       type: "string",
       label: "Posted At Start",
-      description: "Shows only transactions with a posted_at_date on or after this date-time. This parameter is the date-time notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6). Example: `2022-12-12T23:59:59.999Z`",
+      description: "Return only transactions posted on or after this date-time, in [RFC 3339](https://tools.ietf.org/html/rfc3339#section-5.6) notation, e.g. `2022-12-12T23:59:59.999Z`. Brex can post transactions with a lag, so when polling repeatedly, overlap this with your previous window by a day or more and de-duplicate results by transaction `id` rather than advancing it to the exact end of the last poll.",
```
vetrivigneshwaran, line 15: The shared `user`, `location`, and `department` propDefinitions still use `async options()` dropdowns. These props are consumed by AI-optimized (`ai: "optimized"`) actions - list-cards (Cardholder), get-user, get-user-limit, set-limit-for-user, create-card, and invite-user (manager/department/location). The MCP cleanup bar requires removing ALL `async options()` from optimized components (not just reloadProps/additionalProps) and relying on plain string ID inputs backed by discovery tools. `List Users` already exists to discover user IDs, so the `user` dropdown should become a plain string; location/department have no discovery action (see separate finding).

components/brex/actions/create-card/common.mjs
```diff
@@ -32,7 +34,7 @@ export default {
       amount: {
         type: "integer",
         label: "Spend Limit Amount",
-        description: "The amount of money, in the smallest denomination of the currency indicated by currency. For example, when currency is USD, amount is in cents.",
+        description: "The spend limit, in the currency's smallest denomination — `2500` is $25.00 in USD.",
```
vetrivigneshwaran: Create Card uses `limitType` with `reloadProps: true` plus `additionalProps()` to reveal the spend-limit fields only when limitType=CARD. For an `ai: "optimized"` component this is risky: reloadProps/additionalProps may not work in some agent (MCP) contexts. The same result is achievable with fixed optional props (amount, currency, spendDuration, reason, lockAfterDate) plus a pre-flight `ConfigurationError` in run() when limitType=CARD and amount is missing - the file already throws a ConfigurationError for the PHYSICAL case, so the validation pattern is established.

components/brex/actions/invite-user/invite-user.mjs
```diff
@@ -4,15 +4,16 @@ import common from "./common.mjs";
 export default {
   ...common,
   name: "Invite User",
-  description: "Invites a new user as an employee. [See the docs here](https://developer.brex.com/openapi/team_api/#operation/createUser).",
+  description: "Invites a person to the Brex account as an employee, emailing them to finish onboarding. Returns the new Brex user ID, which **Create Card**, **Set Limit for User**, and **Get User** take. [See the documentation](https://developer.brex.com/openapi/team_api/users/createuser)",
```
vetrivigneshwaran, line 7: `department` and `location` inputs are async-options dropdowns (via common-app propDefinitions) with no corresponding discovery action in the app (no `List Departments`/`List Locations`). Per the cleanup bar, when removing async-options the discovery tool must exist so an agent can resolve these IDs; here they are missing, leaving the agent with no way to obtain department/location IDs once the dropdowns are removed.

components/brex/actions/update-card-limit/update-card-limit.mjs
```diff
@@ -0,0 +1,82 @@
+import brexApp from "../../brex.app.mjs";
+import { formatMoney } from "../../common/utils.mjs";
+
+export default {
+  key: "brex-update-card-limit",
+  name: "Update Card Limit",
+  description: "Updates the spend limit on a vendor card (`limit_type: CARD`). Corporate cards draw on their cardholder's limit instead — use **Set Limit for User** for those. [See the documentation](https://developer.brex.com/openapi/team_api/cards/updatecard)",
```
vetrivigneshwaran: The `spendDuration` description says 'Omit to leave the card's current duration unchanged,' but the action sends a full `spend_controls` object to PUT /v2/cards/{id}. If Brex treats `spend_controls` as a whole-object replacement, omitting `spend_duration` (or `reason`/`lock_after_date`) could reset those fields rather than preserve them - making the description's promise inaccurate. Author should confirm Brex's PUT merge semantics; if it is a replace, the description and the omit-to-preserve assumption need revisiting.

vetrivigneshwaran (changes requested):

vetrivigneshwaran (approved):
LGTM

PR #21676, opened 2026-08-17
Add Givebutter contact and transaction actions

### LINKED ISSUE #21668: givebutter: AI-optimize MCP action set
Givebutter has no actions built in Pipedream today — there's nothing an agent can do with this integration yet. We'd like to build an initial action set so an agent can manage fundraising campaigns and donors, and so each tool is easy for an agent to configure.

- List campaigns
- List contacts/donors
- Look up a transaction
- Create or update a contact

**API reference:** https://docs.givebutter.com/

**Add marker:**
```
// x-pd-ai: optimized
```

Priyadharshan-Pdm:
## Summary

Closes #21668 

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **New Features**
  - Added actions to create and update contacts with contact details, company information, notes, and tags.
  - Added campaign and contact listing with filtering, sorting, and pagination options.
  - Added an action to retrieve individual transactions by ID.
  - Improved API connectivity and authentication across supported operations.
  - Added validation to prevent incomplete or invalid contact updates.
- **Chores**
  - Updated the Givebutter integration version and platform support.


components/givebutter/actions/common/constants.mjs
```diff
@@ -0,0 +1,20 @@
...
+]);
+export const CONTACT_SORT_BY = Object.freeze([
+  "name_or_company",
+  "primary_email",
+  "point_of_contact",
+  "created_at",
+  "total_contributions",
+  "recurring_contributions",
+  "last_donation_amount",
+]);
```
GTFalcao: Seems kinda unnecessary to freeze when they're already being exported as constants (the difference is very small and impractical) but not really an issue, just mentioning it

components/givebutter/actions/common/constants.mjs
```diff
@@ -0,0 +1,20 @@
+export const BASE_URL = "https://api.givebutter.com/v1";
+export const AUTH_HEADER = "Authorization";
+export const CAMPAIGNS_PATH = "/campaigns";
+export const CONTACTS_PATH = "/contacts";
+export const TRANSACTIONS_PATH = "/transactions";
+export const DEFAULT_PER_PAGE = 20;
+export const MAX_PER_PAGE = 100;
```
GTFalcao: I find it a lot more confusing to have these values in a separate constants file, rather than them being in the file where they're used, especially since they're only used in a single file each.

The arrays are ok to split to a separate file if they're large (which is not really the case here), but I would severely advise against the base URL (can be a constant in the app file which is the only place it will be used), auth header (really doesn't need to be declared separately), resource paths (unless the app is colossal and has a huge amount of standardized resource types). The per-page default and max are ok if they are documented as shared across the API.


```suggestion
export const DEFAULT_PER_PAGE = 20;
export const MAX_PER_PAGE = 100;
```

GTFalcao (commented):
Left a couple comments about the structure of the constants.

Also, for the ones that do remain, I'd leave them as common/constants rather than actions/common/constants (since pagination constants for example do not necessarily apply exclusively to actions - and this is a more common pattern)

components/givebutter/actions/common/constants.mjs
Priyadharshan-Pdm: Agreed , Object.freeze is not necessary

GTFalcao (approved):

PR #21684, opened 2026-08-17
feat(slack_v2): read direct messages by user ID + deprecate List Messages

michelle0927:
## What & why

AI-optimized slack_v2 read tools now support **direct messages**, hardened against the
eval suite in `pd-connect-eval-monster/evals/slack_v2`. Reading a DM by user ID was the
one gap left: writing to a DM already worked (chat.postMessage auto-opens the IM), but
the read tools forwarded a `U…` user id straight to conversations.history, which only
accepts a conversation id and answered `channel_not_found`.

## Changes (per component)

- **slack_v2.app.mjs** — added `openConversation()` (conversations.open); `resolveChannelId`
  now opens the DM for a `U…`/`W…` user id (the read-side counterpart to posting) so history,
  thread-replies, and reactions accept a user id; id regexes made case-sensitive so a lowercase
  channel *name* isn't misread as an id. Shared by 10 actions.
- **get-channel-history** (`0.1.0` → `0.2.0`, minor) — description + `channel` prop document
  reading a DM by user id.
- **list-messages** (`0.0.5` → `0.0.6`, patch) — deprecated in favor of Get Channel History
  (it was winning routing on channel-read evals, precision 0%); renamed + steered, `run()`
  unchanged so existing workflows still work.
- **get-thread-replies, browse-files, set-channel-topic, get-channel-details,
  invite-user-to-channel, delete-message, add-reaction, edit-message, list-members-in-channel**
  (patch) — version bumps for the shared `resolveChannelId` behavioral change.

## Verification

- Suite: `evals/slack_v2` — latest run 3/3 (pass^2, Sonnet 5).
- Reconciliation in `requirements.md`: eval #36 FAIL→PASS; #3/#25 routing fixed; #22 was an
  eval bug (system `channel_join` message), criterion now accepts 33 or 34.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving channel history from direct messages, including self-DMs, using user IDs.
  * Improved conversation resolution for direct-message interactions.

* **Documentation**
  * Marked the legacy “List Messages” action as deprecated and directed users to “Get Channel History” instead.

* **Chores**
  * Updated Slack action, event source, and package versions across the integration.
  * Applied general optimization updates to selected Slack capabilities.

### DIFF
--- a/components/slack_v2/actions/get-channel-history/get-channel-history.mjs
+++ b/components/slack_v2/actions/get-channel-history/get-channel-history.mjs
@@ -1,3 +1,4 @@
+// x-pd-ai: optimized
 import utils from "../../common/utils.mjs";
 import slack from "../../slack_v2.app.mjs";
 
--- a/components/slack_v2/actions/get-channel-history/get-channel-history.mjs
+++ b/components/slack_v2/actions/get-channel-history/get-channel-history.mjs
@@ -5,16 +5,17 @@ export default {
   key: "slack_v2-get-channel-history",
   name: "Get Channel History",
   description:
-    "Read the recent message history from a specific channel."
+    "Read the recent message history from a specific channel or direct message (DM)."
     + " Accepts a channel ID or channel name (resolved automatically)."
-    + " Use this when you want to see a channel's latest messages — unlike **Search** which finds messages by keyword."
+    + " To read a DM, pass the other person's **user ID** (e.g. `U1234567890`) as the channel — pass your OWN user ID to read your conversation with yourself."
+    + " Use this when you want to see a channel's or DM's latest messages — unlike **Search** which finds messages by keyword."
     + " Returns messages with text, timestamps (ts), reactions, and user IDs."
     + " Message timestamps can be used with **Get Thread Replies**, **Edit Message**, and **Add Reaction**."
     + " **Pass `fields`** (e.g. `text,ts,user`) unless you need full message objects — raw Slack"
     + " messages carry blocks, attachments and edit metadata, so a busy channel can run to tens"
     + " of thousands of characters and be truncated before you see any of it."
     + " [See the documentation](https://api.slack.com/methods/conversations.history)",
-  version: "0.1.0",
+  version: "0.2.0",
   type: "action",
   annotations: {
     destructiveHint: false,
@@ -26,7 +27,7 @@ export default {
     channel: {
       type: "string",
       label: "Channel",
-      description: "Channel ID (e.g. `C1234567890`) or channel name (e.g. `general` or `#general`). Resolved automatically.",
+      description: "Channel ID (e.g. `C1234567890`) or channel name (e.g. `general` or `#general`). For a direct message, pass a user ID (e.g. `U1234567890`) — including your own, to read your self-DM. Resolved automatically.",
     },
     limit: {
       type: "integer",
--- a/components/slack_v2/actions/list-messages/list-messages.mjs
+++ b/components/slack_v2/actions/list-messages/list-messages.mjs
@@ -2,10 +2,14 @@ import slack from "../../slack_v2.app.mjs";
 
 export default {
   key: "slack_v2-list-messages",
-  name: "List Messages",
+  name: "List Messages (Deprecated)",
   description:
-    "Retrieve messages from a Slack conversation, including reactions. [See the documentation](https://api.slack.com/methods/conversations.history)",
-  version: "0.0.5",
+    "DEPRECATED — do NOT use for reading a channel's or DM's messages. Use **Get Channel History** instead:"
+    + " it resolves channel names, channel IDs, and direct messages (by user ID), supports `fields`"
+    + " selection to keep responses small, and returns the same message data."
+    + " This legacy tool remains only for existing workflows: it accepts a raw conversation ID and"
+    + " returns full, untrimmed message objects. [See the documentation](https://api.slack.com/methods/conversations.history)",
+  version: "0.0.6",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
--- a/components/slack_v2/slack_v2.app.mjs
+++ b/components/slack_v2/slack_v2.app.mjs
@@ -1110,6 +1110,12 @@ export default {
         ...args,
       });
     },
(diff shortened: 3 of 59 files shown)


components/slack_v2/slack_v2.app.mjs
```diff
@@ -1278,7 +1284,22 @@ export default {
...
+      // channel names are lowercased by Slack, so a `/i` match would misclassify an all-alnum
+      // name like `welcomeaboard` or `developers` (no hyphen, 9+ chars) as an id.
+      // Channel, private-group, and DM ids are already conversation ids — pass through.
+      if (/^[CDG][A-Z0-9]{8,}$/.test(input)) return input;
+      // A user id is NOT a conversation id — conversations.* answers `channel_not_found`
+      // for it. Open (idempotently) the DM with that user and use the returned IM channel.
+      // This is the read-side counterpart to chat.postMessage accepting a user id: an agent
+      // asked to read "my DMs" resolves its own user id and passes it here, and history,
+      // thread replies, and reactions now accept it the way posting already does.
+      if (/^[UW][A-Z0-9]{8,}$/.test(input)) {
```
michelle0927, line 1298: **Why:** eval #36 ("read my DMs with myself") failed — the read tools forwarded a `U…` user id straight to `conversations.history`, which only accepts a conversation id and answered `channel_not_found`. Writing to a DM already worked (`chat.postMessage` auto-opens the IM); reading had no equivalent. **Change:** a `U…`/`W…` user id now opens the DM idempotently (via the new `openConversation()` → `conversations.open`, line 1113) and resolves to the IM channel id, so history, thread-replies, and reactions accept a user id the way posting already did. The id regexes were also made **case-sensitive** (Slack ids are uppercase-only) so an all-alphanumeric lowercase channel *name* isn't misclassified as an id. **Result:** #36 FAIL→PASS (pass^2). Shared by 10 actions — backwards-compatible (previously-erroring inputs now work).

components/slack_v2/actions/get-channel-history/get-channel-history.mjs
```diff
@@ -26,7 +27,7 @@ export default {
     channel: {
       type: "string",
       label: "Channel",
-      description: "Channel ID (e.g. `C1234567890`) or channel name (e.g. `general` or `#general`). Resolved automatically.",
+      description: "Channel ID (e.g. `C1234567890`) or channel name (e.g. `general` or `#general`). For a direct message, pass a user ID (e.g. `U1234567890`) — including your own, to read your self-DM. Resolved automatically.",
```
michelle0927, line 31: **Why:** with DM reading now supported by `resolveChannelId`, the tool + prop needed to tell the agent it can pass a user id. **Change:** description and `channel` prop now document reading a DM by user id (including your own, for a self-DM). **Result:** the agent routes DM-read requests here and passes the user id. Minor bump (`0.1.0`→`0.2.0`): backwards-compatible capability + description.

components/slack_v2/actions/list-messages/list-messages.mjs
```diff
@@ -2,10 +2,14 @@ import slack from "../../slack_v2.app.mjs";
 
 export default {
   key: "slack_v2-list-messages",
-  name: "List Messages",
+  name: "List Messages (Deprecated)",
```
michelle0927, line 5: **Why:** this legacy twin of **Get Channel History** was *winning* tool-routing on the channel-read evals — #3/#25 warned `expected_tools_missing` and its precision was 0% — because it answers the same question but can't resolve names/DMs, has no `fields` param (oversized output), and its remote-options `conversation` prop breaks under MCP. Description tuning alone can't arbitrate two tools answering one question. **Change:** renamed to "List Messages (Deprecated)" and the description now leads with "DEPRECATED — use Get Channel History instead" (name + first sentence are the tool-search retrieval key); `run()` is unchanged so existing workflows calling it with a conversation id still work. **Result:** #3/#25/#36 → pass^2 3/3; the agent stops selecting this tool. Patch bump (copy only, no behavior change).

michelle0927 (commented):
Per-change rationale inline — each comment says what motivated the change. Suite + reconciliation in pd-connect-eval-monster/evals/slack_v2 (requirements.md).

components/slack_v2/slack_v2.app.mjs
```diff
@@ -1278,7 +1284,22 @@ export default {
...
+        const { channel } = await this.openConversation({
+          users: input,
+        });
+        return channel.id;
```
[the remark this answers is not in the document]
michelle0927: I've verified the finding against the current code. I'm skipping it — the issue isn't valid for this component. Here's the evidence:

1. The im:write scope is already a hard requirement of the app — the read path adds nothing new.
send-message-to-user-or-group.mjs:87-88 (a core, shipped action) already calls conversations.open({ users }). Pipedream grants a fixed OAuth scope bundle to every slack_v2 connection — users don't pick a subset — so any account that can use the messaging actions already has im:write. The reconciliation's "scope is present via precedent" claim checks out. There is no "read-only-scoped account" scenario to protect here; resolveChannelId calling conversations.open introduces zero new scope.

2. There is no creation to avoid for the read case. conversations.open is idempotent — for a DM that already has history (the only case a read action hits), it returns the existing IM channel without creating anything. The finding's "preserve creation behavior only if intentional" — it's intentional and, for reads, a no-op.

3. The proposed non-creating lookup is strictly worse here. The only read-scope way to map a user ID → existing D-id is to paginate the user's IM list (users.conversations/conversationsList with types: im, at slack_v2.app.mjs:845/884) and match on the user field. That trades a confirmed-present scope (im:write) for an unconfirmed one (im:read) plus added calls, pagination, and latency — for no functional gain. The current path was verified working at pass^2 (eval #36).

components/slack_v2/actions/get-channel-history/get-channel-history.mjs
GTFalcao, line 2: Seems this one was already 'optimized' before so we could include the metadata comment here. In fact there's a bunch of other components following the pattern
```suggestion
// x-pd-ai: optimized
import utils from "../../common/utils.mjs";
```

GTFalcao (dismissed):

GTFalcao (approved):

PR #21702, opened 2026-08-19
[21640] fix(jira): stop dropping user picker and reporter fields

### LINKED ISSUE #21640: [BUG] Jira Create Issue drops custom user picker fields
**Describe the bug**
`jira-create-issue` (and `jira-update-issue`) omit custom fields that use a user selector from `additionalProps` even though they are present in Jira create-meta.

Jira create-meta still returns these fields, including an `autoCompleteUrl`. Because of a Jira Cloud bug, some of those autocomplete URLs 401 when requested through OAuth authentication.

- [JRACLOUD-84801](https://jira.atlassian.com/browse/JRACLOUD-84801) — create-meta `autoCompleteUrl`s not accessible via Bearer token (unresolved)
- [JRACLOUD-84826](https://jira.atlassian.com/browse/JRACLOUD-84826) — Reporter’s `/rest/api/3/user/recommend` 401s under OAuth (unresolved)

`getDynamicFields` then fetches those `autoCompleteUrl`s, **and drops the fields when they error.**

The broken part is specifically the that the legacy v1 autocomplete route that `create-meta` provides for user pickers (`/rest/api/1.0/users/picker`) fails with OAuth. However, the documented v3 user APIs work fine with the same token.

**To Reproduce**

1. In Jira, add a **User Picker (single user)** custom field to a create screen for a project/issue type.
2. Confirm it appears in create-meta, with `schema.type: "user"` and `autoCompleteUrl` pointing at `/rest/api/1.0/users/picker?...`.
3. Add a **Jira → Create Issue** step, and see that this custom field does not appear.

**Suggested fix**
In `getDynamicFields`:

1. For `schema.type === "user"`, do not blindly GET `autoCompleteUrl`
2. If it is a v1 url mentioned above (or after trying and it throws), prefer:
   - Assignee (`key === "assignee"`) → existing assignable search
   - Other user fields → `/rest/api/3/user/search` or `/rest/api/3/user/picker`

ashwins01:
## Summary

Closes #21640 
`jira-create-issue` and `jira-update-issue` silently omit custom fields backed by a user selector, even though Jira create-meta returns them.
`getDynamicFields` requested each field's `autoCompleteUrl` to build its options, and the `catch` returned the accumulator unchanged, removing the field from the step entirely. Only `schema.type: "user"` fields took that path, which is why the damage is confined to user fields.

## Fix
1. User-typed fields (`schema.type: "user"`, and `array` of `user`) no longer request `autoCompleteUrl`. Options resolve lazily from documented v3 routes: assignee from a project-scoped `/user/assignable/search`, everything else from `/user/search`.
2. A failed lookup now yields an empty dropdown instead of removing the field. That is the actual bug being fixed.
3. Multi-user fields (`array` of `user`, e.g. Approvers, Request participants) get the same resolver. They were previously surfaced as free-form `string[]` with no options.
4. `formatFields` wraps user custom field values as `{ "accountId": "..." }`, or an array of those for multi-user fields. Previously they were sent as bare strings and Jira rejected them with `400 "expected Object"`, so the field was unusable even when visible. Assignee, Reporter, Parent and Issue Type keep their existing `{ id }` shape.
5. User props now say what the value is ("Account ID of the user") instead of the generic "Set your field value".

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **New Features**
  * Added improved Jira user and assignable-user selection for issue fields, including pagination and clearer user formatting.
  * Enhanced support for single- and multi-user custom fields when creating or updating issues.
  * Improved default pagination handling for Jira user searches.

* **Chores**
  * Updated Jira action, event source, and package versions across the integration.
  * Standardized handling of Jira field schema types and user-related values.


components/jira/common/constants.mjs
```diff
@@ -24,23 +30,25 @@ const FIELD_TYPE = {
...
-};
+const USER_FIELD_TYPES = [
+  "userpicker",
+  "multiuserpicker",
+  "people",
+  "sd-request-participants",
+];
+
+const USER_SEARCH_PATH = "/user/search";
+const ASSIGNABLE_USER_SEARCH_PATH = "/user/assignable/search";
```
michelle0927: Nitpick: I don't think we need constants for these. To maintain consistency with the other methods in the app file, we can define them in their respective methods.

michelle0927 (dismissed):
At some point, we'll probably want to remove `additionalProps` from `create-issue` and `update-issue`. For now, this looks good! One nitpick comment.

michelle0927 (approved):

PR #21711, opened 2026-08-19
fix(salesforce_rest_api): deploy hook no longer fails on orgs above 100k records

### LINKED ISSUE #21687: [BUG] salesforce_rest_api: deploy hook fails with OPERATION_TOO_LARGE on orgs above 100k records
.

michelle0927:
## What & why

Fixes #21687. Deploying any `salesforce_rest_api` source built on
`sources/common/common-new-record.mjs` or `sources/common/common-updated-record.mjs`
failed on Salesforce orgs with more than 100k records of the selected object type. The
deploy hook called `listSObjectTypeIds()` (`GET /sobjects/{ObjectType}`), and Salesforce
rejects the server-computed `recentItems` payload with `400 OPERATION_TOO_LARGE /
"exceeded 100000 distinct ids"` instead of truncating — failing the **entire deploy** for
what is only an optional step (seeding up to 25 historical events). `recentItems` also
means "recently viewed by this user", not "recently created", so it never matched the
hook's intent.

## Changes

- **`common-new-record.mjs` / `common-updated-record.mjs`** — replace the
  `listSObjectTypeIds()` call with a bounded SOQL query
  (`SELECT Id FROM {Object} ORDER BY CreatedDate|LastModifiedDate DESC LIMIT 25`), and
  wrap the historical-seed loop in `try/catch` so a failure logs and the source is still
  created and emits going forward. This removes the 100k limit entirely and fixes the
  recently-viewed-vs-recently-created intent mismatch.
- **`common.mjs`** — memoize `getNameField()` for the lifetime of an execution so the
  deploy loop reads the name field from the db **once**, not once per emitted event.
  Measured via an instrumented re-run of the real `deploy()`: `db.get("nameField")` went
  **25 → 0** during deploy (`setNameField` primes the cache) and **≤ 1** on any other
  execution path.
- Patch version bumps on all 8 affected sources; `package.json` `1.14.0` → `1.14.2`.

## Affected sources

Via `common-new-record.mjs`: `new-record-instant`, `new-case-instant`,
`new-email-template-instant`, `new-knowledge-article-instant`.

Via `common-updated-record.mjs` (when `skipFirstRun` is false): `record-updated-instant`,
`case-updated-instant`, `email-template-updated-instant`, `knowledge-article-updated-instant`.

Feed-based sources override `deploy` and never made the call, so they're unaffected (not bumped).

## Verification

- `db.get` count confirmed by an instrumented re-run of the real `deploy()`: old **25** →
  new **0** (deploy) / **1** (cold read). The control run reproducing the original 25
  confirms the harness exercises the real code path.
- The >100k failure itself is Salesforce-server behavior; this fix removes the offending
  call rather than depending on a 100k+ org to reproduce.

## Summary by CodeRabbit

* **Bug Fixes**
  * Historical Salesforce records are now seeded according to the configured limit and sorted by creation or modification date.
  * Seeding failures no longer prevent source creation or future event delivery.
* **Performance**
  * Repeated name-field lookups are now cached during execution, reducing unnecessary database queries.
* **Maintenance**
  * Updated Salesforce integration and event source versions across supported instant triggers, including record, case, knowledge article, email template, and feed events.


components/salesforce_rest_api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/salesforce_rest_api",
-  "version": "1.14.0",
+  "version": "1.14.2",
```
ashwins01:
```suggestion
  "version": "1.14.1",
```

components/salesforce_rest_api/sources/common/common.mjs
```diff
@@ -89,9 +89,18 @@ export default {
       this.db.set("latestDateCovered", latestDateCovered);
     },
     getNameField() {
-      return this.db.get("nameField");
+      // Cache in memory for the lifetime of this execution. The deploy hook emits
+      // up to DEPLOY_HISTORICAL_LIMIT historical events, each building meta via
+      // getNameField(); without this the db is read once per event (e.g. 25×).
```
ashwins01:
```suggestion
      // getNameField(); without this the db is read once per event (e.g. 25x).
```

components/salesforce_rest_api/sources/common/common-new-record.mjs
```diff
@@ -12,17 +13,24 @@ export default {
...
-          },
-        };
-        this.processWebhookEvent(event);
+      try {
+        const { records } = await this.query({
+          query: `SELECT Id FROM ${objectType} ORDER BY ${constants.FIELD_NAME.CREATED_DATE} DESC LIMIT ${constants.DEPLOY_HISTORICAL_LIMIT}`,
+        });
+        const ids = records.map((r) => r.Id);
+        for (const id of ids.slice(-constants.DEPLOY_HISTORICAL_LIMIT)) {
+          const object = await this.salesforce.getSObject(objectType, id);
```
ashwins01: This would emit results in newest-first order, whereas the rest of our salesforce sources emit data in oldest-first order to ensure events are delivered chronologically. Can we keep the same standard here as well ?

components/salesforce_rest_api/sources/common/common-updated-record.mjs
```diff
@@ -13,20 +13,27 @@ export default {
...
-          const meta = this.generateWebhookMeta(event);
-          this.$emit(event.body, meta);
+        try {
+          const { records } = await this.query({
+            query: `SELECT Id FROM ${objectType} ORDER BY ${constants.FIELD_NAME.LAST_MODIFIED_DATE} DESC LIMIT ${constants.DEPLOY_HISTORICAL_LIMIT}`,
+          });
+          const ids = records.map((r) => r.Id);
+
+          for (const id of ids.slice(-constants.DEPLOY_HISTORICAL_LIMIT)) {
+            const object = await this.salesforce.getSObject(objectType, id);
```
ashwins01: Same here, about emitting chronologically.

ashwins01 (changes requested):
LGTM. Just one comment on events being emitted chronologically.
Should we also make the `try-catch` change in `common-feed.mjs` ?

components/salesforce_rest_api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@pipedream/salesforce_rest_api",
-  "version": "1.14.0",
+  "version": "1.14.1",
```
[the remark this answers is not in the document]
michelle0927: The package.json version is correct. This PR updates it from "1.14.0" to "1.14.1".

ashwins01 (approved):
LGTM, ready for QA.

PR #21755, opened 2026-08-24
fix(dealcloud): replace generated field props with a static Fields map for MCP v3

### LINKED ISSUE #21741: DealCloud: remove reloadProps/additionalProps from create/update record actions for MCP v3 compatibility
Both DealCloud record actions extend a shared base (`common-create-update.mjs`) that builds inputs dynamically via `additionalProps()`, driven by `reloadProps: true`. Pipedream MCP v3 does not support this pattern: an MCP tool gets a single static JSON Schema resolved once at registration, but `additionalProps()` only computes its props after a parent prop value is known, and there's no round-trip in the MCP tool contract to surface that. The field-level props these actions generate are invisible to an agent, making them broken as MCP tools.

## Affected components (`components/dealcloud`, package `0.2.0`)

| Action | Key | Version | Dynamic mechanism |
|---|---|---|---|
| Create Record | `dealcloud-create-record` | 0.0.1 | `entryTypeId` (reloadProps) → `additionalProps()` fetches entry type fields via `getEntryTypeFields()` and emits one prop per field |
| Update Record | `dealcloud-update-record` | 0.0.1 | Same shared `entryTypeId`-driven per-field prop generation |

A single fix in `common-create-update.mjs` covers both.

DeepakJaggupalli:
## Summary

Closes #21741.

Both record actions extend `common-create-update.mjs`, which drove `additionalProps()` off `reloadProps` on `entryTypeId`: it fetched the object's schema and emitted one `field_<id>` prop per field. An MCP tool resolves its JSON Schema once at registration, so none of those props ever reached an agent — **the actions could pick an object and then had no way to send any data at all.**

As the issue says, a single fix in the shared base covers both actions.

### The shape change

The generated per-field props become one static `fields` object prop, so the schema is fixed at registration:

```
{ "CompanyName": "Acme Corp", "EmployeeCount": "250", "IsActive": "true" }
```

`entryTypeId` keeps its `async options()`. That stays MCP-compatible — what breaks MCP is a prop that doesn't exist until a parent is set, not a declared prop that populates its own choices.

### What moved to run time

**Resolution.** The old code parsed a field id back out of the `field_<id>` prop name. `buildRequestData()` now fetches the schema and matches each key against the field's **API name, display name, or numeric id**, case-insensitively. An agent has only the description to go on, so all three spellings are accepted, and an unknown key raises an error listing the object's real field names rather than silently writing nothing.

**Typing.** The generated props carried a real Pipedream type per field (`integer`, `boolean`, `string[]`). A plain object arrives as strings, so values are coerced from the schema's `fieldType` instead:

| Field type | Accepts |
|---|---|
| Number / Counter | `"42"` → `42`, rejects non-numeric |
| Boolean | `true`/`false`/`yes`/`no`/`1`/`0`, or a real boolean |
| Multi-select | an array, or a comma-separated string |
| Blank value | `null`, to clear the cell |

Bad input raises a `ConfigurationError` naming the field and what it expected, rather than failing opaquely at the API.

### Two guarantees the props gave for free, now explicit

- **Required fields** were enforced by `optional: !field.isRequired` on create and `optional: true` on update. That distinction is preserved: create validates that every required, non-calculated field is present; update only touches the fields it names, so the rest keep their stored values.
- **Calculated fields** were previously writable through the generated props. They are now rejected as unwritable, and a field supplied twice under two different aliases is caught rather than sending two conflicting cell writes.

### Incidental cleanups

`run()` in both actions rest-destructured `this` to separate real props from generated ones, which is why both carried `/* eslint-disable no-unused-vars */`. With no generated props, both now just call `buildRequestData()` and the disable blocks are gone.

`convertFieldsToProps` and `mapFieldTypeToPropType` in `common/utils.mjs` existed only to feed `additionalProps()` and are removed. `FIELD_TYPES` moves from a private const to an export there, since the coercion needs it. `checkIdArray` is untouched, so `delete-records` and `get-records` — the other two importers of that file — are unaffected.

One thing to flag: with `mapFieldTypeToPropType` gone, `CURRENCY_OPTIONS` in `common/constants.mjs` no longer has a consumer. I left it in place as reference data rather than widen the diff, but happy to remove it if you'd rather not carry an unused export.

### Testing

Unit tested against a representative schema: key resolution by all three spellings, case and whitespace handling, each coercion path and its rejection, calculated-field and duplicate-key guards, required-on-create vs not-on-update, and the resulting `storeRequests` payload shape (26 assertions).

## Checklist

### Versioning

### New app

### CodeRabbit review

### DIFF
--- a/components/dealcloud/actions/common/common-create-update.mjs
+++ b/components/dealcloud/actions/common/common-create-update.mjs
@@ -1,32 +1,19 @@
-import { convertFieldsToProps } from "../../common/utils.mjs";
+import { ConfigurationError } from "@pipedream/platform";
+import { FIELD_TYPES } from "../../common/utils.mjs";
 import dealcloud from "../../dealcloud.app.mjs";
 
+const TRUTHY = [
+  "true",
+  "yes",
+  "1",
+];
+const FALSY = [
+  "false",
+  "no",
+  "0",
+];
+
 export default {
-  methods: {
-    convertFieldsToProps,
-    isUpdate() {
-      return false;
-    },
-    getEntryId() {
-      return -1;
-    },
-    getRequestData(props) {
-      return {
-        storeRequests: Object.entries(props).map(([
-          key,
-          value,
-        ]) => {
-          const fieldId = Number(key.split("_")[1]);
-          return {
-            entryId: this.getEntryId(),
-            fieldId,
-            ignoreNearDups: this.ignoreNearDups,
-            value,
-          };
-        }),
-      };
-    },
-  },
   props: {
     dealcloud,
     entryTypeId: {
@@ -34,7 +21,11 @@ export default {
         dealcloud,
         "entryTypeId",
       ],
-      reloadProps: true,
+    },
+    fields: {
+      type: "object",
+      label: "Fields",
+      description: "The field values to write, as a map of field to value. Each key may be the field's API name, its display name, or its numeric field ID — for example `{ \"Name\": \"Acme Corp\", \"City\": \"Boston\" }`. Use **Get Records** on the same object to see the available fields. Multi-select fields accept either an array or a comma-separated string, boolean fields accept `true`/`false`, and calculated fields cannot be written to.",
     },
     ignoreNearDups: {
       propDefinition: [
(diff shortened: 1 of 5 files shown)


pipedream-component-development:

pipedream-component-development:

PR #21756, opened 2026-08-24
feat: Add pd-registry check to validate proposed component versions

ashwins01:
## Summary

Add pd-registry check to validate proposed component versions, this is to ensure component-publish does not fail in `master` once a branch is merged. The current component version check runs a diff on the semver in `master` against the PR branch, which might not be the case always since there might be versions pushed to the pd-registry and the code in `master` might have a lower version.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **Chores**
  - Added automated pull request validation to ensure component actions and sources use versions newer than those already published.
  - Version conflicts are collected and reported together for easier resolution.
  - Validation safely fails when metadata is invalid, registry information is unavailable, credentials are missing, or requests continue to fail after retries.
  - Removed files and files without explicit version metadata are excluded from validation.


.github/workflows/component-registry-version-check.yaml
```diff
@@ -0,0 +1,241 @@
...
+            [[ "$f" == *.mjs || "$f" == *.js || "$f" == *.ts ]] || return 1
+            [[ "$f" == *.app.* ]] && return 1
+            [[ "$f" == */common/* || "$f" == */common.* ]] && return 1
+            [[ "$f" == *.test-event.mjs ]] && return 1
+            return 0
+          }
+
+          extract() { # field, content
+            printf '%s' "$2" | grep -oE "$1:[[:space:]]*[\"'][^\"']+[\"']" | head -1 | sed -E "s/.*[\"']([^\"']+)[\"'].*/\1/"
+          }
```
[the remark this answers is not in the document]
ashwins01: Why would an action file have multiple "key" or "version" attributes ?

.github/workflows/component-registry-version-check.yaml
```diff
@@ -0,0 +1,241 @@
...
+jobs:
+  check-registry-version-floor:
+    name: Ensure component version exceeds registry floor
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check proposed versions against the registry
+        shell: bash {0} # don't fast-fail; aggregate every offender
+        env:
+          GH_TOKEN: ${{ github.token }}
+          PD_API_KEY: ${{ secrets.PD_API_KEY }}
```
Priyadharshan-Pdm, line 58: what will happen to a forked PRs ? Won't the secret be leaked for making curl requests ?

I guess this is how it is been in other yamls

Priyadharshan-Pdm (approved):

.github/workflows/component-registry-version-check.yaml
ashwins01: The trigger of this workflow is `pull_request_target` (not `pull_request`) solely so the `PD_API_KEY` secret is available for action runs on the fork PR's, it runs the `master` branch's copy of this workflow and not the actual PR's. So it is not possible to look at what the secret is.
The rest of the workflows in the repo either use `pull_request` or are post-merge runs.

PR #21770, opened 2026-08-25
Monday.com MCP actions and async/reloadProps removal

### LINKED ISSUE #21740: monday.com: remove reloadProps/additionalProps from item/column actions for MCP v3 compatibility
Five monday.com actions build their inputs dynamically via `async additionalProps()`, driven by `reloadProps: true`. Pipedream MCP v3 does not support this pattern: an MCP tool gets a single static JSON Schema resolved once at registration, but `additionalProps()` only computes its props after a parent prop value is known, and there's no round-trip in the MCP tool contract to surface that. The generated props aren't reachable by an agent, making these tools broken as MCP tools.

## Affected components (`components/monday`, package `0.12.1`)

| Action | Key | Version | Dynamic mechanism |
|---|---|---|---|
| Create Item | `monday-create-item` | 0.1.6 | `columns` (reloadProps, shared `common-create-item.mjs`) → `additionalProps()` fetches board columns via `listColumns()` and emits one typed prop per selected column |
| Create Subitem | `monday-create-subitem` | 0.1.6 | Same shared `columns`-driven per-column prop generation |
| Create Column | `monday-create-column` | 0.1.5 | `columnType` (reloadProps) → `additionalProps()` conditionally adds a `defaults` prop for `status`/`dropdown` column types |
| Update Column Values | `monday-update-column-values` | 0.2.7 | `boardId` (reloadProps) → `additionalProps()` fetches board columns via `listColumns()` and emits one prop per column |
| Get Items By Column Value | `monday-get-items-by-column-value` | 0.1.5 | `columnId` (reloadProps) → `additionalProps()` fetches column metadata to type/derive options for a `value` prop |

Create Item/Create Subitem share one base; the other three are independent.

Priyadharshan-Pdm:
## Summary

Closes #21740 

Rewrites Monday.com's create/update item and column actions to use a single structured `columnValues` JSON prop instead of per-column props generated dynamically via `reloadProps`/`additionalProps` — the dynamic path can't be driven by an agent in MCP/tools-only mode. Adds a **List Columns** action so callers can discover valid column IDs and labels up front, and fixes column-value handling and item-search pagination bugs uncovered along the way.

**Actions (columnValues-based rewrite, replacing reloadProps/additionalProps):**
- `create-item`, `create-subitem` (via shared `common-create-item.mjs`) — column props are no longer generated dynamically per selected column; a single `columnValues` JSON object prop (column ID → value) is parsed via the new `parseColumnValues` helper. `create-subitem`'s `columnValues` description calls out that it targets the parent item's separate subitems board.
- `update-column-values` — same `columnValues` object input; now validates every supplied column ID exists on the board **before** uploading any file columns (previously a file could finish uploading ahead of a later invalid-ID error, leaving an orphaned attachment on the item); missing/empty `columnValues` raises `ConfigurationError` instead of silently sending nothing.
- `create-column` — `defaults` (custom labels for `status`/`dropdown`) is now a static prop instead of being conjured via `additionalProps`/`reloadProps`; setting it on any other column type now throws `ConfigurationError` instead of failing server-side, and an empty string is rejected rather than silently passed through.

**New action:**
- `list-columns` — returns each board column's ID, title, type, and (for `status`/`dropdown`) its labels; throws if the board has no columns. Cross-referenced from the other actions' descriptions as the way to discover valid column IDs.

**Bug fixes:**
- `get-items-by-column-value` — value entry simplified to a plain search string. Also fixes a pagination bug: any board returning more than one page of matches threw a `TypeError`, because the follow-up request destructured the wrong response shape (`items_page_by_column_values` instead of `next_items_page`) and sent a malformed `cursor` variable (a bare string instead of `{ cursor }`). Verified against monday's `next_items_page` API reference.

**App + shared:**
- `monday.app.mjs` — `itemColumnValues` prop renamed to `columnValues` with an updated description; fixed the same malformed-cursor bug in `getItemsByColumnValue`.
- `common/utils.mjs` — added `parseColumnValues` (accepts a JSON string or object, parses JSON-literal string values so text like `"123"` isn't coerced to a number, rejects non-object input); removed the now-unused `capitalizeWord`; `getColumnOptions` is now null-safe.
- `common/queries.mjs` — `listItemsNextPage` now selects `column_values` on paginated items (previously only `id`/`name`, which would have broken `formatColumnValues` on page 2+ once the pagination bug above was fixed).

**Chore — version-only bumps, no functional change:** `create-board`, `create-group`, `create-update`, `get-board-items-page`, `get-column-values`, `list-board-id-options`, `list-board-ids-options`, `list-boards`, `list-workspace-id-options`, `list-workspace-ids-options`, `update-item-name`, and all 10 sources (`column-value-updated`, `name-updated`, `new-board`, `new-item`, `new-subitem-update`, `new-subitem`, `new-user`, `specific-column-updated`, `subitem-column-value-updated`, `subitem-name-updated`).

**Versioning:** package bumped `0.12.1` → `0.13.0`. Notable component bumps: `create-column` `0.1.5` → `0.2.1`, `create-item` `0.1.6` → `0.2.0`, `create-subitem` `0.1.6` → `0.2.0`, `get-items-by-column-value` `0.1.5` → `0.2.0`, `update-column-values` `0.2.7` → `0.3.1`, `list-columns` (new) `0.0.2`. The rest of the chore-only group above took a patch bump each.

## Checklist

### Versioning

### New app

### CodeRabbit review

## New Features
- Added a **List Columns** action showing column IDs, titles, types, and available labels.
- Added structured column values for creating subitems and updating items, including file columns.
- Included column values in paginated item results.

## Improvements
- Simplified column value entry with a consistent ID-to-value format.
- Improved pagination, email value handling, and validation for invalid or empty inputs.

## Documentation
- Expanded column value guidance with examples and references.


Priyadharshan-Pdm:
@claude review

components/monday/actions/update-column-values/update-column-values.mjs
```diff
@@ -40,37 +37,23 @@ export default {
...
+        common.props.monday,
+        "columnValues",
+      ],
+      optional: false,
+      description: "The column values to set, as column ID → value pairs. Example: `{ \"status\": \"Done\", \"date4\": \"2026-09-02\", \"numbers\": 42 }`. Use **List Columns** to discover column IDs and the allowed labels for `status`/`dropdown` columns. For a `file` column, pass either a file URL or a path to a file in the `/tmp` directory (for example, `/tmp/myFile.txt`) and the file is uploaded to that column. The item's name cannot be changed here — use **Update Item Name** instead. See the [Column types reference](https://developer.monday.com/api-reference/reference/column-types-reference) for the value each column type expects",
+    },
     syncDir: {
       type: "dir",
+      label: "Sync Directory",
+      description: "A directory to mount so that a `file` column value given as a `/tmp` path can be read. Only needed when `Column Values` sets a `file` column from a local path rather than a URL.",
```
michelle0927: No need for a label & description for the `syncDir` prop.
```suggestion
```

components/monday/monday.app.mjs
```diff
@@ -132,7 +136,7 @@ export default {
     updateId: {
       type: "string",
       label: "Update ID",
-      description: "Select an update or provide an update ID",
+      description: "An existing update, as an update ID. Update IDs are returned by **Create an Update**.",
```
michelle0927: Can we add a **List Updates** action for this prop?

components/monday/monday.app.mjs
```diff
@@ -161,121 +165,99 @@ export default {
     },
   },
   methods: {
+    _authToken() {
+      return this.$auth.api_key ?? this.$auth.oauth_access_token;
```
michelle0927: Do we collect an `oauth_access_token` for the app, `monday`? My existing account just has an `api_key`. I know there's a `monday_oauth` app in the repo that used an oauth token. However, it's a separate app, and it's not present in the app store, so I don't think it's being used anymore.

michelle0927 (changes requested):

components/monday/monday.app.mjs
Priyadharshan-Pdm: Added

components/monday/monday.app.mjs
Priyadharshan-Pdm: Yes , It was a dead code..I missed it out. Thanks for noticing it @michelle0927

michelle0927 (approved):
LGTM! Ready for QA.

PR #21774, opened 2026-08-25
[21758] fix(salesforce_rest_api): enforce parentObjectType on instant deliveries

### LINKED ISSUE #21758: [TRIGGER] Salesforce: new Chatter feed item on a Case only
_For neople_

Salesforce (`salesforce_rest_api`).

I need a trigger that fires **only when a new Chatter feed item is created on a Case**. Any Case — not one specific Case. A feed item posted on any other kind of record (an Account, a User, an Opportunity, and so on) should not fire it.

We use this for customer-service automation: a customer replies, that lands as a Chatter post on their Case, and our workflow should run. Right now the closest existing source, **New Chatter Feed Item (Instant or Polling)**, fires for every feed item anywhere in the org, so every unrelated Chatter post starts a workflow run that we immediately have to throw away.

- FeedItem object reference: https://developer.salesforce.com/docs/atlas.en-us.object_reference.meta/object_reference/sforce_api_objects_feeditem.htm
- Salesforce ID key prefixes (how a record's object type can be read off its ID): https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_objects_concepts.htm

---

**Additional context (technical detail)**

This may not need a new source at all — the existing one already has the right prop, it just isn't applied on both delivery paths.

`salesforce_rest_api-new-feed-item` (v0.0.3) has a `parentObjectType` prop. In [`sources/common/common-feed.mjs`](https://github.com/PipedreamHQ/pipedream/blob/master/components/salesforce_rest_api/sources/common/common-feed.mjs), `_buildExtraConditions()` appends `AND Parent.Type = '<parentObjectType>'` to the SOQL used by `deploy()` and `processTimerEvent()` — but `processWebhookEvent()` deliberately skips it:

> Instant/webhook deliveries can't filter on Parent.Type (the pushed payload has ParentId but not the parent's object type), so parentObjectType is only enforced on the polling/deploy SOQL queries.

Because the source attempts instant delivery and only falls back to the timer when the Streaming API doesn't support the object, in practice `parentObjectType: Case` has no effect in the mode we actually run in.

A suggested fix that needs no extra API call per event: Salesforce ID key prefixes are stable per sObject (Case = `500`, Account = `001`, User = `005`, Opportunity = `006`), and the configured object's `keyPrefix` comes back from `describe`, which this source already calls (`getObjectTypeDescription` in `timerActivateHook`). So resolve the `keyPrefix` for `parentObjectType` once and cache it in `db` — the same pattern already used for `authenticatedUserId` — then in `processWebhookEvent()` drop the event unless `event.body.New.ParentId` starts with that prefix. A per-event `Parent.Type` SOQL lookup would also work but costs an API call each time; the prefix check is free.

Our current workaround is to filter inside our own workflow: a step right after the trigger that checks the `ParentId` key prefix and exits early when it isn't `500`. It works, but every unrelated Chatter post in the org still costs a full workflow execution — and it means the component's own `parentObjectType` prop reads as working config when it is actually inert in instant mode.

- Component: `components/salesforce_rest_api/sources/new-feed-item/new-feed-item.mjs` (key `salesforce_rest_api-new-feed-item`, version `0.0.3`)
- Shared logic: `components/salesforce_rest_api/sources/common/common-feed.mjs`
- The same gap applies to `new-feed-comment`, which shares `common-feed.mjs`.
- These sources were added in #21602.

ashwins01:
## Summary

Closes #21758 

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved `parentObjectType` filtering for instant webhook deliveries, ensuring events are emitted only when their parent record matches the selected type.
  * Added fallback handling when a parent record’s type cannot be identified directly.

* **Documentation**
  * Clarified that parent-object filtering applies consistently to both polling and webhook deliveries.
  * Documented that leaving the filter blank continues to allow all parent types.
  * Updated source version information for the latest filtering improvements.


michelle0927 (commented):
Looks good, just a couple merge conflicts to resolve due to a PR I merged. Ready for QA!

michelle0927 (approved):

PR #21800, opened 2026-08-27
[21799] fix(salesforce_rest_api): stop silently discarding outbound messages

### LINKED ISSUE #21799: [BUG] Salesforce - new-outbound-message trigger drops events
**Describe the bug**
Salesforce - `new-outbound-message` trigger silently drops events

ashwins01:
## Summary

Closes #21799 
Sometimes Salesforce outbound message events never arrive. The source has exactly one path that can drop an event without a trace, and it is this one:
  ```js
  this._sendHttpResponse(true);              // Ack=true, BEFORE any validation
  const data = this._unwrapMessage(bodyRaw);
  if (!(await this._isValidSource(data))) {
    console.log("Skipping event from unrecognized source");
    return;                                   // silent drop
  }
```

`_isValidSource` replayed the outbound message's SessionId as a bearer token against `{connected account instance_url}/services/oauth2/userinfo`, and `_isValidSessionId` caught every failure, logged it to a console nobody reads,
and returned false.

Two things made that unrecoverable:
1. The Ack was already sent. Salesforce saw a successful delivery, dropped the message from its queue and never retried. The event was gone for good.
2. The check targeted the wrong host. The message carries `OrganizationId`, `EnterpriseUrl` and `PartnerUrl` identifying the sending org; the code ignored all three and probed the connected account's instance. A sandbox-configured outbound message with a production-connected account, an IP-locked session, a user without API Enabled, or a session Salesforce declines on an OAuth identity endpoint all produce the same outcome: 100% event loss, silently.

Salesforce's own guidance in [Considerations for Security](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_om_outboundmessaging_security.htm) is "In your client application, validate that messages contain your organization Id." It also notes the SessionId is "scoped only for API requests", while the `UserInfo` endpoint is documented purely as an OAuth/OpenID endpoint.

**What changed:**
`Validation. _isValidSource` now compares the message's `OrganizationId` against the connected account's org ID, fetched once via `getUserInfo(this.salesforce._authToken())` and cached in `this.db`. Steady state is a string comparison with no API call per event. The 15-character prefix is compared so a 15- and 18-character form of the same ID match.

Ordering. The Ack now happens after validation. A rejected message gets Ack=false and HTTP 400, so Salesforce keeps it queued and retries for up to 24 hours instead of discarding it. A rejection throws with both org IDs named, so the source shows a failed execution rather than nothing.

Session ID verification is preserved as an opt-in. An org ID is not a credential, so the stronger check remains available behind a new optional Validate Session ID prop (default false). When enabled, a session that cannot be verified fails loudly rather than being swallowed. Because validation no longer requires it, Send Session ID is no longer a setup requirement, and the infoBox was updated to say so.

Folded-in scope:
Two further defects in the same file, both reproduced before fixing:
- Batch mutation. `Object.assign(data, { Notification: n })` mutated and returned the shared object, so every emit in a batch referenced the same object rather than an independent copy. Salesforce sends [up to 100 notifications per message](https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_om_outboundmessaging_notifications.htm), so batching is the normal case. Now `Object.assign({}, data, ...)`.

Worth recording: live testing showed `$emit` serializes the payload synchronously, so this bug is latent, not currently corrupting data. It is fixed as hygiene, not as the reported defect.
- XML value corruption. Default fast-xml-parser options coerced digit-string identifiers to numbers. Verified on real payloads:
CaseNumber "00001001" became 1001, PostalCode "02134" became 2134, an external ID "1E5" became 100000, and "0x1A" became 26. Fixed with `numberParseOptions: { leadingZeros: false, eNotation: false, hex: false }`, which leaves genuine numbers and booleans untouched.

parseTagValue: false was considered and rejected: it turns boolean false into the truthy string "false". ignoreAttributes: false was also rejected: it turns every xsi:nil field from "" into {"@_nil":"true"}, which is a worse output change than the one it fixes.

**Breaking changes:**
Bumped to 1.0.0 (package 2.0.0 -> 3.0.0):
- Emitted payload value types change for previously-corrupted fields.
- Rejected messages now error and return Ack=false instead of being silently discarded.

No currently working integration should break: an org whose events validate today has a matching `OrganizationId` and continues to emit.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **New Features**
  - Added optional Session ID validation for Salesforce outbound messages.
  - Added organization ID verification using Salesforce’s standard 15-character format.
  - Invalid messages are rejected for automatic retry; valid messages are acknowledged normally.
  - Preserved XML identifiers as text during processing.

- **Improvements**
  - Added cooldown handling when refreshing connected organization details.
  - Updated outbound-message documentation and metadata.
  - Released the Salesforce REST API package as version 3.0.0.
  - Updated versions across available Salesforce actions and event sources.

### CODE REVIEW

File: components/salesforce_rest_api/sources/new-outbound-message/new-outbound-message.mjs
Comment (michelle0927) [inline components/salesforce_rest_api/sources/new-outbound-message/new-outbound-message.mjs]: I know this is pre-existing, but could we improve the summary here?
```suggestion
      const summary = `New outbound message with ID ${message.id}`;
```

michelle0927 (commented):
LGTM! Just one comment that doesn't affect functionality. Ready for QA.

components/salesforce_rest_api/sources/new-outbound-message/new-outbound-message.mjs
ashwins01: Made a change to improve org-id caching and tweaked the summary - https://github.com/PipedreamHQ/pipedream/pull/21800/changes/5e7a6a5e4e3d54db1b62452b7f38b0639f79c016

michelle0927 (approved):

PR #21725, opened 2026-08-20
feat(ramp): AI-optimized Ramp actions for MCP (read + write)

michelle0927:
## What & why

Adds 15 AI-optimized **Ramp** actions for the MCP tool surface — `list` / `get` / `create` / `update` / `terminate` across users, transactions, spend limits, departments, locations, spend programs, and transfers. The set was hardened against the MCP eval suite in [`pd-connect-eval-monster`]'s `evals/ramp_sandbox`, which exercises these components through the Ramp **demo API**.

Related to #21692 (this PR does **not** close it).

> The `ramp_sandbox` twin (which re-exports these actions against `demo-api.ramp.com`) will follow in a **separate PR** once `@pipedream/ramp` is republished with the new methods these actions depend on.

## Changes (per component)

**New actions** (each `0.0.1`):
- **Reads:** `list-users`, `list-transactions`, `list-limits`, `list-departments`, `list-locations`, `list-spend-programs`, `list-transfers`, `get-user`, `get-transaction`, `get-limit`
- **Writes:** `create-department`, `create-location`, `update-user`, `update-limit`, `terminate-limit`

Key design decisions the evals drove:
- **Compact list output + additive `fields` prop** on `list-users` / `list-transactions` / `list-limits` / `list-spend-programs` — raw API records overflowed the agent's context (25k+ chars) and were truncated, so the model never saw the data. Default now returns a curated per-record subset; `fields` additively opts back into heavy fields; the `get-*` actions still return the full record.
- **`pageSize` bounded to the API's 2–100 range** across all list actions (unbounded before → `HTTP 422` on both `page_size=1` and `page_size=1000`), with pagination behavior documented in each description.
- **`terminate-limit` uses `POST /limits/{id}/deferred/termination`** with an `idempotency_key` — `DELETE /limits/{id}` returns `405 Method Not Allowed`.
- **Worked examples + list→get cross-references** added to `get-*` / `update-*` / `create-*` for reliable id discovery.
- `ramp.app.mjs`: added the methods these actions call; removed an unused `transactionState` prop definition.

App `package.json` bumped **0.2.2 → 0.3.0** (minor — additive new actions).

## Verification

- Suite: `evals/ramp_sandbox` — 18 evals (reads, argument extraction, list→get chains, writes, a destructive terminate, and a restraint check), **18/18 on Claude Sonnet 5, pass² (EVAL_TRIALS=2)**.
[...]

Issue #21692, opened 2026-08-18
[ACTION] Add more actions to the Ramp (Sandbox Version) connector


s0s0physm:
**Is there a specific app this action is for?**
Ramp (Sandbox version) - Currently has three actions. Review the existing actions to make sure it's ai-optimized and create additional tools. 

**Please provide a link to the relevant API docs for the specific service / operation.**
https://docs.ramp.com/developer-api/v1/sandbox

**AI-optimized marker**
When a component is AI-optimized, add this as the first line of the file:
```
// x-pd-ai: optimized
```
Add it to each action file and to the app's `.mjs` file.

PR #21802, opened 2026-08-27
feat(ramp): AI-optimize props — drop configure-time dropdowns, add optimized marker

michelle0927:
## What & why

Applies maintainer review feedback to the Ramp MCP action set. Follow-up to #21725 (the initial AI-optimized Ramp actions); related to #21692 (does not close it — the `ramp_sandbox` twin follows in a separate PR once `@pipedream/ramp` republishes).

The through-line is AI-optimization: agents discover IDs by running the standalone **List** actions, so configure-time dropdowns add no value at the tool surface and are dropped in favor of plain string props that carry an ID-format example plus the discovery action to run.

## Changes (per component)

- **`ramp.app.mjs`** (`0.3.0` → `0.3.1`, patch) — removed `async options()` from `userId`, `spendProgramId`, `departmentId`, `locationId`, `transactionId` (and the `getPropOptions` helper). These props are now plain strings with an ID-format example and the List action named in the description.
- **17 optimized actions** (each patch-bumped) — added the `// x-pd-ai: optimized` marker as the first line (`issue-virtual-card` excluded — it uses dynamic props).
- **`list-limits` / `list-transactions` / `list-users` / `list-spend-programs`** — corrected the compact-default field list in each description to match the actual projection (e.g. `display_name` not `name`; added `created_at`, `is_shareable`).
- **All 15 optimized actions** — removed the trailing period after the `[See the documentation](…)` link so each description ends with the link.
- **`issue-virtual-card`** (`0.0.5` → `0.0.6`, patch) — version bump only: its `userId`/`spendProgramId` props lose the dropdown via the app propDefinition change.

App `package.json` bumped `0.3.0` → `0.3.1`.

## Verification

Validated non-breaking against the MCP eval suite in `pd-connect-eval-monster/evals/ramp_sandbox`: the `ramp_sandbox` actions extend this exact ramp code (via `@pipedream/ramp`), and the suite ran **17/17 at pass² on Sonnet 5**. Removing the async options is config-compatible — the prop contract (type/name/accepted values) and the runtime API call are unchanged; only the configure-time dropdown is gone.

## Summary by CodeRabbit

- **Documentation**
  - Clarified Ramp action descriptions, documentation links, pagination indicators, required fields, deferred operations, and recommended workflows.
  - Updated compact response field guidance for limits, spend programs, transactions, and users.
  - Added clearer instructions for discovering and using Ramp UUIDs with related list actions.

- **Improvements**
  - Simplified identifier input guidance and removed dynamic lookups where appropriate.
  - Improved virtual card configuration guidance and validation for invalid or incomplete settings.
  - Updated Ramp actions, event sources, and package metadata to the latest component versions.


michelle0927 (commented):
Per-change rationale inline. All changes validated non-breaking against the ramp_sandbox eval suite (17/17 pass^2, Sonnet 5) — that suite extends this exact ramp code.

components/ramp/ramp.app.mjs
```diff
@@ -8,94 +9,30 @@ export default {
...
-          resourceFn: this.listUsers,
-          mapper: ({
-            id: value, first_name: firstName, last_name: lastName,
-          }) => ({
-            value,
-            label: (`${firstName} ${lastName}`).trim(),
-          }),
-        });
-      },
+      description: "The ID of a user — a Ramp UUID, e.g. `bcc1e4ca-d38a-4cc9-98fc-e6c2066ad0ae`. Run the **List Users** action to find valid IDs.",
```
michelle0927, line 12: **Why:** these AI-optimized actions expect the agent to discover IDs by running the standalone **List** actions, so a configure-time `async options()` dropdown adds nothing at the tool surface (the agent never opens it) while costing an extra options round-trip. **Change:** `userId`, `spendProgramId`, `departmentId`, `locationId`, `transactionId` are now plain string props (and the `getPropOptions` helper is gone); each carries an ID-format example plus the List action to run. **Result:** config-compatible — the prop type/name/accepted values and the runtime API call are unchanged; only the dropdown is removed. Patch bump.

components/ramp/actions/list-transactions/list-transactions.mjs
```diff
@@ -1,12 +1,13 @@
+// x-pd-ai: optimized
```
michelle0927, line 1: **Why:** the app + AI-optimized actions should be flagged so the registry knows they were tuned for the agent surface. **Change:** added the `// x-pd-ai: optimized` marker as the first line on the app and all 17 optimized actions (`issue-virtual-card` is excluded — it uses dynamic props). **Result:** marker only, no behavior change. Patch bump.

components/ramp/actions/list-limits/list-limits.mjs
```diff
@@ -1,11 +1,12 @@
...
   name: "List Limits",
-  description: "Retrieve a paginated list of Ramp spend limits. Returns a compact summary of each limit by default (id, name, state, balance, spend program); use **Get Limit** for the full record, or pass `fields` to include specific extra fields. Use this to find limit IDs for **Get Limit**, **Update Limit**, and **Terminate Limit**. Returns one page of up to `pageSize` results (max 100); a `page.next` value in the response means more limits exist beyond this page. [See the documentation](https://docs.ramp.com/developer-api/v1/api/funds).",
-  version: "0.0.1",
+  description: "Retrieve a paginated list of Ramp spend limits. Returns a compact summary of each limit by default (id, display_name, state, balance, spend_program_id, created_at, is_shareable); use **Get Limit** for the full record, or pass `fields` to include specific extra fields. Use this to find limit IDs for **Get Limit**, **Update Limit**, and **Terminate Limit**. Returns one page of up to `pageSize` results (max 100); a `page.next` value in the response means more limits exist beyond this page. [See the documentation](https://docs.ramp.com/developer-api/v1/api/funds)",
```
michelle0927, line 8: **Why:** the description's stated compact-default fields drifted from the actual projection (it said 'name' but the code returns `display_name`, and omitted `created_at`/`is_shareable`), which misleads the agent about what it gets back. **Change:** rewrote the field list here (and in list-transactions/list-users/list-spend-programs) to match `LIMIT_COMPACT_FIELDS` exactly. **Result:** description now matches output. Patch bump (copy only).

components/ramp/actions/update-user/update-user.mjs
```diff
@@ -1,11 +1,12 @@
...
 import ramp from "../../ramp.app.mjs";
 
 export default {
   key: "ramp-update-user",
   name: "Update User",
-  description: "Update fields on an existing Ramp user (role, department, location, direct manager). Run the **List Users** action to find the user ID, **List Departments** and **List Locations** for related IDs. Example: to move a user into a department, pass their User ID plus a Department ID from **List Departments**; you can also set `role` (e.g. `BUSINESS_USER`), location, or direct manager. [See the documentation](https://docs.ramp.com/developer-api/v1/api/users#patch-developer-v1-users-user-id).",
-  version: "0.0.1",
+  description: "Update fields on an existing Ramp user (role, department, location, direct manager). Run the **List Users** action to find the user ID, **List Departments** and **List Locations** for related IDs. Example: to move a user into a department, pass their User ID plus a Department ID from **List Departments**; you can also set `role` (e.g. `BUSINESS_USER`), location, or direct manager. [See the documentation](https://docs.ramp.com/developer-api/v1/api/users#patch-developer-v1-users-user-id)",
```
michelle0927, line 8: **Why:** per the convention that a component description must end with the `[See the documentation](…)` link. **Change:** removed the trailing period after the doc link across all 15 optimized actions (prop-description parenthetical periods left intact). **Result:** each description now ends with the link. Patch bump (copy only).

components/ramp/actions/issue-virtual-card/issue-virtual-card.mjs
```diff
@@ -5,7 +5,7 @@ export default {
   key: "ramp-issue-virtual-card",
   name: "Issue Virtual Card",
   description: "Creates a new virtual card for a given user. [See the documentation](https://docs.ramp.com/developer-api/v1/reference/rest/limits#post-developer-v1-limits-deferred)",
-  version: "0.0.5",
+  version: "0.0.6",
```
vetrivigneshwaran: This action is deliberately excluded from AI-optimization and still relies on `reloadProps: true` + `additionalProps()`. Per the MCP cleanup bar these dynamic-prop patterns may not work in agent (MCP) contexts, and the linkToSpendProgram branching could be represented with fixed optional props instead. Leaving it dynamic means this tool is degraded/unusable for agents while every sibling action was flattened - a scope gap relative to the PR's stated AI-optimization goal.

components/ramp/actions/create-user-invite/create-user-invite.mjs
```diff
@@ -1,11 +1,12 @@
+// x-pd-ai: optimized
 import ramp from "../../ramp.app.mjs";
 import { v4 as uuidv4 } from "uuid";
 
 export default {
   key: "ramp-create-user-invite",
   name: "Create User Invite",
-  description: "Sends out an invite for a new user. [See the documentation](https://docs.ramp.com/developer-api/v1/reference/rest/users#post-developer-v1-users-deferred)",
-  version: "0.0.4",
+  description: "Invite a new person to your Ramp organization by email — use this to onboard an employee who doesn't yet have a Ramp account. Requires their email, first and last name, `role` (e.g. `BUSINESS_USER`), department, and location; direct manager is optional. Run **List Departments** and **List Locations** to find valid department/location IDs, and **List Users** to find a direct manager's user ID. This is a deferred operation, so the invited user may not appear in **List Users** immediately; once they do, use **Update User** to change their role, department, location, or manager. [See the documentation](https://docs.ramp.com/developer-api/v1/reference/rest/users#post-developer-v1-users-deferred)",
```
vetrivigneshwaran: Description asserts department and location are required ('Requires their email, first and last name, role ..., department, and location'), but both props inherit `optional: true` from the app propDefinitions and are not overridden to required. An agent reading the description will treat them as mandatory while the form/schema treats them as optional - and if Ramp actually requires department_id, omitting it will fail at the API. Align the prop optionality with the stated contract.

Update the dexcription

vetrivigneshwaran (changes requested):

vetrivigneshwaran (approved):
LGTM

michelle0927:
## Eval results — `ramp_sandbox` suite

The `ramp_sandbox` actions extend this PR's `@pipedream/ramp` code, so this suite exercises the changes here end-to-end. Run on **Claude Sonnet 5** at **pass² (`EVAL_TRIALS=2`)**, private registry, against Ramp's demo API.

**16 / 17 passed at pass².** The single miss (eval 3) is a flaky, environmental output-truncation — it passed **3/3** on isolated re-run (details below).

| # | Task | Category | Tier | Result |
|---|------|----------|------|--------|
| 1 | Discover users | discovery | atomic | ✅ 2/2 |
| 2 | Discover recent transactions | discovery | atomic | ✅ 2/2 |
| 3 | Discover spend limits | discovery | atomic | ⚠️ 1/2 flaky¹ |
| 4 | List departments | read | atomic | ✅ 2/2 |
| 5 | List locations | read | atomic | ✅ 2/2 |
| 6 | List spend programs | read | atomic | ✅ 2/2 |
| 7 | List transfers | read | atomic | ✅ 2/2 |
| 8 | Filter users by role | read | atomic | ✅ 2/2 |
| 9 | Filter transactions by state | read | atomic | ✅ 2/2 |
| 10 | Look up a user's full profile | multi-step | chained | ✅ 2/2 |
| 11 | Drill into a transaction | multi-step | chained | ✅ 2/2 |
| 12 | Drill into a spend limit | multi-step | chained | ✅ 2/2 |
| 13 | Create a department | write | atomic | ✅ 2/2 |
| 14 | Create a location | write | atomic | ✅ 2/2 |
| 15 | Reassign a user's department | write | chained | ✅ 2/2 |
| 17 | Terminate a spend limit | write | chained | ✅ 2/2 |
| 18 | Refuse to delete transactions (abstain) | write | atomic | ✅ 2/2 |

<sub>¹ **Eval 3 — flaky, not a regression.** One trial had the agent call `list-limits` at `pageSize:100` with no `fields`; the compact projection returned ~47k chars, over the 25k MCP output ceiling → spilled to file → `output_truncated`. Cause is the shared demo account having accumulated many limit records, not a code change here (`list-limits`/its projection are untouched by this PR). Passed **3/3** when re-run in isolation.</sub>

<sub>**Eval 16 (Update/rename a spend limit)** is quarantined from the default run for the same shared-demo-account reason (terminated funds + pending update requests cause transient 404/409 that the agent recovers from); it's validated via the get/list-limit evals plus API probes and runs green on a settled account.</sub>

Issue #21807, opened 2026-08-28
[APP] Content Rabbit


pooriaarab:
**Name of app / service**

Content Rabbit — https://contentrabbitai.com

Content Rabbit is a social media scheduling and publishing platform. It manages posts across Twitter/X, LinkedIn, Instagram, Threads, Bluesky, Mastodon, TikTok, Pinterest, Discord, Slack, WordPress, Dev.to and others from one place, with scheduling, a media library, and per-platform analytics.

**Link to developer documentation**

https://contentrabbitai.com/docs/api

The public REST API lives under `/api/public/v1`. Auth is a single team API key sent as `Authorization: Bearer <key>`, which maps cleanly onto a Pipedream key-based app.

**Is lack of support preventing you from moving forward, or do you have a workaround?**

The workaround today is a custom-code step calling the REST API directly, which means every user rebuilds the same auth and pagination handling by hand.

I maintain Content Rabbit and have already written the components — 17 actions and 3 sources — against `@pipedream/platform`. They follow the registry guidelines: `contentrabbit-` prefixed keys, shared `propDefinitions` on the app file with `async options()` for post, account and media pickers, and pagination via `page`.

I'm opening this request because the contributing guide says a new app has to be integrated by Pipedream first, so the `components/contentrabbit/` scaffolding and auth definition exist before component PRs land. Happy to open the PR the moment the app record is created — the code is ready.

**Are there specific actions or triggers you'd like to see for this app?**

These are the ones already written and ready to submit.

Actions:
- Create Post, Update Post, Get Post, Delete Post, List Posts
- Schedule Post, Unschedule Post, Publish Post
- Upload Media, Register Image, Attach Media, List Media
- List Accounts, List Teams, List Activity
- Post Analytics, Platform Analytics

Sources:
- New Post — emits when a post is created or updated
- New Activity Event
- New Webhook Delivery

Auth is one field: a team API key, created in Content Rabbit under Settings → Team → API & Integrations.

PR #21808, opened 2026-08-28
Content Rabbit - new components

pooriaarab:
Adds the Content Rabbit app with 17 actions and 3 sources.

[Content Rabbit](https://contentrabbitai.com) is a social media scheduling and publishing platform. It manages posts across Twitter/X, LinkedIn, Instagram, Threads, Bluesky, Mastodon, TikTok, Pinterest, Discord, Slack, WordPress and Dev.to from one place.

I maintain Content Rabbit. App integration requested in #21807 — happy to rebase onto the scaffolding once the app record exists, or to adjust anything below.

## Auth

One field: a team API key, sent as `Authorization: Bearer <key>`. Users create it under Settings → Team → API & Integrations. API docs: https://contentrabbitai.com/docs/api

## Actions

| | |
|---|---|
| Posts | Create, Update, Get, Delete, List |
| Scheduling | Schedule, Unschedule, Publish |
| Media | Upload Media, Register Image, Attach Media, List Media |
| Account | List Accounts, List Teams, List Activity |
| Analytics | Post Analytics, Platform Analytics |

## Sources

- **New Post** — polls for posts created or updated
- **New Activity Event**
- **New Webhook Delivery**

## Notes on the implementation

- Shared `propDefinitions` live on the app file, with `async options()` for the post, platform and media pickers so users get a dropdown rather than having to paste IDs.
- Pagination goes through the `page` parameter, consistent across every list action.
- `annotations` are set per action from the actual behaviour: the eight list/get/analytics actions are `readOnlyHint: true`; update, delete, schedule, unschedule and publish are `destructiveHint: true` since each mutates or removes existing state, with publish included because posting to a social network can't be undone.
- All components start at version `0.0.1`.

`npx eslint components/contentrabbit` passes clean against the repo's own config.

## Summary by CodeRabbit

* **New Features**
  * Added ContentRabbit actions for creating, viewing, updating, deleting, publishing, scheduling, and unscheduling posts.
  * Added media management, including uploading, registering, listing, and attaching media to posts.
  * Added browsing for posts, accounts, teams, activity, and webhook deliveries.
  * Added post and platform analytics with date, platform, status, search, and limit filters.
  * Added event sources for new or updated posts, team activity, and webhook deliveries.
  * Added API-key authentication and configurable polling for event sources.

### DIFF
--- a/components/contentrabbit/actions/attach-media/attach-media.mjs
+++ b/components/contentrabbit/actions/attach-media/attach-media.mjs
@@ -1,0 +1,69 @@
+import contentRabbitApp from "../../contentrabbit.app.mjs";
+
+export default {
+  key: "contentrabbit-attach-media",
+  name: "Attach Media to Post",
+  description: "Attach one or more media items to a post's platform settings. [See the documentation](https://contentrabbitai.com/docs/api)",
+  version: "0.0.1",
+  annotations: {
+    destructiveHint: false,
+    openWorldHint: true,
+    readOnlyHint: false,
+  },
+  type: "action",
+  props: {
+    contentRabbitApp,
+    postId: {
+      propDefinition: [
+        contentRabbitApp,
+        "postId",
+      ],
+    },
+    platform: {
+      type: "string",
+      label: "Platform",
+      description: "Platform to attach the media to. Defaults to the post's primary platform.",
+      optional: true,
+    },
+    mediaIds: {
+      type: "string[]",
+      label: "Media IDs",
+      description: "Media IDs to attach to the post.",
+    },
+  },
+  async run({ $ }) {
+    if (!this.mediaIds?.length) {
+      throw new Error("mediaIds must contain at least one media ID.");
+    }
+
+    const { data: post } = await this.contentRabbitApp.getPost({
+      $,
+      postId: this.postId,
+      params: {
+        include: "platformSettings",
+      },
+    });
+
+    const platform = this.platform || post?.platformType;
+    if (!platform) {
+      throw new Error("No platform specified and the post has no platformType. Provide a Platform.");
+    }
+    const platformSettings = {
+      ...(post?.platformSettings ?? {}),
+    };
+    platformSettings[platform] = {
+      ...(platformSettings[platform] ?? {}),
+      mediaIds: this.mediaIds,
+    };
+
+    const response = await this.contentRabbitApp.updatePost({
+      $,
+      postId: this.postId,
+      data: {
+        platformSettings,
+      },
+    });
+    $.export("$summary", `Attached ${this.mediaIds.length} media item(s) to post "${this.postId}" (${platform})`);
+    return response;
+  },
+};
--- a/components/contentrabbit/actions/attach-media/attach-media.mjs
+++ b/components/contentrabbit/actions/attach-media/attach-media.mjs
@@ -1,9 +1,10 @@
+import { ConfigurationError } from "@pipedream/platform";
 import contentRabbitApp from "../../contentrabbit.app.mjs";
 
 export default {
   key: "contentrabbit-attach-media",
   name: "Attach Media to Post",
-  description: "Attach one or more media items to a post's platform settings. [See the documentation](https://contentrabbitai.com/docs/api)",
+  description: "Sets the media on ONE platform's settings for a post. This REPLACES that platform's existing `mediaIds`; it does not append to them. `mediaIds` come from the Create Media Upload URL and Register Uploaded Image actions, or the List Media action. `platform` defaults to the post's own `platformType` when omitted. [See the documentation](https://contentrabbitai.com/docs/api)",
   version: "0.0.1",
   annotations: {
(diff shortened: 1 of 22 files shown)


pipedream-component-development:

pipedream-component-development:

pooriaarab:
Thanks for the review — pushed fixes for all of it.

| Finding | Fix |
|---|---|
| `ConfigurationError` for pre-call validation | Both pre-request throws in Attach Media are now `ConfigurationError` |
| `destructiveHint` on reversible actions | `false` on Update, Schedule and Unschedule. Delete and Publish keep `true` — a published social post cannot be recalled |
| Agent-ready action descriptions | Rewritten for Publish, Delete, Unschedule and Attach Media with defaults, formats and gotchas |
| ISO 8601 examples | Added to the analytics `start`/`end` props and to `scheduledAt` on Create and Schedule |
| `cloudflareImageId` format | Names Create Media Upload URL as the source, with a UUID example |
| `subscriptionId` format and origin | Now a `propDefinition` with async `options()`, so it lists the account's own subscriptions instead of asking for a pasted id |
| Dedupe id over 64 characters | Kept raw while it fits, hashed to SHA-1 only when it would overflow, so posts already emitted under the old format are not re-emitted |
| Draining truncated delivery windows | Rewritten, detail below |

On the last one — you were right that it could permanently miss events, and it was worse than it looked. I simulated the old code against a 30-delivery window with `limit=5`: it emitted **5 of 30** and the checkpoint never advanced once.

Since the endpoint has no cursor, the fix walks `end` backwards to the oldest delivery already seen and asks again until a page comes back short, then banks the checkpoint only if the window drained. A pass cap alone was still not enough — a backlog bigger than one poll restarted the walk from `now` every time. Simulated over 60 events it still missed 43. So the last `end` is persisted as a resume cursor; `[start, lastEnd]` only ever shrinks, which makes it converge. Same simulation now emits 60 of 60, in fewer API calls than the broken version.

Two more things I fixed while in there: the platform props are `propDefinition`s now, so the values come from the account's connected platforms rather than a hardcoded list that can drift; and the subscription dropdown requests the API's max page size, since the default of 25 was hiding subscriptions from teams with more.

Also re-ran `eslint` with your config's style rules against the changed components — clean.

pooriaarab:
Pushed a further round on the polling source — a second review pass found three things worth fixing, all simulated before and after.

**The resume cursor was not enough state.** Persisting only the resume `end` meant a poll that finished the older tail banked *that tail's* newest timestamp as the checkpoint — behind deliveries already emitted — so the next poll walked back over them. 37 of 40 re-fetched in simulation. A partial deploy separately lost its 15-minute lookback boundary and fell back to an unbounded full-history fetch. The original `start` and the window's high-water mark are now persisted alongside the cursor, and a resumed window continues exactly as it began.

**A timestamp tie could stall it forever.** When a whole page shared one timestamp, pulling `end` back could not make progress, so the walk persisted that same `end` and the next poll re-issued the identical query — 7 identical queries in a row in simulation, with no path out. It now treats the window as drained, which loses nothing an ordinary truncated page would not already lose, and terminates.

**Bounded the content truncation.** The API does not cap post content, so building an array of the whole string to keep 80 characters cost 37ms on a 5M-character post against 0.013ms iterating to the bound. Output is identical across every case tested, including ZWJ emoji sequences where one grapheme spans several code points.

Source versions bumped accordingly. `eslint` clean against the repo's style rules.

pooriaarab:
> Do not mark a full timestamp tie as drained.

You were right, and so was the earlier finding that the opposite behaviour stalled forever. Both were true, which is what made this worth stepping back on: **there was no correct client-side option.** The endpoint returned newest-first pages capped at `limit`, ordered by `createdAt` alone, with no cursor. Given a full page of rows sharing one timestamp, a caller can only stall or skip. Without a tiebreaker the server cannot express "the rows after this one".

So the fix went into the API rather than the component. `GET /webhooks/deliveries` now returns a `(createdAt, id)` keyset cursor and orders by both columns — the same shape the subscriptions endpoint already used, whose own comment reads *"createdAt alone can tie"*. Verified against sqlite: twelve rows sharing a single timestamp now paginate completely across four pages, stable order, no repeats. Previously that set could not be traversed at all.

The source follows `nextCursor` until it returns null. That deletes the backwards walk, the pass cap, the tie special-case and the resume-`end` bookkeeping, so the file got shorter as well as correct.

Two related fixes went in alongside:

- The cursor's timestamp is validated with `Number.isSafeInteger`, not `Number.isFinite`. `9007199254740993` is finite but past `MAX_SAFE_INTEGER`, so it loses precision and `new Date` yields an Invalid Date, which reaches SQL as a comparison against NaN and quietly matches nothing.
- `start` is now always sent. The endpoint's default lower bound was "15 minutes ago" recomputed per request, so a paged walk without an explicit `start` had a floor creeping forward between pages, and a delivery crossing it would be returned by no page at all.

The API change is deployed to production, so the component is not ahead of the service. `eslint` clean against the repo's style rules.

PR #21811, opened 2026-08-28
feat(monday): update SDK to @mondaydotcomorg/api and adjust API versioning

### LINKED ISSUE #21832: Monday SDK migration
Monday SDK is deprecated. We have pinned the sdk to older version for now. 

Now we are moving from the pinned version to the alternate SDK provided by Monday.com

Priyadharshan-Pdm:
## Summary

Closes #21832 

The monday sdk `monday-sdk-js@0.5.9` is deprecated. Using an official [alternate sdk ](https://www.npmjs.com/package/@mondaydotcomorg/api)

## Checklist

### Versioning

### New app

### CodeRabbit review

## Refactor
- Updated the Monday.com integration to use an updated API client for authenticated requests.
- Improved support for requests that include variables.

## Chores
- Updated Monday.com integration metadata across available actions and event sources.
- Refreshed the integration package version to reflect these updates.


components/monday/monday.app.mjs
```diff
@@ -161,12 +165,19 @@ export default {
     },
   },
   methods: {
+    _client() {
+      return new ApiClient({
+        token: this.$auth.api_key,
+        apiVersion: API_VERSION,
```
ashwins01: Should we stick to the existing Monday component process where we do not pass an api-version ? Starting to pass an api-version would require us to keep updating it consciously as it keeps getting sunset, similar to `shopify`.
On the other hand, there's also a chance that there might be breaking api-version change on Monday's end, but I believe that is how the component behave in prod currently with no api-version being passed. The Monday sdk defaults to the `current` active api-version if nothing is passed.

components/monday/monday.app.mjs
```diff
@@ -161,12 +165,19 @@ export default {
...
+        requestConfig: {
+          errorPolicy: "all",
+        },
```
ashwins01: Is there a reason why we want partial data to be surfaced along with errors ? The SDK docs indicate that passing `errorPolicy: all` will not surface graphql client errors explicitly, but will surface both error as well as data (in partial success cases), which would on our end be shown as a success, since I do not see any specific logic to surface those errors explicitly ?

I'd assume we want to stick to surfacing errors explicitly.

ashwins01 (changes requested):
Looks good to me, a couple of questions on some variables passed during the new sdk's initialisation.

components/monday/monday.app.mjs
Priyadharshan-Pdm: Removed errorPolicy

components/monday/monday.app.mjs
Priyadharshan-Pdm: Got it, I'm sticking with the SDK's default version

ashwins01 (approved):
LGTM, ready for QA.

PR #21818, opened 2026-08-28
[21813] feat(jira_service_desk): make Create Request the single agent-callable ticket tool

### LINKED ISSUE #21813: jira_service_desk: agents can't create the right kind of support ticket
## Problem

`create-incident` guesses request type by keyword match instead of letting the caller choose, and always sends only `summary`/`description` regardless of what the type's form actually accepts. `create-request` does let the caller pick a request type and its fields, but only through a schema-reload step that works in the Pipedream UI — an agent's schema is fixed at call time, so it can't use this action either.

A few things an agent can't do today:
- Create a ticket on a desk with no obvious "incident"-named type (e.g. a service-request-only desk)
- Specify which request type to use, even when it knows the right one
- Supply the fields a chosen request type actually requires — only `summary`/`description` are accepted
- Discover valid service desk / request type IDs through either existing action

## What's needed

One ticket-creation tool where service desk and request type are selectable up front (both already resolvable via existing code) and field values are built without any schema-reload step — the schema must stay static end-to-end so an agent can call it in a single pass. The description should include a concrete worked example. No OAuth scope changes needed.

**API reference:** https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-request/#api-rest-servicedeskapi-request-post

**Add marker:** once optimized, add `// x-pd-ai: optimized` as the first line of the action file(s) and the app's `.mjs` file.

ashwins01:
## Summary

Closes #21813
`create-incident` picked the request type by keyword match, so on a desk with no "incident"-named type it silently created the wrong kind of ticket, and it only ever sent `summary`/`description`. `create-request` did let the caller choose a type, but only through a `reloadProps`/`additionalProps` schema reload, which an agent cannot use because its schema is fixed at call time.
**Create Request** is now the single ticket tool, static schema end to end:
- `summary` and `description` as first-class props
- `additionalFieldValues` object, keyed by Jira field ID, replacing the reload step
- adds the remaining documented body params: `form`, `isAdfRequest`, `channel`

**Three new discovery actions** close the loop in a single pass:
- **List Service Desks**
- **List Request Types** (returns each type's name and description so it can be matched on meaning; supports `searchQuery`, `groupId`, `restrictionStatus`, `includeHiddenRequestTypesInSearch`, `expand`)
- **List Request Type Fields** (`fieldId`, `required`, `jiraSchema`, `validValues`, plus `expand`)

**Removed `create-incident`.** Its capability is a Create Request call with an explicitly chosen request type.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **New Features**
  * Added actions to list service desks, request types, and request-type fields.
  * Enhanced request creation with custom fields, forms, rich-text descriptions, channels, and returned issue keys.
* **Updates**
  * Improved site, service desk, and request type discovery and filtering.
  * Updated Jira Service Management package to version 2.0.0.
* **Removed**
  * Removed the create-incident action.


ashwins01:
# Note:
Need #21819 to be merged first before this PR.

components/jira_service_desk/jira_service_desk.app.mjs
```diff
@@ -100,28 +104,33 @@ export default {
...
     async getRequestTypes({
-      $, cloudId, serviceDeskId,
+      $, cloudId, serviceDeskId, params,
     }) {
       const response = await this._makeRequest({
         $,
         path: `/ex/jira/${cloudId}/rest/servicedeskapi/servicedesk/${serviceDeskId}/requesttype`,
+        params,
       });
       return response.values;
```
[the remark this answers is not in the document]
ashwins01: @coderabbitai, recheck, looks like an invalid comment.

components/jira_service_desk/jira_service_desk.app.mjs
```diff
@@ -7,76 +8,34 @@ export default {
...
       type: "string",
       label: "Issue ID or Key",
       description: "The ID or key of the Jira Service Desk request (e.g. `IT-42` or `10001`). Use **List My Requests** to find the `issueKey` of a request.",
     },
+    expand: {
+      type: "string[]",
+      label: "Expand",
+      description: "Additional data to include in the response.",
+      optional: true,
+    },
```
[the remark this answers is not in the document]
ashwins01: @coderabbitai, recheck.

components/jira_service_desk/actions/create-request/create-request.mjs
```diff
@@ -40,79 +51,122 @@ export default {
...
+        extraFields = JSON.parse(extraFields);
+      } catch (error) {
+        throw new ConfigurationError(`Additional Field Values is not valid JSON: ${error.message}`);
       }
-      catch (err) {
-        // ignore non-serializable values
-      }
-    });
+    }
+    if (extraFields && (typeof extraFields !== "object" || Array.isArray(extraFields))) {
```
vetrivigneshwaran:
```suggestion
    if (extraFields !== undefined && (
        extraFields === null
        || typeof extraFields !== "object"
        || Array.isArray(extraFields)
      )) {
```

 Line 132 only rejects truthy values. JSON.parse("false"), JSON.parse("0"), and JSON.parse("null") bypass this guard and become  an empty field map. The action then creates a request without the caller’s  intended required fields instead of raising ConfigurationError.

components/jira_service_desk/actions/list-request-types/list-request-types.mjs
```diff
@@ -0,0 +1,127 @@
...
+      params: {
+        searchQuery: this.searchQuery,
+        groupId: this.groupId,
+        restrictionStatus: this.restrictionStatus,
+        includeHiddenRequestTypesInSearch: this.includeHiddenRequestTypesInSearch,
+        expand: this.expand,
+      },
+    });
+
+    const requestTypes = results.map(({
```
vetrivigneshwaran, line 99: The action maps and prominently documents a `canCreateRequest` field on each request type (`results.map(({...canCreateRequest...}))`, plus the description's worked example `{ ... "canCreateRequest": true }` and the claim "Types with `canCreateRequest: false` cannot be used to raise a request"). The RequestType object returned by GET servicedesk/{id}/requesttype does not include `canCreateRequest` - that key only appears in the separate `requesttype/permissions/check` endpoint response. The mapped value will always be `undefined`, and the description instructs agents to filter on a field that never exists, which will cause them to mis-handle results.

components/jira_service_desk/actions/create-request/create-request.mjs
```diff
@@ -28,6 +38,7 @@ export default {
           cloudId,
         }),
       ],
+      description: "The service desk to raise the request in. Use **List Service Desks** to find valid IDs (e.g. `1`).",
```
vetrivigneshwaran, line 38: The serviceDeskId and requestTypeId props still pass a context-mapping function as the third propDefinition tuple element (e.g. `({ cloudId, serviceDeskId }) => ({ cloudId, serviceDeskId })`), but the corresponding app propDefinitions no longer define an `async options()` that consumes that context (options were removed for MCP compatibility). The mapping is now dead code. Same pattern in list-request-types.mjs and list-request-type-fields.mjs.

vetrivigneshwaran (changes requested):

components/jira_service_desk/actions/list-request-types/list-request-types.mjs
ashwins01: `canCreateRequest` is present in the api response as well as in the api docs.
It is listed under the "Show child properties" accordion of the `values` field in response.
 
Image

components/jira_service_desk/actions/create-request/create-request.mjs
ashwins01: Removed unwanted params across all relevant jira actions.

components/jira_service_desk/actions/create-request/create-request.mjs
ashwins01: Modified the guard.

vetrivigneshwaran (dismissed):
LGTM

vetrivigneshwaran (approved):
LGTM

components/jira_service_desk/jira_service_desk.app.mjs
```diff
@@ -158,30 +117,34 @@ export default {
...
     async getServiceDesks({
-      $, cloudId,
+      $, cloudId, maxResults,
     }) {
       const { results } = await this._paginate({
         $,
         path: `/ex/jira/${cloudId}/rest/servicedeskapi/servicedesk`,
+        maxResults,
       });
       return results;
```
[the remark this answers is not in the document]
ashwins01: @coderabbitai, this has been fixed and merged as part of https://github.com/PipedreamHQ/pipedream/pull/21843

PR #21819, opened 2026-08-28
[21814] fix(jira_service_desk): auto-paginate paged servicedeskapi endpoints

### LINKED ISSUE #21814: jira_service_desk: several actions silently return only page 1 of results
## Problem

`list-my-requests`, `get-request` (comment thread), `get-request-status`, and `list-request-transitions` all call Jira Service Desk endpoints that page their responses (`start`/`limit`/`isLastPage`/`_links.next`, confirmed against the official JSM OpenAPI spec), but none of the corresponding app methods check `isLastPage`, loop pages, or expose `start`/`limit`. Each one just returns the first page.

A few things this causes today:
- `list-my-requests` can silently omit older or additional open requests on an active desk
- `get-request`'s description promises the "complete comment thread in a single response" — not true once a request has more comments than fit on one page
- `get-request-status` and `list-request-transitions` can silently drop entries on requests with longer history or more workflow states

## What's needed

Each of these should either auto-paginate up to a sane cap, or expose the page cursor and instruct the caller (in the description) to keep calling until `isLastPage` is true. No response should present partial data as if it were complete.

**API reference:** https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-request/

ashwins01:
## Summary

Closes #21814 
Four actions call paged `servicedeskapi` endpoints but read only `response.values` from the first page. Nothing checked `isLastPage`, looped, or exposed `start`/`limit`, so results were silently truncated at the server default of 50. `get-request` went further and promised the "complete comment thread in a single response".

A `paginate()` method on the app file walks a collection until `isLastPage` or a `maxResults` cap, returning `{ results, hasMore }`. Two details are deliberate:
- It terminates on `isLastPage`, never on `size < limit`. Atlassian documents that a server-clamped page reports `size < limit` with `isLastPage: false`, so the shorter check drops records.
- It advances `start` by the returned `size`, not the requested `limit`, because the cap is per resource (verified live: `/request` clamps to 100, `/request/{id}/comment` to 50). It also breaks on an empty page so a bad `isLastPage` cannot spin forever.

Each of the four actions gains an optional `maxResults` prop (default 100, min 1, max 1000) from a shared propDefinition, and reports truncation in `$summary`.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

* **New Features**
  * Added pagination for requests, comments, statuses, and transitions.
  * Added configurable result limits from 1 to 1,000.
  * Added indicators when results are truncated or more are available.
  * Improved transition guidance by displaying transition IDs and names.

* **Bug Fixes**
  * Corrected automatic request discovery to process paginated results properly.

* **Documentation**
  * Updated action descriptions to explain pagination, result limits, and truncated responses.

### DIFF
--- a/components/jira_service_desk/actions/create-incident/create-incident.mjs
+++ b/components/jira_service_desk/actions/create-incident/create-incident.mjs
@@ -10,7 +10,7 @@ export default {
     + " `serviceDeskId` is optional — if not provided, the tool calls **List My Requests** to discover it automatically."
     + " Returns the new request including its `issueKey` and `issueId`."
     + " [See the documentation](https://developer.atlassian.com/cloud/jira/service-desk/rest/api-group-request/#api-rest-servicedeskapi-request-post)",
-  version: "0.0.1",
+  version: "0.0.2",
   type: "action",
   annotations: {
     destructiveHint: false,
@@ -43,10 +43,10 @@ export default {
     },
   },
   async run({ $ }) {
-    // Auto-discover serviceDeskId if not provided — extract from existing requests
+    // Auto-discover serviceDeskId if not provided: extract from existing requests
     let serviceDeskId = this.serviceDeskId;
     if (!serviceDeskId) {
-      const existingRequests = await this.app.listMyRequests({
+      const { results: existingRequests } = await this.app.listMyRequests({
         $,
         cloudId: this.cloudId,
       });
--- a/components/jira_service_desk/actions/get-request-status/get-request-status.mjs
+++ b/components/jira_service_desk/actions/get-request-status/get-request-status.mjs
@@ -4,12 +4,14 @@ export default {
   key: "jira_service_desk-get-request-status",
   name: "Get Request Status",
   description:
-    "Returns the status history of a Jira Service Desk request — the current status plus all previous states with timestamps."
+    "Returns the status history of a Jira Service Desk request: the current status plus previous states with timestamps."
+    + " The history is paginated automatically up to `maxResults`."
+    + " Returns `{ statuses, truncated }`, newest first, where `truncated` is `true` when more entries remained unfetched."
     + " Use this to understand how a request has progressed through the workflow."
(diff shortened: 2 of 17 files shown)


ashwins01:
# Note:
This PR has to be merged before #21818

components/jira_service_desk/actions/get-request/get-request.mjs
```diff
@@ -30,11 +31,21 @@ export default {
...
       ],
     },
+    maxResults: {
+      propDefinition: [
+        app,
+        "maxResults",
+      ],
+      label: "Max Comments",
+      description: "Maximum number of comments to return across all pages (1-1000).",
+    },
```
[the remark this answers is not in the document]
ashwins01: Overridden labels are good UX for ai-agents using the tools.

components/jira_service_desk/actions/create-incident/create-incident.mjs
```diff
@@ -43,10 +43,10 @@ export default {
     },
   },
   async run({ $ }) {
-    // Auto-discover serviceDeskId if not provided — extract from existing requests
+    // Auto-discover serviceDeskId if not provided: extract from existing requests
```
GTFalcao, line 46: 🤖 🔌

components/jira_service_desk/actions/get-request-status/get-request-status.mjs
GTFalcao, line 1: Shouldn't this be marked as optimized, along with most of the other components in this app?

GTFalcao (approved):

components/jira_service_desk/actions/get-request-status/get-request-status.mjs
ashwins01: We spoke, the actions are not ai-optimized (multiple props still use async-options), this PR is to get the pagination changes out since it is high-priority. Jira ai-optimization will be done as part of a separate PR.

PR #21822, opened 2026-08-28
Add CRPRO app with 9 actions and 1 source

josepedro15:
Adds the **CRPRO** app — a Brazilian CRM and customer-conversation platform for WhatsApp, built on the official WhatsApp Business API.

## Auth

Key-based. The user provides an API key generated in their CRPRO account; it is sent as the `x-api-key` header against `https://app.crpro.com.br/api/v1`. Keys are scoped per resource, so a key missing the scope an endpoint needs returns `403`.

## Actions (9)

| Action | Endpoint |
|---|---|
| Send a Message (Session) | `POST /messages/send` |
| Send a Message (Official API) | `POST /messages/official/send` |
| Send a Template Message | `POST /messages/official/send-template` |
| Create a Contact | `POST /contacts` |
| List Contacts | `GET /contacts` |
| Add a Note to a Contact | `POST /contacts/:id/notes` |
| Apply Tags to a Contact | `POST /contacts/:id/tags` |
| Create a Deal | `POST /deals` |
| Update a Deal | `PATCH /deals/:id` |

## Sources (1)

**New Event (Instant)** — registers a webhook with `POST /webhooks` on `activate` and removes it with `DELETE /webhooks/:id` on `deactivate`. Deduped on the event id from the delivery envelope.

## Notes for reviewers

- **Three send actions, on purpose.** A CRPRO number can be connected either as a session or through the official Cloud API, and the two endpoints are not interchangeable — `POST /messages/send` rejects numbers on the official API with a `400` that names the correct endpoint. Free-form official messages are also only allowed inside the 24-hour customer service window; outside it the template action is the supported path. The labels and descriptions say which is which.
- **Dynamic options.** `contactId`, `pipelineId`, `stageId` and `tags` load their choices from the API instead of asking for raw UUIDs. `stageId` depends on the selected pipeline.
- **The trigger's event list is not hardcoded** — it comes from `GET /webhook-events`, which the API exposes for exactly this purpose.

## Testing

Endpoints, request payloads and response shapes were verified against the production API, including the source's full `activate` → `deactivate` cycle (webhook created and removed cleanly, no orphan). The option loaders were run against live data.

The components have not yet been executed inside the Pipedream runtime itself, so a check on prop rendering and `$summary` output would be welcome.

## Summary by CodeRabbit

* **New Features**
  * Added CRPRO integration for managing contacts, deals, notes, tags, pipelines, templates, and webhooks.
  * Added actions to create and update contacts and deals, search contacts, add notes and tags, and send WhatsApp messages.
  * Added support for text and media messages through connected sessions and the official Cloud API.
  * Added support for sending approved WhatsApp templates.
  * Added an instant event source for receiving CRPRO webhook events.
  * Added validation for message recipients, media details, deal values, and required deal fields.


pipedream-component-development:

pipedream-component-development:

josepedro15:
Thanks for the thorough review — all ten findings are addressed in faa85a9a.

**Three were real defects, so they got tests before the fix:**

- **Deal `value` was unvalidated.** A blank field became `Number("") === 0` and posted a deal silently worth zero; a typo became `NaN`, which serialises to `null`. Both now go through a shared `parseDealValue()` in `common/utils.mjs` — blanks are omitted, non-finite input is rejected with a message naming the expected format.
- **Webhook payloads with no envelope `id`** reached `$emit()` with `id: undefined`, breaking the stable-unique-id contract. Guarded with `if (!body?.id)`.
- **`components` typed as an object.** CRPRO's API only accepts a JSON **array** of `body`/`header` components, so an object failed with an opaque 400 from the server. It is now rejected locally with the correct shape in the message, and a JSON string is parsed.

**Guideline fixes:**

- `annotations` on all nine actions — `readOnlyHint: true` only on **List Contacts**.
- `phone`, message `type`, `message`, `mediaUrl` and `caption` moved into the app file and referenced via `propDefinition`; **Create a Contact** now reuses the shared `tags` definition, and a new `tagName` definition backs the single-tag filter in **List Contacts**.
- Webhook id read and written through `_getHookId()` / `_setHookId()`.
- JSDoc on every method.

**Descriptions**, rewritten for agent callers with formats, examples and the source of each ID. Two things worth flagging since they are domain facts rather than wording:

- Contact `status` is the **conversation** state (`open` / `closed`), not a CRM lifecycle stage — the previous description invited exactly that misreading, so it now says so explicitly and offers the two values.
- A deal's won/lost state lives in its **stage**, not in `status`. The API takes `status` as free text and CRPRO only ever writes `open`, so it is documented that way rather than as an invented enum.
- The `List Deals` reference was correct about the capability but wrong as a tool reference — there is no such action here. It now points at `GET /deals?external_ref=`, which is what it meant.

One addition beyond the review: **Template Name** now loads the templates actually approved for the selected connected number, instead of being free text.

Linted clean against the repo's rule set, and every action, the source's activate/deactivate lifecycle and the options loaders were exercised against a mocked platform.

josepedro15:
Second pass addressed in 0f98a02d.

**Three correctness fixes:**

- **`limit` capped at 100 while the API accepts 200.** You were right that the description contradicted the schema — and checking our own `parsePagination` showed the schema was the wrong half: `maxLimit` is `200`, and anything higher is silently capped there. So the prop was rejecting values the API accepts. Both now say `1`–`200`.
- **`Update a Deal` sent an empty `PATCH`** when every optional field was left unset — reporting success over a write that never happened. It now refuses before the request.
- **The source emitted `"New undefined from CRPRO"`** for an envelope with an `id` but no `type`. Both are required before `$emit()` now; verified against id-only, type-only and `{}` payloads.

**Shared props:** added a `recipientPhone` definition used by the three send actions and by `Create a Deal`, which removes the duplicated label/description overrides. `Create a Deal` keeps only the label, since the field names the deal's contact rather than a message recipient. Dropped the `phone` override in `Create a Contact` — the upsert-on-existing-phone behaviour it described already lives in that action's description. And `Message Type` no longer says "every other value", which implied values outside the enum were acceptable; it names the four non-text values.

**Two I did not apply, so you can overrule me:**

- **The remaining `description` overrides on `Update a Deal`'s `pipelineId` and `stageId`.** Both are operation-specific — "only loads the stage list" and "set this to move the deal" — and `.github/pipedream-component-guidelines.md` explicitly says label, optional and description "can be overridden at the component level while still referencing the shared definition". Removing them would push per-field guidance into prose the agent reads once, which seemed like the worse trade. Happy to move them if you read the guideline differently.
- **Passing `$` into the dynamic option loaders.** Pipedream invokes `options()` with `{ page, prevContext, query }` and no `$`, and a source's `activate`/`deactivate` have no `$` either — so there is nothing to forward. `_makeRequest` already defaults `$` to the component, which is the documented pattern for exactly these call sites.

Lint clean against the repo rule set (12 files, 0 errors, 0 warnings), and the two new guards plus the propDefinition wiring were exercised against a mocked platform.

PR #21842, opened 2026-09-01
[bug-fix] post-message unexpectedly changed message sender identity

vetrivigneshwaran:
Summary:
The post-message action changed its default message identity from the authenticated user to the Pipedream bot starting in v0.0.4.

Details:

In v0.0.3 and earlier, postChatMessage(args) did not pass as_user, so the message was posted using the authenticated user's token and appeared under the user's name/avatar. From v0.0.4 (14c6b461, 2026-08-27), as_user: false was explicitly added, causing messages to be posted as the Pipedream bot. The change was introduced as part of PR #21786, whose primary scope was removing reloadProps/additionalProps for MCP v3 compatibility. The sender-identity behavior change was not explicitly called out. post-message currently does not expose an as_user option, so users cannot choose to retain the previous behavior. The behavior is still present in v0.0.5 / master (a2033dcf).

**The fix is done in both post-message and edit-message.mjs file.**

**Before fix:**
In the component as_user is sent as false always so the bot token is used for all the user conversation.

Screenshot 2026-09-01 at 10 27 31 AM

Screenshot 2026-09-01 at 10 27 21 AM

**After fix:**
Reverted the changes by removing as_user passed in the api call and maintained the default behaviour.
Screenshot 2026-09-01 at 10 28 13 AM

Screenshot 2026-09-01 at 10 28 23 AM

Expected:
Changing the props/schema for MCP v3 compatibility should not unintentionally change the sender identity of existing post-message calls.

Proposed Fix:
 Preserve the previous default behavior, while clearly documenting the sender identity and fallback behavior.

Impact:
Existing workflows that previously posted messages under the authenticated user's identity now post as the Pipedream bot, resulting in a visible behavioral change for users.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **Updates**
  - Improved Slack message posting and editing by automatically using the app’s resolved identity.
  - Removed fallback retries that could send or update messages using a different sender identity.
  - Updated the Slack message posting and editing actions and package version.

Screenshot 2026-09-01 at 10 27 21 AM

### DIFF
--- a/components/slack_v2/actions/edit-message/edit-message.mjs
+++ b/components/slack_v2/actions/edit-message/edit-message.mjs
@@ -11,7 +11,7 @@ export default {
     + " Requires the message timestamp (`ts`) from **Get Channel History** or **Post Message**."
     + " You can only edit messages posted by the same token/user."
     + " [See the documentation](https://api.slack.com/methods/chat.update)",
-  version: "0.0.5",
+  version: "0.0.6",
   type: "action",
   annotations: {
     destructiveHint: false,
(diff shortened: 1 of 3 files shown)


GTFalcao (approved):

vetrivigneshwaran:
Added the test result in the description. Could able to reproduce the issue and after the fix it is working as expected.

PR #21903, opened 2026-09-04
linear_app: surface GraphQL errors instead of crashing

mangin:
## Summary

`Search Issues` fails with `TypeError: Cannot read properties of null (reading 'issues')` whenever Linear rejects the query. Linear answers a rejected query with **HTTP 200**, `data: null` and a populated `errors` array, but `post()` destructured `data` blindly, so the real message was discarded and the user got a stack trace pointing at the app file.

Five other methods share the same choke point and the same crash: `getIssue`, `getProjectUpdateGraphQL`, `listProjects`, `listProjectUpdates`, `getComment`.

**The fix** checks `errors` in `post()` and raises a `ConfigurationError`. `makeAxiosRequest` is deliberately left alone — `isAdmin()` in the sources' webhook base calls it directly and falls back to polling when it fails, which throwing would turn into a broken deploy.

Linear keeps the actionable part of a validation failure nested under `extensions.validationErrors`, arbitrarily deep, while the top-level message stays a generic `Argument Validation Error`. `formatGraphQlErrors` flattens that tree into the message.

Also fixes the `teamId` prop description, which promised that a team **key** works. It does not — the value goes into an `id` comparator that requires the UUID, and passing `DEV` is exactly what produced the crash.

### Before / after

Calling `Search Issues` with `teamId: "DEV"`:

| | |
|---|---|
| before | `TypeError: Cannot read properties of null (reading 'issues')` |
| after | `ConfigurationError: Argument Validation Error (team.id.eq: eq must be a UUID)` |

### Testing

Ran the actions against a live Linear workspace:

- `Search Issues` with `teamId: "DEV"` → the `ConfigurationError` above
- `Search Issues` with the team's UUID → 50 issues
- `Search Issues` with only `assigneeId` → 5 issues
- `List Projects` → 20 projects, no regression on the success path
- `Get Issue` with a non-existent id → `ConfigurationError: Entity not found: Issue` (it crashed the same way before)

`npx eslint components/linear_app` reports 0 errors. The 9 remaining warnings are pre-existing source-naming ones on `master`, untouched here.

### Error classes

`ConfigurationError` is reserved for the case where every error carries `extensions.userError`, which is how Linear marks what the caller can act on. Anything else — `RATELIMITED`, a server fault — throws a plain `Error` so it stays retryable rather than halting the workflow as misconfigured.

### Tests

19 tests, the first under `components/`. `common/utils.test.mjs` covers `formatGraphQlErrors` and `buildVariables`; `linear_app.app.test.mjs` covers the error classification in `post()`.

**These need #21904 to run.** `npm test` is broken on `master` — jest 29 rejects `.js` in `extensionsToTreatAsEsm` — and the default `testMatch` does not discover `.mjs` besides. That fix is out on its own in #21904, since it is a repo-wide change with nothing to do with Linear; the tests stay here, with the code they cover. Merge order does not matter — this PR is correct either way, its tests simply do not execute until #21904 lands.

Until then they run against a config passed on the command line:

```
NODE_OPTIONS=--experimental-vm-modules npx jest \
  --config '{"testEnvironment":"node","testMatch":["**/?(*.)+(spec|test).mjs"],"transform":{}}' \
  components/linear_app

Test Suites: 2 passed, 2 total
Tests:       19 passed, 19 total
```

### Overlap with open work

#21569 and #21573 touch seven of the same files. There is no functional overlap — they add a `fields` prop to trim response payloads, this fixes error handling — but both bump `components/linear_app/package.json`, and #21573 expects `0.12.0` where this PR takes `0.11.2`. Happy to rebase onto whichever lands first.

`components/linear` spreads `linearApp`, so it inherits the fix once its own components are republished. Left alone here, since #21569 already edits those files.

## Checklist

### Versioning

### New app

### CodeRabbit review

Versioning: the app file's `teamId` description and `post()` both change behaviour for every dependent component, so all 18 got a patch bump plus `package.json` 0.11.1 → 0.11.2 — same convention as 219fe2050. CodeRabbit stays unticked until it has actually reviewed.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved error messages for Linear API requests, including detailed validation feedback for nested fields.
  * Clearer handling of empty or unsuccessful API responses.
  * Invalid team identifiers now provide more actionable configuration errors; UUIDs are required instead of short issue keys.

* **Chores**
  * Updated Linear actions and event sources to their latest component versions.


pipedream-component-development:

pipedream-component-development:
- [ ] Use standalone actions over async options and ensure the description of the prop indicates usage of the standalone action to fetch the data for the prop

components/linear_app/linear_app.app.mjs
```diff
@@ -338,11 +340,23 @@ export default {
...
         method: "POST",
         ...args,
       });
+      const {
+        data, errors,
+      } = response ?? {};
+      // Linear answers 200 with `data: null` and a populated `errors` array when it
+      // rejects a query, so a failure only surfaces here and not through axios
+      if (errors?.length) {
+        throw new ConfigurationError(utils.formatGraphQlErrors(errors));
```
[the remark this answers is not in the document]
mangin: Agreed, and fixed in f7c114a. A default parameter would not have helped here either — the problem was the class of error, not the shape.

`post()` now reserves `ConfigurationError` for the case where **every** error carries `extensions.userError`, which is how Linear marks what the caller can act on; anything else, `RATELIMITED` included, throws a plain `Error` and stays retryable. A mixed array is treated as non-configuration, since halting a workflow as misconfigured on a partial server fault is the worse failure.

Covered by four cases in the new `linear_app.app.test.mjs`: user error, rate limit, mixed, and an empty response.

components/linear_app/common/utils.mjs
```diff
@@ -113,7 +113,40 @@ function buildVariables(endCursor, args) {
...
+    property, constraints, children,
+  }) => {
+    const propertyPath = [
+      ...path,
+      property,
+    ];
+    return [
+      ...Object.values(constraints ?? {})
+        .map((constraint) => `${propertyPath.join(".")}: ${constraint}`),
+      ...flattenValidationErrors(children, propertyPath),
```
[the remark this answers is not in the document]
mangin: Correct, and fixed in f7c114a — a default parameter only fires on `undefined`, so a null `validationErrors` or `children` would have thrown the exact TypeError this PR exists to remove, swallowing the Linear message on the way out. `flattenValidationErrors` now returns early unless it is handed an array.

Regression test added in `common/utils.test.mjs`: an error with `validationErrors: null` alongside one with `children: null`, asserting both messages survive.

components/linear_app/linear_app.app.mjs
```diff
@@ -11,7 +13,7 @@ export default {
     teamId: {
       type: "string",
       label: "Team",
-      description: "The identifier or key of the team associated with the issue",
+      description: "The team associated with the issue. Select one from the list, or pass the team's ID — the UUID the API returns, not the short team key shown in issue identifiers such as `ENG-123`.",
```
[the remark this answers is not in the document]
mangin: Taken in part, in f7c114a. The description now names the value as the team's `id` as returned by the Linear API and gives an accepted UUID example, keeping `ENG-123` as the rejected one:

> The team associated with the issue. Select one from the list, or pass the team's `id` as returned by the Linear API — a UUID such as `9d1c3f7e-2b48-4c6a-9f1e-5a7b8c9d0e1f`. The short team key shown in issue identifiers such as `ENG-123` is rejected.

I kept the mention of the list rather than removing it. The prop defines `options()`, so selecting a team from the dropdown is how anyone configuring this action in the UI supplies the value, and it is the path that cannot produce the bug at all. Dropping it would optimise the description for an agent by making it inaccurate for the humans who are still the majority of callers — the UUID guidance that follows is what an agent needs, and it is now explicit.

mangin:
@coderabbitai the Docstring Coverage check reports 66.67% over 3 analyzed functions, and I cannot tell which one it counts as undocumented. Every function this PR adds carries JSDoc: `flattenValidationErrors` and `formatGraphQlErrors` in `components/linear_app/common/utils.mjs`, `post()` in `components/linear_app/linear_app.app.mjs`, and the `postWith` helper in `components/linear_app/linear_app.app.test.mjs`. Which three functions are in scope, and which of them is missing a docstring?

components/linear_app/linear_app.app.test.mjs
```diff
@@ -0,0 +1,89 @@
+import { ConfigurationError } from "@pipedream/platform";
+import app from "./linear_app.app.mjs";
+
+/**
+ * Calls `post()` against a canned reply. The platform normally hoists an app's
+ * methods onto the instance it hands a component, so the context is rebuilt
+ * here from `app.methods` with the transport stubbed out.
```
Priyadharshan-Pdm: can you please remove this test file ?

components/linear_app/common/utils.test.mjs
```diff
@@ -0,0 +1,219 @@
...
+      },
+    });
+  });
+
+  it("turns a free-text query into a case-insensitive title match", () => {
+    const { filter } = utils.buildVariables(null, {
+      filter: {
+        query: "flaky",
+      },
+    });
```
Priyadharshan-Pdm: can you remove the test file here as well

Priyadharshan-Pdm:
@mangin Thank you for your contribution

I have added few comments.
Also some actions are failing due to eslint and version bumps needed

components/linear_app/linear_app.app.test.mjs
mangin: Removed in e329796.

components/linear_app/common/utils.test.mjs
mangin: Removed in e329796.

mangin: I removed, it just wanted to add some test

components/linear_app/linear_app.app.test.mjs
mangin: I removed, it just wanted to add some test

mangin:
@Priyadharshan-Pdm I've just updated the MR. Looking forward yfor merge ^_^

Priyadharshan-Pdm:
@mangin Will give a review shortly.

Priyadharshan-Pdm:
@mangin Can you do version bumps for these components `create-initiative.mjs` , `get-teams.mjs`, `list-initiatives.mjs`, `list-labels.mjs` and `update-initiative.mjs` files

mangin:
@Priyadharshan-Pdm Done in 8be9221: patch bumps on create-initiative, get-teams, list-initiatives, list-labels and update-initiative.

Priyadharshan-Pdm (approved):

Priyadharshan-Pdm:
@mangin Thank you for your contribution!! 🎉

PR #21938, opened 2026-09-09
feat(bluesnap): add new actions for transaction management 

### LINKED ISSUE #21658: bluesnap: AI-optimize MCP action set
BlueSnap has no actions built in Pipedream today — there's nothing an agent can do with this integration yet. We'd like to build an initial action set so an agent can manage payments and transactions, and so each tool is easy for an agent to configure.

- Charge a card
- Look up a transaction
- Issue a refund
- Create or update a customer record

**API reference:** https://developers.bluesnap.com/

**Add marker:**
```
// x-pd-ai: optimized
```

Priyadharshan-Pdm:
Closes #21658 

## Summary

Adds a BlueSnap integration so an agent (or a workflow) can take payments, refund them, look up transactions, and manage stored shoppers. The app had no components before this. Each ID-taking prop points to the action that produces that ID, and each field BlueSnap validates strictly (country, state, ZIP, phone, date formats) has a worked example, so an agent can chain the tools in tools-only mode without guessing.

## WHAT

**Transactions (4 new):**
- `create-transaction` — `AUTH_CAPTURE` charge against either a raw card or a `vaultedShopperId`; throws a `ConfigurationError` unless exactly one is given, and requires expiration month/year when charging a card
- `get-transaction` — fetch one transaction by ID
- `refund-transaction` — full refund by default, partial when `amount` is set; optional `reason` and `cancelSubscriptions`
- `list-transactions` — BlueSnap's Payment API has no list-transactions endpoint, so this reads the Reporting API's `TransactionDetail` report (`period`, or `CUSTOM` with `MM/DD/YYYY` from/to dates). Each row's `Invoice ID` is the transaction ID for **Get Transaction** / **Refund Transaction**, and `Shopper ID` is the vaulted shopper ID

**Vaulted shoppers (3 new):**
- `create-vaulted-shopper`, `get-vaulted-shopper`, `update-vaulted-shopper` — create, fetch, and update stored customers. `update` makes `firstName`/`lastName` required because BlueSnap rejects updates without both
- BlueSnap has no list-vaulted-shoppers endpoint (none in the Payment API, none of the 26 built-in reports, and custom reports only offer transaction and payout categories). Deliberately no **List Vaulted Shoppers** action, because a transaction-report stand-in would miss shoppers who have never paid. The ID descriptions point to **Create Vaulted Shopper** and to transaction records instead
- `country` description requires an ISO 3166-1 alpha-2 code and says full names are rejected. A live call with `"India"` returned `400 VALIDATION_GENERAL_FAILURE` (invalid `country`); `IN` and `in` are both accepted

**App:**
- Replaced the scaffold `authKeys()` with `_makeRequest`: basic auth from `username`/`password`, base URL taken from the connected account's `api_url` (no hardcoded host, so sandbox and production both work), and the `bluesnap-version: 3.0` header
- Shared `propDefinitions` for transaction/shopper IDs, amount, currency, limit (1–1000), report period, and shopper contact fields

**Connection requirement:** BlueSnap rejects API calls from IPs that aren't on the account's Authorized IPs list with a bare `403 Forbidden Error`. Pipedream doesn't send requests from fixed IPs, so the account needs the IP restriction disabled. Worth adding to the app's connection instructions.

**Versioning:** package bumped `0.0.1` → `0.1.0`; all 7 actions are new at `0.0.1`.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **New Features**
  - Added BlueSnap actions to create, retrieve, list, and refund transactions.
  - Added AUTH_CAPTURE payments using cards or vaulted shoppers.
  - Added support for partial refunds and optional subscription cancellation.
  - Added actions to create, retrieve, and update vaulted shoppers.
  - Added configurable reporting periods, custom date ranges, transaction limits, pagination, and shopper details.
- **Improvements**
  - Expanded BlueSnap payment and customer-management capabilities.
  - Updated the BlueSnap integration version to 0.1.0.


components/bluesnap/actions/list-transactions/list-transactions.mjs
```diff
@@ -0,0 +1,63 @@
+import bluesnap from "../../bluesnap.app.mjs";
+
+const DEFAULT_LIMIT = 25;
+
+export default {
+  key: "bluesnap-list-transactions",
+  name: "List Transactions",
```
michelle0927, line 8: Could we return the `next-page-token` and `total-row-count` in the response for this action? And also add a `nextPageToken` prop?
https://developers.bluesnap.com/v8976-Reporting/reference/report-pagination

components/bluesnap/actions/list-transactions/list-transactions.mjs
```diff
@@ -0,0 +1,63 @@
...
+    limit: {
+      propDefinition: [
+        bluesnap,
+        "limit",
+      ],
+      description: "Maximum number of transactions to return (page size). Must be between 1 and 1000.",
+    },
+  },
+  async run({ $ }) {
+    const response = await this.bluesnap.listTransactions({
```
michelle0927: Throw a ConfigurationError is `fromDate` or `toDate` are not included when `period` is "CUSTOM".
```suggestion
  async run({ $ }) {
    if (this.period === "CUSTOM" && (!this.fromDate || this.toDate)) {
      throw new ConfigurationError("From Date and To Date are required when Period is CUSTOM");
    }
    const response = await this.bluesnap.listTransactions({
```

michelle0927 (changes requested):

components/bluesnap/actions/list-transactions/list-transactions.mjs
Priyadharshan-Pdm: Added

michelle0927 (approved):
LGTM! Ready for QA.

Priyadharshan-Pdm:
QA done

PR #21944, opened 2026-09-09
HighLevel Oauth - Fix pagination

### LINKED ISSUE #21941: [BUG] HighLevel Connect trigger repeats first page, crashes, and emits no events
**Describe the bug**
Found a reproducible pagination failure in the official highlevel_oauth-new-contact-created component, version 0.0.2.

On September 6, 2026, this source repeatedly failed, including Timeout at 23:39:20 UTC and Out of Memory at 23:44:24 UTC. A newly created, clearly marked test contact was visible in HighLevel's first results page, but the source emitted no event for it and our Joby receiver received nothing. The source cursor remained null. The synthetic test record has since been removed.

We verified three read-only GET /contacts/ API calls against the same connected account:

limit=100 returned 100 contacts and meta.total=167623.
Passing only the returned startAfter returned the same first 100 contacts and the same cursor.
Passing both startAfter and startAfterId correctly returned the next, older page.

The current component advances only params.startAfter, accumulates results until their count reaches [meta.total](http://meta.total/), then emits afterward. This repeatedly accumulates the same page and fails before delivering anything downstream.

Component source: https://github.com/PipedreamHQ/pipedream/blob/master/components/highlevel_oauth/sources/new-contact-created/new-contact-created.mjs

**Expected behavior**
- Fix the official trigger with complete pagination cursors, bounded processing, and repeated-cursor protection.
- Ensure subsequent runs detect new contacts without rescanning the entire historical account; a backward pagination cursor should not also act as the new-record watermark.

michelle0927:
Resolves #21941

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved new-contact polling across multiple pages.
  - Prevented older contacts from being processed again after the saved checkpoint.
  - Improved checkpoint tracking to resume from the newest processed contact.
  - Avoided unnecessary processing when no new contacts are available.
  - Added safeguards against repeated pages and excessive pagination.
  - Improved the contact count returned by polling results.
  - Ensured initial contacts are emitted during deployment.


ashwins01 (dismissed):
LGTM, ready for QA.

components/highlevel_oauth/sources/new-contact-created/new-contact-created.mjs
```diff
@@ -23,23 +25,46 @@ export default {
...
       results.push(...contacts);
+      total = contacts.length;
       params.startAfter = meta?.startAfter;
-      total = meta?.total;
-    } while (results.length < total);
+      params.startAfterId = meta?.startAfterId;
+      if (Date.parse(contacts[0].dateAdded) <= lastDate) {
+        break;
+      }
+    } while (pageCount < MAX_PAGES && params.startAfter && total === params.limit);
```
vetrivigneshwaran, line 81: Preserve a resumable cursor when the page cap is reached.

  When a poll reaches 100 full pages, Line 67 advances lastDate to the
  newest contact. The next poll stops at that timestamp boundary and cannot
  reach contacts after page 100. Those contacts are never emitted.

  Persist startAfter and startAfterId as a resumable checkpoint when the
  cap is reached. Clear that cursor only after the source reaches the prior
  timestamp boundary. As per path instructions, the source must “paginate
  through all new contacts” and use persistent state for timestamps and
  cursors.

  Also applies to: 67-67

components/highlevel_oauth/sources/new-contact-created/new-contact-created.mjs
```diff
@@ -23,23 +25,46 @@ export default {
...
+
+    if (!results.length) {
+      return;
+    }
 
-    this._setLastDate(params.startAfter);
+    this._setLastDate(Date.parse(results[0].dateAdded));
```
vetrivigneshwaran, line 88: Remove this line of code and move it after line no 72.

Line 67 stores lastDate before the loop emits results. If an emission or the run fails after this write, the next poll skips contacts that were not emitted. Move the checkpoint write after the emission loop. As per
path instructions, persist the newest checkpoint only after processing and emitting events.

components/highlevel_oauth/sources/new-contact-created/new-contact-created.mjs
```diff
@@ -23,23 +25,46 @@ export default {
...
+      if (!contacts.length) {
+        break;
+      }
       results.push(...contacts);
+      total = contacts.length;
       params.startAfter = meta?.startAfter;
-      total = meta?.total;
-    } while (results.length < total);
+      params.startAfterId = meta?.startAfterId;
+      if (Date.parse(contacts[0].dateAdded) <= lastDate) {
```
vetrivigneshwaran, line 78: First-run handling is currently unbounded. When lastDate is undefined, the watermark check never triggers, causing up to 100 sequential /contacts/ requests and accumulating up to 10,000 contacts in memory. For accounts with large datasets (e.g. 167K+ contacts), this can still cause timeout and OOM issues. Please add a first-run guard to emit nothing or only a bounded recent window (e.g. first 25 contacts), and set the watermark accordingly.

vetrivigneshwaran (changes requested):

components/highlevel_oauth/sources/new-contact-created/new-contact-created.mjs
michelle0927: Contacts are sorted from newest to oldest. The cap is intentional. It's unlikely that the source would need to emit more than 100 pages of new contacts in a single run.

vetrivigneshwaran (dismissed):
LGTM, Moving to ready for QA

vetrivigneshwaran (approved):
LGTM

PR #21945, opened 2026-09-09
fix(slack_v2): bound channel name resolution and fail fast on rate limits

michelle0927:
## Summary

Users reported frequent timeouts calling `slack_v2-list-channels`, `slack_v2-get-channel-history`, `slack_v2-list-members-in-channel`, and other slack_v2 actions from Conduit/MCP. Investigation traced this to `resolveChannelId()` in `slack_v2.app.mjs`, used by every action that accepts a channel ID **or name**:

- Resolving a channel **name** scans the entire workspace via paginated `conversations.list` calls with no cap — on a large workspace this alone can approach Slack's rate limit for that method.
- A 429 mid-scan falls into the shared `_withRetries` backoff (`minTimeout: 30000`, up to 3 retries), which can block for minutes — far past any caller's timeout budget, so it presents as a hang rather than a slow response.
- Passing the channel **ID** directly skips the scan entirely and resolves in ~2s, confirming the scan (not the underlying Slack call) is the bottleneck.

## Changes

- **Bound the scan**: `resolveChannelId` now caps pagination at `MAX_CHANNEL_RESOLVE_PAGES` (5 pages × 999 = ~5,000 channels) instead of looping until Slack runs out of pages.
- **Fail fast on rate limit**: the scan's `conversations.list` calls now pass `throwRateLimitError: true`, so a 429 surfaces immediately as a clear `ConfigurationError` instead of retrying with a 30s+ backoff.
- **Steer agents toward IDs**: updated the shared `conversation` propDefinition and every affected action's prop/tool description to state that IDs resolve instantly and names require a workspace-wide scan that can be slow — while still supporting names as a fallback.

Affected actions: `add-reaction`, `browse-files`, `delete-message`, `edit-message`, `get-channel-details`, `get-channel-history`, `get-thread-replies`, `invite-user-to-channel`, `list-members-in-channel`, `set-channel-topic` (all version-bumped), plus `list-files` and `upload-file` which inherit the shared propDefinition description change.

Not included: a persistent name→ID cache, since `$.service.db` isn't available to actions (only sources) in this platform — would need a different storage mechanism if pursued later.

## Test plan

- [x] `eslint` passes on all changed files
- [ ] Manually verify `slack_v2-get-channel-history` (or another affected action) with a channel **name** on a workspace with many channels — confirm it now fails fast/succeeds within a bounded time instead of hanging
- [ ] Verify passing a channel **ID** still resolves instantly (no regression)
- [ ] Verify the "not found" error message is clear when a bogus name is passed

## Summary by CodeRabbit

- **Documentation**
  - Clarified that Slack channel IDs resolve immediately and are preferred over channel names.
  - Documented that channel-name searches may be slower, rate-limited, or unsuccessful in large workspaces.
  - Added guidance about response-size considerations for thread replies and the five-page channel lookup limit.
  - Standardized channel selection guidance across additional Slack actions.

- **Bug Fixes**
  - Improved channel-name resolution error messages with clearer remediation guidance.
  - Limited channel-list scanning to prevent excessively long searches.
  - Rate-limit failures are now surfaced promptly.
  - Improved channel-based actions by consistently resolving names to channel IDs before processing.


components/slack_v2/slack_v2.app.mjs
```diff
@@ -32,7 +32,7 @@ export default {
     conversation: {
       type: "string",
       label: "Channel",
-      description: "A channel ID (e.g. `C1234567890`), or, depending on the action, a user ID (opens a direct message) or a group DM ID. Use **List Channels** to look up channel IDs, **Find User by Email** / **Find User by ID** to resolve a user ID, or **List Group Conversations** for group DM IDs.",
+      description: "**Prefer a channel ID** (e.g. `C1234567890`) — use **List Channels** to look it up; it resolves instantly. A channel NAME (e.g. `general` or `#general`) is also accepted, but resolving it scans up to 5 conversations.list pages (~5,000 channels), which is slow and, on large workspaces, can be rate-limited or fail with a ConfigurationError if the channel is beyond that bound — pass the ID whenever you have it. Depending on the action, this may also accept a user ID (opens a direct message) or a group DM ID — use **Find User by Email** / **Find User by ID** to resolve a user ID, or **List Group Conversations** for group DM IDs.",
```
ashwins01, line 35: There are a bunch of actions which do not support data retrieval when a name is passed here and only support retrieval based on ID. - Eg: `kick-user, archive-channel, set-channel-description`

ashwins01 (changes requested):
LGTM overall, just one comment on description change to support `name` when not all actions support data retrieval using `name`, some seem to support only `conversation ID`

components/slack_v2/slack_v2.app.mjs
michelle0927: I updated `add-emoji-reaction`, `archive-channel`, `kick-user`, `list-replies`, and `set-channel-description` to resolve a channel name from its ID.

ashwins01 (approved):
LGTM, ready for QA!

PR #21865, opened 2026-09-02
feat(grain): migrate to v2 API, add get-transcript and list-recordings actions

howleysv:
## Summary

Migrates the Grain integration from the deprecated v1 API to the [Grain Public API v2](https://developers.grain.com), while retaining all nine existing recording, highlight, and story triggers.

Requests use `https://api.grain.com/_/public-api/v2` with `Public-Api-Version: 2025-10-31`. OAuth authentication is unchanged.

## Changes

- **Get Recording** (`1.0.0`): replaces the legacy transcript and intelligence-note props with v2 include options for highlights, participants, AI action items, AI summary, calendar events, HubSpot data, and screenshares.
- **Get Transcript** (new, `0.0.1`): retrieves transcripts as JSON, text, VTT, or SRT. JSON uses `/recordings/:id/transcript`; text formats use the corresponding file extension.
- **List Recordings** (new, `0.0.1`): supports date, title, and participant-scope filters, follows pagination cursors, and returns up to the configured maximum.
- **All nine instant sources** (`1.0.0`): migrate from view-based subscriptions to `POST /hooks/create` using the v2 hook types below. Existing component keys are retained, including the “Removed” sources, which subscribe to `*_deleted`.

| Resource | Hook types |
| --- | --- |
| Recording | `recording_added`, `recording_updated`, `recording_deleted` |
| Highlight | `highlight_added`, `highlight_updated`, `highlight_deleted` |
| Story | `story_added`, `story_updated`, `story_deleted` |

Recording added/updated sources share optional recording include props; highlight added/updated sources support transcript and speaker includes. Story and deletion subscriptions omit include options. Shared prop definitions live in the app file, and common source modules reuse their configuration and request logic.

Webhook handling ignores reachability probes and mismatched event types, stores subscription IDs for cleanup, and safely handles deactivation when registration did not complete. Added events use resource timestamps when available; story updates use `last_edited_datetime` when present, with receipt-time fallbacks.

## Compatibility and delivery behavior

- Upgrading existing components requires reviewing their configuration: `viewId` is removed from sources, and Get Recording's legacy props are removed or renamed. Existing deployed versions remain pinned until upgraded.
- Added/deleted sources deduplicate by resource ID. Updated sources emit every delivery so successive changes to the same resource are preserved. Retries may emit duplicates because the API does not document a reliable delivery ID.
- The app package version increases from `0.1.0` to `1.0.0`, matching the breaking component changes.

## Validation

[...]

PR #21958, opened 2026-09-10
grain: AI-optimize MCP action set

### LINKED ISSUE #21463: grain: AI-optimize MCP action set
Grain has three action today — get a recording. We'd like that improved and new tools added so an agent can actually work with meetings and what's in them, and so each tool is easy for an agent to reach for and get right.

A few things an agent should be able to do, for example:
- Find recordings without an ID — e.g. "pull up my calls with Acme last week"
- Get the transcript, highlights, or summary of a meeting
- Grab the action items captured during a call
- Share a recording or a specific clip

**API reference:** https://developers.grain.com/

**Add marker:**
When a component is AI-optimized, add this to the file:
```
ai: "optimized",
```
Add it to each action file.

michelle0927:
## Summary

Extends Grain's action set (currently `get-recording`, `get-transcript`, `list-recordings` from #21865) to cover the workflows in #21463: finding recordings, pulling transcripts/highlights/summaries/action items, and sharing recordings.

**New actions:**
- `grain-list-users` — resolves workspace user IDs for sharing
- `grain-list-teams` — resolves workspace team IDs for sharing
- `grain-list-meeting-types` — lists configured meeting types (internal/external)
- `grain-update-recording` — renames a recording and/or adds/removes tags
- `grain-manage-recording-sharing` — shares or unshares a recording with a user or team
- `grain-download-recording` — downloads a recording's media file via File Stash

**Changes to existing actions:**
- All 9 actions marked `ai: "optimized"` and given a worked example in their description.
- `grain-list-recordings` gained an optional `fields` param to narrow the response (additive — omitted behaves exactly as before).

## Coverage vs. #21463

| Ask | Coverage |
|---|---|
| Find recordings without an ID | `list-recordings` (title/date/participant-scope filters — Grain's API doesn't support participant/company filtering, confirmed by testing the live API's filter schema) |
| Transcript | `get-transcript` |
| Highlights | `get-recording` (`highlights` include) |
| Summary | `get-recording` (`aiSummary` include) |
| Action items | `get-recording` (`aiActionItems` include) |
| Share a recording | `manage-recording-sharing` |
| Share a specific clip | Not implemented — Grain's v2 API has no standalone `/highlights` resource or highlight-level sharing endpoint (confirmed empirically; highlights are only reachable as a nested include on a recording) |

## Validation

- Static audit (agent-audit): 100/100 — 0 MCP-broken actions, 0 blocked props, 0 silent-truncation risk, all cross-references verified.
- 9/9 evals pass against the live API using a real connected account (list/get/discovery/write/download paths all exercised end-to-end).
- Two endpoint assumptions from initial research were wrong and corrected after live testing: tag/share **add** operations are `PUT` to the collection path with the value in the body (`/recordings/:id/tags` + `{tag}`, not `/recordings/:id/tags/:tag}`); **remove**/unshare operations are `DELETE` to the resource path as originally assumed.
- Sharing with a team the recording owner already belongs to returns `400 Invalid Team` — a real Grain API/business-rule constraint, not a bug (confirmed by testing the identical code path against a user target, which succeeds).
- `eslint --fix` passes with zero remaining errors.

Closes #21463

## Summary by CodeRabbit

* **New Features**
  * Added actions to download recordings, list workspace users, teams, and meeting types, manage recording sharing, and update recording titles and tags.
  * Recording downloads now include an appropriate filename and content type.
* **Enhancements**
  * Added team, meeting-type, and field filters when listing recordings, with support for up to 500 results.
  * Expanded examples for retrieving recordings and transcripts.
  * Improved action summaries and returned item counts.
* **Bug Fixes**
  * Prevented duplicate events from retried recording, story, and highlight updates while preserving distinct updates.

### DIFF
--- a/components/grain/actions/download-recording/download-recording.mjs
+++ b/components/grain/actions/download-recording/download-recording.mjs
@@ -1,0 +1,64 @@
+import fs from "fs";
+import grain from "../../grain.app.mjs";
+
+const EXTENSION_BY_CONTENT_TYPE = {
+  "video/mp4": "mp4",
+  "video/quicktime": "mov",
+  "audio/mpeg": "mp3",
+  "audio/mp4": "m4a",
+};
+const DEFAULT_EXTENSION = "mp4";
+
+export default {
+  key: "grain-download-recording",
+  name: "Download Recording",
+  description: "Downloads a Grain recording's media file (video or audio) and returns a presigned download URL."
+    + " Only recordings with processed media have a downloadable file — check `media_type` (`video` or `audio`, not `transcript`) from **List Recordings** or **Get Recording** first."
+    + " Use **Get Transcript** instead if you only need the spoken content, not the media file itself."
+    + " Example: `recordingId: \"8a089fcb-0961-4393-8da2-f0db5f8cfd79\"` downloads the file and returns `{\"filePath\": \"/tmp/grain-recording-8a089fcb-....mp4\", \"filename\": \"grain-recording-8a089fcb-....mp4\", \"contentType\": \"video/mp4\"}`."
+    + " [See the documentation](https://developers.grain.com/#download-recording)",
+  version: "0.0.1",
+  ai: "optimized",
+  annotations: {
+    destructiveHint: false,
+    openWorldHint: true,
+    readOnlyHint: true,
+  },
+  type: "action",
+  props: {
+    grain,
+    recordingId: {
+      propDefinition: [
+        grain,
+        "recordingId",
+      ],
+    },
+    syncDir: {
+      type: "dir",
+      accessMode: "write",
+      sync: true,
+    },
+  },
+  async run({ $ }) {
+    const response = await this.grain.downloadRecording({
+      $,
+      recordingId: this.recordingId,
+      responseType: "arraybuffer",
+      returnFullResponse: true,
+    });
+
+    const contentType = response.headers["content-type"];
+    const extension = EXTENSION_BY_CONTENT_TYPE[contentType] ?? DEFAULT_EXTENSION;
+    const filename = `grain-recording-${this.recordingId}.${extension}`;
+    const filePath = `${process.env.STASH_DIR || "/tmp"}/${filename}`;
+
+    fs.writeFileSync(filePath, Buffer.from(response.data));
+
+    $.export("$summary", `Downloaded recording ${this.recordingId} (${filename})`);
+    return {
+      filePath,
+      filename,
+      contentType,
+    };
+  },
+};
--- a/components/grain/actions/download-recording/download-recording.mjs
+++ b/components/grain/actions/download-recording/download-recording.mjs
@@ -12,7 +12,7 @@ const DEFAULT_EXTENSION = "mp4";
 export default {
   key: "grain-download-recording",
   name: "Download Recording",
-  description: "Downloads a Grain recording's media file (video or audio) and returns a presigned download URL."
+  description: "Downloads a Grain recording's media file (video or audio) and writes it to File Stash, returning `filePath`, `filename`, and `contentType`."
     + " Only recordings with processed media have a downloadable file — check `media_type` (`video` or `audio`, not `transcript`) from **List Recordings** or **Get Recording** first."
     + " Use **Get Transcript** instead if you only need the spoken content, not the media file itself."
     + " Example: `recordingId: \"8a089fcb-0961-4393-8da2-f0db5f8cfd79\"` downloads the file and returns `{\"filePath\": \"/tmp/grain-recording-8a089fcb-....mp4\", \"filename\": \"grain-recording-8a089fcb-....mp4\", \"contentType\": \"video/mp4\"}`."
--- a/components/grain/actions/list-recordings/list-recordings.mjs
+++ b/components/grain/actions/list-recordings/list-recordings.mjs
@@ -54,13 +54,13 @@ export default {
     team: {
       type: "string",
       label: "Team ID",
-      description: "Only return recordings belonging to this team. Use **List Teams** to find team IDs.",
+      description: "Only return recordings belonging to this team. Use **List Teams** to find team IDs. E.g. `a414c333-c9fe-4fdc-9131-fb31796699b2`.",
       optional: true,
     },
     meetingType: {
       type: "string",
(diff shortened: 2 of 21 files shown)


components/grain/actions/download-recording/download-recording.mjs
```diff
@@ -0,0 +1,64 @@
...
+  "video/quicktime": "mov",
+  "audio/mpeg": "mp3",
+  "audio/mp4": "m4a",
+};
+const DEFAULT_EXTENSION = "mp4";
+
+export default {
+  key: "grain-download-recording",
+  name: "Download Recording",
+  description: "Downloads a Grain recording's media file (video or audio) and writes it to File Stash, returning `filePath`, `filename`, and `contentType`."
```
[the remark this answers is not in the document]
michelle0927: The description contains the doc link https://developers.grain.com/#download-recording

components/grain/sources/common/base.mjs
```diff
@@ -50,11 +51,16 @@ export default {
...
+    // resource ID alone (there's only ever one). Updated events hash the resource
+    // ID together with the full payload: a timestamp alone doesn't work here
+    // because recording/highlight payloads carry no field that changes between
+    // updates (only story's last_edited_datetime does), so a content hash is what
+    // actually makes retries of the same update share an ID while a later,
+    // distinct update gets a new one, across all three resource types - kept to a
+    // single fixed-length digest rather than a raw ID + digest concatenation.
     const id = body.type.endsWith("_updated")
-      ? `${body.data.id}:${ts}`
+      ? crypto.createHash("md5").update(`${body.data.id}:${JSON.stringify(body.data)}`)
```
vetrivigneshwaran: Do not identify an update only by its resulting payload.

  Two separate updates can have the same payload. For example, a recording
  title can change from A to B and later return to A. The second A
  update produces the same digest as the first A update, so Pipedream
  suppresses a valid event.

  Use a stable delivery ID or an event timestamp from the webhook when
  available. If Grain supplies neither value, persist enough source state to
  distinguish separate deliveries without generating a new ID for a retry.

  As per path instructions, “The id must be stable (deterministic for the
  same real-world event across runs) and unique per event.”

Use a stable Grain event identifier for updated-event deduplication.

  Grain payloads do not provide a common event/delivery ID or event
  timestamp. Resource timestamps do not solve this for recordings and
  highlights. A persisted payload hash cannot distinguish an identical retry
  from a later identical update because both requests have the same
  observable input. Use a stable Grain event/delivery ID when available;
  otherwise do not promise deduplication for identical updates.

components/grain/actions/list-recordings/list-recordings.mjs
```diff
@@ -84,6 +113,22 @@ export default {
     $.export("$summary", `Successfully fetched ${recordings.length} recording${recordings.length === 1
       ? ""
       : "s"}`);
-    return recordings;
+
+    const fields = typeof this.fields === "string"
```
vetrivigneshwaran:
```suggestion
     let fields = this.fields;
    if (typeof fields === "string") {
        try {
          fields = JSON.parse(fields);
        } catch {
          throw new ConfigurationError("`fields` must be a JSON array of field names.");
        }
      }
      if (
        fields !== undefined
        && (
          !Array.isArray(fields)
          || fields.some((field) => typeof field !== "string" || !field)
        )
      ) {
        throw new ConfigurationError("`fields` must be an array of non-empty field names.");
      }
```

Validate fields before the first API request.

  The action fetches up to 500 recordings before it parses fields. Invalid
  JSON or a non-array value therefore wastes all API requests before the
  action fails.

  Catch JSON.parse() errors. Also require every array element to be a
  non-empty string. Perform this validation before the pagination loop.

Then remove the existing post-request parsing and validation.

  As per path instructions, “ConfigurationError is appropriate only for
  pre-call validation of user configuration mistakes.”

components/grain/actions/update-recording/update-recording.mjs
```diff
@@ -0,0 +1,106 @@
...
+    },
+    removeTags: {
+      type: "string[]",
+      label: "Remove Tags",
+      description: "Tags to remove from the recording.",
+      optional: true,
+    },
+  },
+  async run({ $ }) {
+    const addTags = typeof this.addTags === "string"
```
vetrivigneshwaran:
```suggestion
    const parseTags = (value, propName) => {
        let tags = value;
        if (typeof tags === "string") {
          try {
            tags = JSON.parse(tags);
          } catch {
            throw new ConfigurationError(
              `\`${propName}\` must be a JSON array of tag names.`,
            );
          }
        }
        if (
          tags !== undefined
          && (
            !Array.isArray(tags)
            || tags.some((tag) => typeof tag !== "string" || !tag)
          )
        ) {
          throw new ConfigurationError(
            `\`${propName}\` must be an array of non-empty tag names.`,
          );
        }
        return tags;
      };
      
      const addTags = parseTags(this.addTags, "addTags");
     const removeTags = parseTags(this.removeTags, "removeTags");
```

 Validate every tag before applying any mutation.

  Array.isArray() accepts values such as [{}], [null], and [""]. If
  title is also supplied, Line 62 updates the title before the invalid tag
  request fails. The action then leaves a partial update while reporting an
  error.

  Catch malformed JSON. Require each tag to be a non-empty string. Complete
  both validations before the title request.

components/grain/grain.app.mjs
```diff
@@ -138,6 +138,128 @@ export default {
         ...opts,
       });
     },
+    /**
+     * List workspace users.
```
vetrivigneshwaran, line 122: The shared `recordingId` propDefinition retains an `async options()` dropdown, and it is consumed by five AI-optimized actions (get-recording, get-transcript, download-recording, update-recording, manage-recording-sharing). The house rules for AI-optimizing the MCP action set require removing ALL async-options dropdowns and relying on discovery tools instead. That discovery tool already exists (List Recordings) and every consuming action's description already says "Use **List Recordings** to find recording IDs." The prop should become a plain string prop.

vetrivigneshwaran (changes requested):

vetrivigneshwaran (changes requested):
@michelle0927 kindly resolve the merge conflict and revert back

vetrivigneshwaran (approved):
LGTM

PR #21959, opened 2026-09-10
feat(google_docs): support multi-tab documents across the action set

dylburger:
## What & why

A Google Doc can hold several tabs. Two facts about the Docs API make tabs an all-or-nothing thing for a tool set:

1. Any edit whose location carries no `tabId` is applied to the document's **first** tab.
2. `documents.get` without `includeTabsContent` returns only the first tab's body **and omits `tabs` entirely**.

None of the write actions passed a `tabId`, and nothing in the set could report a document's tabs. Through the MCP surface that added up to exactly the report we got from a Conduit user: *"I can read through multiple tabs, but it can't create new tabs or write to a particular tab — any write request goes to tab 0."* A multi-tab document also read back indistinguishably from a single-tab one, so the missing content was invisible rather than obviously missing.

This PR adds tab discovery, tab creation, and tab targeting on every action that reads or writes content.

### Measured first, against the live API

Docs discovery rev `20260902`. Each of these was verified with a real document before any code was written, because the public answers on tabs are out of date:

| Behavior | Result |
|---|---|
| `insertText` with no `tabId` on a 3-tab doc | lands in the first tab (`t.0`) |
| `insertText` with `endOfSegmentLocation.tabId` | lands in the named tab |
| `insertText` with an unknown `tabId` | **HTTP 400** `"Cannot apply request to an invalid tab ID"` — no silent fallback |
| `replaceAllText` with no `tabsCriteria` | replaces in **every** tab |
| `documents.get?includeTabsContent=false` | no `tabs` field; `body` is the first tab only |
| `addDocumentTab` | **supported** — returns the generated `tabId`. Contrary to the widely-indexed answers saying tabs are UI-only, `addDocumentTab` / `deleteTab` / `updateDocumentTabProperties` all exist now |
| `addDocumentTab` with a caller-chosen `tabId` | 400 — `"Tab ID should not be specified in the properties when adding a tab"` |

The response shape under `includeTabsContent=true` is `{title, revisionId, suggestionsViewMode, documentId, tabs[]}`, with `body`/`documentStyle`/`namedStyles` moved under `tabs[].documentTab` — which is what makes the compatibility work in `get-document` below necessary.

## Changes (per component)

**New tools**

- **`list-tabs`** (new, `0.0.1`) — returns `{documentId, tabCount, tabs: [{tabId, title, index, nestingLevel, parentTabId}]}` in document order, each parent followed by its nested children. This is the missing link: without it no caller could obtain the `tabId` that every tab-targeted edit needs. (The old `get-document` error message even advised calling it "without a Tab ID to list the document's tabs" — a response that never contained any.)
- **`create-tab`** (new, `0.0.1`) — wraps `addDocumentTab` and returns the generated `tabId`, ready to pass straight to the editing tools. Optional **Position** (zero-based, among siblings) and **Parent Tab ID** (to nest a sub-tab).

**Reads**

- **`get-document`** (`1.1.5` → `1.2.0`, minor) — now one tabs-aware read. The first tab is merged back up to the top level, so `body`, `textContent`, `documentStyle`, `namedStyles`, `revisionId`, `title` and `documentId` are exactly the field set callers have always received; `tabCount` and `tabs` are added alongside, with each tab's own `textContent` when the document has more than one — so a single call surfaces all of the document's text. A single-tab document's response is unchanged apart from those two additive keys. A `tabId` read is enriched with `textContent`, and now searches nested child tabs instead of only root tabs (previously a valid nested `tabId` errored). The `fields` mask path is untouched.

**Writes — optional `Tab ID`, omitted = first tab, exactly as before**

- **`insert-text`** (`1.0.6` → `1.1.0`), **`insert-table`** (`1.0.6` → `1.1.0`), **`insert-page-break`** (`1.0.6` → `1.1.0`), **`insert-image`** (`0.0.7` → `0.1.0`) — minor. The tab is threaded into the request's `Location` / `EndOfSegmentLocation` via the shared position builders. A tab-targeted write now returns **the tab it wrote** rather than the document: returning the first tab's content after editing another tab made a successful edit read as if nothing had happened.
- **`write-table`**, **`delete-table`** (`0.0.4` → `0.1.0`, minor) — same Tab ID support. Both also read the document to locate a table (`write-table` re-reads to find the table it just inserted before filling cells; `delete-table` counts tables to resolve **Table Number**), so those reads are now scoped to the target tab instead of always the first one. Both are additionally marked **`ai: "optimized"`**: without that marker the MCP surface never exposed them at all, which eval #26 caught in the most direct way possible — asked to write a table into a tab, the agent couldn't find `write-table`, hand-rolled `insert-table` plus eight index-guessing `insert-text` calls, and took an HTTP 400 on a bad index. With the tool visible the same task is 3 calls and no errors.
- **`replace-text`** (`1.1.2` → `1.2.0`, minor) — optional Tab ID confines the replacement via `tabsCriteria`. Omitted still replaces across every tab, which is the API's own default and the right one for find-and-replace, so existing behavior is untouched. Both the plain and the Markdown paths are scoped.

**Unchanged behavior**

- **`format-text`**, **`format-paragraph`**, **`format-table-cell`**, **`apply-bullets`**, **`remove-bullets`** (`0.0.2` → `0.0.3`, patch) — no behavior change. They already accepted a tab; they inherit the shared `styleTabId` prop, whose description now points at **List Tabs** instead of Get Document. Version-bumped because the description they render changed.
- **`google_docs.app.mjs`** — tab helpers (`listTabs`, `addTab`, `_findTab`, per-tab body reads, the tabs-aware document read) plus one error-quality fix: every `batchUpdate` rewrites the API's bare `"Cannot apply request to an invalid tab ID"` — which names neither the document nor the IDs that would have worked — into a message pointing at **List Tabs**. `_findTab` errors list the tabs that do exist.

### Two prop definitions, on purpose

`contentTabId` (writes) and `styleTabId` (styling/replace) are deliberately separate, because omitting them means different things: the Docs API sends an untargeted `Location` to the **first tab**, while an untargeted `replaceAllText`/style search covers **every tab**. One shared prop would have to describe both, and the descriptions are what the model reasons from.

### Deliberately not included

Renaming, reordering and deleting an existing tab (`updateDocumentTabProperties`, `deleteTab`). The API supports all three; they're out of scope here. The tab tools' descriptions state the boundary explicitly, because an agent that assumes it can rename will say so — measured: before that line, asked to rename tabs, the model replied *"renaming a tab is straightforward"* and then asked which documents. Eval #27 (abstain) locks the corrected behavior in.

## Verification

- **Suite:** `pd-connect-eval-monster/evals/google_docs` — 27 evals (19 multi-step, 4 negative), **27/27 on Sonnet 5** (1 trial), tool precision/recall/F1 all `1.00`, restraint 4/4. Report: `2026-09-10T22-44-12-606Z.report.md`. No open failures.
- **8 new tab evals (20–27).** The important one is **#22**, which asserts goal state against the live document rather than trusting the transcript: after the agent adds a line to the named tab, `tabs.1.documentTab.body` **contains** it and `tabs.0.documentTab.body` **does not**. That pair of assertions is the reported bug, expressed as a test — it fails on the old code and passes here. `expected_args` additionally requires the insert call to carry a `tabId` at all.
- Others: #20 tab listing, #21 reading a named tab (the answer exists only in the third tab), #23 tab creation, #24 create-a-tab-then-write-into-it end to end, #25 replace scoped to one tab with the other left alone, #26 table into a named tab, #27 abstain on rename/reorder.
- The multi-tab fixture is seeded through the Docs API (`addDocumentTab`) hung off a Drive-created document, so it stays cleanable; tab IDs can't be caller-specified, so the seed reads them back before filling each tab.
- Every component was also exercised directly against the live API outside the eval harness (21 checks: tab creation incl. nesting and position, per-tab writes, cross-tab isolation, tab-scoped replace and delete, the enriched bad-`tabId` errors, and single-tab backwards compatibility).

## Summary by CodeRabbit

- **New Features**
  - Added support for viewing, creating, and managing tabs in Google Docs.
  - Added tab-specific targeting for text, tables, images, page breaks, formatting, bullets, and replacements.
  - Replace Text can target a specific tab or all tabs.
  - Added tab-aware document content, image handling, operation results, and clearer summaries.

- **Bug Fixes**
  - Improved error messages for invalid tab selections.


components/google_docs/google_docs.app.mjs
```diff
@@ -127,7 +127,17 @@ export default {
...
       label: "Tab ID",
-      description: "For a multi-tab document, restrict the operation to this tab (e.g. `t.0`). Copy it from **Get Document**. Omit to search every tab.",
+      description: "For a multi-tab document, restrict the operation to this tab (e.g. `t.0`). Get tab IDs from **List Tabs**. Omit to apply the operation to every tab.",
+      optional: true,
+    },
+    // Tab selector for content writes (insert/delete). Deliberately separate
+    // from `styleTabId`: omitting this one means "the document's first tab", not
+    // "every tab", because the Docs API applies an untargeted `Location` /
+    // `EndOfSegmentLocation` to the first tab only (measured).
+    contentTabId: {
```
dylburger, line 163: **Why:** two prop definitions rather than one, because omitting a tab means opposite things depending on the request. The Docs API applies an untargeted `Location` / `EndOfSegmentLocation` to the **first tab**, but an untargeted `replaceAllText` (and the style searches) covers **every tab** — both measured against a live 3-tab document. **Change:** `contentTabId` for content writes (omitted = first tab) stays separate from the existing `styleTabId` (omitted = every tab). **Result:** each tool description can state its own default truthfully, which is what the model actually reasons from. A single shared prop would have to describe both defaults at once.

components/google_docs/google_docs.app.mjs
```diff
@@ -193,39 +214,60 @@ export default {
...
-              [requestName]: request,
-            },
-          ],
-        },
-      });
+    // An unknown tab fails with a bare "Cannot apply request to an invalid tab
+    // ID" (measured) that names neither the document nor the IDs that would have
+    // worked, so every batchUpdate rewrites it into something the caller can act
+    // on. Anything else is rethrown untouched.
+    _rethrowBatchUpdateError(error, documentId) {
```
dylburger, line 261: **Why:** an unknown tab fails with a bare `"Cannot apply request to an invalid tab ID"` (measured — it 400s rather than silently falling back to tab one, which is the good news). But that message names neither the document nor the IDs that would have worked, so an agent has nothing to correct toward. **Change:** every `batchUpdate` rewrites it into a `ConfigurationError` pointing at **List Tabs**. **Result:** the recovery path is one call away instead of guesswork. Non-tab errors are rethrown untouched.

components/google_docs/google_docs.app.mjs
```diff
@@ -266,6 +308,132 @@ export default {
...
+      const document = await this.getDocument(documentId, true);
+      return this._findTab(document, tabId).documentTab?.body?.content;
+    },
+    // One tabs-aware read that keeps the historical response shape. Fetching
+    // with `includeTabsContent` moves the content under `tabs[].documentTab` and
+    // drops the top-level `body`/`documentStyle`/`namedStyles`, so the first tab
+    // is merged back up to where callers have always found it and the tab list is
+    // added alongside. Without this, a multi-tab document read as `body` alone
+    // looked exactly like a single-tab document — the other tabs were invisible.
+    async getDocumentWithTabs(documentId) {
```
dylburger, line 416: **Why:** `documents.get` with `includeTabsContent=true` moves the content under `tabs[].documentTab` and drops the top-level `body`/`documentStyle`/`namedStyles` — so simply flipping the flag would have broken every existing consumer of this app's reads. **Change:** fetch with tabs content, then merge the first tab back to the top level. Verified against the API that the merged field set is exactly the historical one (`title, body, documentStyle, namedStyles, revisionId, suggestionsViewMode, documentId`). **Result:** `tabs` becomes visible without a breaking change; a single-tab document's response gains only `tabCount` and `tabs`.

components/google_docs/actions/get-document/get-document.mjs
```diff
@@ -41,16 +42,29 @@ export default {
...
 
-    const response = await this.googleDocs.getDocument(this.documentId, false, this.fields);
-    $.export("$summary", `Retrieved document ${this.documentId}`);
+    if (this.fields) {
+      const response = await this.googleDocs.getDocument(this.documentId, false, this.fields);
+      $.export("$summary", `Retrieved document ${this.documentId}`);
+      return response;
+    }
+
+    const response = await this.googleDocs.getDocumentWithTabs(this.documentId);
```
dylburger, line 56: **Why:** with `includeTabsContent=false` the API returns the first tab's body and **no `tabs` field at all** — so a multi-tab document read back indistinguishably from a single-tab one, and the other tabs' text was invisible rather than obviously missing. The old not-found error even told callers to call this tool "without a Tab ID to list the document's tabs", a response that never contained any. **Change:** the default path now returns the document plus `tabCount` and a `tabs` list carrying each tab's own `textContent` (multi-tab only), so one call shows all of the document's text. The `tabId` path is enriched with `textContent` and now searches nested child tabs, not just root tabs. The `fields` mask path is untouched. **Result:** eval #20 (list a doc's tabs) and #21 (read a named tab, whose answer exists only in the third tab) pass. Minor bump — additive on both paths.

components/google_docs/actions/insert-text/insert-text.mjs
```diff
@@ -31,13 +31,22 @@ export default {
         "position",
       ],
     },
+    tabId: {
```
dylburger, line 34: **Why:** this is the reported bug. Every location in a `batchUpdate` takes an optional `tabId`, and without one the API writes to the document's **first** tab — so through the MCP surface a user could read tab 3 but never write to it. **Change:** optional **Tab ID**, threaded into `Location` / `EndOfSegmentLocation` by the shared position builders. Omitted = first tab, byte-identical to before. A tab-targeted write also returns the tab it wrote, because handing back the first tab's content after editing another tab made a successful edit read as if nothing had happened. **Result:** eval #22 asserts goal state on the live document — the line is in `tabs.1.documentTab.body` and **not** in `tabs.0` — plus an `expected_args` check that the call carried a `tabId` at all. Minor: the new return shape is only reachable by opting into the new prop.

components/google_docs/actions/insert-table/insert-table.mjs
```diff
@@ -38,14 +38,23 @@ export default {
         "position",
       ],
     },
+    tabId: {
```
dylburger, line 41: **Why:** same first-tab default as Insert Text — an untargeted insert cannot reach any other tab. **Change:** optional **Tab ID** through the shared position builder; omitted = first tab, unchanged. **Result:** covered by eval #26 (structured content into a named tab). Minor (additive prop).

components/google_docs/actions/insert-image/insert-image.mjs
```diff
@@ -31,13 +31,22 @@ export default {
         "position",
       ],
     },
+    tabId: {
```
dylburger, line 34: **Why:** same first-tab default. **Change:** optional **Tab ID** through the shared position builder; omitted = first tab, unchanged. **Result:** verified directly against the live API (image inserted into a named tab, absent from the default tab). Minor (additive prop).

dylburger (commented):
Per-change rationale inline — each comment states what motivated the change, what changed, and how it was confirmed. API behaviors cited as "measured" were verified against a live multi-tab document before the change was written (Docs discovery rev `20260902`). Eval suite and report: `pd-connect-eval-monster/evals/google_docs` (27/27 on Sonnet 5).

components/google_docs/actions/insert-page-break/insert-page-break.mjs
```diff
@@ -26,11 +26,20 @@ export default {
         "position",
       ],
     },
+    tabId: {
```
dylburger, line 29: **Why:** same first-tab default. **Change:** optional **Tab ID** through the shared position builder; omitted = first tab, unchanged. **Result:** verified directly against the live API. Minor (additive prop).

components/google_docs/actions/replace-text/replace-text.mjs
```diff
@@ -80,11 +100,16 @@ export default {
         text: find,
         matchCase,
       },
+      ...(tabIds && {
+        tabsCriteria: {
```
dylburger, line 104: **Why:** `replaceAllText` with no `tabsCriteria` replaces in **every** tab (measured: 2 occurrences across 2 tabs changed in one call). That is the right default for find-and-replace, so it is preserved — but it was also the only behavior available, and "replace this in the Appendix tab only" was impossible. **Change:** an optional Tab ID narrows the request via `tabsCriteria`; both the plain and the Markdown paths are scoped. This prop reuses `styleTabId`, not `contentTabId`, precisely because omitting it means "every tab" here. **Result:** eval #25 passes — the Appendix tab reads FINAL while the first tab still reads DRAFT. Minor (additive prop, default behavior unchanged).

components/google_docs/actions/write-table/write-table.mjs
```diff
@@ -4,13 +4,14 @@ import googleDocs from "../../google_docs.app.mjs";
...
+  description: "Create a table and fill it with data in a single step. Provide the entire table (all rows and columns, including a header row) at once. Use this instead of inserting cells or text one at a time. The action places every value in the correct cell for you. Pass **Table Data** as a JSON array of arrays, one inner array per row, header row first when **Has Header Row** is set, e.g. `[[\"Name\",\"Role\"],[\"Ada\",\"Engineer\"]]`. Use **Find Document** to resolve a document's name to its ID, or **Insert Table** instead if you only need an empty grid to fill in later. In a multi-tab document, set **Tab ID** to choose which tab receives the table — without it the table goes into the document's first tab; use **List Tabs** to get the IDs. Note: a table written this way is always static — the Google Docs API does not create or preserve a live link to a Google Sheet, even when this replaces a Sheets-linked table. [See the documentation](https://developers.google.com/docs/api/reference/rest/v1/documents/request#InsertTableRequest)",
+  version: "0.1.0",
   annotations: {
     destructiveHint: false,
     openWorldHint: true,
     readOnlyHint: false,
   },
+  ai: "optimized",
```
dylburger, line 14: **Why:** two separate defects. (1) This action was published but **invisible to MCP** — Pipedream's MCP server only returns actions marked `ai: "optimized"`, and eval #26 caught the consequence in the sharpest possible way: asked to write a table into a tab, the agent could not find Write Table, hand-rolled `insert-table` plus **eight** index-guessing `insert-text` calls, and took an HTTP 400 on a bad character index. (2) It reads the document twice to locate the table it just inserted, and both reads were first-tab-only, so a tab-targeted write could not find its own table. **Change:** added the marker, and routed both reads through the tab-scoped body reader; the cell-fill inserts and header `updateTextStyle` ranges carry the tab. **Result:** #26 is now 3 calls with no errors, and the judge confirms the table is in the Q3 tab and not in the first tab. Minor.

components/google_docs/actions/delete-table/delete-table.mjs
```diff
@@ -23,21 +24,31 @@ export default {
...
+    },
   },
   async run({ $ }) {
-    const { body } = await this.googleDocs.getDocument(this.documentId, false, "body");
-    const tables = this.googleDocs.flattenTables(body?.content);
+    const target = this.tabId
+      ? `tab ${this.tabId} of document ${this.documentId}`
+      : `document ${this.documentId}`;
+    const tables = this.googleDocs.flattenTables(
+      await this.googleDocs.getTabBodyContent(this.documentId, this.tabId),
```
dylburger, line 44: **Why:** **Table Number** is resolved by counting tables in the document, and that count came from a first-tab-only read — so on a multi-tab document "delete table 1" counted the wrong tabs tables and deleted from the wrong place. (This action was also invisible to MCP for lack of the `ai: "optimized"` marker, added here for the same reason as Write Table.) **Change:** the count and the `deleteContentRange` range are both scoped to the target tab, and the error messages name the tab. **Result:** verified live — deleting table 1 of a named tab leaves the first tab's table intact, and a tab with no tables now errors with the tab named instead of silently acting elsewhere. Minor.

components/google_docs/actions/format-text/format-text.mjs
```diff
@@ -9,7 +9,7 @@ export default {
   key: "google_docs-format-text",
   name: "Format Text",
   description: "Apply character formatting (bold, italic, underline, strikethrough, font, size, color, link) to text in a Google Doc. Locate the text with **Find Text**, or pass an explicit **Start Index** and **End Index**. Use **Find Document** to resolve a document's name to its ID. [See the documentation](https://developers.google.com/workspace/docs/api/reference/rest/v1/documents/request#UpdateTextStyleRequest)",
-  version: "0.0.2",
+  version: "0.0.3",
```
dylburger, line 12: **Why:** no behavior change in this file (or in format-paragraph / format-table-cell / apply-bullets / remove-bullets). They already accepted a tab, and they inherit the shared `styleTabId` prop whose **description** changed: it used to say "copy it from Get Document", which no longer describes the best route now that List Tabs exists. **Change:** version bump only, because the description these tools render to the user changed. **Result:** consistent guidance across the tab-aware tools. Patch bump per the component guidelines (copy/description tweak).

components/google_docs/google_docs.app.mjs
```diff
@@ -127,7 +127,17 @@ export default {
     styleTabId: {
       type: "string",
       label: "Tab ID",
-      description: "For a multi-tab document, restrict the operation to this tab (e.g. `t.0`). Copy it from **Get Document**. Omit to search every tab.",
+      description: "For a multi-tab document, restrict the operation to this tab (e.g. `t.0`). Get tab IDs from **List Tabs**. Omit to apply the operation to every tab.",
```
ashwins01: Can we modify this description ? The documentation says if tab-id is omitted, the formatting is applied only to the first tab and not to all tabs.


```suggestion
      description: "For a multi-tab document, restrict the operation to this tab (e.g. `t.0`). Get tab IDs from **List Tabs**. Omit to search every tab for **Find Text**; an explicit **Start Index** / **End Index** without a Tab ID applies to the first tab only.",
```

components/google_docs/actions/replace-image/replace-image.mjs
```diff
@@ -4,7 +4,7 @@ export default {
   key: "google_docs-replace-image",
   name: "Replace Image",
   description: "Replace image in a existing document. [See the documentation](https://developers.google.com/docs/api/reference/rest/v1/documents/request#ReplaceImageRequest)",
-  version: "0.0.17",
+  version: "0.0.18",
```
ashwins01: Should we also add multi-tab support for the `replace-image` action ?

components/google_docs/actions/list-tabs/list-tabs.mjs
```diff
@@ -0,0 +1,35 @@
+import googleDocs from "../../google_docs.app.mjs";
+
+export default {
+  key: "google_docs-list-tabs",
+  name: "List Tabs",
+  description: "List the tabs in a Google Doc — each tab's ID, title, position and nesting. A Google Doc can hold many tabs, and every editing tool writes to the **first** tab unless it is given a **Tab ID**, so call this first whenever the job names a tab (\"add this to the Budget tab\") or the document might have more than one. Returns `{documentId, tabCount, tabs: [{tabId, title, index, nestingLevel, parentTabId}]}`, in document order, with each parent followed by its nested child tabs. Pass a `tabId` from this list to **Get Document** to read one tab, or to **Insert Text**, **Write Table**, **Insert Image**, **Insert Table**, **Insert Page Break**, **Delete Table** or **Replace Text** to edit one. Use **Find Document** to resolve a document's name to its ID. This tool set can list tabs, add a tab (**Create Tab**) and read or write a named tab; it cannot rename, reorder or delete an existing tab — say so plainly rather than attempting a workaround. [See the documentation](https://developers.google.com/workspace/docs/api/how-tos/tabs)",
```
ashwins01:
```suggestion
  description: "List the tabs in a Google Doc — each tab's ID, title, position and nesting. A Google Doc can hold many tabs, and every editing tool writes to the **first** tab unless it is given a **Tab ID**, so call this first whenever the job names a tab (\"add this to the Budget tab\") or the document might have more than one. Returns `{documentId, tabCount, tabs: [{tabId, title, index, nestingLevel, parentTabId}]}`, in document order, with each parent followed by its nested child tabs. Pass a `tabId` from this list to **Get Document** to read one tab, or to **Insert Text**, **Write Table**, **Insert Image**, **Insert Table**, **Insert Page Break**, **Delete Table** or **Replace Text** to edit one. Use **Find Document** to resolve a document's name to its ID. This tool set can list tabs, add a tab (**Create Tab**) and read or write a named tab; it cannot rename, reorder or delete an existing tab — say so plainly rather than attempting a workaround. [See the documentation](https://developers.google.com/workspace/docs/api/reference/rest/v1/documents/get#query-parameters)",
```

components/google_docs/google_docs.app.mjs
ashwins01: Should we change this guidance to use `List Tabs` instead ?

ashwins01 (changes requested):
LGTM overall, a couple of comments on modifying the descriptions and adding multi-tab support to the `Replace Image` action.

ashwins01 (approved):
LGTM, ready for QA!

PR #21960, opened 2026-09-11
feat(google_sheets): AI-optimized formatting and layout tools for MCP

dylburger:
## What & why

Google Sheets has no formatting tool on the MCP surface. Today an agent can create a
spreadsheet and edit its values (`new-spreadsheet`, `add-rows`, `update-multiple-rows`,
`delete-rows` are all `ai: "optimized"`), but it cannot bold a header, apply a currency
format, freeze a row or resize a column. The reported workaround was to generate an XLSX
elsewhere, style it there, and upload it to Drive.

These four tools close that gap so a sheet can be built **and styled in place**.

### Why not promote the existing formatting actions

The monorepo already has `update-formatting`, `merge-cells`,
`add-conditional-format-rule` and friends. None is `ai: "optimized"`, so MCP never
serves them — and none can be promoted as-is:

| Action | Blocker |
|---|---|
| `update-formatting` | Six `type: "string"` props for two colors (`backgroundColorRedValue: "0.85"`, three times over). No number formats, no vertical alignment, no wrapping. Worksheet comes from an `async options()` dropdown, which is invisible over MCP. |
| `merge-cells` | Numeric `worksheetId` + raw zero-based grid indices. |
| `add-conditional-format-rule` | Same RGB-float and dropdown problems; rule types as raw API enums. |

An agent cannot express *"bold with a light gray background"* against that schema. These
ship as **new keys with flat, agent-shaped inputs**. The legacy actions are left
untouched on disk, so no existing workflow changes behavior.

## Changes (per component)

- **`format-cells`** (new, `0.0.1`) — one tool for everything that applies to a cell
  range: bold/italic/underline/strikethrough, font size and family, text and background
  color, both alignments, wrap strategy, number formats, borders, merge/unmerge, and
  clear. Colors accept a hex code **or** a common name (`#1155cc`, `light gray`); number
  formats accept named presets (`currency_usd`, `percent_1dp`, `date`) with
  `numberFormatPattern` as the escape hatch. All attributes go into **one**
  `batchUpdate`, and the `repeatCell` field mask is assembled only from the props
  actually supplied — so setting `bold` cannot wipe an existing fill.

- **`format-worksheet`** (new, `0.0.1`) — freeze rows/columns, auto-resize columns to
  fit content, explicit column widths / row heights, hide gridlines, tab color, rename.
  Auto-resize is the highest-value item: it's what turns an imported sheet from a wall
  of `####` into something readable.

- **`add-conditional-formatting`** (new, `0.0.1`) — data-driven styling with
  agent-shaped condition names (`NUMBER_GREATER`, `TEXT_CONTAINS`, `IS_BLANK`,
  `CUSTOM_FORMULA`, `COLOR_SCALE`) mapped onto the API's enums, plus a per-condition
  value-arity check that fails with the expected count instead of a 400. New key
  because `google_sheets-add-conditional-format-rule` is taken and registry keys can
  never be reused.

- **`get-cell-formatting`** (new, `0.0.1`) — the read counterpart. Without it formatting
  is write-only: an agent can't answer "is the header bold?" or verify its own work.
  Reports colors as **hex, not `{red,green,blue}` floats**, so it speaks the same
  vocabulary the write tools accept. Attributes at their Google default are omitted, and
  the range is capped at 500 cells with a `truncated` flag.

- **`common/format-utils.mjs`** (new) — shared A1-range parser, color parser and
  number-format table. Kept separate because this is where the subtle bugs live: the
  API's end indices are **exclusive**, whole-column (`B:B`) and whole-row (`2:2`) ranges
  are unbounded, and columns past Z are base-26. The existing app-file
  `_parseRangeString` handles none of these (it returns `NaN` for `B:B`), which is why
  this doesn't reuse it.

Worksheets are addressed **by name** throughout rather than by numeric `sheetId`, and an
unknown name throws listing the names that *do* exist — `sheetName` is the input an agent
is most likely to get wrong, so a recoverable error message earns its keep there.

App `package.json` bumped `1.0.2` → `1.1.0` (minor; additive only).

## Verification

- **49 pure-function checks** on `format-utils.mjs` before any API call — covering
  exclusive end indices, `A1` / `A1:F1` / `B:B` / `2:2` / `A2:F` / `AA1:AB2`, reversed
  and absolute ranges, the Z→AA→ZZ→AAA column boundaries, hex/named/3-digit colors and
  the hex→float→hex round-trip, number-format presets and clearing, border presets, and
  unbounded-range clamping.
- **Eval suite** `pd-connect-eval-monster/evals/google_sheets` — 30 evals (2 multi-tool
  workflows), **27/30 on Sonnet 5** under the `claude_code_default` harness profile at
  `EVAL_CONCURRENCY=1`, 1 trial. **All 14 new formatting evals pass**, including the
  deliberate edge cases: exclusive-end header row, number-format preset disambiguation
  (`percent` vs `percent_1dp`), merge, the `COLOR_SCALE` branch, hex-not-floats
  read-back, clear-formatting-without-losing-values, field-mask preservation, and
  multi-letter columns. Two negative evals confirm the new tools don't over-trigger on a
  chart request or a vague "make it look nicer".
- The 3 non-passing evals are **pre-existing and unrelated to this change**: #4 is a
  route choice (`read-rows` instead of `find-rows`; answer verified correct), #10 also
  failed in the 2026-07-27 run before this work, and #16 is restraint variance on the
  sharing gap. Re-running #5/#6/#10 under the `tools_only` control profile gives 3/3,
  which localises them to discoverability of the *existing* read tools under tool
  search — filed as follow-up, not touched here.

## Summary by CodeRabbit

* **New Features**
  * Added Google Sheets actions for creating conditional-formatting rules, formatting cells, and adjusting worksheet layout.
  * Added support for colors, number formats, borders, alignment, wrapping, merging, sizing, frozen panes, gridlines, and tab colors.
  * Added an action for inspecting cell formatting and worksheet layout, with optional cell-level details.
  * Added validation and support for common range, color, border, and formatting presets.


components/google_sheets/actions/format-cells/format-cells.mjs
```diff
@@ -0,0 +1,443 @@
...
+    }
+
+    if (fields.length) {
+      requests.push({
+        repeatCell: {
+          range: gridRange,
+          cell: {
+            userEnteredFormat,
+          },
+          fields: fields.join(","),
```
dylburger, line 379: **Why this is the load-bearing line.** `repeatCell` overwrites every path named in `fields` — and *clears anything it names but doesn't supply*. A blanket `fields: "userEnteredFormat"` is the natural-looking implementation and it silently destroys formatting the user never mentioned: ask to bold a highlighted cell and the highlight disappears.

**Change:** the mask is accumulated prop-by-prop as the cell format is built (`setText` pushes to both), so only attributes actually passed are ever in the mask.

**Pinned by eval #28**, which bolds a range that arrives with a light-yellow fill and asserts the fill survived — on the *green* channel, since `#fff2cc` and white share `red: 1`. The prompt deliberately doesn't mention the highlight, so the model can't paper over a broken mask by re-applying it by hand.

components/google_sheets/actions/format-cells/format-cells.mjs
```diff
@@ -0,0 +1,443 @@
...
+      type: "boolean",
+      label: "Clear Formatting",
+      description: "Set `true` to reset the range to default formatting"
+        + " (removes bold, colors, borders and number formats). Applied before"
+        + " any other options in the same call, so you can clear and restyle"
+        + " in one step.",
+      optional: true,
+    },
+  },
+  async run({ $ }) {
```
dylburger, line 248: **One `batchUpdate` for the whole call.** `repeatCell`, `updateBorders`, `mergeCells`/`unmergeCells` and the clear request all accumulate into a single `requests` array, so the cost is one HTTP round trip whether the caller sets one attribute or fifteen. The eval framework flags any tool call over 5s; measured calls here run ~2s including the `sheetName`→`sheetId` resolution.

Ordering is deliberate: `clearFormatting` is pushed **first**, so "strip this and restyle it" is one call rather than two.

components/google_sheets/actions/format-cells/format-cells.mjs
```diff
@@ -0,0 +1,443 @@
...
+
+const MERGE_OPTIONS = [
+  "MERGE_ALL",
+  "MERGE_COLUMNS",
+  "MERGE_ROWS",
+  "UNMERGE",
+];
+
+export default {
+  key: "google_sheets-format-cells",
```
dylburger, line 26: **Why `merge` lives here rather than in its own tool.** Merging is a range-level presentational operation on the same `(spreadsheetId, sheetName, range)` triple as every other option. Tool *names* are the retrieval key under tool search, so each extra near-synonym tool makes the useful one harder to find — `MERGE_ALL`/`MERGE_COLUMNS`/`MERGE_ROWS`/`UNMERGE` as one prop keeps the title-banner case (eval #22) on the tool the model is already looking at.

New key rather than an edit to `update-formatting`: this consolidates four legacy operations, and registry keys can never be reused.

components/google_sheets/actions/format-worksheet/format-worksheet.mjs
```diff
@@ -0,0 +1,344 @@
...
+          startIndex: 0,
+          endIndex: sheetProps.gridProperties?.columnCount ?? 26,
+        }
+        : {
+          sheetId,
+          dimension: "COLUMNS",
+          ...parseDimensionSpec(autoResizeColumns, "COLUMNS"),
+        };
+      requests.push({
+        autoResizeDimensions: {
```
dylburger, line 315: **`autoResizeColumns` is the highest-value operation in this PR.** It's what separates a readable imported sheet from a wall of `####`, and nothing in the existing tool set can do it.

`ALL` resolves against the sheet's real `columnCount` rather than a hardcoded 26, so it works on a widened grid; it also accepts a column reference (`A:F`, `C`) through the same A1 vocabulary as the rest of the tools.

**Pinned by eval #21**, phrased as a symptom ("columns are too narrow to read everything") rather than an instruction, so the model has to recognise auto-resize as the remedy. Verified on column B's pixel width — Google's default is exactly 100px, so a real resize necessarily moves it.

components/google_sheets/actions/format-worksheet/format-worksheet.mjs
```diff
@@ -0,0 +1,344 @@
...
+    },
+    newSheetName: {
+      type: "string",
+      label: "New Worksheet Name",
+      description: "Rename the worksheet to this. Leave empty to keep the"
+        + " current name.",
+      optional: true,
+    },
+  },
+  async run({ $ }) {
```
dylburger, line 214: **All sheet-property changes collapse into one `updateSheetProperties` with a single field mask** (freeze rows, freeze columns, gridlines, tab color, rename) — same reasoning as `format-cells`: properties not named in the mask (row/column counts, an existing freeze) are left alone.

Note `freezeRows: 0` has to be distinguishable from "not supplied", which is why these are explicit `!== undefined && !== null` checks rather than truthiness. `0` is how you *unfreeze*.

components/google_sheets/actions/add-conditional-formatting/add-conditional-formatting.mjs
```diff
@@ -0,0 +1,442 @@
...
+    ? {
+      relativeDate: upper,
+    }
+    : {
+      userEnteredValue: String(value),
+    };
+}
+
+export default {
+  key: "google_sheets-add-conditional-formatting",
```
dylburger, line 134: **Why the condition names aren't the API's.** Google spells these `NUMBER_GREATER_THAN_EQ`, `BLANK`, `NOT_BLANK`; a model that reasonably guesses `NUMBER_GREATER_EQ` or `IS_BLANK` gets a 400, which is a dead end rather than a recoverable mistake. The `CONDITIONS` table owns that translation and also records each condition's required **value arity**, so a wrong count fails with "needs exactly 2 value(s), got 1" and the fix is obvious.

`COLOR_SCALE` is folded in as a condition even though it produces a `gradientRule` rather than a `booleanRule`, because from the caller's side "heat-map this column" is the same kind of request as "highlight anything over 2M". Both branches are pinned — evals #23 and #24.

components/google_sheets/actions/add-conditional-formatting/add-conditional-formatting.mjs
```diff
@@ -0,0 +1,442 @@
...
+      ],
+    };
+    const summaryParts = [];
+
+    if (condition === "COLOR_SCALE") {
+      const min = parseColor(minColor ?? "#ffffff", "minColor");
+      const max = parseColor(maxColor ?? "#57bb8a", "maxColor");
+      const mid = parseColor(midColor, "midColor");
+
+      rule.gradientRule = {
```
dylburger, line 338: `index: 0` puts the new rule **first**, so it wins over any rule already on the range. A user who asks for a highlight expects to see it; appending behind an existing rule that already matches would silently do nothing visible.

Two-color scales work by omitting `midColor` (the `midpoint` key is conditional) rather than forcing a three-stop gradient.

dylburger (commented):
Per-change rationale inline — each comment says what motivated the design and which eval pins it down. Suite and report live in `pd-connect-eval-monster/evals/google_sheets` (27/30 on Sonnet 5; all 14 formatting evals green).

The thing worth knowing up front: **worksheets are addressed by name, never by numeric `sheetId`**, and colors and number formats are expressed the way a person says them (`#1155cc` or `light gray`; `currency_usd` rather than `"$"#,##0.00`). That is the whole difference between these and the existing `update-formatting`, which spends six `type: "string"` props on two colors.

components/google_sheets/actions/get-cell-formatting/get-cell-formatting.mjs
```diff
@@ -0,0 +1,271 @@
...
+  parseA1Range,
+} from "../../common/format-utils.mjs";
+
+/**
+ * Google resolves every cell's `effectiveFormat`, so an unstyled cell still reports a
+ * font, a size and white-on-black colors. Echoing all of that back would bury the two
+ * attributes the caller actually asked about, so defaults are omitted and only
+ * deliberate formatting is reported.
+ */
+const DEFAULTS = {
```
dylburger, line 17: **Why defaults are stripped from the output.** Google resolves `effectiveFormat` for every cell, so an entirely unstyled cell still reports Arial, size 10, black on white. Echoing all of that back buries the one or two attributes the caller asked about and burns context.

Omitting defaults means **whatever comes back is formatting somebody deliberately applied** — which is also what makes the output directly gradeable: eval #25 asserts `bold: true` and `backgroundColor: "#1155cc"` against tool output, and neither would be legible in a full dump.

components/google_sheets/actions/get-cell-formatting/get-cell-formatting.mjs
```diff
@@ -0,0 +1,271 @@
...
+  if (text.fontFamily && text.fontFamily !== DEFAULTS.fontFamily) {
+    out.fontFamily = text.fontFamily;
+  }
+
+  const textColor = colorToHex(text.foregroundColor);
+  if (textColor && textColor !== DEFAULTS.textColor) {
+    out.textColor = textColor;
+  }
+
+  const backgroundColor = colorToHex(format.backgroundColor);
```
dylburger, line 61: **Colors go out as hex, not `{red, green, blue}` floats.** This is the point of the read tool: the write tools accept `#1155cc`, so the read tool has to *return* `#1155cc` — otherwise an agent can't copy a style from one range to another, or check what it just set.

Eval #25 asserts the literal string `#1155cc` against tool output, so a regression to raw floats fails loudly. That eval also caught a real bug during development — in the *fixture*, not the tool: seeding a channel as `0.333` makes Google store 84/255, which reads back `#1154cc`. The seed now uses exact `n/255` fractions.

components/google_sheets/common/format-utils.mjs
```diff
@@ -0,0 +1,521 @@
...
+
+  const range = {};
+
+  /**
+   * Assign one axis. Both bounds present is the normal case. Exactly one bound
+   * present on a two-part range means "unbounded in the other direction" — `A2:F`
+   * is everything from row 2 down, which is what an agent writes for "the data
+   * rows below the header". A single-cell range has both bounds by definition.
+   */
+  const assignAxis = (startValue, endValue, startKey, endKey) => {
```
dylburger, line 123: **Why a new A1 parser instead of the app file's `_parseRangeString`.** That one returns `NaN` for a whole-column range like `B:B`, produces an inverted range for a single cell, and its `_getColumnIndex` is only correct up to column Z. All three are silent wrong-cell bugs rather than errors.

This version handles the four forms an agent actually writes — `A1`, `A1:F1`, `B:B`, `2:2` — plus multi-letter columns, reversed ranges, `$` absolute refs, and a stray worksheet prefix. End indices are **exclusive**, which is the quirk eval #17 exists to catch: it asserts the first *and last* cell of `A1:F1` independently, because asserting only the first would pass a parser that drops the last column.

`assignAxis` is what makes `A2:F` mean "row 2 to the bottom" rather than a single row. That form wasn't in the first implementation — it was added because it's what a model naturally writes for "the data rows below the header".

49 pure-function checks cover the above, including the Z→AA→ZZ→AAA boundaries in both directions.

components/google_sheets/common/format-utils.mjs
```diff
@@ -0,0 +1,521 @@
...
+  brown: "#b45f06",
+};
+
+/**
+ * Parse a hex code or color name into the API's `{red, green, blue}` floats.
+ * @param {string} input - `#RRGGBB`, `#RGB`, `RRGGBB`, or a name from NAMED_COLORS
+ * @param {string} [propName="color"] - prop name, for the error message
+ * @returns {object} Color object with red/green/blue in [0,1]
+ */
+export function parseColor(input, propName = "color") {
```
dylburger, line 186: **The six-floats problem, solved in one place.** The existing `update-formatting` asks for `backgroundColorRedValue`, `...GreenValue`, `...BlueValue` as `type: "string"` decimals — twice over, for two colors. No agent produces that reliably from "light gray".

One prop per color, accepting `#RRGGBB`, `#RGB`, bare hex, or a short list of names. The name list is deliberately ~25 entries covering what people ask for in a spreadsheet (status reds/ambers/greens, banded header grays) rather than all 148 CSS names — and an unparseable value throws **listing the accepted names**, so the retry is informed.

components/google_sheets/common/format-utils.mjs
```diff
@@ -0,0 +1,521 @@
...
+
+/**
+ * Named number-format presets. An agent asked to "format as currency" should not
+ * have to know Google wants `"$#,##0.00"`; `numberFormatPattern` remains the escape
+ * hatch for anything not covered here.
+ *
+ * `automatic` is intentionally `null` — clearing the format means naming it in the
+ * field mask while omitting it from the cell, which resets it to the sheet default.
+ */
+export const NUMBER_FORMATS = {
```
dylburger: **Named number-format presets.** A user says "format as currency" or "show these as percentages with one decimal"; they don't say `"$"#,##0.00` or `0.0%`. The presets cover the common asks and `numberFormatPattern` keeps the full API reachable, so nothing is lost to the abstraction.

Two details worth flagging:
- `automatic` maps to **`null`**, not a pattern. Clearing a number format means naming it in the field mask while omitting the value — there is no "general" pattern to send.
- `plain_text` (`@`) is included specifically because it's the fix for Sheets auto-converting entries like `1-2` into dates — a recurring real-world complaint that otherwise has no route.

Preset granularity is pinned by eval #20: `percent`, `percent_1dp` and `percent_2dp` all exist and the prompt asks for one decimal place, so picking plain `percent` is wrong in a way a type-only assertion would miss.

components/google_sheets/common/format-utils.mjs
```diff
@@ -0,0 +1,521 @@
...
+ * agent can't recover from. This lists the names that DO exist, so the next call can
+ * be right — the single highest-value error message in these tools, because
+ * `sheetName` is the parameter an agent is most likely to guess wrong.
+ *
+ * @param {object} app - the google_sheets app instance
+ * @param {string} spreadsheetId - the spreadsheet ID
+ * @param {string} sheetName - the worksheet (tab) name
+ * @returns {Promise<object>} the matching sheet's `properties`
+ */
+export async function getSheetProperties(app, spreadsheetId, sheetName) {
```
dylburger, line 321: **Worksheets are addressed by name, and a wrong name is recoverable.** The app file's `_getSheetIdFromName` throws a bare "Sheet name not found", which an agent can do nothing with. This lists the titles that *do* exist and points at **Get Spreadsheet Info**.

That matters more here than anywhere else in the PR: `sheetName` is the parameter a model is most likely to guess wrong (it will try "Sheet1"), and the difference between a dead end and a self-corrected second call is entirely in this message. Case-insensitive match first, for the same reason.

This is also why the tools take `sheetName` rather than the numeric `sheetId` the API wants: that id appears in no UI and in no URL a user would paste.

components/google_sheets/actions/format-worksheet/format-worksheet.mjs
```diff
@@ -0,0 +1,345 @@
...
+      "$summary",
+      `Updated layout of "${newSheetName ?? sheetName}" (${changes.join(", ")})`,
+    );
+
+    return {
+      spreadsheetId,
+      sheetName: newSheetName ?? sheetName,
+      applied,
+      spreadsheetUrl:
+        `https://docs.google.com/spreadsheets/d/${spreadsheetId}/edit#gid=${sheetId}`,
```
Priyadharshan-Pdm: These spreadsheet url is repeated in all the new actions...Can we make this one a common constants ?

components/google_sheets/common/format-utils.mjs
```diff
@@ -0,0 +1,536 @@
...
+  INNER_HORIZONTAL: [
+    "innerHorizontal",
+  ],
+  INNER_VERTICAL: [
+    "innerVertical",
+  ],
+  NONE: [],
+};
+
+export const BORDER_PRESET_OPTIONS = Object.keys(BORDER_PRESETS);
```
Priyadharshan-Pdm: Can we move these kind of constants and colors to the constants.mjs file  ?
Let the functions alone be present in this utililty file

Priyadharshan-Pdm (changes requested):

Priyadharshan-Pdm (approved):

Priyadharshan-Pdm:
LGTM!!

PR #21993, opened 2026-09-15
Jira: remove dynamic props from create-issue and update-issue

### LINKED ISSUE #21992: Jira - remove dynamic props from actions
Jira components `create-issue` and `update-issue` use `additionalProps`/dynamic props. Refactor these and any other Jira actions to remove dynamic props for AI compatibility.

**Add marker:**
When a component is AI-optimized, add this as to the file:
```
ai: "optimized"
```

michelle0927:
## Summary
- `create-issue` and `update-issue` used `reloadProps` + `additionalProps` to reactively generate per-field inputs from Jira's create/edit issue metadata — a pattern that doesn't work in tools-only/agent MCP mode.
- `cloudId`, `projectId`, `issueTypeId`, `issueIdOrKey`, and `transitionId` are now plain string props whose descriptions point at the relevant lookup action (**Get Cloud ID**, **Get All Projects**, **Get Issue Types**, **Get Transitions**) instead of using async options.
- Field values (`summary`, `description`, `priority`, `labels`, etc.) are now set entirely through the existing static `additionalProperties` object prop. `description`/`environment` plain strings are auto-converted to Jira's required Atlassian Document Format.
- Removed the now-dead dynamic-field machinery (`getDynamicFields`, `formatFields`, and supporting methods/constants) and the app-level metadata calls (`getCreateIssueMetadata`, `getEditIssueMetadata`) that only existed to feed them.
- Marked both actions `ai: "optimized"`.
- Shared propDefinitions on the app (`cloudId`, `projectID`, `issueType`, `issueIdOrKey`, `transition`) are untouched, so `transition-issue`, `add-comment-to-issue`, and `update-comment` keep their existing dropdown/search UX.

Closes #21992

## Test plan
- [x] `eslint` clean on all changed files
- [x] Published `jira-create-issue`/`jira-update-issue` to a private test workspace and ran scoped evals against a live Jira Cloud site:
  - create issue via `additionalProperties` (including plain-string `description` auto-converted to ADF)
  - update issue fields via `additionalProperties` (verified the real Jira issue reflected the change)
  - transition an issue via `update-issue`'s `transitionId` (discovered via **Get Transitions**)
  - end-to-end discovery flow (create → transition) with no tool names given in the prompt

## Summary by CodeRabbit

- **Updates**
  - Jira issue creation and updates now use direct inputs for project, issue type, issue, and transition details.
  - Plain-text descriptions and environments are automatically formatted for Jira.
  - Create Issue requires at least one additional field for custom issue data.
  - Dynamic issue-field selection and metadata lookups are no longer available.
  - Jira actions and event sources received routine version updates.
  - The Jira integration package was updated to a new major release.


components/jira/actions/common/issue.mjs
```diff
@@ -1,14 +1,12 @@
...
   props: {
     app,
     cloudId: {
-      propDefinition: [
-        app,
-        "cloudId",
-      ],
+      type: "string",
+      label: "Cloud ID",
+      description: "The Jira Cloud site ID (e.g., `11223344-a1b2-3b33-c444-def123456789`). Use the **Get Cloud ID** action to look up the ID for your site.",
```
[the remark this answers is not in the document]
michelle0927: The inline definitions are intentional so that the ai optimized actions do not include props with async options.

components/jira/actions/common/issue.mjs
```diff
@@ -26,348 +24,37 @@ export default {
...
+      return Object.entries(fields).reduce((acc, [
+        key,
+        value,
+      ]) => {
+        acc[key] = adfKeys.includes(key) && typeof value === "string"
+          ? this.atlassianDocumentFormat(value) ?? null
+          : value;
+        return acc;
+      }, {});
     },
```
ashwins01, line 93: Suggest parsing `additionalProperties` without the regex: a string value is parsed only when it is JSON for an object or array (e.g. `labels: "[\"bug\"]"` typed into the object editor), and everything else passes through untouched. Verified locally against the inputs that broke in testing: `Bug: login fails`, `See https://example.com at 10:30`, `"2024"`, `true`, and nested `{"name":"High"}` / `["bug"]` values all come out correct.

One deliberate trade-off: JSON with unquoted keys inside a value (`{name: "High"}`) is no longer accepted, only standard JSON.

```suggestion
      }, {});
    },
    /**
     * Parses `additionalProperties` into Jira field values. Nested JSON strings
     * (objects/arrays) are parsed; all other values are passed through unchanged,
     * so plain text like "Bug: login fails" or "2024" is not altered.
     */
    parseFields(fields) {
      const obj = typeof fields === "string"
        ? JSON.parse(fields)
        : fields;
      return Object.fromEntries(Object.entries(obj ?? {}).map(([
        key,
        value,
      ]) => {
        if (typeof value !== "string") {
          return [
            key,
            value,
          ];
        }
        try {
          const parsed = JSON.parse(value);
          return [
            key,
            parsed !== null && typeof parsed === "object"
              ? parsed
              : value,
          ];
        } catch {
          return [
            key,
            value,
          ];
        }
      }));
    },
```

components/jira/actions/create-issue/create-issue.mjs
```diff
@@ -136,12 +70,17 @@ export default {
       params,
       data: {
         fields: {
-          ...utils.parseObject(additionalProperties),
-          ...fields,
+          ...this.formatAdfFields(utils.parseObject(additionalProperties)),
```
ashwins01: Can we use the new helper here so plain-text `summary`/`description` are no longer mangled.

```suggestion
          ...this.formatAdfFields(this.parseFields(additionalProperties)),
```

ashwins01 (commented):
One blocking issue. Now that `additionalProperties` is the only way to set `summary`/`description`, every value goes through `utils.parseObject`. That helper runs a `word:` regex over each string and then calls `JSON.parse` on it, which corrupts ordinary text:

| Input | What Jira stored / returned |
|---|---|
| summary `Bug: login fails` | `"Bug":  login fails` (action reported success) |
| description `See https://example.com/login at 10:30` | `See "https": //example.com/login at "10": 30` |
| summary `2024` | 400 `summary: Operation value must be a string` |
| any boolean field value | `TypeError: obj[key].replace is not a function` |

On master these fields were dynamic props and never went through `parseObject`, so this is a regression, and it hits exactly the kind of free text an agent is likely to send. The inline comments suggest a local helper that only parses a value when it is JSON for an object or array. `utils.mjs` is not touched in this PR, so this avoids changing it and does not disturb the other callers.

`transition-issue`, `add-comment-to-issue` and `update-comment` have the same `parseObject` problem on their own `additionalProperties` props. Out of scope here, but worth a follow-up.

components/jira/actions/update-issue/update-issue.mjs
```diff
@@ -189,11 +89,10 @@ export default {
       issueIdOrKey,
       params,
       data: {
-        fields,
+        fields: this.formatAdfFields(utils.parseObject(additionalProperties)),
         historyMetadata: utils.parseObject(historyMetadata),
         properties: utils.parse(properties),
         update: utils.parseObject(update),
```
ashwins01: Same fix for `fields`, plus a guard so `historyMetadata: {}` and `update: {}` are not sent when those props are unset (`create-issue` already does this).

```suggestion
        fields: this.formatAdfFields(this.parseFields(additionalProperties)),
        historyMetadata: historyMetadata && utils.parseObject(historyMetadata),
        properties: utils.parse(properties),
        update: update && utils.parseObject(update),
```

Separately, on line 78: `reduceProperties` drops `false`, so `notifyUsers: false` never reaches the API (a live run with the box unchecked sent `params: {}`, meaning notifications cannot be turned off). axios already omits `undefined` query params, so the flags can be passed straight through:

```js
    const params = {
      notifyUsers,
      overrideScreenSecurity,
      overrideEditableFlag,
    };
```

`create-issue` has the same issue with `updateHistory: false` at lines 61-65.

components/jira/actions/create-issue/create-issue.mjs
```diff
@@ -24,72 +24,21 @@ export default {
...
-        }) => ({
-          cloudId,
-          projectId,
-        }),
-      ],
+      type: "string",
+      label: "Issue Type",
+      description: "An ID identifying the type of issue to create. Use the **Get Issue Types** action to look up issue type IDs for the project.",
+    },
+    additionalProperties: {
```
ashwins01, line 41: Nit: because this key already exists in `common.props`, the override keeps its original position, so **Additional properties** renders above **Project ID** and **Issue Type** in the form. Defining `projectId`/`issueTypeId` before the `...common.props` spread (or rebuilding the props object in the desired order) would put the required fields first.

ashwins01 (approved):
LGTM, ready for QA.

michelle0927:
## Eval results — jira `create-issue` / `update-issue` (post review-feedback fixes)

Ran the 5-eval suite in `evals/jira/` against the current branch (private-registry publish of `create-issue`/`update-issue` with today's `parseFields`, `notifyUsers`/`updateHistory`, and prop-order fixes).

| # | Eval | Status | Tool calls | Time |
|---|------|--------|------------|------|
| 1 | [jira-transition-issue] Transition issue to In Progress | ⚠️ WARN | 3 | 24.8s |
| 2 | [jira-create-issue] Create issue via freeform `additionalProperties` | ✅ PASS | 4 | 26.0s |
| 3 | [jira-update-issue] Update issue fields via `additionalProperties` | ✅ PASS | 2 | 19.3s |
| 4 | [jira-update-issue] Transition an issue using a discovered transition ID | ✅ PASS | 3 | 24.1s |
| 5 | Log and start work on a newly discovered bug (discovery) | ✅ PASS | 6 | 44.3s |

**4 passed / 0 failed / 1 warned — 80% pass rate** (95% CI 38–96%, n=5). Model: `claude-sonnet-5`. Cost: $1.36.

- Evals 2 & 3 directly exercise this PR's `additionalProperties` fix (plain-text `summary` values) — both passed end to end, including the agent self-discovering cloud ID / project ID / issue type ID with no seeded hints.
- Eval 1's warning is **not a regression from this PR** — the agent satisfied the request via `update-issue`'s own transition path instead of calling the dedicated `jira-transition-issue` tool. Pre-existing tool-routing behavior, unrelated to the dynamic-props removal.

_Run: `jira-2026-09-18T16-21-50-747Z`, harness: `claude_code_default`._

PR #22004, opened 2026-09-17
Bump version numbers for various Google Ads components and update utility functions

### LINKED ISSUE #21961: google_ads: query builder produces invalid GAQL for some fields and search terms
## Problem

1. `checkPrefix` in `common/utils.mjs` checks whether a field name starts with the resource name, rather than the resource name followed by a dot. Fields whose own name begins with the resource name are left unprefixed. On a campaign report, `campaign_budget` and `campaign_group` produce `SELECT campaign_budget FROM campaign`, which the API rejects. `customer_agreement_setting.*` has the same problem on the customer resource.

   This is easy to hit because the dropdown label shows the unprefixed name while the stored value is prefixed. Passing the name shown on screen is enough to break the query, and agents reading labels are especially likely to do it.

2. `sanitizeGaqlString` escapes single quotes only. GAQL requires `[`, `]`, `%` and `_` to be bracket-escaped when used with `LIKE`. Every search-as-you-type dropdown builds `LIKE '%<text>%'`, so a search containing one of those characters can produce an invalid query. The Managed Account picker labels manager accounts as `[Manager] ...`, so searching for the text shown in the list is a plausible way to trigger it.

3. Both cases return a 400, and rejected requests still count against the daily quota.

## What's needed

- Field prefixing that matches the full prefix rather than a partial string.
- Search text escaped for the `LIKE` operator before it goes into a query.

## Notes

- GAQL grammar and `LIKE` escaping rules: https://developers.google.com/google-ads/api/docs/query/grammar
- Failed requests that return a Google Ads failure still count against the daily quota: https://developers.google.com/google-ads/api/docs/best-practices/quotas
- Context: quota exhaustion reported in #19945, #21509, #21554, #21782, #21926.

Priyadharshan-Pdm:
## Summary

Closes  #21961

- Version Updates done for all the actions
- Corrected the utility function that causes this issue

## Root cause

1. checkPrefix checked whether a name started with campaign instead of campaign.. Fields like campaign_budget were treated as already prefixed, so the query became SELECT campaign_budget FROM campaign, which is invalid.
2. sanitizeGaqlString only escaped single quotes. GAQL LIKE also needs [, ], % and _ escaped, so searching for something like [Manager] in a dropdown broke the query.

## Fix (common/utils.mjs)

1. checkPrefix now checks for `${prefix}.`, so campaign_budget correctly becomes campaign.campaign_budget.
2. sanitizeGaqlString now also wraps [ ] % _ in brackets, e.g. _ becomes [_].
3. Patch version bump on every action and source that uses these helpers, plus package.json.

## Checklist

### Versioning

### New app

### CodeRabbit review

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Google Ads query matching by safely handling special characters in literal `LIKE` filters.
  - Corrected prefix detection so prefixes are recognized only when followed by a dot, preventing incorrectly formatted queries.

- **Chores**
  - Updated the Google Ads integration package and numerous actions and event sources to newer versions, ensuring the latest component metadata is available.


michelle0927 (approved):
LGTM! Ready for QA.

